PR #51, opened 2020-08-18
Sample App

fjnoyp:
New PR for the Sample App with all categories integrated and using Dart public packages from pub.dev

- Auth
- Analytics
- Storage

Basic image picker app.

*Issue #, if available:*

*Description of changes:*


README.md
```diff
@@ -26,6 +26,9 @@ We are iterating and looking for feedback and collaboration, so please [**let us
  - Predictions
  - Datastore
  - Hub Events (Listening to the Amplify events)
+ - iOS Error Events in Amplify Analytics
+
+> Amplify for Flutter is in preview, which should be considered an alpha version and not recommended for production deployments. We are iterating and looking for feedback and collaboration, so please [**LET US KNOW YOUR FEEDBACK**](https://github.com/aws-amplify/amplify-flutter/issues/5) on our direction and roadmap.
```
jamesonwilliams, line 31: Is this going to be rebased on top of https://github.com/aws-amplify/amplify-flutter/pull/38/files? It appears to contain duplicate changes.

fjnoyp: Hi @jamesonwilliams this is a new PR.  By making our github repo public, we had to recreate all of our original PRs and delete the old ones (I'll delete pull 38 that you reference right now).

fjnoyp: I've looked at and implemented all your comments from the previous, closed PR

packages/examples/sample_app/README.md
```diff
@@ -0,0 +1,16 @@
+# sample_app
```
haverchuck, line 1: Should we update this with instructions for running the sample app?

README.md
```diff
@@ -26,6 +26,9 @@ We are iterating and looking for feedback and collaboration, so please [**let us
  - Predictions
  - Datastore
  - Hub Events (Listening to the Amplify events)
```
haverchuck, line 28: A bit off topic, but we currently listen for Auth Hub events... can we clarify that Hub events would be needed for storage and analytics, but are present for Auth?

packages/examples/sample_app/lib/Views/ImageUploader.dart
```diff
@@ -0,0 +1,41 @@
...
+
+class ImageUploader extends StatelessWidget {
+  void _upload(BuildContext context) async {
+    try {
+      print('In upload');
+      // Uploading the file with options
+      File local = await FilePicker.getFile(type: FileType.image);
+      local.existsSync();
+      final key = new DateTime.now().toString();
+      Map<String, String> metadata = <String, String>{};
```
haverchuck, line 15: Is the metadata really necessary for the sample app?

packages/examples/sample_app/lib/main.dart
```diff
@@ -0,0 +1,85 @@
...
+        storagePlugins: [storage],
+        analyticsPlugins: [analytics]);
+
+    // Initialize AmplifyFlutter
+    await amplify.configure(amplifyconfig);
+    setState(() {
+      _isAmplifyConfigured = true;
+    });
+
+    AnalyticsEvent event = AnalyticsEvent("image_url_retrieved");
```
haverchuck, line 62: The amplify-android/ios libs store signIn data so that the user's session persists upon app exit and restart.  We should demonstrate the ability to check the session at app startup, and enter a signedIn state depending on the result.

You can do something like this:

```
    var isSignedIn = false;
   ... 
    try {
      isSignedIn = await _isSignedIn();
    } on AuthError catch(e) {
      print("User is not signed in.");
    }
   ...
    Future<bool> _isSignedIn() async {
      final session = await Amplify.Auth.fetchAuthSession(
        options: CognitoSessionOptions(getAWSCredentials: false)
      );
      return session.isSignedIn;
    }
```

haverchuck: You don't necessarily need to pass in the options parameter.  I left it in to educate the user.

packages/examples/sample_app/README.md
fjnoyp: Yeah good call, updated it.

README.md
fjnoyp: done

 - Hub Events for Storage and Analytics (but are implemented on Auth)

packages/examples/sample_app/lib/Views/ImageUploader.dart
fjnoyp: I think it's good to include, it's an example on how to use it for the developer.

packages/examples/sample_app/lib/main.dart
fjnoyp: Added and tested your suggested code changes.

PR #53, opened 2020-08-19
feat(amplify_auth_cognito): getCurrentUser

haverchuck:
*Description of changes:*
1. Adds getCurrentUser method 
2. Fixes bug wherein fetchSession was not throwing error if user was signedout but awsCredentials were not requested.


packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterAuthIUser.kt
```diff
@@ -0,0 +1,32 @@
+/*
```
jamesonwilliams, line 1: The file is named `...IUser...`, but the class is `FlutterAuthUser`. Is the `I` intentional?

packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthCognitoPluginTest.kt
```diff
@@ -293,6 +293,28 @@ class AmplifyAuthCognitoPluginTest {
...
 
+    @Test
+    fun getCurrentUser_returnsSuccess() {
+        // Arrange
+        doAnswer { invocation: InvocationOnMock ->
+            plugin.prepareUpdatePasswordResult(mockResult)
+            null as Void?
+        }.`when`(mockAuth).getCurrentUser()
+
+        val data: HashMap<*, *> = hashMapOf(
```
jamesonwilliams, line 304: Does it need to be a mutable map? And does it need to be a mutable map of the hash variety? Why not a default immutable
```suggestion
        val data: Map<*, *> = mapOf(
```

Related: [a discussion on `hashMapOf` vs `mutableMapOf`](https://stackoverflow.com/a/51563494/695787)

packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthCognitoPluginTest.kt
```diff
@@ -293,6 +293,28 @@ class AmplifyAuthCognitoPluginTest {
...
+                "userSub" to "userSub"
+        )
+        val arguments: HashMap<String, Any> = hashMapOf("data" to data)
+        val call = MethodCall("getCurrentUser", arguments)
+
+        // Act
+        plugin.onMethodCall(call, mockResult)
+
+        // Assert
+        verify(mockResult, times(1)).success(ArgumentMatchers.any<LinkedTreeMap<String, Any>>());
```
jamesonwilliams, line 315: `times(1)` is actually the default verification. Unless you add it for self-documenting, DAMP purposes, you do just:
```
verify(mock).method(...)
```

Oh also you don't need the trailing `;`, in Kotlin.

packages/amplify_auth_cognito/example/lib/main.dart
```diff
@@ -346,6 +346,24 @@ class _MyAppState extends State<MyApp> {
...
 
+  void _getCurrentUser() async {
+    setState(() {
+      error = "";
+      exceptions = [];
+    });
+    try {
+      AuthUser res = await Amplify.Auth.getCurrrentUser();
+      print(res);
+    } on AuthError catch (e) {
```
jamesonwilliams, line 357: I'd encourage you to not use single letter variable names anywhere. They're very hard to search for with e.g. `less`, `vim`, even IDEs. How about `error`, or `authError`, for this example. But, same general advice around the code base.

packages/amplify_auth_cognito/example/lib/main.dart
```diff
@@ -346,6 +346,24 @@ class _MyAppState extends State<MyApp> {
...
+      setState(() {
+        error = e.cause;
+        e.exceptionList.forEach((el) {
```
jamesonwilliams, line 360: `el` is too short -- but the real problem is that it looks like __E__exception __L__ist, which is a collection of many things. Since this is a lambda argument, iterating over the list, this variable should be _singular_ in some way. Perhaps `forEach((exception) {`, etc.

packages/amplify_auth_cognito/ios/Classes/SwiftAuthCognito.swift
```diff
@@ -284,7 +286,12 @@ public class SwiftAuthCognito: NSObject, FlutterPlugin, FlutterStreamHandler {
                 flutterResult(sessionData.toJSON())
             } else {
                 let sessionData = try FlutterFetchSessionResult(res: result)
-                flutterResult(sessionData.toJSON())
+                if (sessionData.isSignedIn == true) {
```
jamesonwilliams, line 289: Hm, do you need the explicit `== true`?

packages/amplify_auth_cognito/ios/Classes/SwiftAuthCognito.swift
```diff
@@ -293,6 +300,22 @@ public class SwiftAuthCognito: NSObject, FlutterPlugin, FlutterStreamHandler {
       }
     }
     
+    private func onGetCurrentUser(flutterResult: @escaping FlutterResult) {
+      do {
+        guard let user = Amplify.Auth.getCurrentUser() else {
+           // TODO: Use the AuthPluginErrorConstants
+           throw AuthError.signedOut(
+           "You are currently signed out.",
+           "Please sign in and reattempt the operation.")
```
jamesonwilliams, line 309:
```suggestion
               "You are currently signed out.",
               "Please sign in and reattempt the operation."
           )
```

packages/amplify_auth_cognito/ios/Classes/SwiftAuthCognito.swift
```diff
@@ -293,6 +300,22 @@ public class SwiftAuthCognito: NSObject, FlutterPlugin, FlutterStreamHandler {
...
+        }
+        let userData = FlutterAuthUserResult(res: user)
+        flutterResult(userData.toJSON())
+        
+      } catch {
+          self.handleAuthError(error: error as! AuthError, flutterResult: flutterResult,  msg: FlutterAuthErrorMessage.GET_CURRENT_USER.rawValue)
+        }
```
jamesonwilliams, line 316:
```suggestion
      } catch {
          self.handleAuthError(
              error: error as! AuthError,
              flutterResult: flutterResult,
              msg: FlutterAuthErrorMessage.GET_CURRENT_USER.rawValue
          )
      }
```

packages/amplify_auth_cognito/lib/amplify_auth_cognito.dart
```diff
@@ -86,6 +86,11 @@ class AmplifyAuthCognito extends AuthPluginInterface {
     return res; 
   }
 
+  Future<AuthUser> getCurrentUser({AuthUserRequest request}) async {
+    final res = await _instance.getCurrentUser();
```
jamesonwilliams, line 90: Do you need the `res` or can you just do like `return (await _instance.getCurrentUser());` or some such?

packages/amplify_auth_cognito/test/amplify_auth_cognito_getCurrentUser_test.dart
```diff
@@ -0,0 +1,71 @@
+/*
```
jamesonwilliams, line 1: Given that this file is already in `amplfy_auth_cognito/test`, do you need to include `amplify_auth_cognito` in the filename _again_? -- general question around naming for any/all test files.

packages/amplify_auth_cognito/test/amplify_auth_cognito_getCurrentUser_test.dart
```diff
@@ -0,0 +1,71 @@
...
+import 'package:flutter/services.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
+import 'package:amplify_core/amplify_core.dart';
+
+void main() {
+  const MethodChannel authChannel = MethodChannel('com.amazonaws.amplify/auth_cognito');
+  const MethodChannel coreChannel = MethodChannel('com.amazonaws.amplify/core');
+
+  Amplify amplify = new Amplify();
```
jamesonwilliams, line 25:
```suggestion
  Amplify amplify = Amplify();
```

packages/amplify_auth_cognito/test/amplify_auth_cognito_getCurrentUser_test.dart
```diff
@@ -0,0 +1,71 @@
...
+  setUp(() {
+    authChannel.setMockMethodCallHandler((MethodCall methodCall) async {
+      switch(testCode) {
+        case 1:
+          return {
+            "username": "testUser",
+            "userSub": "testSub"
+          };
+          case 2:
+            return throw PlatformException(code: "AMPLIFY_EXCEPTION", message: "AMPLIFY_GET_CURRENT_USER_FAILED", details: {} );
```
jamesonwilliams, line 41: Do you need to specify `details: {}`? In the definition of the `PlatformExcetpion`, can `{}` be a default value assigned to the optional `details` argument?

packages/amplify_auth_cognito/test/amplify_auth_cognito_getCurrentUser_test.dart
```diff
@@ -0,0 +1,71 @@
...
+  tearDown(() {
+    authChannel.setMockMethodCallHandler(null);
+    coreChannel.setMockMethodCallHandler(null);
+  });
+
+  test('getCurrentUser request returns an AuthUser', () async {
+    testCode = 1;
+    await amplify.addPlugin(authPlugins: [auth]);
+    await amplify.configure("{}");
+    expect(await Amplify.Auth.getCurrrentUser(), isInstanceOf<AuthUser>());
```
jamesonwilliams, line 58: Three awaits in a row - can you chain these with `.then(` clauses, instead, and call `await` once?

packages/amplify_auth_cognito/test/amplify_auth_cognito_getCurrentUser_test.dart
```diff
@@ -0,0 +1,71 @@
...
+  });
+
+  test('fetchSession thrown PlatFormException results in AuthError', () async {
```
jamesonwilliams, line 61:
```suggestion
  test('PlatformException in getCurrentUser surfaces as AuthError', () async {
```

packages/amplify_auth_plugin_interface/lib/src/Session/AuthUser.dart
```diff
@@ -0,0 +1,25 @@
...
+
+import 'package:flutter/foundation.dart';
+
+
+class AuthUser {
+  String userId;
+  String username;
+  AuthUser({@required this.userId, @required this.username}) {
+    this.username = username;
+   }
```
jamesonwilliams, line 24:
```suggestion
  }
```

packages/amplify_core_plugin_interface/lib/amplify_auth_category.dart
```diff
@@ -64,6 +64,11 @@ class AuthCategory {
     return plugins[0].confirmPassword(request: request);
   }
 
+  Future<AuthUser> getCurrrentUser() {
+    var request = AuthUserRequest();
+    return plugins[0].getCurrentUser();
```
jamesonwilliams, line 69: Let's suppose there's no plugin. Do you need to validate that `plugins[0]` exists? Or will that throw a range error? Or will `getCurrentException()` throw an error because it is invoked on a null reference?

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterAuthIUser.kt
haverchuck: Fixed.

packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthCognitoPluginTest.kt
haverchuck: I think this could a Map as you say, but I'm going to leave this for a larger refactor of the tests.  I need to look at the Flutter method channel before making this change.

packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthCognitoPluginTest.kt
haverchuck: Removed the comma.  I prefer to leave the explicit "1" and not rely on knowledge of defaults (for this particular project).

packages/amplify_auth_cognito/example/lib/main.dart
haverchuck: Agreed... but again this will require several changes across this file for consistency, so I'm going to do it in a separate PR.

packages/amplify_auth_cognito/example/lib/main.dart
haverchuck: Same as previous comment.

packages/amplify_auth_cognito/ios/Classes/SwiftAuthCognito.swift
haverchuck: Fixed.

packages/amplify_auth_cognito/ios/Classes/SwiftAuthCognito.swift
haverchuck: This is definitely more readble, but I think I should put this in a style PR rather than this one (again, for consistency)

packages/amplify_auth_cognito/lib/amplify_auth_cognito.dart
haverchuck: You could return directly, but I prefer this way for debugging purposes.

packages/amplify_auth_cognito/test/amplify_auth_cognito_getCurrentUser_test.dart
haverchuck: No, I think we can make that change to all the files (but it's going to make this PR appear much larger since it will be interpreted as deleted/new files, so I'm going to wait).

packages/amplify_auth_cognito/test/amplify_auth_cognito_getCurrentUser_test.dart
haverchuck: Yeah, this can just be null: https://api.flutter.dev/flutter/services/PlatformException-class.html

packages/amplify_auth_cognito/test/amplify_auth_cognito_getCurrentUser_test.dart
haverchuck: What's the difference?

packages/amplify_core_plugin_interface/lib/amplify_auth_category.dart
haverchuck: This will throw a range error - we need to discuss this as a team because we still closing on our strategy for allowing multiple plugins. I fixed this temporarily.

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterAuthErrorMessage.kt
```diff
@@ -72,6 +72,11 @@ enum class FlutterAuthFailureMessage {
       return "AMPLIFY_CONFIRM_PASSWORD_FAILED"
     }
   },
+  CURRENT_USER{
```
Ashish-Nanda, line 75: Not important right now, but you can change this to enums similar to how I've done in the storage category. You do not need to override as well.

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
```diff
@@ -233,6 +233,27 @@ class AmplifyAuthCognitoMethodChannel extends AmplifyAuthCognito {
...
+        },
+      );
+      res = _formatAuthUserResponse(data);
+      return res;
+    } on PlatformException catch(e) {
+      _throwError(e);
+    } on AuthError catch(e) {
+      throw(e);
+    }
+    return res;
```
Ashish-Nanda, line 254: There is already a return above. Is this line getting executed?

packages/amplify_auth_plugin_interface/lib/src/Session/AuthUser.dart
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+import 'package:flutter/foundation.dart';
+
+
+class AuthUser {
+  String userId;
+  String username;
+  AuthUser({@required this.userId, @required this.username}) {
+    this.username = username;
```
Ashish-Nanda, line 23: What is this line for?

packages/amplify_auth_plugin_interface/lib/src/Session/AuthUserOptions.dart
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+class AuthUserOptions {
+
+  const AuthUserOptions();
+
```
Ashish-Nanda, line 18: extra whitespace

packages/amplify_auth_plugin_interface/lib/src/Session/AuthUserRequest.dart
```diff
@@ -0,0 +1,25 @@
...
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+import './AuthUserOptions.dart';
+
+
```
Ashish-Nanda, line 17: extra white space. Consider using format on save with the dart plugin in VS Code.

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
haverchuck: This is to stop the compiler from complaining about not getting a return of the expected type.

packages/amplify_auth_plugin_interface/lib/src/Session/AuthUser.dart
haverchuck: good catch- fixed it.

packages/amplify_auth_plugin_interface/lib/src/Session/AuthUserOptions.dart
haverchuck: fixed.

packages/amplify_auth_plugin_interface/lib/src/Session/AuthUserRequest.dart
```diff
@@ -0,0 +1,25 @@
...
+class AuthUserRequest {
+  AuthUserOptions options;
```
Ashish-Nanda, line 19: How does this member get initialized or used?

haverchuck: It does not right now, but I'm thinking it might once there is additional work in native libraries. Just including it for future use.

packages/amplify_auth_plugin_interface/lib/src/Session/AuthUserRequest.dart
haverchuck: fixed.

packages/amplify_auth_cognito/test/amplify_auth_cognito_getCurrentUser_test.dart
jamesonwilliams: Great question. I had to re-asses my assumptions here.

[This guy](https://blog.logrocket.com/promise-chaining-is-dead-long-live-async-await-445897870abc/), at least, has the opposite advice. His blog post is more well thought out than my passing comment, so let's go with that.

PR #66, opened 2020-08-19
fix(amplify_auth_cognito): gradle cognito

haverchuck:
*Issue #, if available:*
Fixing gradle version of cognito gradle dep.


packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthCognitoPluginTest.kt
```diff
@@ -38,7 +38,7 @@ class AmplifyAuthCognitoPluginTest {
     private val signUpStep = AuthNextSignUpStep(AuthSignUpStep.CONFIRM_SIGN_UP_STEP, emptyMap(), codeDeliveryDetails)
     private val signInStep = AuthNextSignInStep(AuthSignInStep.CONFIRM_SIGN_IN_WITH_SMS_MFA_CODE, emptyMap(), codeDeliveryDetails)
     private val resetStep = AuthNextResetPasswordStep(AuthResetPasswordStep.CONFIRM_RESET_PASSWORD_WITH_CODE, emptyMap(), codeDeliveryDetails)
-    private val mockSignUpResult = AuthSignUpResult(false, signUpStep)
+    private val mockSignUpResult = AuthSignUpResult(false, signUpStep, null)
```
jamesonwilliams, line 41: This was a topic of active debate recently. I believe @TrekSoft wants to send a non-null user object back after sign up, so that the customer can understand what user was created, before sign-in.

Issue #136, opened 2020-10-06
Missing Signup options cause NoSuchMethodError


JamesMcIntosh:
**Describe the bug**
`AuthCategory.signUp()` parameter named `options` is treated as mandatory but is not a required field.

**To Reproduce**
Execute the following code and 
```
final SignUpResult result = await Amplify.Auth.signUp(
  username: "username",
  password: "password",
);
```
`SignUpRequest` is created within the `signUp` method which has mandatory parameter named `options` (which is null set to null). This causes the exception

```NoSuchMethodError: The method 'serializeAsMap' was called on null.```

See: [SignUpRequest Line 29](https://github.com/aws-amplify/amplify-flutter/blob/562c009884316ca5fd1823aaab35f01fb1dab675/packages/amplify_auth_plugin_interface/lib/src/SignUp/SignUpRequest.dart#L29)
```
pendingRequest['options'] = options.serializeAsMap();
```

**Fix**

1. Update method signature in `AuthCategory.signUp` to make options a required field
```
Future<SignUpResult> signUp({@required String username, @required password, @required SignUpOptions options}
```
2. Update method signature in `AuthCategory.signUp` to give options a default value
```
Future<SignUpResult> signUp({@required String username, @required password, SignUpOptions options = const SignUpOptions()}
```
3. Null check options in `SignUpRequest.serializeAsMap()`
```
if (options != null) {
  pendingRequest['options'] = options.serializeAsMap();
}
```

**Comment by haverchuck** [comment]
@JamesMcIntosh We can't make the options field required at this time, but we ARE implementing a fix to serializeAsMap call as you suggested. The result will be a response from the platform code indicating that this request is malformed, which should hopefull be more helpful.

**Comment by JamesMcIntosh** [comment]
@haverchuck Sounds good as long as the platform code error explains that it's the option field missing

**Comment by Ashish-Nanda** [comment]
@JamesMcIntosh A fix for this bug was released today in version `0.0.1-dev.3`, so will go ahead and close it out. Please give it a try and let us know if you are still facing issues, and we can look into it.

**Comment by haverchuck** [comment]
@JamesMcIntosh I don't think we are currently surfacing the specific fields that are missing, but we are looking at large, general improvement to our error handling, so please be patient.  We are aware that this (and other) error visibility issues exist.

**Comment by JuanDavid1996** [comment]
Hi, I update amplify_auth_cognito to ^ 0.2.1, then I made this solution
```
[...]

PR #137, opened 2020-10-07
fix(amplify_auth_cognito) AuthError instead of serialization error

haverchuck:
*Issue #, if available:*
#136 

*Description of changes:*
Returns 'AMPLIFY_REQUEST_MALFORMED' AuthError instead of obtuse error for serialization on null


packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterSignUpRequest.kt
```diff
@@ -74,17 +74,19 @@ data class FlutterSignUpRequest(val map: HashMap<String, *>) {
 
     companion object {
         fun validate(req : HashMap<String, *>?): Boolean {
-            var valid: Boolean = true;
             if (req == null) {
-                valid = false;
+                return false;
+            }
+            if (req.get("options") == null) {
```
Ashish-Nanda, line 80: It looks like options is required. Should we not add the `@required` annotation in the API as well to make it apparent to customers, or pass a default value if possible.
eg.
```
Future<SignUpResult> signUp({@required String username, @required password, @required SignUpOptions options}
```
This is in line with suggestions 1 and 2 in #136

Ashish-Nanda:
The change looks good. 
I think one additional step as mentioned above would be to add `@required` or a default value for SignUpOptions in the customer facing API to reduce the chance of a malformed request error being thrown in the platform code.

haverchuck:
@Ashish-Nanda Making options required would be problematic for future plugins using the interface.

Ashish-Nanda (approved):
Discussed offline. Looks good for now.

PR #170, opened 2020-10-30
test(amplify_auth_cognito): confirmSignIn unit test

haverchuck:
*Description of changes:*
Adds unit test for confirmSignIn method. Introduces assertions as a way to test serialization over the method channel.

### DIFF
--- a/packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
+++ b/packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
@@ -21,18 +21,12 @@ import 'package:amplify_core/amplify_core.dart';
 
 extension IsEqual on SignInResult {
   bool isMostlyEqual(SignInResult comparator) {
-    if(
-      comparator.isSignedIn == isSignedIn &&
+    return comparator.isSignedIn == isSignedIn &&
       comparator.nextStep.signInStep == nextStep.signInStep &&
       comparator.nextStep.additionalInfo.length == nextStep.additionalInfo.length &&
       comparator.nextStep.codeDeliveryDetails.destination == nextStep.codeDeliveryDetails.destination &&
       comparator.nextStep.codeDeliveryDetails.attributeName == nextStep.codeDeliveryDetails.attributeName &&
-      comparator.nextStep.codeDeliveryDetails.deliveryMedium == nextStep.codeDeliveryDetails.deliveryMedium
-    ) {
-      return true;
-    } else {
-      return false;
-    }
+      comparator.nextStep.codeDeliveryDetails.deliveryMedium == nextStep.codeDeliveryDetails.deliveryMedium;
   }
 }
 
--- a/packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
+++ b/packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
@@ -16,7 +16,13 @@
 import 'package:flutter/services.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
+import 'package:amplify_auth_cognito/method_channel_auth_cognito.dart';
 import 'package:amplify_core/amplify_core.dart';
+
+import 'package:mockito/mockito.dart';
+
+class MockAuthChannel extends Mock implements AmplifyAuthCognitoMethodChannel {}
+
 
 void main() {
   const MethodChannel authChannel = MethodChannel('com.amazonaws.amplify/auth_cognito');
@@ -31,18 +37,24 @@ void main() {
 
   setUp(() {
     authChannel.setMockMethodCallHandler((MethodCall methodCall) async {
-      switch(testCode) {
-        case 1:
-          return {
-            "isSignedIn": false,
-            "nextStep": {
-              "signInStep": "DONE",
-              "codeDeliveryDetails":  { "atttibuteName": "email" }
-            }
-          };
-        case 2:
-          return throw PlatformException(code: "AMPLIFY_EXCEPTION", message: "AMPLIFY_SIGNIN_FAILED", details: {} );
-      } 
+      if (methodCall.method == "confirmSignIn") {
+        assert(methodCall.arguments["data"] is Map);
+        assert(methodCall.arguments["data"]["confirmationCode"] is String);
+        switch(testCode) {
+          case 0:
+            return {
+              "isSignedIn": false,
+              "nextStep": {
+                "signInStep": "DONE",
+                "codeDeliveryDetails":  { "atttibuteName": "email" }
+              }
+            };
+          case 1: 
+            return throw PlatformException(code: "AMPLIFY_EXCEPTION", message: "AMPLIFY_CONFIRM_SIGNIN_FAILED", details: {} );
+        }
+      } else {
+        return true;
+      }     
     });
     coreChannel.setMockMethodCallHandler((MethodCall methodCall) async {
       return true;
@@ -54,29 +66,24 @@ void main() {
     coreChannel.setMockMethodCallHandler(null);
   });
 
-
-  test('signUp request returns a SignInResult', () async {
-    testCode = 1;
+  test('confirmSignIn request returns a SignInResult', () async {
     await amplify.addPlugin(authPlugins: [auth]);
     await amplify.configure("{}");
-    expect(await Amplify.Auth.signIn(username: 'testUser', password: '123'), isInstanceOf<SignInResult>());
-  });
-
-  test('signIn request nextStep casts to AuthNextSignStep and AuthNextStep', () async {
-    testCode = 1;
-    var res = await Amplify.Auth.signIn(username: 'testUser', password: '123');
+    var res = await Amplify.Auth.confirmSignIn(confirmationValue: "iAmLegit");
+    expect(res, isInstanceOf<SignInResult>());
+    expect(res.isSignedIn, false);
     expect(res.nextStep, isInstanceOf<AuthNextSignInStep>());
-    expect(res.nextStep, isInstanceOf<AuthNextStep>());
   });
 
   test('signIn thrown PlatFormException results in AuthError', () async {
-    testCode = 2;
+    testCode = 1;
     AuthError err;
-   try {
-     await Amplify.Auth.signIn(username: 'testUser', password: '123');
-   } on AuthError catch (e) {
+    try {
+      await Amplify.Auth.confirmSignIn(confirmationValue: "iAmNotLegit");
+    } catch (e) {
       err = e;
     } 
-    expect(err.cause, "AMPLIFY_SIGNIN_FAILED");
+    expect(err.cause, "AMPLIFY_CONFIRM_SIGNIN_FAILED");
+    expect(err, isInstanceOf<AuthError>());
   });
 }
--- a/packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
+++ b/packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
@@ -20,6 +20,7 @@ import 'package:amplify_auth_cognito/method_channel_auth_cognito.dart';
 import 'package:amplify_core/amplify_core.dart';
 
 extension IsEqual on SignInResult {
+  // This method only checks the length of the additionalInfo field, and the values of all other fields
   bool isMostlyEqual(SignInResult comparator) {
     return comparator.isSignedIn == isSignedIn &&
       comparator.nextStep.signInStep == nextStep.signInStep &&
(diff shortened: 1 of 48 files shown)


packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
```diff
@@ -0,0 +1,84 @@
...
+  const MethodChannel coreChannel = MethodChannel('com.amazonaws.amplify/core');
+
+  Amplify amplify = new Amplify();
+  AmplifyAuthCognito auth = AmplifyAuthCognito();
+
+  TestWidgetsFlutterBinding.ensureInitialized();
+
+  int testCode = 0;
+
+  setUp(() {
```
Amplifiyer, line 33: We should avoid the per test setup in the "before each setup() method" and do the setup in the test itself. This method should be reserved for setup that is common across all tests.

We should do this across all auth test.

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
```diff
@@ -0,0 +1,84 @@
...
+    coreChannel.setMockMethodCallHandler(null);
+  });
+
+  test('confirmSignIn request returns a SignInResult', () async {
+    await amplify.addPlugin(authPlugins: [auth]);
+    await amplify.configure("{}");
+    var res = await Amplify.Auth.confirmSignIn(confirmationValue: "iAmLegit");
+    expect(res, isInstanceOf<SignInResult>());
+    expect(res.isSignedIn, false);
+    expect(res.nextStep, isInstanceOf<AuthNextSignInStep>());
```
Amplifiyer, line 70: Can we instead assert against a full response object that we expect.

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
```diff
@@ -0,0 +1,107 @@
...
+            "isSignedIn": false,
+            "nextStep": {
+              "signInStep": "DONE",
+              "codeDeliveryDetails":  {
+                "deliveryMedium": "EMAIL",
+                "attributeName": "email",
+                "destination": "test@test.test"
+              }
+            }
+          };
```
jamesonwilliams, line 58: Fixes to indentation:
```suggestion
  test('confirmSignIn request returns a SignInResult', () async {
    authChannel.setMockMethodCallHandler((MethodCall methodCall) async {
      if (methodCall.method == "confirmSignIn") {
        assert(methodCall.arguments["data"] is Map);
        assert(methodCall.arguments["data"]["confirmationCode"] is String);
        return {
          "isSignedIn": false,
          "nextStep": {
            "signInStep": "DONE",
            "codeDeliveryDetails":  {
              "deliveryMedium": "EMAIL",
              "attributeName": "email",
              "destination": "test@test.test"
            }
          }
        };
```

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
```diff
@@ -0,0 +1,107 @@
...
+        signInStep: "DONE",
+      )
+    );
+    var res = await Amplify.Auth.confirmSignIn(confirmationValue: "iAmLegit");
+    expect(res.isSignedIn, equals(expectation.isSignedIn));
+    expect(res.nextStep.signInStep, equals(expectation.nextStep.signInStep));
+    expect(res.nextStep.additionalInfo, equals(expectation.nextStep.additionalInfo));
+    expect(res.nextStep.codeDeliveryDetails.attributeName, equals(expectation.nextStep.codeDeliveryDetails.attributeName));
+    expect(res.nextStep.codeDeliveryDetails.deliveryMedium, equals(expectation.nextStep.codeDeliveryDetails.deliveryMedium));
+    expect(res.nextStep.codeDeliveryDetails.destination, equals(expectation.nextStep.codeDeliveryDetails.destination));
```
jamesonwilliams, line 84: instead of having many smaller `expect` checks, can you create an expected `res`, and then do a single `expect(expectedRes, equals(actualRes));`?

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
```diff
@@ -0,0 +1,107 @@
...
+ authChannel.setMockMethodCallHandler((MethodCall methodCall) async {
+    if (methodCall.method == "confirmSignIn") {
+      assert(methodCall.arguments["data"] is Map);
+      assert(methodCall.arguments["data"]["confirmationCode"] is String);
+      return throw PlatformException(code: "AMPLIFY_EXCEPTION", message: "AMPLIFY_CONFIRM_SIGNIN_FAILED", details: {} );
+    } else {
+      return true;
+    }     
+  });
+    AuthError err;
```
jamesonwilliams, line 98: Indentation:

```suggestion
  test('signIn thrown PlatFormException results in AuthError', () async {
    authChannel.setMockMethodCallHandler((MethodCall methodCall) async {
      if (methodCall.method == "confirmSignIn") {
        assert(methodCall.arguments["data"] is Map);
        assert(methodCall.arguments["data"]["confirmationCode"] is String);
        return throw PlatformException(code: "AMPLIFY_EXCEPTION", message: "AMPLIFY_CONFIRM_SIGNIN_FAILED", details: {} );
      } else {
        return true;
      }     
    });
    AuthError err;
```

jamesonwilliams (approved):

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
haverchuck: I don't this think this works in Dart due to object hashCodes.

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
haverchuck: Yes - although I still check the values individually.

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
haverchuck: I've made this change for the confirmSignIn tests.

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
jamesonwilliams: Gotcha, I think this is the same situation as occurs in Java, then, where you have to implement your own equals and hashCode? https://stackoverflow.com/a/52602964/695787

Amplifiyer: You could use extensions (even in the test classes) to add a `toString()` or `equals()` method if you don't want to pollute data classes.

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
haverchuck: Fixed this on my branch.

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
haverchuck: Fixed on my branch.

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
haverchuck: @Amplifiyer @jamesonwilliams - I tried my hand at an extension,

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
```diff
@@ -19,6 +19,23 @@ import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
...
 
+extension IsEqual on SignInResult {
+  bool isMostlyEqual(SignInResult comparator) {
+    if(
+      comparator.isSignedIn == isSignedIn &&
+      comparator.nextStep.signInStep == nextStep.signInStep &&
+      comparator.nextStep.additionalInfo.length == nextStep.additionalInfo.length &&
+      comparator.nextStep.codeDeliveryDetails.destination == nextStep.codeDeliveryDetails.destination &&
+      comparator.nextStep.codeDeliveryDetails.attributeName == nextStep.codeDeliveryDetails.attributeName &&
+      comparator.nextStep.codeDeliveryDetails.deliveryMedium == nextStep.codeDeliveryDetails.deliveryMedium
```
Amplifiyer, line 30: nit: you can just return this statement.

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
```diff
@@ -0,0 +1,118 @@
...
+    authChannel.setMockMethodCallHandler(null);
+    coreChannel.setMockMethodCallHandler(null);
+  });
+
+  test('confirmSignIn request returns a SignInResult', () async {
+
+    authChannel.setMockMethodCallHandler((MethodCall methodCall) async {
+      if (methodCall.method == "confirmSignIn") {
+        assert(methodCall.arguments["data"] is Map);
+        assert(methodCall.arguments["data"]["confirmationCode"] is String);
```
Amplifiyer, line 64: We can do more detailed assertions here to ensure that the method channel was called with right parameters (testing serialization in dart)

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
haverchuck: ack, yeah good call.

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
haverchuck: These actually are the expected parameters - it's just one attribute for this method (still using a "data" map in anticipation of adding lambda metadata down the road).

haverchuck: @Amplifiyer ^^

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
```diff
@@ -19,6 +19,23 @@ import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
 import 'package:amplify_auth_cognito/method_channel_auth_cognito.dart';
 import 'package:amplify_core/amplify_core.dart';
 
+extension IsEqual on SignInResult {
+  bool isMostlyEqual(SignInResult comparator) {
```
jamesonwilliams, line 23: Which is/are the field(s) that are being ignored? Can you annotate this somewhere?

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
```diff
@@ -19,6 +19,23 @@ import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
...
+      comparator.isSignedIn == isSignedIn &&
+      comparator.nextStep.signInStep == nextStep.signInStep &&
+      comparator.nextStep.additionalInfo.length == nextStep.additionalInfo.length &&
+      comparator.nextStep.codeDeliveryDetails.destination == nextStep.codeDeliveryDetails.destination &&
+      comparator.nextStep.codeDeliveryDetails.attributeName == nextStep.codeDeliveryDetails.attributeName &&
+      comparator.nextStep.codeDeliveryDetails.deliveryMedium == nextStep.codeDeliveryDetails.deliveryMedium
+    ) {
+      return true;
+    } else {
+      return false;
```
jamesonwilliams: Can you just:
```
return expr1 &&
    expr2;
```
instead of using the `if`?

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
Amplifiyer: No I mean asserting the actual values rather than the type. e.g.
`assert("iAmLegit", methodCall.arguments["data"]["confirmationCode"]); `

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
haverchuck: No fields are being ignored - the additionalInfo attribute is just a map of key values, so it could take a bit of work to recursively check for equality within the potential maps it may contain.  For the purposes of this test, I think checking the length is adequate. I left a comment.

Amplifiyer (approved):

PR #175, opened 2020-11-05
feat(amplify_datastore): Android Config ModelSchema 

fjnoyp:

**Config for Datastore Android**

This PR adds config support for Datastore in Android working with latest "Flutter" branch on Amplify Android. 

The datastore plugin interface module contains the various Dart classes for representing ModelSchema and generating ModelSchema from Codegen Models.  These classes contain methods for serializing and deserializing with the native layer. 

The datastore plugin contains native kotlin code with dedicated Kotlin classes for representing ModelSchema classes received from Dart.  These classes then provide utility methods to provide the underlying native level types so that a proper native ModelSchema can be generated. 

Unit tests have been added for Dart and Kotlin ModelSchema serialization and deserialization as well as a minor test in Query to check that nested models are loaded correctly. 

I removed the unit tests for Delete for now as @haverchuck has recreated them in a separate branch. 

**Notes on switching amplify-android library**
Within the amplify-android project you wish to use
Run in amplify-anrdoid library ./gradlew build && ./gradlew publishToMavenLocal

**Steps before final merge**

- [ ] Update Pubspec.yaml to point to versions instead of local plugins

- [ ] Update gradle to use public amplify-android

- [ ] Remove mavenLocal in gradle


packages/amplify_datastore/android/src/main/java/com/amplifyframework/datastore/generated/model/AmplifyModelProvider.java
```diff
@@ -0,0 +1,53 @@
...
+   */
+  @Override
+   public Set<Class<? extends Model>> models() {
+    final Set<Class<? extends Model>> modifiableSet = new HashSet<>(
+          Arrays.<Class<? extends Model>>asList(Blog.class, Post.class, Comment.class, PostTest.class)
+        );
+    
+        return Immutable.of(modifiableSet);
+        
+  }
```
Amplifiyer, line 42: Why do we have this in our main repo?

packages/amplify_datastore/android/src/main/java/com/amplifyframework/datastore/generated/model/Blog.java
```diff
@@ -0,0 +1,175 @@
...
+import com.amplifyframework.core.model.query.predicate.QueryField;
+
+import static com.amplifyframework.core.model.query.predicate.QueryField.field;
+
+
+
+/** This is an auto generated class representing the Blog type in your schema. */
+@SuppressWarnings("all")
+@ModelConfig(pluralName = "Blogs")
+public final class Blog implements Model {
```
Amplifiyer, line 23: Same. Is this example code?

packages/amplify_datastore/android/src/main/java/com/amplifyframework/datastore/generated/model/AmplifyModelProvider.java
fjnoyp: Hey see the comments above - these were added for the sake of testing since we we didn't have a save method and I used the model builder methods directly instead to make nested models.  I have a note above to remove these before the actual merge.

fjnoyp: Jameson's example approach to writing nested serialized models is working as well but requires some extra steps / hard coding, so for the sake of easier testing I've kept them here for now.

jamesonwilliams: ^ I use the same "mixed" approach in the Android library tests. Either save Java models, query serialized data -- or save serialized data, query Java models. I don't actually have a test right now to _both_ save and query serialized data.

fjnoyp: Java models have been removed.

packages/amplify_datastore/android/src/main/java/com/amplifyframework/datastore/generated/model/Blog.java
fjnoyp: Java test models have been removed.

packages/amplify_core/pubspec.yaml
```diff
@@ -10,7 +10,8 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  amplify_core_plugin_interface: ^0.0.1-dev.2
+  amplify_core_plugin_interface:
+    path: ../amplify_core_plugin_interface
```
jamesonwilliams, line 14: Does it make sense to do this for all of the dependencies in this list?

packages/amplify_datastore/android/build.gradle
```diff
@@ -52,8 +54,8 @@ android {
 
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
-    implementation "com.amplifyframework:aws-datastore:1.1.3"
-    implementation "com.amplifyframework:aws-api-appsync:1.1.3"
+    implementation "com.amplifyframework:aws-datastore:main"
+    implementation "com.amplifyframework:aws-api-appsync:main"
```
jamesonwilliams, line 58: I wonder if there is any way that we can centralize all of these `main` references, around the codebase, into a single `amplify-flutter`-side variables somewhere.

packages/amplify_datastore_plugin_interface/test/amplify_modelschema_to_map_test.dart
```diff
@@ -0,0 +1,226 @@
...
+    Map<String, dynamic> map = commentSchema.toMap();
+
+    expect(map, {
+      'name': 'Comment',
+      'pluralName': 'Comments',
+      'authRules': [],
+      'fields': {
+        'id': {
+          'name': 'id',
+          'type': {'fieldType': 0, 'ofModelName': ''},
```
jamesonwilliams, line 77: Hm, the integers for the field types is weird. Is it not possible to use a label of some kind, instead?

packages/amplify_datastore_plugin_interface/lib/src/types/models/uuid.dart
```diff
@@ -0,0 +1,24 @@
...
+ * permissions and limitations under the License.
+ */
+
+import 'package:uuid/uuid.dart';
+
+class UUID {
+  static final _internal = Uuid();
+
+  static String getUUID() {
+    return _internal.v1();
```
jamesonwilliams, line 22: Android uses [V4 UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)). (See also, https://github.com/aws-amplify/amplify-android/issues/846.)

packages/amplify_datastore_plugin_interface/lib/src/types/models/model_association.dart
```diff
@@ -14,17 +14,19 @@
...
-  const ModelAssociation({
-    this.name,
-    this.targetName,
-    this.associatedName,
-    this.associatedType,
+  ModelAssociation({
+    this.associationType,
+    this.targetName = "",
+    this.associatedName = "",
+    this.associatedType = "",
```
jamesonwilliams, line 29: empty strings feel like the wrong default. If these are not specified, they should probably be nullable.

packages/amplify_datastore_plugin_interface/lib/src/types/models/auth_rule.dart
```diff
@@ -20,23 +20,22 @@ import 'package:collection/collection.dart';
...
-      this.identityClaim,
-      this.groupClaim,
-      this.groups,
-      this.groupsField,
-      this.operations});
+      this.ownerField = "",
+      this.identityClaim = "",
+      this.groupClaim = "",
+      this.groups = const [],
+      this.groupsField = "",
```
jamesonwilliams, line 37: Same here, I think null is probably a better default for the strings.

packages/amplify_datastore/lib/method_channel_datastore.dart
```diff
@@ -56,23 +55,25 @@ class AmplifyDataStoreMethodChannel extends AmplifyDataStore {
...
   @override
   Future<T> delete<T extends Model>(T model, {QueryPredicate when}) async {
-     try {
+    try {
       var modelJson = model.toJson();
       final Map<dynamic, dynamic> serializedResult =
-      await _channel.invokeMapMethod('delete', <String, dynamic>{
-        'modelName': model.instanceType.modelName(),
+          await _channel.invokeMapMethod('delete', <String, dynamic>{
+        'modelName': model.getInstanceType().modelName(),
```
jamesonwilliams, line 62: this indentation looks screwy

packages/amplify_datastore/example/lib/main.dart
```diff
@@ -152,6 +179,7 @@ class _MyAppState extends State<MyApp> {
     );
   }
 
+  /*
   ModelSchema createTestSchema() {
```
jamesonwilliams, line 183: Can you delete this stuff, now?

packages/amplify_datastore/example/lib/codegen/PostOld.txt
```diff
@@ -0,0 +1,150 @@
+import 'package:flutter/foundation.dart';
```
jamesonwilliams, line 1: Looks like this file should be deleted

packages/amplify_datastore/example/lib/codegen/Post.dart
```diff
@@ -0,0 +1,215 @@
...
+  rating: Int!
+  created: AWSDate!
+  blogID: ID!
+  blog: Blog @connection(fields: ["blogID"])
+  comments: [Comment] @connection(keyName: "byPost", fields: ["id"])
+}
+ */
+
+@immutable
+class Post extends Model {
```
jamesonwilliams, line 37: Seems like there are a lot of redundant models in this PR. Can you just use the `testData` ones across the various modules?

packages/amplify_datastore/android/src/test/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyModelSchemaTest.kt
```diff
@@ -0,0 +1,518 @@
...
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
```
jamesonwilliams, line 14: Looks like you're missing a `package` declaration.

packages/amplify_datastore/android/src/test/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyModelSchemaTest.kt
```diff
@@ -0,0 +1,518 @@
...
+                modelSchema.convertToNativeModelSchema(),
+                native.build()
+        )
+
+    }
+
+
+
+
+
```
jamesonwilliams, line 263: I would go through this PR and remove any extra lines. You should have at most one empty line around the code.  And `}` should almost never be preceded by an empty line.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterSerializedModel.kt
```diff
@@ -15,15 +15,18 @@
...
 data class FlutterSerializedModel(val serializedModel: SerializedModel) {
-    private val id: String = serializedModel.id
-    private val serializedData: Map<String, Any> = parseAnyMap(serializedModel.serializedData)
 
-    private val modelName: String = serializedModel.modelName
+    private val serializedData: Map<String, Any> = parseSerializedDataMap(serializedModel.serializedData)
+
+    // ignored fields
+    private val id: String = serializedModel.id
+    private val modelName: String = parseModelName(serializedModel.modelName) // ModelSchema -> SerializedModel should always have a name
```
jamesonwilliams, line 29: `serialiazedModel.modelName` can currently be null, in the case of a nested model (child).

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterModelField.kt
```diff
@@ -25,11 +24,11 @@ data class FlutterModelField(val map: Map<String, Any>) {
 
     // Type of the field is the data type of the instance variables
     // of the Model class.
-    private val type: String = map["type"] as String
+    private val type: FlutterModelFieldType = FlutterModelFieldType( map["type"] as Map<String, Any>)
 
     // The type of the field in the target. For example: type of the
     // field in the GraphQL target.
-    private val targetType: String? = map["targetType"] as String?
+    // private val targetType: String? = map["targetType"] as String?
```
jamesonwilliams, line 31:
```suggestion
```

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterAuthRule.kt
```diff
@@ -16,10 +16,74 @@
...
+
+        if(!identityClaim.isNullOrEmpty()){
+            builder.identityClaim(identityClaim)
+        }
+
+        if(!groupClaim.isNullOrEmpty()){
+            builder.groupClaim(groupClaim)
+        }
+
+        if(!groupsField.isNullOrEmpty()){
```
jamesonwilliams, line 78: In the various places where you check `isNullOrEmpty()`, I'm not sure you have to. If you pass a null value to the build methods, won't it just store null (which it was going to do, anyway)? 

You might simplify several files by removing all of the `if` checks (assuming that you indeed can do so.)

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterAuthRule.kt
```diff
@@ -16,10 +16,74 @@
...
 import com.amplifyframework.core.model.AuthRule
+import com.amplifyframework.core.model.AuthStrategy
+import com.amplifyframework.core.model.ModelField
+import com.amplifyframework.core.model.ModelOperation
 
 data class FlutterAuthRule(val map: Map<String, Any>) {
+    
+    private val authStrategy : AuthStrategy =
+            indexToAuthStrategy(map["authStrategy"] as Int)
+    private val ownerField : String? = map["ownerField"] as String?
```
jamesonwilliams, line 27: In all of these cases, I don't think you need to say `String?` twice. Can't you just do:
```kotlin
private val ownerField = map["ownerField"] as String?
```

Also, if any of these are required, can you lose the `?`?

packages/amplify_core/android/build.gradle
```diff
@@ -43,6 +45,6 @@ android {
 }
 
 dependencies {
-    implementation 'com.amplifyframework:core:1.1.3'
```
haverchuck, line 46: Please remove these changes to the build.gradle.

packages/amplify_core/pubspec.yaml
haverchuck: This should not be necessary now that melos is incorporated.

packages/amplify_core/android/build.gradle
```diff
@@ -4,6 +4,7 @@ version '1.0-SNAPSHOT'
 buildscript {
     ext.kotlin_version = '1.3.50'
     repositories {
+        mavenLocal()
```
haverchuck, line 7: Please remove the references to mavenLocal.

packages/amplify_datastore/android/build.gradle
```diff
@@ -4,6 +4,7 @@ version '1.0-SNAPSHOT'
 buildscript {
     ext.kotlin_version = '1.3.50'
     repositories {
+        mavenLocal()
```
haverchuck, line 7: Please remove reference to mavenLocal.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/FlutterModelProvider.kt
```diff
@@ -19,14 +19,15 @@ import com.amplifyframework.core.model.Model
...
+import java.util.Collections
 
 class FlutterModelProvider private constructor() : ModelProvider {
     private val modelSchemaMap: MutableMap<String, ModelSchema> =
             HashMap()
 
     // TODO change it to return an empty set once it's properly supported in the amplify-android library
     override fun models(): Set<Class<out Model?>>? {
-        return null
+        return Collections.emptySet()
```
haverchuck, line 30: We already [made this change](https://github.com/aws-amplify/amplify-flutter/blob/fb16d0597c36559a20a1d9801bf0fe3c2106ebc8/packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/FlutterModelProvider.kt#L29).

Maybe you need to rebase this branch?

Please remove the TODO comment, though.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterAuthRule.kt
```diff
@@ -16,10 +16,74 @@
...
+    private val identityClaim : String? = map["identityClaim"] as String?
+    private val groupClaim : String? = map["groupClaim"] as String?
+    private val groups : List<String> = map["groups"] as List<String>
```
haverchuck, line 30: You are sure that `groups` doesn't need the optional `?`

packages/amplify_datastore/android/src/test/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePluginTest.kt
```diff
@@ -56,22 +58,54 @@ class AmplifyDataStorePluginTest {
...
+
     private val amplifySuccessResults = mutableListOf<SerializedModel>(
             SerializedModel.builder()
                     .serializedData(
                             mapOf("id" to "4281dfba-96c8-4a38-9a8e-35c7e893ea47",
-                                  "title" to "Title 1",
-                                  "rating" to 4))
-                    .modelName("Post")
-                    .id("4281dfba-96c8-4a38-9a8e-35c7e893ea47")
+                                    "title" to "Title 1",
```
haverchuck, line 99: We've made some of these changes already -  you might need to rebase.

packages/amplify_datastore/example/lib/codegen/ModelProvider.dart
```diff
@@ -0,0 +1,21 @@
+import 'package:amplify_datastore_example/codegen/Blog.dart';
```
haverchuck, line 1: Need license information.

packages/amplify_datastore/example/lib/codegen/Comment.dart
```diff
@@ -0,0 +1,121 @@
+import 'package:flutter/foundation.dart';
```
haverchuck, line 1: Need license information.

packages/amplify_datastore/example/lib/codegen/Blog.dart
```diff
@@ -0,0 +1,120 @@
+import 'package:flutter/foundation.dart';
```
haverchuck, line 1: Need license information.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterModelFieldType.kt
```diff
@@ -0,0 +1,92 @@
+package com.amazonaws.amplify.amplify_datastore.types.model
```
haverchuck, line 1: Need license information.

packages/amplify_datastore_plugin_interface/lib/src/types/models/model_association.dart
fjnoyp: The strings are empty because null is not serializable on the method channel.  

When serializing data between Dart and Android, I decided to include all keys.  If a key is "null" I just send an empty string instead.  This is to simplify the code.  

What are your thoughts on that general approach @jamesonwilliams ?

packages/amplify_datastore/example/lib/codegen/PostAuthComplex.dart
```diff
@@ -0,0 +1,123 @@
+import 'package:flutter/foundation.dart';
```
haverchuck, line 1: Need license information.

packages/amplify_datastore/lib/method_channel_datastore.dart
haverchuck: I definitely think you need to rebase- this is using the older delete interface.

packages/amplify_datastore/lib/method_channel_datastore.dart
```diff
@@ -56,23 +55,25 @@ class AmplifyDataStoreMethodChannel extends AmplifyDataStore {
...
         'model': modelJson,
         'queryPredicate': when?.serializeAsMap(),
       });
 
-      return model.instanceType.fromJson(new Map<String, dynamic>.from(serializedResult["serializedData"]));
+      return model
```
haverchuck, line 67: I don't think there should be any changes to the 'return' of the delete method, since it should now be void.

packages/amplify_datastore/test/amplify_datastore_delete_test.dart
```diff
@@ -53,14 +58,18 @@ void main() {
...
-    var json = await getJsonFromFile('delete_api/response/1_deleted_result.json');
+    var json =
+        await getJsonFromFile('delete_api/response/1_deleted_result.json');
 
     dataStoreChannel.setMockMethodCallHandler((MethodCall methodCall) async {
       return json;
     });
 
-    Post instance = Post(title: json["serializedData"]["title"], rating: json["serializedData"]["rating"], created: DateTime.parse(json["serializedData"]["created"]), id: json["id"]);
+    Post instance = Post(
```
haverchuck, line 68: This is all old code.

haverchuck (changes requested):
Looks great so far - please see comments.

packages/amplify_datastore/android/build.gradle
fjnoyp: I don't think we will be keeping these reference changes for the final release - this is just temporary.

packages/amplify_core/pubspec.yaml
fjnoyp: This is a temporary change for testing and will be changed for the final release.

packages/amplify_datastore_plugin_interface/test/amplify_modelschema_to_map_test.dart
fjnoyp: It's not possible, we're just creating a simple map using inline declaration for dart.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterSerializedModel.kt
fjnoyp: Indeed, the null value causes issues when serializing between Kotlin to Dart ...  Because of this I ensure that all values are empty strings rather than null.

packages/amplify_datastore_plugin_interface/lib/src/types/models/uuid.dart
fjnoyp: updated to v4

packages/amplify_datastore_plugin_interface/lib/src/types/models/auth_rule.dart
fjnoyp: See comment above.  null will break platform serialization

packages/amplify_datastore/android/build.gradle
jamesonwilliams: The problem is independent of whatever this string happens to say. `1.1.3` should be updated to the latest version sooner than later. It'd be advantageous to do that change in one location. Also, the developer use case of writing `main` would be facilitated.

packages/amplify_datastore_plugin_interface/lib/src/types/models/auth_rule.dart
jamesonwilliams: Ah yes, makes sense.

packages/amplify_datastore_plugin_interface/lib/src/types/models/model_association.dart
Amplifiyer: What's the complications in null checks in de/serialization?

jamesonwilliams: I understand why you've done it this way, now, thanks. I think that works. The other approach you could consider would be to not send keys associated to null values. Is there any downside to that approach?

That is, if you have:
```kotlin
val map = mapOf("key1" to null, "key2" to "Hamburger", "key3" to 5)
```
You can either do:
```kotlin
dispatchToMethodChannel(
    map.filterValues { it != null }
)
```
or:
```kotlin
dispatchToMethodChannel(
    map.mapValues { if (it == null) "" else it }
)
```

This probably isn't our most pressing issue right now though. Feel free to ignore in favor of higher value work items. :-D

packages/amplify_datastore/android/build.gradle
fjnoyp: Right that makes sense.  Sounds good to consider unified version name variable for a future PR.

packages/amplify_datastore_plugin_interface/lib/src/types/models/model_association.dart
fjnoyp: Thanks for that suggestion @jamesonwilliams.  I don't think that Dart has such elegant syntax for map comprehensions like Kotlin, but I found a similar way in Dart here: https://stackoverflow.com/questions/21130889/how-can-i-filter-a-map-access-the-entries

I agree it's not our most pressing issue and I don't see a strong pull to either solution beyond possibly adding more burden on what needs to be passed back and forth on the method channel.  

The benefit I see is simpler code on the Dart side that doesn't need to worry about checking for null values being sent or stored.  Also, I remember speaking with an Android developer and they told me that passing null or empty values to the Kotlin builders wasn't the ideal thing to do.  Can you confirm if that's true @jamesonwilliams ? 

@Amplifiyer it's not a big complication to do those null checks, but they do require extra checks for every single field which can make our methods larger.

packages/amplify_datastore/lib/method_channel_datastore.dart
fjnoyp: rebase done

packages/amplify_datastore/example/lib/main.dart
fjnoyp: yes, I remove it - thanks!

packages/amplify_datastore/example/lib/codegen/PostOld.txt
fjnoyp: file deleted

packages/amplify_datastore/example/lib/codegen/Post.dart
fjnoyp: That's a bit tough and I don't know a way to do that without more research on this topic. 

The main issue is that the test models exist within the test folder which cannot be accessed from the separate example project.  In the opposite direction, the test is not able to access contents within the example folder.  This is likely because both are not packages so they aren't designed to be used elsewhere. 

I think I could create the codegen files within the package itself, but that would mean providing it within our public package which might be confusing to our end users.  

We could possibly create an entirely different package and include them as dependencies in our pubspec.yaml (as a dev dependency for tests and as a regular dependency for the example app)  

I see that as a possible solution.  However I don't see this as a major issue yet and I think we might want to put more time into finding a better solution before comitting more time to a specific solution.

packages/amplify_datastore/android/src/test/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyModelSchemaTest.kt
fjnoyp: Is that necessary as this is a test class?  Another kotlin test (AmplifyDataStorePluginTest.kt) also does not have a package declaration.

packages/amplify_datastore/android/src/test/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyModelSchemaTest.kt
fjnoyp: Sounds good thanks

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterAuthRule.kt
fjnoyp: How would you recommend we do that? 

The builder API does not accept nullable variables.

ex: 
        @NonNull
        public AuthRule.Builder groupClaim(@NonNull String groupClaim) {
            this.groupClaim = (String)Objects.requireNonNull(groupClaim);
            return this;
        }

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterAuthRule.kt
fjnoyp: Sounds good, I've updated all Kotlin files in types/model to not have these redundant type declarations

Yes that is true.  Those fields I found to be optional

packages/amplify_core/android/build.gradle
fjnoyp: Removing these will prevent us from testing on Jameson's flutter ampilfy-android branch.  Can we just keep this step for the final merge?

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/FlutterModelProvider.kt
fjnoyp: removed the comments.  rebase has happened

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterAuthRule.kt
fjnoyp: You're right it can be optional as well, added the ? to groups

packages/amplify_datastore/lib/method_channel_datastore.dart
fjnoyp: rebased and used your version instead

packages/amplify_datastore/test/amplify_datastore_delete_test.dart
fjnoyp: rebased and used your version instead

packages/amplify_core/pubspec.yaml
haverchuck: I think these can be removed now.  Melos solves this for local testing.

packages/amplify_core/android/build.gradle
haverchuck: Ok - As long as we remember to remove them.

haverchuck (approved):

packages/amplify_datastore_plugin_interface/lib/src/types/models/uuid.dart
```diff
@@ -0,0 +1,24 @@
...
+
+import 'package:uuid/uuid.dart';
+
+class UUID {
+  static final _internal = Uuid();
+
+  static String getUUID() {
+    return _internal.v4();
+  }
+}
```
Amplifiyer, line 24: Do we need this utility? Why not just use `Uuid().v4()` in the code?

packages/amplify_datastore_plugin_interface/lib/src/types/query/query_field.dart
```diff
@@ -24,8 +27,9 @@ part 'query_sort.dart';
 
 class QueryField {
   final String fieldName;
+  final ModelFieldType fieldType;
 
-  const QueryField(this.fieldName);
+  const QueryField({@required this.fieldName, this.fieldType});
```
Amplifiyer, line 32: Will this impact the queryPredicate generation DX?

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterAuthRule.kt
Amplifiyer: I feel in some cases it adds to the readability, e.g. when a variable is receiving value from a complex lambda expressions. To understand the type of the variable one would have to fully read and understand the complex transformations happening in the lambdas. I encountered that while reviewing this PR. That's just my perception though, happy to discuss.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterSerializedModel.kt
Amplifiyer: Somehow I'm not able to be convinced that this reason is strong enough to justify default empty strings. Will sync up offline over this.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterAuthRule.kt
jamesonwilliams: Haha, ah right. I remember making a fuss to ensure it was that way, too. Fair enough. :-D

packages/amplify_datastore/android/src/test/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyModelSchemaTest.kt
jamesonwilliams: Both Java and Kotlin appear to _permit_ source files to compiled without `package` declarations:

In [Kotlin](https://kotlinlang.org/spec/packages-and-imports.html#packages-and-imports):
> an absence of a package header in a file means it belongs to the special root package.

In [Java](https://docs.oracle.com/javase/tutorial/java/package/createpkgs.html):
> If you do not use a package statement, your type ends up in an unnamed package. Generally speaking, an unnamed package is only for small or temporary applications or when you are just beginning the development process. Otherwise, classes and interfaces belong in named packages.

However, the practice is so rare an unconventional that I had to look it up. I do not recommend omitting a `package` header.

packages/amplify_datastore_plugin_interface/lib/src/types/models/model_association.dart
jamesonwilliams: Ah, true @fjnoyp, sorry, I was in Kotlin-mode, but this is clearly Dart.

I guess Dart 2.0 has a `removeWhen` function:
```dart
someMap.removeWhere((key, value) => value == null);
```

packages/amplify_datastore_plugin_interface/lib/src/types/models/uuid.dart
jamesonwilliams: The Amplify data modeling suite does include an entity called an `ID`. To strongly model that type, it _could_ make sense to have a thing called an `ID`. (After all, we have things such as `Temporal.DateTime`, etc., that come from the modeling SDL.)

Today the Android code base just uses `UUID.randomUUID().toString()` in a variety of places. There's nothing really wrong with that, but it would a hair cleaner to encapsulate the UUID details into the Amplify modeling type (an "`ID`", whatever that is.)

packages/amplify_datastore/android/src/test/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyModelSchemaTest.kt
```diff
@@ -0,0 +1,339 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import com.amazonaws.amplify.amplify_datastore.*
```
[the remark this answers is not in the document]
jamesonwilliams: There are a number of style bloopers around the native code.

I think the best course of action at this phase would be to incrementally on-board individual style checks, in upcoming PRs. This way, you could address repository-wide style issues one PR at a time. For example, you might start by onboarding [a rule about concurrent empty lines](https://github.com/aws-amplify/amplify-android/blob/main/configuration/checkstyle-rules.xml#L54-L72).

packages/amplify_datastore/example/lib/codegen/Post.dart
jamesonwilliams: > We could possibly create an entirely different package and include them as dependencies in our pubspec.yaml (as a dev dependency for tests and as a regular dependency for the example app)

This is precisely what the Android code base does. We have a module called `testmodels`, that we don't publish, but that gets consumed by the unit and integration tests of several other modules.

I do recommend this approach.

packages/amplify_datastore/example/lib/main.dart
```diff
@@ -92,7 +116,8 @@ class _MyAppState extends State<MyApp> {
     });
 
     (await Amplify.DataStore.query(Post.classType,
-            where: Post.ID.ne("e25859fc-e254-4e8b-8cae-62ccacce4097")))
+            where: QueryField(fieldName: "post.id")
```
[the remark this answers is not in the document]
jamesonwilliams: @Amplifiyer this is an open issue on Android and iOS, as well. It is mentioned [in the public documentation, here](https://docs.amplify.aws/lib/datastore/data-access/q/platform/android#predicates):

> Note: when constructing predicates, static `QueryField` instances such as `Post.RATING` do not own any information about the model to which the field belongs. In order to avoid any ambiguity between field names which are used across multiple models, it is recommended to construct a custom instance of `QueryField` in the form of `QueryField.field("{model-name}.{field-name}")` (i.e. `field("post.rating")`).

Please see https://github.com/aws-amplify/amplify-android/issues/908 for some additional discussion and a possible resolution.

TLDR: it's an easy bug to fix, by updating your Dart codegen. The generated `QueryField`s in your model should say `ModelName_id` instead of just `id`. Android and iOS still need this fix.

packages/amplify_datastore_plugin_interface/lib/src/types/models/model_field_definition.dart
```diff
@@ -0,0 +1,144 @@
...
+    this.type,
+    this.isRequired,
+    this.isArray = false,
+    this.association,
+    this.authRules = const [],
+  });
+
+  static ModelFieldDefinition field(
+      {@required QueryField key,
+      bool isRequired = true,
```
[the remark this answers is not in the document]
jamesonwilliams: FWIW, [Android uses **false** as the default](https://github.com/aws-amplify/amplify-android/blob/main/core/src/main/java/com/amplifyframework/core/model/ModelField.java#L229-L230). (Might be worth checking the other defaults, too.)

packages/amplify_datastore/example/lib/main.dart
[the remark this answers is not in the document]
Amplifiyer: Thanks Jameson, I believe its all the more reasons for us to fix it right away since we are in the process of implementing the codegen.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterAuthRule.kt
fjnoyp: Sure let's discuss this.  The types in the Kotlin ModelSchema classes are not very complex and you can easily see their type by seeing our casting statement right after (as Int, as String, etc)

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterSerializedModel.kt
fjnoyp: After speaking with @Amplifiyer we have decided to stop sending blank values in the Dart-> Kotlin map

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterAuthRule.kt
```diff
@@ -16,10 +16,74 @@
...
+
+    fun indexToAuthStrategy(index: Int) : AuthStrategy {
+        return when(index){
+            0 -> AuthStrategy.OWNER
+            1 -> AuthStrategy.GROUPS
+            2 -> AuthStrategy.PRIVATE
+            3 -> AuthStrategy.PUBLIC
+            else -> throw Exception("indexToAuthStrategy - invalid index supplied: $index")
+        }
+    }
```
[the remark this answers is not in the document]
fjnoyp: done

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterModelFieldType.kt
```diff
@@ -0,0 +1,107 @@
...
+    fun getTargetType() : String {
+        if(!ofModelName.isNullOrEmpty()) return ofModelName;
+        else return getJavaClass().simpleName;
+    }
+
+    fun getModelName() : String? {
+        return ofModelName;
+    }
+
+    fun getModelTargetType(): String {
```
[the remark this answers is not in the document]
fjnoyp: Field was duplicated. Removed the extra one.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterModelFieldType.kt
```diff
@@ -0,0 +1,107 @@
...
+            ModelFieldTypeEnum.int -> "Integer"
+            ModelFieldTypeEnum.double -> "Double"
+            ModelFieldTypeEnum.date -> "AWSDate"
+            ModelFieldTypeEnum.dateTime -> "AWSDateTime"
+            ModelFieldTypeEnum.time -> "AWSTime"
+            ModelFieldTypeEnum.timestamp -> "AWSTimestamp"
+            ModelFieldTypeEnum.bool -> "Boolean"
+            ModelFieldTypeEnum.enumeration -> "String"
+            ModelFieldTypeEnum.model -> ofModelName!!
+            ModelFieldTypeEnum.collection -> ofModelName!!
```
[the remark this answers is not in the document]
fjnoyp: yes.  I confirmed with @jamesonwilliams that if it's a Model the target type has to be the model name.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterModelFieldType.kt
```diff
@@ -0,0 +1,107 @@
...
+
+    // Name of the field is the name of the instance variable
+    // of the Model class.
+    private val ofModelName = map["ofModelName"] as String?
+
+    fun getFieldType() : ModelFieldTypeEnum {
+        return fieldType;
+    }
+
+    // Android specific support
```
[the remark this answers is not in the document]
fjnoyp: yes I removed it thanks!

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterModelSchema.kt
```diff
@@ -18,23 +18,20 @@ package com.amazonaws.amplify.amplify_datastore.types.model
...
-            (map["associations"] as Map<String, Any>?)?.mapValues { entry ->
-                FlutterModelAssociation(entry.value as Map<String, Any>)
-            }
-    private val indexes: Map<String, FlutterModelIndex>? =
-            (map["indexes"] as Map<String, Any>?)?.mapValues { entry ->
-                FlutterModelIndex(entry.value as Map<String, Any>)
+    private val associations =
+            ( fields ).filterKeys { key -> fields[key]?.getModelAssociation() != null }
+                    .mapValues { entry ->
+                entry.value.getModelAssociation()!!
```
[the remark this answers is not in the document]
fjnoyp: Not too sure why we would throw an exception here.  Associations is an optional value so if it's not provided the value can be null and can be ignored.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterSerializedModel.kt
```diff
@@ -32,11 +35,21 @@ data class FlutterSerializedModel(val serializedModel: SerializedModel) {
...
+    private fun parseSerializedDataMap(serializedData: Map<String, Any>): Map<String, Any> {
+
+        if(serializedData == null) return mapOf<String,Any>()
+
         return serializedData.mapValues {
             when (val value: Any = it.value) {
                 is Temporal.DateTime -> value.format()
+                is Model? -> FlutterSerializedModel(value as SerializedModel).toMap()
                 // TODO add for other complex objects that can be returned or be part of the codegen model
+                // It seems we can ignore collection types for now as we aren't returning lists of Models in hasMany relationships
```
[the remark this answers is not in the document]
fjnoyp: @jamesonwilliams mentioned that only belongsTo/hasOne nested model relationships are returned for Android and that multiple models will not be returned.  Thus it is a known limitation.  

I will check with Daniel Rochetti for iOS side.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterModelFieldType.kt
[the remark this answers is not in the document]
Amplifiyer: I meant for the collection field type, line 88. Is it always of type model?

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterSerializedModel.kt
```diff
@@ -32,11 +35,21 @@ data class FlutterSerializedModel(val serializedModel: SerializedModel) {
...
 
-    private fun parseAnyMap(serializedData: Map<String, Any>): Map<String, Any> {
+    private fun parseModelName(modelName: String?) : String{
+        return if(modelName.isNullOrEmpty()) ""
+        else modelName!!
+    }
+
+    private fun parseSerializedDataMap(serializedData: Map<String, Any>): Map<String, Any> {
+
+        if(serializedData == null) return mapOf<String,Any>()
```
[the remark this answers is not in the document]
fjnoyp: yes you are correct.  I switched it to throw an exception now.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/FlutterModelProvider.kt
```diff
@@ -19,14 +19,14 @@ import com.amplifyframework.core.model.Model
...
     private val modelSchemaMap: MutableMap<String, ModelSchema> =
             HashMap()
 
-    // TODO change it to return an empty set once it's properly supported in the amplify-android library
     override fun models(): Set<Class<out Model?>>? {
-        return java.util.Collections.emptySet()
+        return Collections.emptySet()
     }
 
     override fun version(): String {
```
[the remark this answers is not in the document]
fjnoyp: Now sending model provider version from dart to Kotlin. 

Also changed the way that config works for the Datastore.  We send a ModelProviderInterface instead of the List<ModelSchema> 

As the config method signature has changed, @Amplifiyer you might want to look at those changes too (it's in a separate commit labeled "Send ModelProviderVersion".

packages/amplify_datastore/android/src/test/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyModelSchemaTest.kt
```diff
@@ -0,0 +1,339 @@
...
+import org.mockito.Mockito.verify
+import org.mockito.invocation.InvocationOnMock
+import org.robolectric.RobolectricTestRunner
+
+// test that map received then transformed to flutter equivalent
+// generates appropriate native equivalent
+
+
+
+@RunWith(RobolectricTestRunner::class)
```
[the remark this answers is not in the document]
fjnoyp: lines removed.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterModelFieldType.kt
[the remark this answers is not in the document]
fjnoyp: yes for the separate getJavaClass that will return List::class.java but the target type will still be the underlying model.

packages/amplify_datastore/android/src/test/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyModelSchemaTest.kt
```diff
@@ -0,0 +1,339 @@
...
+                                        "association" to mapOf(
+                                                "associationType" to 0,
+                                                "targetName" to "",
+                                                "associatedName" to "blog",
+                                                "associatedType" to "Blog"
+                                        ),
+                                        "authRules" to emptyList<Any>()
+                                )
+                        )
+                )
```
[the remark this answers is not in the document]
fjnoyp: Using the same file as dart unit tests would probably require us to make a shared json file.  This would require investigating serializing from json to the same map format in Kotlin.  

I've looked online for some tools but don't know which one will actually work as expected and I'm hesitant to spend too much time on this point right now. 

@jamesonwilliams do you have any suggestions for easily loading a json file as a Kotlin map?

packages/amplify_datastore/android/src/test/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyModelSchemaTest.kt
```diff
@@ -0,0 +1,339 @@
...
+        // Verify result
+        assertEquals(
+                modelSchema.convertToNativeModelSchema(),
+                blogSchema
+        )
+
+    }
+
+    @Test
+    fun test_CommentSchema() {
```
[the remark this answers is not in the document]
fjnoyp: updated names of unit tests to describe what's actually being tested

packages/amplify_datastore/android/src/test/kotlin/com/amazonaws/amplify/amplify_datastore/PostSchemaData.kt
```diff
@@ -0,0 +1,232 @@
...
+                                    .name("HasMany")
+                                    .associatedName("post")
+                                    .associatedType("Post")
+                                    .build()
+
+            ))
+        .build()
+
+val test: String? = null; 
+    val postAuthComplexSchema = ModelSchema.builder()
```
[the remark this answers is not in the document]
fjnoyp: removed val test

packages/amplify_datastore/android/src/test/kotlin/com/amazonaws/amplify/amplify_datastore/PostSchemaData.kt
```diff
@@ -0,0 +1,232 @@
...
+                    "post" to
+                            ModelAssociation.builder()
+                                    .name("BelongsTo")
+                                    .targetName("postID")
+                                    .build()
+            ))
+            .build()
+
+
+    val blogSchema = ModelSchema.builder()
```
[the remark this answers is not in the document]
fjnoyp: yes file renamed thanks

packages/amplify_datastore/example/lib/main.dart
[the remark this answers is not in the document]
fjnoyp: Will speak with codegen team about fixing this.  In the meantime see the changes I made to the codegen models to use `modelname.id` instead of just `id`

packages/amplify_datastore_plugin_interface/lib/src/types/models/model.dart
```diff
@@ -34,6 +47,34 @@ enum ModelOperation { CREATE, UPDATE, DELETE, READ }
 abstract class ModelType<T extends Model> {
   const ModelType();
 
+  T fromSerializedMap(Map<String, dynamic> serializedMap) {
+    Map<String, dynamic> jsonMap = {};
+    serializedMapToJson(jsonMap, serializedMap);
```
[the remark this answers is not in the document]
fjnoyp: sounds good, updated method signature

packages/amplify_datastore_plugin_interface/lib/src/types/models/model_field_definition.dart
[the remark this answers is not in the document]
fjnoyp: I used ios defaults which is for ModelFieldNullability = required.  In the codegen models isRequired is always set, so the default value should not be used.

packages/amplify_datastore_plugin_interface/lib/src/types/models/model_field_definition.dart
```diff
@@ -0,0 +1,144 @@
...
+    // Extra code needed due to lack of nullability support
+    String associatedName;
+    String associatedType;
+
+    if (associatedKey != null) {
+      associatedName = associatedKey.fieldName;
+
+      if (associatedKey.fieldType != null) {
+        associatedType = associatedKey.fieldType.ofModelName;
+      }
```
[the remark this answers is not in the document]
fjnoyp: somewhat though not entirely.  I updated it to this: 
```

associatedName = associatedKey?.fieldName; 
    associatedType = associatedKey?.fieldType != null ? associatedKey.fieldType.ofModelName : ofModelName;
```

packages/amplify_datastore_plugin_interface/lib/src/types/models/model_field_type.dart
```diff
@@ -0,0 +1,96 @@
...
+
+  const ModelFieldType(this.fieldType, {this.ofModelName = ""});
+
+  static ModelFieldType from(Type type) {
+    // ETC TODO
+  }
+
+  //isArray
+  //isMap (would have key value pair)
+
```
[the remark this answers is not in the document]
fjnoyp: removed extra comment, thanks

packages/amplify_datastore_plugin_interface/lib/src/types/models/uuid.dart
fjnoyp: I think that would expose the implementation details too much especially if we wanted to change the way in which we generate a UUID.  I prefer now to call directly third party code in this way.

packages/amplify_datastore_plugin_interface/lib/src/types/query/query_field.dart
fjnoyp: I added this for our own internal usage so it won't affect the DX

packages/amplify_core/pubspec.yaml
fjnoyp: removed

packages/amplify_datastore/android/src/test/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyModelSchemaTest.kt
fjnoyp: Sounds good, I've added package declaration

packages/amplify_datastore/android/src/test/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyModelSchemaTest.kt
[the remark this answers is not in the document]
Amplifiyer: Query tests use extensive use of reading from json files and converting to a kotlin map. Is it not sufficient?

packages/amplify_core/android/build.gradle
```diff
@@ -43,6 +43,6 @@ android {
 }
 
 dependencies {
-    implementation 'com.amplifyframework:core:1.1.3'
+    implementation 'com.amplifyframework:core:main'
```
Amplifiyer, line 46: We should remove this change and upgrade to 1.4.1 like other places. When we do that, we probably have to update all the other plugins to rely on the same version until we can abstract the version to one common place.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -80,22 +77,37 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
...
-        } else {
+
+        if( !request.containsKey("modelSchemas") || !request.containsKey("modelProviderVersion") || request["modelSchemas"] !is List<*>){
             prepareError(flutterResult,
-                         Exception(
-                                 FlutterDataStoreFailureMessage.AMPLIFY_QUERY_REQUEST_MALFORMED.toString()),
-                         FlutterDataStoreFailureMessage.ERROR_CASTING_INPUT_IN_PLATFORM_CODE.toString())
+                    Exception(
+                            FlutterDataStoreFailureMessage.AMPLIFY_QUERY_REQUEST_MALFORMED.toString()),
+                    FlutterDataStoreFailureMessage.ERROR_CASTING_INPUT_IN_PLATFORM_CODE.toString())
```
Amplifiyer, line 85: We should fix this and use the correct error code.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/FlutterModelProvider.kt
Amplifiyer: I still see this change.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterModelSchema.kt
```diff
@@ -18,32 +18,33 @@ package com.amazonaws.amplify.amplify_datastore.types.model
 import com.amplifyframework.core.model.ModelSchema
 
 data class FlutterModelSchema(val map: Map<String, Any>) {
-    val name: String = map["name"] as String
-    private val pluralName: String = map["pluralName"] as String
-    private val authRules: List<FlutterAuthRule>? =
+    val name : String = map["name"] as String
+    private val pluralName : String? = map["pluralName"] as String?
+    private val authRules : List<FlutterAuthRule>? =
```
Amplifiyer, line 23: Any reasons for adding a space before the colon in declaration? It is counter coding convention https://kotlinlang.org/docs/reference/coding-conventions.html#colon

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterModelSchema.kt
```diff
@@ -18,32 +18,33 @@ package com.amazonaws.amplify.amplify_datastore.types.model
...
 
     fun convertToNativeModelSchema(): ModelSchema {
         var builder: ModelSchema.Builder = ModelSchema.builder()
                 .name(name)
                 .pluralName(pluralName)
-                .fields(fields.mapValues { entry ->
-                    entry.value.convertToNativeModelField()
-                })
+
+        if(!fields.isNullOrEmpty()){
```
Amplifiyer, line 42: Can fields be ever empty though? By design we require an Id field always.

packages/amplify_datastore_plugin_interface/lib/src/types/models/model.dart
```diff
@@ -13,27 +13,88 @@
...
+      if (value is Map) {
+        Map<String, dynamic> subMap = {};
+        mapSoFar[key] = subMap;
+        serializedMapToJson(subMap, new Map<String, dynamic>.from(value));
+      } else {
+        mapSoFar[key] = value;
+      }
+    });
+  }
+   */
```
Amplifiyer, line 96: Safe to remove?

packages/amplify_datastore_plugin_interface/lib/src/types/models/model_association.dart
```diff
@@ -14,14 +14,18 @@
...
-  final String associatedType;
+  ModelAssociationEnum associationType;
+  String targetName; // opt (used in belongsTo)
+  String associatedName; // opt (used in hasMany/hasOne)
+  String associatedType; // opt (used in hasMany/hasOne)
 
-  const ModelAssociation({
-    this.name,
+  ModelAssociation({
+    this.associationType,
```
Amplifiyer, line 28: can we not have the const constructor or final fields?

Amplifiyer:
Overall looks really good,  just some minor comments

packages/amplify_core/android/build.gradle
fjnoyp: Sounds good updated all android/build.gradle plugins to use 1.4.1

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
fjnoyp: thanks my bad, using AMPLIFY_CONFIGURE_REQUEST_MALFORMED now

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterModelSchema.kt
fjnoyp: No good reason, thank you for pointing that out.  I've updated the method signatures and typing to follow that format.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterModelSchema.kt
fjnoyp: Right, I've made it non nullable now.  I had not considered the id field.

packages/amplify_datastore_plugin_interface/lib/src/types/models/model.dart
fjnoyp: yes

packages/amplify_datastore_plugin_interface/lib/src/types/models/model_association.dart
fjnoyp: thank you, you're correct we should have that - I had removed it previously when the code was different.  I went through other files in models folder and changed to const as well

fjnoyp:
@Amplifiyer I've responded to your latest comments - they are in the latest commit on this PR. 

One last thing to do before final merge is to fix the unit tests - they appear to be broken due to queryField switch.

packages/amplify_analytics_pinpoint/android/build.gradle
```diff
@@ -47,8 +47,8 @@ android {
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
 
-    implementation 'com.amplifyframework:aws-analytics-pinpoint:1.0.0'
-    implementation 'com.amplifyframework:aws-auth-cognito:1.0.0'
+    implementation 'com.amplifyframework:aws-analytics-pinpoint:1.4.1'
```
jamesonwilliams, line 50: Anything special about this version?

https://mvnrepository.com/artifact/com.amplifyframework/core

jamesonwilliams (approved):

packages/amplify_core_plugin_interface/lib/amplify_datastore_category.dart
```diff
@@ -32,7 +32,7 @@ class DataStoreCategory {
       // Extra step to configure datastore specifically.
       // Note: The native datastore plugins are not added
       // in the `onAttachedToEngine` but rather in the `configure()
-      await plugin.addModelSchemas(modelSchemas: plugin.modelSchemas);
+      await plugin.addModelSchemas(modelProvider: plugin.modelProvider);
```
Amplifiyer, line 35: I think this needs to be renamed as well.

PR #181, opened 2020-11-10
feat(amplify_datastore) delete API for iOS

haverchuck:
*Description of changes:*
Adds delete API method for iOS.


packages/amplify_datastore/ios/Classes/FlutterDataStoreErrorMessage.swift
```diff
@@ -20,4 +20,5 @@ enum FlutterDataStoreErrorMessage: String {
     case INVALID_STATE = "AMPLIFY_INVALID_STATE"
     case MALFORMED = "AMPLIFY_REQUEST_MALFORMED"
     case QUERY_FAILED = "AMPLIFY_DATASTORE_QUERY_FAILED"
+    case DELETE_FAILED = "AMPLIFY_DATASTORE_DELETE_FAILED"
```
royjit, line 23: enum cases are normally named in lowerCamelCase in swift. 

```suggestion
    case deleteFailed = "AMPLIFY_DATASTORE_DELETE_FAILED"
```

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
```diff
@@ -131,6 +133,57 @@ public class SwiftAmplifyDataStorePlugin: NSObject, FlutterPlugin {
...
     }
+
+    func onDelete(args: [String: Any], flutterResult: @escaping FlutterResult) {
+        var modelName: String
+        var id: String
+        var rawModel: Dictionary<String, Any>
+        var modelData: SerializedModel
+        
+        do {
+            modelName = args["modelName"] as! String
```
royjit, line 144: It is always best to avoid force casting, if for some reason `args` do not contains the key, this line will crash the app. So to be on the safe side you can guard this as line and return an error if the required information is not present.
```swift
guard let modelName = args["modelName"] as? String, 
           let rawModel = args["model"] as? Dictionary<String, Any>,
           let id = rawModel["id"] as? String else {
      
          flutterResult(error)
          return
}

let modelData = SerializedModel(id: id, map: getJsonValue(rawModel))

do {
...
} catch {

}
           
```

packages/amplify_datastore/ios/Classes/FlutterDataStoreErrorMessage.swift
royjit: If this enum is just used to namespace the error message I think you can do the following as mentioned in [here](https://google.github.io/swift/#apples-api-style-guidelines) 

> Declaring an enum without cases is the canonical way to define a “namespace” to group a set of related declarations, such as constants or helper functions. 

```swift
enum FlutterDataStoreErrorMessage {
    static let casting: String = "ERROR_CASTING_INPUT_IN_PLATFORM_CODE"
    static let sendingError: String = "AMPLIFY_REQUEST_FAILED_TO_SEND"
    ...
}
```

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
```diff
@@ -131,6 +133,57 @@ public class SwiftAmplifyDataStorePlugin: NSObject, FlutterPlugin {
...
+            FlutterDataStoreErrorHandler.prepareError(
+                flutterResult: flutterResult,
+                msg: FlutterDataStoreErrorMessage.MALFORMED.rawValue,
+                errorMap: ["UNKNOWN": "\(error.localizedDescription).\nAn unrecognized error has occurred. See logs for details." ])
+            return
+        }
+                        
+    }
+    
+    private func getJSONValue(_ jsonDict: [String: Any]) -> [String: JSONValue]{
```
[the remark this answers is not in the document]
haverchuck: Done.

packages/amplify_datastore/ios/Classes/DataStoreBridge.swift
```diff
@@ -36,4 +36,26 @@ public class DataStoreBridge {
                           paginate: paginationInput,
                           completion: completion)
     }
+    func onDelete(id: String,
+                   modelData: SerializedModel,
```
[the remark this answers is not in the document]
haverchuck: fixed.

packages/amplify_datastore/ios/Classes/FlutterDataStoreErrorMessage.swift
haverchuck: This is a good suggestion but would require a lot of changes across the library.

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
haverchuck: I ended up using individual guard statements.

royjit (approved):
LGTM

packages/amplify_datastore/ios/Classes/GetJsonValue.swift
```diff
@@ -0,0 +1,29 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import Amplify
+
+func getJSONValue(_ jsonDict: [String: Any]) -> [String: JSONValue]{
+    guard let jsonData = try? JSONSerialization.data(withJSONObject: jsonDict) else {
+        print("JSON error")
+        return [:]
```
jamesonwilliams, line 21: If I'm reading this right, it says:

> If there's a serialization failure, just ignore it, and return an empty map.

Potentially, this is just pushing errors down the road, where there will be more debugging steps needed, to determine their root cause.

Would it not be better to "fail fast" on bad JSON?

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
```diff
@@ -132,6 +134,58 @@ public class SwiftAmplifyDataStorePlugin: NSObject, FlutterPlugin {
...
+                    msg: FlutterDataStoreErrorMessage.MALFORMED.rawValue,
+                    errorMap: ["MALFORMED_REQUEST": "model did not contain an id." ])
+                return
+            }
+
+            let modelData = SerializedModel(id: id, map: getJSONValue(rawModel))
+            
+            try bridge.onDelete(id: id,
+                              modelData: modelData,
+                              modelSchema: flutterModelRegistration.modelSchemas[modelName]!) { (result) in
```
Amplifiyer, line 166: put a guard on the modelSchema to avoid a force cast

Amplifiyer (approved):
Looks good. Just left a minor comment on adding guard for modelSchema retrieval.

packages/amplify_datastore/ios/Classes/GetJsonValue.swift
Ashish-Nanda: Agree with this, I am throwing an error in the save PR around being unable to convert to JSON.

Ashish-Nanda (changes requested):
Same comment as Jameson. Otherwise looks good!

packages/amplify_datastore/ios/Classes/GetJsonValue.swift
haverchuck: Good point- I'll do this.

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
haverchuck: Done.

packages/amplify_datastore/ios/Classes/GetJsonValue.swift
haverchuck: Done.

Ashish-Nanda (approved):
LGTM 🚀

PR #195, opened 2020-11-17
Datastore ios config

fjnoyp:
This is a draft PR for the amplify ios config. 

Currently simple post models and nested models (comment -> post) are tested as working.  There are several things to finalize before this PR is ready for a final review pass: 

- [ ] Amplify ios-android discrepancy with nested model loading 

- [ ] Amplify ios-android discrepancy handling of ID queryfield.  Android uses model.id while ios can only use id. 

- [ ] Unit tests for iOS


packages/amplify_datastore/example/lib/main.dart
```diff
@@ -115,12 +115,14 @@ class _MyAppState extends State<MyApp> {
       postWithCreatedDate += encoder.convert(element.toJson()) + '\n';
     });
 
+    /*
     (await Amplify.DataStore.query(Post.classType,
-            where: QueryField(fieldName: "post.id")
+            where: QueryField(fieldName: "id")
```
Amplifiyer, line 120: This should be changed to Post.ID.ne... as we discussed.

packages/amplify_datastore/example/lib/codegen/Post.dart
```diff
@@ -105,7 +105,7 @@ class Post extends Model {
     buffer.write("id=" + id + ", ");
     buffer.write("title=" + title + ", ");
     buffer.write("rating=" + rating.toString() + ", ");
-    buffer.write("created=" + formatDateToISO8601(created));
+    buffer.write("created=" + created.toString());
```
Amplifiyer, line 108: Are we making this change in the modelgen as well? I preferred ISO8601 format even in toString() since that's what users have to pass in the query predicates `Post.CREATED.eq("2020-02-02T20:20:20-08:00")`

packages/amplify_datastore/example/lib/main.dart
```diff
@@ -158,13 +161,18 @@ class _MyAppState extends State<MyApp> {
     });
   }
 
+  sampleFunction() {
+    print('Function on Click Event Called.');
+    // Put your code here, which you want to execute on onPress event.
+  }
+
```
Amplifiyer, line 168: We should try to avoid all such local testing artifacts.

packages/amplify_datastore/example/lib/main.dart
```diff
@@ -158,13 +161,18 @@ class _MyAppState extends State<MyApp> {
...
   @override
   Widget build(BuildContext context) {
     return MaterialApp(
       home: Scaffold(
           appBar: AppBar(
-            title: const Text('Plugin example app'),
-          ),
+              title: RaisedButton(
+                  child: Text("click here"), onPressed: sampleFunction)),
           body: Center(
```
Amplifiyer, line 176: Same thing about local testing artifacts

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
```diff
@@ -225,23 +231,86 @@ public class SwiftAmplifyDataStorePlugin: NSObject, FlutterPlugin {
             print("Failed to get the datastore plugin \(error)")
             flutterResult(false)
         }
+         */
```
Amplifiyer, line 234: Definitely don't check this in with this PR

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
```diff
@@ -225,23 +231,86 @@ public class SwiftAmplifyDataStorePlugin: NSObject, FlutterPlugin {
...
-                      SerializedModel(map: try getJSONValue(["id": UUID().uuidString,
-                                                             "title": "Title 2",
-                                                             "rating": 3] as [String : Any])),
-                      SerializedModel(map: try getJSONValue(["id": UUID().uuidString,
-                                                             "title": "Title 3",
-                                                             "rating": 2] as [String : Any])),
-                      SerializedModel(map: try getJSONValue(["id": UUID().uuidString,
-                                                             "title": "Title 4"] as [String : Any]))]
         try models.forEach { model in
             try getPlugin().save(model, modelSchema: flutterModelRegistration.modelSchemas["Post"]!) { (result) in
```
Amplifiyer, line 315: Same

packages/amplify_datastore/ios/Classes/types/model/FlutterAuthRule.swift
```diff
@@ -0,0 +1,96 @@
...
+*  http://aws.amazon.com/apache2.0
+*
+* or in the "license" file accompanying this file. This file is distributed
+* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+*/
+
+
+import Foundation
```
Amplifiyer, line 17: extra new line

packages/amplify_datastore/ios/Classes/types/model/FlutterAuthRule.swift
```diff
@@ -0,0 +1,96 @@
...
+    private let groupClaim : String?
+    private let groups : [String]?
+    private let groupsField : String?
+    private var authStrategy : AuthStrategy!
+    private var operations : [ModelOperation]?
+    
+    init(serializedData: [String: Any]) {
+        
+        
+        self.ownerField = serializedData["ownerField"] as? String
```
Amplifiyer, line 33: extra new lines.

packages/amplify_datastore/ios/Classes/types/model/FlutterAuthRule.swift
```diff
@@ -0,0 +1,96 @@
...
+import Amplify
+
+public struct FlutterAuthRule {
+    
+    private let ownerField : String?
+    private let identityClaim : String?
+    private let groupClaim : String?
+    private let groups : [String]?
+    private let groupsField : String?
+    private var authStrategy : AuthStrategy!
```
Amplifiyer, line 27: Just for my understanding, what does `!` means in the var declaration?

packages/amplify_datastore/ios/Classes/types/model/FlutterAuthRule.swift
```diff
@@ -0,0 +1,96 @@
...
+        self.groups = serializedData["groups"] as? [String]
+        
+        self.groupsField = serializedData["groupsField"] as? String
+        
+        self.authStrategy = stringToAuthStrategy(string: serializedData["authStrategy"] as! String)
+
+        self.operations =
+            (serializedData["operations"] as! [String]).map {
+                stringToModelOperation(string: $0)
+        }
```
Amplifiyer, line 48: avoid forcecasts in favor of guards

packages/amplify_datastore/ios/Classes/types/model/FlutterAuthRule.swift
```diff
@@ -0,0 +1,96 @@
...
+        
+    }
+    
+    private func stringToAuthStrategy(string: String) -> AuthStrategy{
```
Amplifiyer, line 52: Space after the `AuthStrategy`

packages/amplify_datastore/ios/Classes/types/model/FlutterAuthRule.swift
```diff
@@ -0,0 +1,96 @@
...
+            case "OWNER":
+                return AuthStrategy.owner
+            case "GROUPS":
+                return AuthStrategy.groups
+            case "PRIVATE":
+                return AuthStrategy.private
+            case "PUBLIC":
+                return AuthStrategy.public
+            default:
+                preconditionFailure("Could not create a AuthStrategy from \(string)")
```
Amplifiyer, line 63: How does this exception get handled?

packages/amplify_datastore/ios/Classes/types/model/FlutterAuthRule.swift
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }
+    
+    private func stringToModelOperation(string: String) -> ModelOperation{
```
Amplifiyer, line 67: Same formatting.

packages/amplify_datastore/ios/Classes/types/model/FlutterAuthRule.swift
```diff
@@ -0,0 +1,96 @@
...
+            case "DELETE":
+                return ModelOperation.delete
+            case "READ":
+                return ModelOperation.read
+            default:
+            preconditionFailure("Could not create a ModelOperation from \(string)")
+        }
+    }
+    
+    public func convertToNativeModelField() -> AuthRule{
```
Amplifiyer, line 82: formatting and the name of the method is wrong.

packages/amplify_datastore/ios/Classes/types/model/FlutterAuthRule.swift
Amplifiyer, line 27: It's weird to mix & match Flutter models and external models this way. If we are creating a layer of model in between (Native and Method Channel), it should be 100% flutter model (preferred way) or not create the middle layer model at all (Not so preferred albeit faster way, see `QueryPredicateBuilder`)

packages/amplify_datastore/ios/Classes/types/model/FlutterAuthRule.swift
Amplifiyer, line 52: Can we rename the variable `string` to be more specific.

packages/amplify_datastore/ios/Classes/types/model/FlutterModelAssociation.swift
```diff
@@ -0,0 +1,46 @@
...
+    private let targetName : String?
+    private let associatedName: String?
+
+    init(serializedData: [String: Any]) {
+        self.associationType = serializedData["associationType"] as! String
+        self.targetName = serializedData["targetName"] as? String
+        self.associatedName = serializedData["associatedName"] as? String
+    }
+    
+    public func convertToNativeModelField() -> ModelAssociation{
```
Amplifiyer, line 33: incorrect method name

packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
```diff
@@ -20,65 +20,88 @@ import Amplify
...
     public let name: String
-    public let type: ModelFieldType
+    public let type: FlutterModelFieldType
     public let isRequired: Bool
     public let isArray: Bool
+    public let association: FlutterModelAssociation?
+    public let authRules: [FlutterAuthRule]?
+    
+    // Currently ignored
     public let attributes: [ModelFieldAttribute] = []
```
Amplifiyer, line 30: then let's delete them?

packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
```diff
@@ -20,65 +20,88 @@ import Amplify
...
         self.name = serializedData["name"] as! String
-        self.type = ModelFieldType.from(type: FlutterModelField.deserializeFieldType(type: serializedData["type"] as! String))
+        
+        self.type = FlutterModelFieldType(serializedData: serializedData["type"] as! [String: Any])
+        
         self.isRequired = serializedData["isRequired"] as! Bool
         self.isArray = serializedData["isArray"] as! Bool
+        
+        self.association = serializedData["association"] != nil ?  FlutterModelAssociation(serializedData: serializedData["association"] as! [String : Any]) :
+            nil
```
Amplifiyer, line 42: avoid force casts throughout.

packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
```diff
@@ -20,65 +20,88 @@ import Amplify
...
+        self.authRules = (serializedData["authRules"] as? [[String:String]])?.map{
+            FlutterAuthRule(serializedData: $0)
+        }
+            
+            /*
+            (serializedData["authRules"] as? [[String:String]]).map({ (<#[[String : String]]#>) -> [String:String] in
+            FlutterAuthRule(serializedData: code)
+        })
+ */
+            //parseAuthRulesMap(authRulesMap: serializedData["authRules"] as? [[String:String]])
```
Amplifiyer, line 53: cleanup?

packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
```diff
@@ -20,65 +20,88 @@ import Amplify
...
+    /*
+    private func parseAuthRulesMap(authRulesMap : [ [ String : String ]]?) -> [FlutterAuthRule]?{
+        if(authRulesMap != nil){
+            var result: [FlutterAuthRule] = [FlutterAuthRule]();
+            
+            for authRuleMap in authRulesMap! {
+                result.append( FlutterAuthRule(serializedData: authRuleMap));
+            }
+            
+            return result;
```
Amplifiyer, line 65: cleanup?

packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
```diff
@@ -20,65 +20,88 @@ import Amplify
...
-        }
-        if type == "DateTime" {
-            return Temporal.DateTime.self
-        }
-        if type == "Time" {
-            return Temporal.Time.self
+    }
+ */
+    
+    private func convertFlutterAuthRules(flutterAuthRules : [FlutterAuthRule]?) -> [AuthRule]?{
```
Amplifiyer, line 73: formatting

packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
```diff
@@ -20,65 +20,88 @@ import Amplify
...
+        if(flutterAuthRules != nil){
+            var result: [AuthRule] = [AuthRule]();
+            
+            for flutterAuthRule in flutterAuthRules! {
+                result.append( flutterAuthRule.convertToNativeModelField() );
+            }
+            
+            return result;
         }
```
Amplifiyer, line 82: use if-let

Amplifiyer (changes requested):
This PR needs some work around cleanup and formatting, I've left comments at some places but not all.

packages/amplify_datastore/example/ios/amplify_datastore_unit_tests/ModelSchemaEquatableExtensions.swift
```diff
@@ -0,0 +1,124 @@
+//
+//  ModelSchemaEquatableExtensions.swift
+//  amplify_datastore_unit_tests
+//
+//  Created by Admin on 11/17/20.
+//
+
```
haverchuck, line 7: We need license headers here.

packages/amplify_datastore/example/lib/main.dart
fjnoyp: ah yes.  changed the code

packages/amplify_datastore/example/lib/codegen/Post.dart
fjnoyp: The modelgen team is using the toString method instead of formatDateToISO8601 

We can switch to the different format you suggested instead then and communicate that change to the modelgen team. 

If we do so could we use this method instead though?  I think it's a bit easier to read: 

`created.toIso8601String()`

fjnoyp:
Praveen: Just for my understanding, what does ! means in the var declaration?

This declaration allows us to call a local method when initializing that variable.  Otherwise there is a compile error

packages/amplify_datastore/ios/Classes/types/model/FlutterAuthRule.swift
fjnoyp: code updated

packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
fjnoyp: what would you recommend as replacement?  It appears guard does not work in this scenario

packages/amplify_datastore/example/lib/codegen/Post.dart
Amplifiyer: If it produces the same output, yes!

Amplifiyer:
> This declaration allows us to call a local method when initializing that variable. Otherwise there is a compile error

But we aren't calling any method in that declaration statement.

packages/amplify_datastore/ios/Classes/types/model/FlutterAuthRule.swift
Amplifiyer:
```swift
guard let authStrategy = stringToAuthStrategy(string: serializedData["authStrategy"] as! String) else {
        preconditionFailure("Could not create a AuthStrategy from serializedData")
}
self.authStrategy = authStrategy
```

cc: @royjit if there is a better way for this.

packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
fjnoyp: yes :)

packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
fjnoyp: yes

packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
fjnoyp: updated

packages/amplify_datastore/ios/Classes/types/model/FlutterAuthRule.swift
fjnoyp: It causes a crash.  However, I added this code primarily to satisfy the compiler.  As these types are generated from enum types in Dart it should never run.

packages/amplify_datastore/ios/Classes/types/model/FlutterAuthRule.swift
fjnoyp: updated thank you

packages/amplify_datastore/ios/Classes/types/model/FlutterAuthRule.swift
fjnoyp: Alright I can just store those as string then and then convert them when the convert to native method is called.

fjnoyp: code change added and tested

packages/amplify_datastore/ios/Classes/types/model/FlutterAuthRule.swift
fjnoyp: renamed to authStrategyString

and below, modelOperationString

packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
fjnoyp: How do you imagine we'd use if-let in this context?  

flutterAuthRules is a input variable - are you suggesting we declare another variable?

packages/amplify_datastore/ios/Classes/types/model/FlutterAuthRule.swift
fjnoyp: hmm were you able to get this code to compile?  I used a slightly modified version but am getting compile errors : Initializer for conditional binding must have Optional type, not 'String'

What is the benefit of this approach? 

`        guard let authStrategy = serializedData["authStrategy"] as! String else { preconditionFailure("Could not create a AuthStrategy from serializedData")
        }
        self.authStrategy = authStrategy`

Amplifiyer: I think you need to change `as!` to `as?` to conditional unwrap/cast if possible else throw exception.
The benefit is graceful exception/error handling than runtime cast failed.

packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
Amplifiyer: yeah basically you can do _I think_

```swift
if let flutterAuthRules = flutterAuthRules {
   ... // doesn't require ! anymore on flutterAuthRules since it's already confirmed to be non nil
}

File: packages/amplify_datastore/example/ios/amplify_datastore_unit_tests/AmplifyModelSchemaUnitTests.swift [hunk 23]
```diff
@@ -0,0 +1,42 @@
+//
+//  AmplifyModelSchemaUnitTests.swift
+//  amplify_datastore_unit_tests
+//
+//  Created by Admin on 11/16/20.
+//
+
```
Amplifiyer, line 7: license information

packages/amplify_datastore/example/ios/amplify_datastore_unit_tests/AmplifyModelSchemaUnitTests.swift
```diff
@@ -0,0 +1,49 @@
...
+    
+    func test_schema_postAuthComplex_with_authRules() throws{
+        let flutterBlogSchema = try FlutterModelSchema( serializedData: schemasMap["PostAuthComplexSchema"] as! [String : Any] ).convertToNativeModelSchema()
+        XCTAssertEqual(SchemaData.PostAuthComplexSchema, flutterBlogSchema)
+    }
+    
+    
+    
+
+}
```
Amplifiyer, line 49: extra lines.

packages/amplify_datastore/example/ios/amplify_datastore_unit_tests/AmplifyModelSchemaUnitTests.swift
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import XCTest
+import Amplify
+import Combine
```
Amplifiyer, line 18: you probably don't need that.

packages/amplify_datastore/example/ios/amplify_datastore_unit_tests/ModelSchemaEquatableExtensions.swift
```diff
@@ -0,0 +1,78 @@
...
+            case (let .collection(collectionOfA), let .collection(collectionOfB)):
+                return collectionOfA == collectionOfB
+            default:
+                return false
+        }
+    }
+}
+
+
+
```
Amplifiyer, line 78: extra lines

packages/amplify_datastore/example/ios/amplify_datastore_unit_tests/resources/SchemaData.swift
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import XCTest
+import Amplify
+import Combine
```
Amplifiyer, line 18: This is not needed here.

packages/amplify_datastore/example/ios/amplify_datastore_unit_tests/resources/SchemaData.swift
```diff
@@ -0,0 +1,81 @@
...
+        authRules: [
+            AuthRule(
+                allow: .owner,
+                ownerField: "owner",
+                identityClaim: "cognito:username"
+                /*
+                operations: [
+                    .read, .delete, .update, .create
+                ]
+                */
```
Amplifiyer, line 71: we are not testing operations here?

packages/amplify_datastore/example/ios/amplify_datastore_unit_tests/ModelSchemaEquatableExtensions.swift
```diff
@@ -0,0 +1,78 @@
...
+    public static func == (lhs: ModelSchema, rhs: ModelSchema) -> Bool {
+        return lhs.name == rhs.name && lhs.pluralName == rhs.pluralName && lhs.authRules == rhs.authRules && lhs.fields == rhs.fields && lhs.attributes == rhs.attributes
+    }
+}
+
+extension AuthRule : Equatable {
+    public static func == (lhs: AuthRule, rhs: AuthRule) -> Bool {
+        return lhs.allow == rhs.allow && lhs.ownerField == rhs.ownerField && lhs.identityClaim == rhs.identityClaim && lhs.groupClaim == rhs.groupClaim && lhs.groups == rhs.groups && lhs.groupsField == rhs.groupsField && lhs.operations == rhs.operations
+    }
+}
```
Amplifiyer, line 29: can we wrap these lines? and other long ones?

packages/amplify_datastore/example/lib/main.dart
```diff
@@ -116,7 +115,7 @@ class _MyAppState extends State<MyApp> {
     });
 
     (await Amplify.DataStore.query(Post.classType,
-            where: QueryField(fieldName: "post.id")
+            where: QueryField(fieldName: Post.ID.fieldName)
```
Amplifiyer, line 118: you can just replace the whole thing with Post.ID which is a `QueryField` itself.

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
```diff
@@ -225,33 +239,11 @@ public class SwiftAmplifyDataStorePlugin: NSObject, FlutterPlugin {
...
-                    print("Save error = \(error)")
-                case .success(let post):
-                    print("Saved post - \(post)")
-                }
-            }
-        }
+    
+    private func getEscapeHatch() -> AWSDataStorePlugin {
+        let plugin = try! Amplify.DataStore.getPlugin(for: "awsDataStorePlugin") as! AWSDataStorePlugin
+        return plugin
```
Amplifiyer, line 246: This is probably not needed.

packages/amplify_datastore/ios/Classes/types/model/FlutterModelSchema.swift
```diff
@@ -21,22 +21,48 @@ struct FlutterModelSchema {
     
     let name: String
     let pluralName: String?
-    let authRules: AuthRules? = nil
-    let fields: FlutterModelFields
+    let authRules: [FlutterAuthRule]?
+    let fields: [String: FlutterModelField]
+    
+    // Not used for now
     let attributes: [ModelAttribute] = []
```
Amplifiyer, line 28: can we remove this then?

packages/amplify_datastore/ios/Classes/types/query/QueryPredicateBuilder.swift
```diff
@@ -19,14 +19,25 @@ import Amplify
...
     
+    // Remove model name if queryPredicate is on id 
+    private static func extractFieldName( field : String? ) -> String? {
+        
+        if(field != nil && field!.hasSuffix(".id")){
+            return "id"
+        }
+        return field;
+    }
+
```
Amplifiyer, line 30: I thought we decided to not add the prefix in dart and add it in android only.

packages/amplify_datastore/ios/Classes/types/model/FlutterModelAssociation.swift
```diff
@@ -0,0 +1,52 @@
...
+* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+*/
+
+
+import Foundation
+import Amplify
+
+
```
Amplifiyer, line 20: extra lines.

Amplifiyer (commented):
Looks good! Some minor formatting comments and questions.

packages/amplify_datastore/ios/Classes/types/model/FlutterModelFieldType.swift
```diff
@@ -0,0 +1,66 @@
...
+        else {
+            throw ModelSchemaError.parse(className: "FlutterModelFieldType", fieldName: "fieldType", desiredType: "String")
+        }
+        self.fieldType = fieldType
+        
+        self.ofModelName = serializedData["ofModelName"] as? String
+        
+    }
+    
+    public func convertToNativeModelField() -> ModelFieldType{
```
royjit, line 37:
```suggestion
    public func convertToNativeModelField() -> ModelFieldType {
```

packages/amplify_datastore/ios/Classes/types/model/FlutterModelFieldType.swift
```diff
@@ -0,0 +1,66 @@
...
+* permissions and limitations under the License.
+*/
+
+
+import Foundation
+import Amplify
+
+public struct FlutterModelFieldType {
+    public let fieldType : String
+    public let ofModelName : String?
```
royjit, line 22: Can we use enum here similar to `ModelFieldType`

packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
```diff
@@ -68,17 +104,39 @@ public struct FlutterModelField {
         //        }
         preconditionFailure("Could not create a ModelFieldType from \(String(describing: type)) MetaType")
     }
-    
-public func convertToNativeModelField() -> ModelField {
+
+    private func convertFlutterAuthRules(flutterAuthRules : [FlutterAuthRule]?) -> [AuthRule]?{
+
+        if(flutterAuthRules != nil){
```
royjit, line 110: `guard else` and return early.

packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
```diff
@@ -20,20 +20,56 @@ import Amplify
...
-        self.isArray = serializedData["isArray"] as! Bool
+    init(serializedData: [String: Any]) throws {
+
+        guard let name = serializedData["name"] as? String
+        else {
+            throw ModelSchemaError.parse(className: "FlutterModelField", fieldName: "name", desiredType: "String")
+        }
+        self.name = name
+
+        if let inputTypeMap = serializedData["type"] as? [String: Any]{
```
royjit, line 37: use guard here as well?

packages/amplify_datastore/ios/Classes/types/model/FlutterAuthRule.swift
```diff
@@ -0,0 +1,95 @@
...
+
+        self.identityClaim = serializedData["identityClaim"] as? String
+        
+        self.groupClaim = serializedData["groupClaim"] as? String
+        
+        self.groups = serializedData["groups"] as? [String]
+        
+        self.groupsField = serializedData["groupsField"] as? String
+
+        guard let authStrategy = serializedData["authStrategy"] as? String
```
royjit, line 41: Usually guard statements are kept at the top of the method. So to return early if fails.

packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
fjnoyp: what would be the benefit of this approach?  This change seems to make the code less clear.

packages/amplify_datastore/example/ios/amplify_datastore_unit_tests/AmplifyModelSchemaUnitTests.swift
fjnoyp: Removed import combine, I believe I need XCTest and Amplify though

packages/amplify_datastore/example/ios/amplify_datastore_unit_tests/resources/SchemaData.swift
fjnoyp: removed XCTest and Combine imports

packages/amplify_datastore/example/ios/amplify_datastore_unit_tests/AmplifyModelSchemaUnitTests.swift
Amplifiyer: git is not very good at rendering the lines I selected for a comment :( It just adds some more lines for context.

packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
Amplifiyer: This ensures that the variable is no longer _nil-able_ before you use it and cannot throw an unhandled exception.

packages/amplify_datastore/example/ios/amplify_datastore_unit_tests/resources/SchemaData.swift
fjnoyp: I updated our testing and equatable framework to allow for this to be properly compared.  Since the list order was arbitrary I needed to add extensions to allow for sorting that array and making the individual model operations comparable.

packages/amplify_datastore/example/ios/amplify_datastore_unit_tests/ModelSchemaEquatableExtensions.swift
fjnoyp: Sounds good, I went through and added line wraps

I looked through some of the model classes and added line wraps especially to the guard statements in init

packages/amplify_datastore/example/lib/main.dart
fjnoyp: Hmm this appears to be outdated, I've already done that update :            
 where: QueryField(fieldName: Post.ID.fieldName)

packages/amplify_datastore/ios/Classes/types/query/QueryPredicateBuilder.swift
fjnoyp: Hmm I thought we decided to have the prefix automatically added in the codegen models

image

packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
fjnoyp: ah ok I made the change

packages/amplify_datastore/ios/Classes/types/query/QueryPredicateBuilder.swift
```diff
@@ -19,14 +19,25 @@ import Amplify
 
 public class QueryPredicateBuilder {
     
+    // Remove model name if queryPredicate is on id 
+    private static func extractFieldName( field : String? ) -> String? {
+        
+        if(field != nil && field!.hasSuffix(".id")){
+            return "id"
+        }
+        return field;
```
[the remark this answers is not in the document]
fjnoyp: hi the code you provided does not compile.  It doesn't seem possible to let queryField and then use queryField in the same expression

image

packages/amplify_datastore/ios/Classes/types/model/FlutterModelSchema.swift
```diff
@@ -21,22 +21,48 @@ struct FlutterModelSchema {
...
+            self.authRules = nil
+        }
+
+        if let inputFieldsMap = serializedData["fields"] as? [String: [String: Any]]{
+            self.fields = try inputFieldsMap.mapValues {
+                try FlutterModelField.init(serializedData: $0)
+            }
+        }
+        else{
+            throw ModelSchemaError.parse(className: "FlutterModelSchema", fieldName: "fields", desiredType: "[String: [String: Any]]")
```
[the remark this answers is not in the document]
fjnoyp: hi could you rephrase your request here?  I don't quite understand what you've written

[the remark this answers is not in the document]
fjnoyp: I see only one guard let statement so I'm not too sure what you're referring to

packages/amplify_datastore/ios/Classes/types/model/FlutterModelFieldType.swift
fjnoyp: Hmm I don't see the benefit to that approach.  This class is meant to take a model and then generate the Amplify-iOS ModelFieldType enum afterwards.  I don't see the benefit of creating yet another enum that will just be converted to the final enum right after

packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
fjnoyp: This code has been updated to : `        if let flutterAuthRules = flutterAuthRules{
`

packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
fjnoyp: Do you know how to use guard if there's an inner statement?

`if let inputTypeMap = serializedData["type"] as? [String: Any]{
            **self.type = try FlutterModelFieldType(serializedData: inputTypeMap)** <-- this causes compile issues if using guard instead of if
        }`

packages/amplify_datastore/ios/Classes/types/query/QueryPredicateBuilder.swift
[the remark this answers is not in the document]
Amplifiyer: use a `,` instead of `&&` may work.

packages/amplify_datastore/ios/Classes/types/query/QueryPredicateBuilder.swift
Amplifiyer: hmm there seems to be some confusion around that. What I recalled, there were two options
1. Add the prefix in dart which will work with android and remove this prefix in iOS
2. Don't add the prefix in dart, only add it in android code, iOS left untouched.

I thought we were going with 2, I'm ok either way

packages/amplify_datastore/example/lib/main.dart
Amplifiyer: but not `where: Post.ID..ne("e25859fc-e254-4e8b-8cae-62ccacce4097")`

Amplifiyer (approved):
Looks good to me, responded to some of the questions. Not a blocker. Ship It

packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
royjit: It will be like this:

```swift
guard let inputTypeMap = serializedData["type"] as? [String: Any] else {
       throw ModelSchemaError.parse(
                 className: "FlutterModelField",
                 fieldName: "type",
                 desiredType: "[String: Any]")
       return
}
self.type = try FlutterModelFieldType(serializedData: inputTypeMap)
```

packages/amplify_datastore/ios/Classes/types/model/FlutterModelFieldType.swift
royjit: By using enum, you can avoid the optional `ofModelName:String?` and assign an associated value of `String` only for `model` and `collection`. This will avoid force casting or nil check later.

But if this class is just storing the json data, I think it is okay to keep like this.

packages/amplify_datastore/ios/Classes/types/model/FlutterModelSchema.swift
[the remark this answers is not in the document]
royjit: I meant create another guard below the first one.
```swift
guard let name = serializedData["name"] as? String else {
    throw ModelSchemaError.parse(className: "FlutterModelSchema", fieldName: "name", desiredType: "String")
 }
guard let inputFieldsMap = serializedData["fields"] as? [String: [String: Any]] else {
     throw ModelSchemaError.parse(className: "FlutterModelSchema", fieldName: "fields", desiredType: "[String: [String: Any]]")
}
self.name = name
self.pluralName = serializedData["pluralName"] as? String
```

packages/amplify_datastore/ios/Classes/types/model/FlutterModelFieldType.swift
```diff
@@ -0,0 +1,66 @@
...
+            case "time":
+                return ModelFieldType.time
+            case "timestamp":
+                return ModelFieldType.timestamp
+            case "bool":
+                return ModelFieldType.bool
+            case "enumeration":
+                return ModelFieldType.string
+            case "model":
+                return ModelFieldType.model(name: ofModelName!)
```
royjit, line 62: Force casting will crash the app if `ofModelName` is nil. Should we throw an error in this case?

packages/amplify_datastore/ios/Classes/types/query/QueryPredicateBuilder.swift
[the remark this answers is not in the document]
fjnoyp: nice that works updated the code

packages/amplify_datastore/ios/Classes/types/model/FlutterModelFieldType.swift
fjnoyp: cool I'll keep it as is for now then

packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
fjnoyp: nice thanks you.  I don't think we need the return (warning shows) so I didn't include that part.

packages/amplify_datastore/example/lib/main.dart
fjnoyp: this is code for that: 

    (await Amplify.DataStore.query(Post.classType,
            where: QueryField(fieldName: Post.ID.fieldName)
                .ne("e25859fc-e254-4e8b-8cae-62ccacce4097")))
        .forEach((element) {
      postWithIdNotEquals += encoder.convert(element.toJson()) + '\n';
    });


It seems we're already using Post.ID.fieldName right?

packages/amplify_datastore/ios/Classes/types/model/FlutterModelFieldType.swift
fjnoyp: Good call, I updated the code to do this then: 

            case "model":
                return ModelFieldType.model(name: ofModelName!)
                guard let ofModelName = ofModelName
                else {
                    throw ModelSchemaError.parse(
                        className: "FlutterModelFieldType",
                        fieldName: "ofModelName",
                        desiredType: "String")
                }
                return ModelFieldType.model(name: ofModelName)
            case "collection" :
                return ModelFieldType.collection(of: ofModelName!)
                guard let ofModelName = ofModelName
                else {
                    throw ModelSchemaError.parse(
                        className: "FlutterModelFieldType",
                        fieldName: "ofModelName",
                        desiredType: "String")
                }
                return ModelFieldType.collection(of: ofModelName)

royjit (approved):
LGTM

PR #196, opened 2020-11-17
Fix(amplify_storage_s3): Compile errors after upgrading amplify-android library

Amplifiyer:
With this change, the whole library builds in android and the tests are successful.

PS: `minSdkVersion` is increased because of increasing >64K methods limit. 


jamesonwilliams (approved):
Looks fine to me. Not sure what caused the change. Asked in Android slack channel.

Issue #178, opened 2020-11-07
[iOS] Amplify.Auth.fetchAuthSession() sessionToken is null


Rayv1:
Hi,

my code to get the session:

```
  Future<bool> getSessionFromAmplify() async {
    try {
      session = await Amplify.Auth.fetchAuthSession(
          options: CognitoSessionOptions(getAWSCredentials: true));

      return true;
    } on AuthError catch (e) {
      return false;
    }
  }

```
The returned CognitoAuthSession has a valid awsAccess & SecretKey but the sessionToken is just null. On Android its working fine. Is this just not implemented, is it a bug or is there an option for getting the sessionToken on iOS?

**Platform**
Amplify Flutter current supports iOS and Android.  This issue is reproducible in (check all that apply):
[] Android
[x] iOS

**Smartphone:**
 - Device: [Simulator iPhone 11 Pro / Hardware iPhone X ]
 - OS: [iOS 14]

Rayv1:
Any Updates on this? Is the behaviour reproducible?

haverchuck:
@Rayv1 Do you have an identity pool configured?

mdridley:
I am having the same issue. Possible that some of the PRs will fix... but even though they are merged, we are two months since the last push.

mdridley:
Latest update still has the issue?

haverchuck:
@mdridley  - What is the result you are seeing from the fetch session request (please redact sensitve information)?

haverchuck:
@Rayv1 Have you been able to test the new version?

Rayv1:
@haverchuck yep, its fixed! Thanks!

Issue #163, opened 2020-10-24
Getting AMPLIFY_FETCH_SESSION_FAILED exception when fetching AWS session credentials


willstepp:
**Describe the bug**

I've been working my way through the documentation at https://docs.amplify.aws/lib/auth/getting-started/q/platform/flutter to set up a Flutter application that will use the AWS Amplify Auth library to authenticate my calls to API Gateway endpoints. Since the Amplify Flutter library does not yet provide functionality for calling API Gateway, I intend to fetch the AWS Credentials manually for now. But when I call the function fetchAuthSession() using the example on this page: https://docs.amplify.aws/lib/auth/access_credentials/q/platform/flutter, it throws an exception: AMPLIFY_FETCH_SESSION_FAILED, and I cannot figure out why. Some additional information: I have confirmed I am signed in before calling this function, and it still happens.

**To Reproduce**

```
try {

CognitoAuthSession sess = await Amplify.Auth.fetchAuthSession(options: CognitoSessionOptions(getAWSCredentials: true));
if (sess.isSignedIn) {
  print("SIGNED IN");
} else {
  print("NOT SIGNED IN");
}

} on AuthError catch (e) {

  print(e);
  print(e.cause);

}
```

**Expected behavior**
I expect to receive a CognitoAuthSession instance that has the AWS Credentials for the current session.

**Platform**
Amplify Flutter current supports iOS and Android.  This issue is reproducible in (check all that apply):
[] Android
[X] iOS

**Smartphone (please complete the following information):**
 - Device: iPhone SE Simulator 
 - OS: macOS Big Sur
 - Browser: Stock Browser
 - Version iOS 14

Please help me if you can. I can provide any additional information. I'd love to use this library, but if I can't get the AWS credentials I'll have to try another route. Thanks!

haverchuck:
@willstepp Is this happening for you on Android as well?

willstepp:
@haverchuck I have since figured this out. There was additional information in the exception for AMPLIFY_FETCH_SESSION_FAILED, which said I had not yet configured an identity pool. Digging into that I learned the option I chose during setup "User Sign-Up & Sign-In only (Best used with a cloud API only)", which I had thought was the correct one for my use case (calling into API Gateway) does not go through the identity pool creation, but the other option does: "User Sign-Up, Sign-In, connected with AWS IAM controls (Enables per-user Storage features for images or other con
[...]

Issue #148, opened 2020-10-11
Cannot get Cognito User Pool Tokens


fpabl0:
**Describe the bug**
My project only uses Cognito User Pools, no IdentityPools, and I want to get cognito user pool tokens using:
```
await Amplify.Auth.fetchAuthSession(
        options: CognitoSessionOptions(getAWSCredentials: true),
);
```
But got error: 
`{localizedErrorMessage: Could not fetch identity Id, AWS Cognito Identity Pool is not configured, recoverySuggestion: Follow the steps to configure AWS Cognito Identity Pool and try again, platform: iOS}
`

**To Reproduce**
1. Run this function:
  ```
Future<void> signIn(String email, String password) async {
    try {
      await Amplify.Auth.signIn(username: email, password: password);
      final resp = await Amplify.Auth.fetchAuthSession(
        options: CognitoSessionOptions(getAWSCredentials: true),
      );
      if (resp.isSignedIn) {
        final sess = resp as CognitoAuthSession;
        print(sess.userPoolTokens.accessToken);
      }
    } on AuthError catch (e) {
      print(e.cause);
      for (final exception in e.exceptionList) {
        print(exception.exception);
        print(exception.detail);
      }
    }
  }
```
**Expected behavior**
Get the cognito user pool tokens without provide an Identity Pool (because I don't need it)

**Platform**
Amplify Flutter current supports iOS and Android.  This issue is reproducible in (check all that apply):
[] Android
[x] iOS (iOS Simulator)

horgag:
try casting with `CognitoAuthSession`. ..

`CognitoAuthSession  res = await Amplify.Auth.fetchAuthSession(options: CognitoSessionOptions(getAWSCredentials: true));`

fpabl0:
> **Describe the bug**
> My project only uses Cognito User Pools, no IdentityPools, and I want to get cognito user pool tokens using:
> 
> ```
> await Amplify.Auth.fetchAuthSession(
>         options: CognitoSessionOptions(getAWSCredentials: true),
> );
> ```
> 
> But got error:
> `{localizedErrorMessage: Could not fetch identity Id, AWS Cognito Identity Pool is not configured, recoverySuggestion: Follow the steps to configure AWS Cognito Identity Pool and try again, platform: iOS} `
> 
> **To Reproduce**
> Steps to reproduce the behavior:
> 
> 1. Run this function:
> 
> ```
> Future<void> signIn(String email, String password) async {
>   try {
[...]

PR #205, opened 2020-11-22
fix(amplify_auth_cognito) guest creds, userpool-only config, and access tokens

haverchuck:
*Issue #, if available:*
#148 
#163 
#178 

*Description of changes:*
This PR addresses several platform-specific discrepencies between Android and iOS which have resulted in bugs.

### DIFF
--- a/packages/amplify_analytics_pinpoint/android/build.gradle
+++ b/packages/amplify_analytics_pinpoint/android/build.gradle
@@ -47,7 +47,7 @@ android {
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     implementation 'com.amplifyframework:aws-analytics-pinpoint:1.+'
-    implementation 'com.amplifyframework:aws-auth-cognito:1.5.0'
+    implementation 'com.amplifyframework:aws-auth-cognito:1.+'
     testImplementation 'junit:junit:4.13'
     testImplementation 'org.mockito:mockito-core:3.1.0'
     testImplementation 'org.mockito:mockito-inline:3.1.0'
--- a/packages/amplify_analytics_pinpoint/android/build.gradle
+++ b/packages/amplify_analytics_pinpoint/android/build.gradle
@@ -47,7 +47,7 @@ android {
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     implementation 'com.amplifyframework:aws-analytics-pinpoint:1.+'
-    implementation 'com.amplifyframework:aws-auth-cognito:1.+'
+    implementation 'com.amplifyframework:aws-auth-cognito:1.5.0'
     testImplementation 'junit:junit:4.13'
     testImplementation 'org.mockito:mockito-core:3.1.0'
     testImplementation 'org.mockito:mockito-inline:3.1.0'
--- a/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
+++ b/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
@@ -358,20 +358,16 @@ public class AuthCognito : FlutterPlugin, ActivityAware, MethodCallHandler {
   private fun onFetchAuthSession (@NonNull flutterResult: Result, @NonNull request: HashMap<String, *>) {
     // TODO: Implement forceRefresh when/if supported by Amplify libs
     var req = FlutterFetchAuthSessionRequest(request)
-    // responsePending used to account for double-call of Consumers
-    var responsePending: Boolean = true
     try {
       Amplify.Auth.fetchAuthSession(
               { result ->
-                if (req.getAWSCredentials /*&& responsePending*/) {
+                if (req.getAWSCredentials) {
                   val cognitoAuthSession = result as AWSCognitoAuthSession
                   when (cognitoAuthSession.identityId.type) {
                     AuthSessionResult.Type.SUCCESS -> {
-                      responsePending = false;
                       prepareCognitoSessionResult(flutterResult, cognitoAuthSession)
                     }
                     AuthSessionResult.Type.FAILURE -> {
-                      responsePending = false;
                       // checking case where user pool data is available but id pool does not exist
                       if (
                             result.identityId.error is AuthException.InvalidAccountTypeException &&
@@ -385,7 +381,7 @@ public class AuthCognito : FlutterPlugin, ActivityAware, MethodCallHandler {
                       }
                     }
                   }
-                } else /*if (responsePending)*/ {
+                } else {
                   val session = result as AuthSession;
                   prepareSessionResult(flutterResult, session)
                 }
--- a/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
+++ b/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
@@ -523,12 +523,8 @@ public class AuthCognito : FlutterPlugin, ActivityAware, MethodCallHandler {
 
   fun prepareSessionResult(@NonNull flutterResult: Result, @NonNull result: AuthSession) {
     var session = FlutterFetchAuthSessionResult(result);
-    if (session.isSignedIn) {
       Handler (Looper.getMainLooper()).post {
         flutterResult.success(session.toValueMap());
       }
-    } else {
-      prepareError(flutterResult, AuthException.SignedOutException(), FlutterAuthFailureMessage.FETCH_SESSION.toString())
-    }
   }
 }
--- a/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
+++ b/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
@@ -372,7 +372,17 @@ public class AuthCognito : FlutterPlugin, ActivityAware, MethodCallHandler {
                     }
                     AuthSessionResult.Type.FAILURE -> {
                       responsePending = false;
-                      prepareCognitoSessionFailure(flutterResult, cognitoAuthSession)
+                      // checking case where user pool data is available but id pool does not exist
+                      if (
+                            result.identityId.error is AuthException.InvalidAccountTypeException &&
+                            result.awsCredentials.error is AuthException.InvalidAccountTypeException &&
+                            result.userPoolTokens.type.toString() == "SUCCESS" &&
+                            result.userSub.type.toString() == "SUCCESS"
+                      ) {
+                        prepareCognitoSessionResult(flutterResult, cognitoAuthSession)
+                      } else {
+                        prepareCognitoSessionFailure(flutterResult, cognitoAuthSession)
+                      }
                     }
                   }
                 } else /*if (responsePending)*/ {
--- a/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
+++ b/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
@@ -358,16 +358,24 @@ public class AuthCognito : FlutterPlugin, ActivityAware, MethodCallHandler {
   private fun onFetchAuthSession (@NonNull flutterResult: Result, @NonNull request: HashMap<String, *>) {
     // TODO: Implement forceRefresh when/if supported by Amplify libs
     var req = FlutterFetchAuthSessionRequest(request)
+    // responsePending used to account for double-call of Consumers
+    var responsePending: Boolean = true
     try {
       Amplify.Auth.fetchAuthSession(
               { result ->
-                if (req.getAWSCredentials) {
+                if (req.getAWSCredentials /*&& responsePending*/) {
                   val cognitoAuthSession = result as AWSCognitoAuthSession
                   when (cognitoAuthSession.identityId.type) {
-                    AuthSessionResult.Type.SUCCESS -> prepareCognitoSessionResult(flutterResult, cognitoAuthSession)
-                    AuthSessionResult.Type.FAILURE -> prepareCognitoSessionFailure(flutterResult, cognitoAuthSession)
+                    AuthSessionResult.Type.SUCCESS -> {
+                      responsePending = false;
+                      prepareCognitoSessionResult(flutterResult, cognitoAuthSession)
+                    }
+                    AuthSessionResult.Type.FAILURE -> {
+                      responsePending = false;
+                      prepareCognitoSessionFailure(flutterResult, cognitoAuthSession)
+                    }
                   }
-                } else {
+                } else /*if (responsePending)*/ {
                   val session = result as AuthSession;
                   prepareSessionResult(flutterResult, session)
                 }
--- a/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterFetchCognitoAuthSessionResult.kt
+++ b/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterFetchCognitoAuthSessionResult.kt
@@ -56,7 +56,7 @@ data class FlutterFetchCognitoAuthSessionResult(private val raw: AWSCognitoAuthS
     }
   }
 
-  //parse userpool tokens
+  //parse credentials
   fun serializeCredentials(res: AuthSessionResult<AWSCredentials>?): Map<String, Any>? {
     var map = res.serializeToMap();
     return if (map != null && map.containsKey("value")) {
--- a/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterFetchCognitoAuthSessionResult.kt
+++ b/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterFetchCognitoAuthSessionResult.kt
@@ -37,8 +37,23 @@ data class FlutterFetchCognitoAuthSessionResult(private val raw: AWSCognitoAuthS
       "identityId" to this.identityId,
       "userSub" to this.userSub,
       "credentials" to this.credentials.serializeToMap(),
-      "tokens" to this.tokens.serializeToMap()
+      "tokens" to serializeTokens(this.tokens)
     )
+  }
+
+  //parse userpool tokens
+  fun serializeTokens(res: AuthSessionResult<AWSCognitoUserPoolTokens>?): Map<String, Any>? {
+    var map = res.serializeToMap();
+    return if (map != null && map.containsKey("value")) {
+      var values = map["value"] as Map<String, Any>
+      if (!values.containsKey("error")) {
+        return values
+      } else {
+        return null
+      }
+    } else {
+      null
+    }
   }
 
   //convert a data class to a map
--- a/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterFetchCognitoAuthSessionResult.kt
+++ b/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterFetchCognitoAuthSessionResult.kt
@@ -26,7 +26,7 @@ var gson = Gson()
 
 data class FlutterFetchCognitoAuthSessionResult(private val raw: AWSCognitoAuthSession) {
   private val isSignedIn: Boolean = raw.isSignedIn
-  private val identityId: String? = raw.identityId.value as String
+  private val identityId: String? = raw.identityId.value as String?
   private val userSub: String? = raw.userSub.value as String?
   private val credentials: AuthSessionResult<AWSCredentials>? = raw.awsCredentials
   private val tokens: AuthSessionResult<AWSCognitoUserPoolTokens>? = raw.userPoolTokens
--- a/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterFetchCognitoAuthSessionResult.kt
+++ b/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterFetchCognitoAuthSessionResult.kt
@@ -27,7 +27,7 @@ var gson = Gson()
 data class FlutterFetchCognitoAuthSessionResult(private val raw: AWSCognitoAuthSession) {
   private val isSignedIn: Boolean = raw.isSignedIn
   private val identityId: String? = raw.identityId.value as String
-  private val userSub: String? = raw.userSub.value as String
+  private val userSub: String? = raw.userSub.value as String?
   private val credentials: AuthSessionResult<AWSCredentials>? = raw.awsCredentials
   private val tokens: AuthSessionResult<AWSCognitoUserPoolTokens>? = raw.userPoolTokens
 
--- a/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterFetchCognitoAuthSessionResult.kt
+++ b/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterFetchCognitoAuthSessionResult.kt
@@ -36,7 +36,7 @@ data class FlutterFetchCognitoAuthSessionResult(private val raw: AWSCognitoAuthS
       "isSignedIn" to this.isSignedIn,
       "identityId" to this.identityId,
       "userSub" to this.userSub,
-      "credentials" to this.credentials.serializeToMap(),
+      "credentials" to serializeCredentials(this.credentials),
       "tokens" to serializeTokens(this.tokens)
     )
   }
@@ -46,10 +46,25 @@ data class FlutterFetchCognitoAuthSessionResult(private val raw: AWSCognitoAuthS
     var map = res.serializeToMap();
     return if (map != null && map.containsKey("value")) {
       var values = map["value"] as Map<String, Any>
-      if (!values.containsKey("error")) {
-        return values
+      return if (!values.containsKey("error")) {
+        values
       } else {
-        return null
+        null
+      }
+    } else {
+      null
+    }
+  }
+
+  //parse userpool tokens
+  fun serializeCredentials(res: AuthSessionResult<AWSCredentials>?): Map<String, Any>? {
+    var map = res.serializeToMap();
+    return if (map != null && map.containsKey("value")) {
+      var values = map["value"] as Map<String, Any>
+      return if (!values.containsKey("error")) {
+        values
+      } else {
+        null
       }
     } else {
       null
--- a/packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthCognitoPluginTest.kt
+++ b/packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthCognitoPluginTest.kt
@@ -347,11 +347,8 @@ class AmplifyAuthCognitoPluginTest {
             "identityId" to "id",
             "userSub" to "sub",
             "credentials" to mapOf(
-                "value" to mapOf(
-                    "accessKey" to "access",
-                    "secretKey" to "secret"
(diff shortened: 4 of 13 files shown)


packages/amplify_auth_cognito/ios/Classes/FlutterFetchCognitoSessionResult.swift
```diff
@@ -30,70 +31,145 @@ struct FlutterFetchCognitoSessionResult {
...
+      try identityProvider.getIdentityId().get()
+    } catch {
+      if case .service(_, _, let errorType) = error as? AuthError {
+        if case .invalidAccountTypeException = errorType as? AWSCognitoAuthError {
+          isUserPoolOnly = true
+        }
+      }
+    }
+    return isUserPoolOnly
+  }
```
Amplifiyer, line 154: The logic is not very clear here from the code an I couldn't understand exactly what `isUserPoolOnly` meant. Documentation here will be helpful and unit tests.

packages/amplify_auth_cognito/lib/src/CognitoSession/AWSCognitoUserPoolTokens.dart
```diff
@@ -23,18 +23,20 @@ class AWSCognitoUserPoolTokens {
...
+      if (Platform.isAndroid) {
+        if (tokens.containsKey("value")) {
+          realTokens = tokens["value"];
+        } else {
+          throw(AmplifyDartExceptions.formatException(methodName: "fetchAuthSession", field: "credentials"));
+        }
       } else {
-        throw(AmplifyDartExceptions.formatException(methodName: "fetchAuthSession", field: "credentials"));
+        realTokens = tokens;
       }
```
Amplifiyer, line 36: Shouldn't we be doing this in the android plugin code? If we intent to keep any logic here we should unit test it.

packages/amplify_auth_cognito/lib/src/CognitoSession/CognitoAuthSession.dart
```diff
@@ -24,4 +24,24 @@ class CognitoAuthSession extends AuthSession {
...
+    if (sessionValues.containsKey("credentials")) {
+      this.credentials = AWSCredentials.init(creds: sessionValues["credentials"]);
+    }
+
+    if (sessionValues.containsKey("tokens")) {
+      var tokenMap = sessionValues["tokens"] as Map;
+      if (!tokenMap.containsKey("error")) {
+        this.userPoolTokens = AWSCognitoUserPoolTokens.init(tokens: sessionValues["tokens"]);
+      }
+    }
```
Amplifiyer, line 45: Can we unit test this?

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
```diff
@@ -364,8 +364,22 @@ public class AuthCognito : FlutterPlugin, ActivityAware, MethodCallHandler {
...
+                    AuthSessionResult.Type.SUCCESS -> {
+                      prepareCognitoSessionResult(flutterResult, cognitoAuthSession)
+                    }
+                    AuthSessionResult.Type.FAILURE -> {
+                      // checking case where user pool data is available but id pool does not exist
+                      if (
+                            result.identityId.error is AuthException.InvalidAccountTypeException &&
+                            result.awsCredentials.error is AuthException.InvalidAccountTypeException &&
+                            result.userPoolTokens.type.toString() == "SUCCESS" &&
+                            result.userSub.type.toString() == "SUCCESS"
```
Amplifiyer, line 376: Is it NPE safe to access all these properties in an event of `AuthSessionResult.Type.FAILURE`

packages/amplify_analytics_pinpoint/android/build.gradle
```diff
@@ -48,7 +48,7 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
 
     implementation 'com.amplifyframework:aws-analytics-pinpoint:1.0.0'
-    implementation 'com.amplifyframework:aws-auth-cognito:1.0.0'
+    implementation 'com.amplifyframework:aws-auth-cognito:1.5.0'
```
fjnoyp, line 51: Do we need to update all other dependencies to 1.5.0 as well?

packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthCognitoPluginTest.kt
```diff
@@ -315,6 +317,58 @@ class AmplifyAuthCognitoPluginTest {
...
+            null as Void?
+        }.`when`(mockAuth).fetchAuthSession(ArgumentMatchers.any<Consumer<AuthSession>>(), ArgumentMatchers.any<Consumer<AuthException>>())
+
+        val data: HashMap<*, *> = hashMapOf(
+                "getAWSCredentials" to true
+        )
+        val arguments: HashMap<String, Any> = hashMapOf("data" to data)
+        val call = MethodCall("fetchAuthSession", arguments)
+        val res = mapOf(
+            "isSignedIn" to true,
```
fjnoyp, line 346: should this be in a separate json file?

packages/amplify_auth_cognito/ios/Classes/FlutterFetchCognitoSessionResult.swift
```diff
@@ -30,70 +31,145 @@ struct FlutterFetchCognitoSessionResult {
...
+  func getUserSub(session: AuthSession) throws -> String? {
+    let guest = isGuest(session: session)
+    var sub: String? = nil
     if let identityProvider = session as? AuthCognitoIdentityProvider {
       do {
         sub = try identityProvider.getUserSub().get()
       } catch {
-         throw error as! AuthError
+        if (error is AuthError) {
+            if case .signedOut = error as! AuthError {
```
fjnoyp, line 67: Can we simplify to: 

if case .signedOut = error as! AuthError && guest {
return nil
}
else{
throw error as! AuthError
}

packages/amplify_auth_cognito/ios/Classes/FlutterFetchCognitoSessionResult.swift
```diff
@@ -30,70 +31,145 @@ struct FlutterFetchCognitoSessionResult {
...
+      do {
+        let creds =  try awsCredentialsProvider.getAWSCredentials().get()
         credentialMap["awsAccessKey"] = creds.accessKey
         credentialMap["awsSecretKey"] = creds.secretKey
+      } catch {
+        throw error as! AuthError
+      }
+      do {
+        let tempCreds = try awsCredentialsProvider.getAWSCredentials().get() as? AuthAWSTemporaryCredentials
+          if ((tempCreds?.sessionKey) != nil) {
```
fjnoyp, line 115: extra () here?

packages/amplify_auth_cognito/ios/Classes/FlutterFetchCognitoSessionResult.swift
```diff
@@ -30,70 +31,145 @@ struct FlutterFetchCognitoSessionResult {
...
+    var result: Dictionary<String, Any> = [:]
+    result["isSignedIn"] = self.isSignedIn
+    result["credentials"] = self.credentials
+    result["identityId"] = self.identityId
+    if (!self.userPoolTokens.isEmpty) {
+        result["tokens"] = self.userPoolTokens
+    }
+    if (self.userSub != "") {
+        result["userSub"] = self.userSub
     }
```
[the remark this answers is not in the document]
haverchuck: These should not be nil.  They will be 'failure' Result objects:
`cognitoTokensResult = (Result<AWSPluginsCore.AuthCognitoTokens, Amplify.AuthError>) failure`

packages/amplify_auth_cognito/ios/Classes/FlutterFetchCognitoSessionResult.swift
haverchuck: I'm adding comments and iOS unit tests.

packages/amplify_analytics_pinpoint/android/build.gradle
jamesonwilliams: Should these not be 1.6.3 as in other parts of the project (DataStore)? The Android team does not generally recommend mixing and matching versions in a consuming project.

packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthCognitoPluginTest.kt
```diff
@@ -315,6 +317,58 @@ class AmplifyAuthCognitoPluginTest {
         verify(mockResult, times(1)).success(ArgumentMatchers.any<LinkedTreeMap<String, Any>>())
     }
 
+    @Test
+    fun fetchSession_returnsSuccess() {
```
jamesonwilliams, line 321: Can you remove the type cues in your `val` assignments?

Java had:
```java
LongTypeName type = new LongTypeName();
```
Kotlin allows us to simplify this, dropping one of the type names, a semicolon, and the `new` keyword:
```kotlin
val type = LongTypeName()
```
The type cue can be useful to disambiguate when necessary. But in other cases, it makes the syntax about as bad (maybe a little worse) than Java had been. The below expression is pretty comparable to in Java, just with `new` swapped for `val`, a `;` swapped for `:`, and the remaining items permuted in their ordering:
```kotlin
val type: LongTypeName = LongTypeName()
```

packages/amplify_auth_cognito/lib/src/CognitoSession/AWSCredentials.dart
```diff
@@ -23,18 +23,18 @@ class AWSCredentials {
...
-    if (Platform.isAndroid) {
-      if (creds.containsKey("value")) {
-        realCreds = creds["value"];
+    if (creds != null) {
+      var realCreds = {};
+      if (Platform.isAndroid) {
+        if (creds.containsKey("value")) {
+          realCreds = creds["value"];
+        }
       } else {
```
jamesonwilliams, line 32: Does Dart provide a switch over the various `Platform` values? Either way, does it make sense to explicitly write `else if (Platform.isIOS)`, here?

packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthCognitoPluginTest.kt
```diff
@@ -315,6 +317,58 @@ class AmplifyAuthCognitoPluginTest {
...
+                    "refreshToken" to "refresh"
+                ),
+                "type" to "SUCCESS"
+            )
+        )
+        // Act
+        plugin.onMethodCall(call, mockResult)
+
+        // Assert
+        verify(mockResult, times(1)).success(res);
```
jamesonwilliams, line 369: `times(1)` is the default behavior, when left unspecified. You can equivalently write this as:

```suggestion
        verify(mockResult).success(res);
```

packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthCognitoPluginTest.kt
```diff
@@ -315,6 +317,58 @@ class AmplifyAuthCognitoPluginTest {
...
+            awsCredentials,
+            userSub,
+            tokens
+        )
+
+        // Arrange
+        doAnswer { invocation: InvocationOnMock ->
+            plugin.prepareCognitoSessionResult(mockResult, mockSession)
+            null as Void?
+        }.`when`(mockAuth).fetchAuthSession(ArgumentMatchers.any<Consumer<AuthSession>>(), ArgumentMatchers.any<Consumer<AuthException>>())
```
jamesonwilliams, line 338: In test code, `any` is a good candidate for static import, IMO.

packages/amplify_auth_cognito/lib/src/CognitoSession/AWSCognitoUserPoolTokens.dart
haverchuck: Fixed this; moved the logic to platform code (Android)

packages/amplify_analytics_pinpoint/android/build.gradle
haverchuck: Done.

packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthCognitoPluginTest.kt
haverchuck: Yes probably, but we can make that change later.

packages/amplify_auth_cognito/ios/Classes/FlutterFetchCognitoSessionResult.swift
haverchuck: I don't think this would be valid swift syntax... I don't think "error as! AuthError" actually works as a boolean.

packages/amplify_auth_cognito/ios/Classes/FlutterFetchCognitoSessionResult.swift
haverchuck: Look like it... thanks. Good catch.

packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthCognitoPluginTest.kt
haverchuck: @jamesonwilliams - I changed all the versions as indicated.  I also removed the typings on the val assignments throughout the whole file, except for one or two instances.

packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthCognitoPluginTest.kt
haverchuck: @jamesonwilliams I prefer to specifiy the value rather than rely on knowledge of default behavior, given that this project involves so may different platforms, testing frameworks, etc.

packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthCognitoPluginTest.kt
haverchuck: @jamesonwilliams I was made this change.

packages/amplify_auth_cognito/lib/src/CognitoSession/AWSCredentials.dart
haverchuck: @jamesonwilliams I moved this logic to the platform code.

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
haverchuck: @jamesonwilliams ^^

jamesonwilliams: I'm not sure; let me check the source. Do you get nullability guides in Android Studio? To be safe, you can use Kotlin's ["safe call operator"](https://kotlinlang.org/docs/reference/null-safety.html#safe-calls). In the worst case, the following should evaluate to `null == "SUCCESS"`, as opposed to throwing a `NullPointerException`:
```kotlin
result?.userSub?.type?.toString() == "SUCCESS"
```

jamesonwilliams: Yea, as far as I can tell, all of your current calls are safe. All of the children of `result` are non-null. The `type` field is non-null, so `toString()` is okay. `error` _might be null_, but that's okay, since it's at the end of your clauses.

packages/amplify_auth_cognito/lib/src/CognitoSession/CognitoAuthSession.dart
haverchuck: I moved this logic to the Android code... but wrote a couple of unit tests anyway.

jamesonwilliams (approved):

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterFetchCognitoAuthSessionResult.kt
```diff
@@ -37,10 +37,25 @@ data class FlutterFetchCognitoAuthSessionResult(private val raw: AWSCognitoAuthS
...
 
+  //parse userpool tokens
+  fun serializeTokens(res: AuthSessionResult<AWSCognitoUserPoolTokens>?): Map<String, Any>? {
+    var map = res.serializeToMap();
+    return if (map != null && map.containsKey("value")) {
+      var values = map["value"] as Map<String, Any>
+      if (!values.containsKey("error")) {
+        return values
+      } else {
+        return null
```
Amplifiyer, line 52: do we need a `return` keyword on these since you are returning the whole `if`

packages/amplify_auth_cognito/example/ios/amplify_auth_cognito_tests/amplify_auth_cognito_tests.swift
```diff
@@ -1165,4 +1164,91 @@ class amplify_auth_cognito_tests: XCTestCase {
...
+                    identityIdResult: .success("testid"),
+                    awsCredentialsResult:  Result<AuthAWSCredentials, AuthError>.success(creds),
+                    // guest access should result in cognito tokens failure (i.e. no UserPool User)
+                    cognitoTokensResult: Result<AuthCognitoTokens, AuthError>.failure(authError)
+                )
+                let sessionData = Result<AuthSession,AuthError>.success(authSession)
+                do {
+                  let fetchSessionData = try FlutterFetchCognitoSessionResult(res: sessionData)
+                  flutterResult(fetchSessionData)
+                } catch {
```
Amplifiyer, line 1186: indent is 2 here

packages/amplify_auth_cognito/ios/Classes/FlutterFetchCognitoSessionResult.swift
```diff
@@ -30,70 +31,147 @@ struct FlutterFetchCognitoSessionResult {
...
+      } catch {
+        throw error as! AuthError
+      }
+      do {
+        let tempCreds = try awsCredentialsProvider.getAWSCredentials().get() as? AuthAWSTemporaryCredentials
+          if (tempCreds?.sessionKey != nil) {
+            credentialMap["sessionToken"] = tempCreds?.sessionKey
+          }
+      } catch {
+        print("Unable to cast session token.")
```
Amplifiyer, line 119: Should we throw here?

packages/amplify_auth_cognito/ios/Classes/FlutterFetchCognitoSessionResult.swift
Amplifiyer: Is this a documented behavior in iOS that won't change in the future?

packages/amplify_auth_cognito/lib/src/CognitoSession/AWSCredentials.dart
Amplifiyer, line 32: I still see this check, is this necessary?

packages/amplify_auth_cognito/ios/Classes/FlutterFetchCognitoSessionResult.swift
```diff
@@ -30,70 +31,147 @@ struct FlutterFetchCognitoSessionResult {
...
+    }
+    do {
+      let credProvider = session as! AuthAWSCredentialsProvider
+      try credProvider.getAWSCredentials().get()
+      credentialsExist = true
+    } catch {
+      credentialsExist = false
+    }
+    return signedOut && credentialsExist
+  }
```
Amplifiyer, line 176: Same, just want to confirm that this is a documented behavior that we can rely on.

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterFetchCognitoAuthSessionResult.kt
haverchuck: Looks like I just needed to lift both returns out of their enclosing 'ifs'.

packages/amplify_auth_cognito/example/ios/amplify_auth_cognito_tests/amplify_auth_cognito_tests.swift
haverchuck: Fixed.

packages/amplify_auth_cognito/lib/src/CognitoSession/AWSCredentials.dart
haverchuck: This is a seperate attribute - but I've made the same fix.

packages/amplify_auth_cognito/ios/Classes/FlutterFetchCognitoSessionResult.swift
haverchuck: Done.

packages/amplify_auth_cognito/ios/Classes/FlutterFetchCognitoSessionResult.swift
royjit: `.invalidAccountTypeException` is also returned if you try to fetch auth session in signedOut state while unauth access is not enabled. Since this is a public error this should not change, but it is not documented. Here are the cases where we throw `invalidAccountTypeException`:

1. Fetching identity id while signedOut with unauth not enabled
1. Fetching AWS credentials while signedOut with unauth not enabled
1. Fetching usersub while signedIn to Cognito Identity Pool
1. Fetching cognito user pool token while signedIn to Cognito Identity Pool
1. Fetching identityId while signedIn to Cognito User pool but CIDP is not configured
1. Fetching AWS credentials while signedIn to Cognito User pool but CIDP is not configured

packages/amplify_auth_cognito/ios/Classes/FlutterFetchCognitoSessionResult.swift
royjit: To check if the current state is guest, you can try this:

```swift
    func isGuest(session: AuthSession) -> Bool {
       // User could be signed in through CIDP or CUP.
        guard !session.isSignedIn else {
            return false
        }
        
       // If signedOut and contains aws credentials, then the user is in guest mode
        if let credProvider = session as? AuthAWSCredentialsProvider {
            return ((try? credProvider.getAWSCredentials().get()) != nil)
        }
        return false
    }
```

packages/amplify_auth_cognito/ios/Classes/FlutterFetchCognitoSessionResult.swift
```diff
@@ -30,70 +31,147 @@ struct FlutterFetchCognitoSessionResult {
...
+    return isUserPoolOnly
+  }
+
+  // Checks if the user is a guest via Identity Pool unauthenticated access
+  func isGuest(session: AuthSession) -> Bool {
+    var signedOut: Bool;
+    var credentialsExist: Bool
+    do {
+      let idp = session as! AuthCognitoIdentityProvider
+      try idp.getUserSub().get()
```
royjit, line 163: User might have signed in through the federatedSignIn() method in escape hatch to CIDP. In which case usersub will return an error. But the user is signed In.

packages/amplify_auth_cognito/ios/Classes/FlutterFetchCognitoSessionResult.swift
```diff
@@ -15,12 +15,13 @@
...
 import Foundation
 import Amplify
+import AmplifyPlugins
 import AWSPluginsCore
 
 struct FlutterFetchCognitoSessionResult {
   var isSignedIn: Bool
-  var userSub: String
-  var identityId: String
+  var userSub: String?
```
royjit, line 23: Instead of returning a nil value can we give an error or exception like iOS stating why the value is not present?

packages/amplify_auth_cognito/ios/Classes/FlutterFetchCognitoSessionResult.swift
haverchuck: I have changed the function name to something more descriptive (but very verbose), and it now checks to see if the user does not have an identity but IS signed in, which would indicate that they've only signed in with a user pool:

```
  // Checks if the user is signed in to a user pool without a CIDP
  func signedInToUserPoolWithoutCIDP(session: AuthSession) -> Bool {
    var userpoolSignInOnly = false;
    do {
      let identityProvider = session as! AuthCognitoIdentityProvider
      try identityProvider.getIdentityId().get()
    } catch {
      if case .service(_, _, let errorType) = error as? AuthError {
        if case .invalidAccountTypeException = errorType as? AWSCognitoAuthError {
            if (session.isSignedIn) {
                userpoolSignInOnly = true
            }
        }
      }
    }
    return userpoolSignInOnly
  }
```

packages/amplify_auth_cognito/ios/Classes/FlutterFetchCognitoSessionResult.swift
haverchuck: This is challenging because the Flutter method channels are only capable of sending back a single response - either a success or an error.  To do what you're suggesting, we would need to send back some sort of generic data structure with data describing the error, and then cast this to an error on the Dart side.  This is something I think we should consider prior to GA, but can't be accomodated as part of this PR.

Amplifiyer (approved):

packages/amplify_auth_cognito/ios/Classes/FlutterFetchCognitoSessionResult.swift
```diff
@@ -30,70 +31,142 @@ struct FlutterFetchCognitoSessionResult {
...
+      } catch {
+        throw error as! AuthError
+      }
+      do {
+        let tempCreds = try awsCredentialsProvider.getAWSCredentials().get() as? AuthAWSTemporaryCredentials
+          if (tempCreds?.sessionKey != nil) {
+            credentialMap["sessionToken"] = tempCreds?.sessionKey
+          }
+      } catch {
+        throw error as! AuthError
```
royjit, line 119: why do we need to force cast here? Can we simply throw the error?

```swift


         let creds =  try awsCredentialsProvider.getAWSCredentials().get()
         credentialMap["awsAccessKey"] = creds.accessKey
         credentialMap["awsSecretKey"] = creds.secretKey

         let tempCreds = try awsCredentialsProvider.getAWSCredentials().get() as? AuthAWSTemporaryCredentials
           if (tempCreds?.sessionKey != nil) {
             credentialMap["sessionToken"] = tempCreds?.sessionKey
           }
   

```

packages/amplify_auth_cognito/ios/Classes/FlutterFetchCognitoSessionResult.swift
```diff
@@ -30,70 +31,142 @@ struct FlutterFetchCognitoSessionResult {
...
+            tokenMap["refreshToken"] = tokens.refreshToken
+        } catch {
+            tokenMap["error"] = "You are currently signed out."
+        }
     }
     return tokenMap;
   }
+
+  // Checks if the user is signed in to a user pool without a CIDP
+  func signedInToUserPoolWithoutCIDP(session: AuthSession) -> Bool {
```
royjit, line 142: Is this more readable? 
```swift

    func signedInToUserPoolWithoutCIDP(session: AuthSession) -> Bool {
        
        // If not signedIn return false
        guard session.isSignedIn else {
            return false
        }
        
        // If get identityID did not fail return false
        guard case .failure(let error) = (session as? AuthCognitoIdentityProvider)?.getIdentityId() else {
            return false
        }
        
        // If returned error is not of .service error, return false
        guard case AuthError.service(_, _, let errorType) = error else {
            return false
        }
        
        // If underlying error is not invalidAccountTypeException return false
        guard case .invalidAccountTypeException = (errorType as? AWSCognitoAuthError) else {
            return false
        }
        
        return true
    }

```

royjit (approved):
LGTM, added some minor comments.

Ashish-Nanda (changes requested):
The diff should be for the change associated with the commit only. Otherwise the commit history is not helpful. Do a `git pull` locally rather than rebase. Then squash and merge.

Ashish-Nanda (approved):
LGTM

Ashish-Nanda (approved):
🚀

PR #215, opened 2020-11-27
Upgrade android to use loose version numbers

Amplifiyer:
So that we get latest and greatest of android without any release. This will be reverted to use fixed/pinned versions in the future.


Ashish-Nanda (approved):
LGTM

jamesonwilliams (approved):

PR #216, opened 2020-11-28
Datastore latest codegen models

fjnoyp:
Fix for saving of Temporal and Enum types for iOS and Android.

The changes here will require a minor update to the way in which codegen models are created. Thus this PR should only be merged when those changes have been confirmed.

Codegen changes required:

Date and Time Fields should be DateTime instead of Date/Time
describeEnum needs to be enumToString instead as a null enum will cause crashes on toJson
make calls to "stringToDateTime" instead of the null check and parse call (messy code)
toString method should use nullable
Timestamp type should be int instead of string
Lastly, we are waiting for Android fixes to allow for Temporal types to be saved properly to test if other code changes are needed.


packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterModelFieldType.kt
```diff
@@ -29,7 +29,8 @@ data class FlutterModelFieldType(val map: Map<String, Any>) {
     private val ofModelName: String? = map["ofModelName"] as String?
 
     fun isEnum(): Boolean {
-        return fieldType == "enumeration"
+        return false;
+        //return fieldType == "enumeration"
```
Amplifiyer, line 33: probably need to revert this change?

packages/amplify_datastore/ios/Classes/types/model/FlutterSerializedModel.swift
```diff
@@ -155,6 +160,21 @@ struct FlutterSerializedModel: Model, JSONValueHolder {
 
                 result[key] = deserializedValue
             }
+            else if case .date = field?.type,
+                case .some(.string(let deserializedValue)) = values[key] {
+
+                result[key] = String(deserializedValue.dropLast())
```
Amplifiyer, line 166: what's the purpose of `dropLast` here? Are we removing the last character off the string value?

packages/amplify_datastore_plugin_interface/lib/amplify_datastore_plugin_interface.dart
```diff
@@ -38,8 +38,9 @@ export 'src/types/models/uuid.dart';
 export 'src/types/models/model_provider.dart';
 export 'src/types/models/auth_rule.dart';
 
-export 'src/types/utils/date_time_parser.dart';
-export 'src/types/utils/enum_parser.dart';
+export 'src/types/temporal/datetime_parse.dart';
+export 'src/types/utils/parsers.dart';
+export 'src/types/temporal/datetime_parse.dart';
```
Amplifiyer, line 43: `datetime_parse.dart` exported twice

Amplifiyer (commented):
> Date and Time Fields should be DateTime instead of Date/Time

I thought we decided against that.

packages/amplify_datastore_plugin_interface/lib/amplify_datastore_plugin_interface.dart
fjnoyp: thanks removed duplicate

packages/amplify_datastore/ios/Classes/types/model/FlutterSerializedModel.swift
fjnoyp: droplast is to remove the last Z character.  not removing it will cause a parse exception in Dart

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterModelFieldType.kt
fjnoyp: spoke with Jameson, isEnum should always be false.

fjnoyp:
Final codegen changes added.  This PR is ready for review.

packages/amplify_datastore/ios/Classes/types/model/FlutterSerializedModel.swift
Amplifiyer: Can you add documentation here about what is the Date value coming here in iOS vs what is expected in Dart (and comparing to what comes in android)? So we accurately document such discrepancies between platforms.

packages/amplify_datastore_plugin_interface/lib/src/types/utils/parsers.dart
```diff
@@ -1,19 +1,21 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import 'package:date_time_format/date_time_format.dart';
+
+String enumToString(Object o) => o.toString().split('.').last;
+
+T enumFromString<T>(String key, List<T> values) =>
+    values.firstWhere((v) => key == enumToString(v), orElse: () => null);
```
Amplifiyer, line 21: We should have unit tests for these.

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/datetime_parse.dart
```diff
@@ -1,22 +1,27 @@
...
+
+// ignore: public_member_api_docs
+extension DateTimeParse on DateTime {
+  String toDateTimeIso8601String() {
+    return DateTimeFormat.format(this, format: DateTimeFormats.iso8601);
+  }
+
+  static DateTime fromString(String input) {
+    return input == null ? null : DateTime.parse(input);
+  }
```
Amplifiyer, line 26: Unit tests for these. These have been changed enough times that I don't feel comfortable shipping without tests to nail down the input and output contract and expectations.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterSerializedModel.kt
```diff
@@ -48,7 +48,10 @@ data class FlutterSerializedModel(val serializedModel: SerializedModel) {
         return serializedData.mapValues {
             when (val value: Any = it.value) {
                 is Temporal.DateTime -> value.format()
+                is Temporal.Date -> value.format()
+                is Temporal.Time -> value.format()
                 is Model -> FlutterSerializedModel(value as SerializedModel).toMap()
+                is Temporal.Timestamp -> value.secondsSinceEpoch
                 // TODO add for other complex objects that can be returned or be part of the codegen model
                 // It seems we can ignore collection types for now as we aren't returning lists of Models in hasMany relationships
                 else -> value
```
Amplifiyer, line 57: Is this captured in the unit tests?

packages/amplify_datastore/ios/Classes/types/model/FlutterSerializedModel.swift
```diff
@@ -155,6 +160,21 @@ struct FlutterSerializedModel: Model, JSONValueHolder {
...
+            else if case .time = field?.type,
+                case .some(.string(let deserializedValue)) = values[key] {
+
+                result[key] = String(deserializedValue.dropLast())
+            }
+            else if case .timestamp = field?.type,
+                case .some(.number(let deserializedValue)) = values[key] {
+                
+                result[key] = NSNumber(value: Int(deserializedValue) )
+            }
```
Amplifiyer, line 177: Are there unit tests that cover these?

jamesonwilliams (approved):
`FlutterModelFieldType` changes look good.

packages/amplify_datastore/ios/Classes/types/model/FlutterSerializedModel.swift
fjnoyp: It only supports for certain formats.  The Date object is returned as yyyy-mm-dd but in order to have Z it would need to be in format yyyy-mm-ddThhZ.  Similar issues for Time.  I think we should drop the Z as Android does not return it in their format.

fjnoyp:
Unit tests updated for Dart and Native.

packages/amplify_datastore/android/src/test/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifySerializedModelTest.kt
```diff
@@ -0,0 +1,79 @@
...
+
+@RunWith(RobolectricTestRunner::class)
+class AmplifySerializedModelTest {
+
+    var serializedModelMaps : Map<String, Any> = (readMapFromFile("model_schema",
+            "serialized_model_maps.json",
+            HashMap::class.java) as HashMap<String, Any>)
+
+    @Test
+    fun test_schema_blog_with_hasMany() {
```
Amplifiyer, line 32: every test has a `hasMany` in it, what are the different things these tests are testing, we should add in their names.

Amplifiyer (approved):
Amazing work @fjnoyp. Ship it!

PR #229, opened 2020-12-03
chore(amplify_datastore): conditionally add API plugin if api configuration exists

Amplifiyer:
*Description of changes:*
This is intended as a workaround until we have a GraphQL API category in flutter which would be added by the flutter users. This whole thing will go away before GA.

### DIFF
--- a/packages/amplify_core/ios/Classes/SwiftCore.swift
+++ b/packages/amplify_core/ios/Classes/SwiftCore.swift
@@ -33,27 +33,26 @@ public class SwiftCore: NSObject, FlutterPlugin {
   public func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
     switch call.method {
       case "configure":
-        do {
-          let arguments = call.arguments as! Dictionary<String, AnyObject>
-          let version = arguments["version"] as! String
-          let configuration = arguments["configuration"] as! String
-          let amplifyConfiguration = try JSONDecoder().decode(AmplifyConfiguration.self, from: configuration.data(using: .utf8)!)
-          onConfigure(result: result, version: version, amplifyConfiguration: amplifyConfiguration)
-            
-        } catch {
-            print("Failed to configure Amplify \(error)")
-            result(FlutterError(code: "AmplifyException",
-                                message: "Failed to Configure Amplify",
-                                details: error.localizedDescription));
-        }
+        let arguments = call.arguments as! Dictionary<String, AnyObject>
+        let version = arguments["version"] as! String
+        let configuration = arguments["configuration"] as! String
+        onConfigure(result: result, version: version, configuration: configuration)
       default:
         result(FlutterMethodNotImplemented)
     }
   }
 
-    private func onConfigure(result: FlutterResult, version: String, amplifyConfiguration: AmplifyConfiguration) {
-      if (!isConfigured) {
+    private func onConfigure(result: FlutterResult, version: String, configuration: String) {
+      if(!isConfigured) {
         do {
+          if let data = configuration.data(using: .utf8) {
+            let configurationDictionary = try JSONSerialization.jsonObject(with: data) as? [String: Any]
+            if(configurationDictionary?["api"] != nil) {
+                // api configuration exists. Let's add the plugin
+                try Amplify.add(plugin: AWSAPIPlugin())
+            }
+          }
+          let amplifyConfiguration = try JSONDecoder().decode(AmplifyConfiguration.self, from: configuration.data(using: .utf8)!)
           AmplifyAWSServiceConfiguration.addUserAgentPlatform(.flutter, version: version)
           try Amplify.configure(amplifyConfiguration)
           isConfigured = true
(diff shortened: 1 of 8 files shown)


jamesonwilliams (commented):
When there is a Flutter GraphQL category, you'll just rely on that being added from Dart, right?

The current workaround is only since there isn't a GraphQL category in Dart, right?

packages/amplify_core/ios/Classes/SwiftCore.swift
```diff
@@ -33,27 +33,26 @@ public class SwiftCore: NSObject, FlutterPlugin {
...
-        } catch {
-            print("Failed to configure Amplify \(error)")
-            result(FlutterError(code: "AmplifyException",
-                                message: "Failed to Configure Amplify",
-                                details: error.localizedDescription));
-        }
+        let arguments = call.arguments as! Dictionary<String, AnyObject>
+        let version = arguments["version"] as! String
+        let configuration = arguments["configuration"] as! String
+        onConfigure(result: result, version: version, configuration: configuration)
```
haverchuck, line 39: +1

Amplifiyer:
> When there is a Flutter GraphQL category, you'll just rely on that being added from Dart, right?
> 
> The current workaround is only since there isn't a GraphQL category in Dart, right?

That is correct, I've added some clarification in the description.

haverchuck (approved):

packages/amplify_core/ios/Classes/SwiftCore.swift
```diff
@@ -33,27 +33,26 @@ public class SwiftCore: NSObject, FlutterPlugin {
...
+      if(!isConfigured) {
         do {
+          if let data = configuration.data(using: .utf8) {
+            let configurationDictionary = try JSONSerialization.jsonObject(with: data) as? [String: Any]
+            if(configurationDictionary?["api"] != nil) {
+                // api configuration exists. Let's add the plugin
+                try Amplify.add(plugin: AWSAPIPlugin())
+            }
+          }
+          let amplifyConfiguration = try JSONDecoder().decode(AmplifyConfiguration.self, from: configuration.data(using: .utf8)!)
```
Ashish-Nanda, line 55: You can reuse the variable `data` above.

packages/amplify_datastore/ios/amplify_datastore.podspec
```diff
@@ -17,7 +17,6 @@ The DataStore module for Amplify Flutter.
   s.dependency 'Flutter'
   s.dependency 'Amplify'
   s.dependency 'AmplifyPlugins/AWSDataStorePlugin'
-  s.dependency 'AmplifyPlugins/AWSAPIPlugin'
   s.platform = :ios, '13.0'
```
Ashish-Nanda, line 20: Not related to this change, but our platform versions are different in `core` and `datastore`. We should align them.

Ashish-Nanda (approved):
Minor comments, but looks good.

I would suggest adding tests for this case, but I think this will be removed before GA when we have the API plugin?

packages/amplify_datastore/ios/amplify_datastore.podspec
Amplifiyer: No, this is by design. Customers not using DataStore have no reason to have a min iOS 13 version.

fjnoyp (approved):
Looks good

packages/amplify_datastore/ios/amplify_datastore.podspec
Ashish-Nanda: Ah ok, I didn't realize it was intentional.

Amplifiyer:
> Minor comments, but looks good.
> 
> I would suggest adding tests for this case, but I think this will be removed before GA when we have the API plugin?

Yes, as mentioned in the description of this PR.

packages/amplify_core/ios/Classes/SwiftCore.swift
```diff
@@ -33,27 +33,26 @@ public class SwiftCore: NSObject, FlutterPlugin {
...
-          let amplifyConfiguration = try JSONDecoder().decode(AmplifyConfiguration.self, from: configuration.data(using: .utf8)!)
-          onConfigure(result: result, version: version, amplifyConfiguration: amplifyConfiguration)
-            
-        } catch {
-            print("Failed to configure Amplify \(error)")
-            result(FlutterError(code: "AmplifyException",
-                                message: "Failed to Configure Amplify",
-                                details: error.localizedDescription));
-        }
+        let arguments = call.arguments as! Dictionary<String, AnyObject>
```
royjit, line 36: nit: These values never fail? ie call.arguments always contains Dictionary? I am concerned with the force casting. 
If this is intended, I guess adding some comments would make it clear.

royjit (approved):
LGTM: Added minor comment

packages/amplify_core/ios/Classes/SwiftCore.swift
Amplifiyer: yeah we need to put guards here. This is an earlier category that we didn't add much error handling. We will be addressing this in future PRs.

PR #247, opened 2020-12-10
Enable code coverage for all platforms

Amplifiyer:


.codecov.yml
```diff
@@ -0,0 +1,21 @@
+codecov:
+  notify:
+    after_n_builds: 1
+
+coverage:
+  status:
+    project:
+      default:
+        target: 70%
+        if_not_found: success
```
haverchuck, line 10: What exactly is 'if_not_found' doing?

melos.yaml
```diff
@@ -26,6 +26,12 @@ scripts:
   test:unit:ios: >
     ./.circleci/test_all_plugins.sh ios-test && exit 0
 
+  coverage:ios: >
+    ./build-support/codecov.sh -F ios-unit-tests
+
+  coverage:flutter: >
+    ./build-support/codecov.sh -F flutter-unit-tests
+
```
haverchuck, line 34: Why not android?

haverchuck (commented):
See 2 inline comments.

melos.yaml
Amplifiyer: This PR is still in progress, was supposed to be only for iOS, but then I got greedy and started adding other 🤷

.codecov.yml
Amplifiyer: This is taken from the [JS repo](https://github.com/aws-amplify/amplify-js/blob/4b5841984727794217295adb37f2f9f7fddb509e/codecov.yml#L12) 

https://docs.codecov.io/docs/commit-status#if_not_found

> success
: the status will pass if there is no report for the head. Use this on commits / PRs where you won't be uploading coverage but still want codecov status checks to pass. 

I guess it makes sense to start with this?

jamesonwilliams (approved):
Looks good!

Did you intend to pin the Android library in this change?

Most trivial of nitpicks, but `android` would be better written as `Android` throughout the comments.

Amplifiyer:
> Did you intend to pin the Android library in this change?

Yeah, we were using some not so public APIs in our unit tests which [got changed](https://github.com/aws-amplify/amplify-android/compare/release_v1.6.6...release_v1.6.7#diff-5404f4b479adeba20bc093eed11a3116b45789dd8b7bf12e0dff095f28df5873L72-R75) in android library and broke the unit tests. Since this change will be released close to GA, decided to pin the versions.

> Most trivial of nitpicks, but `android` would be better written as `Android` throughout the comments.

Changed android to Android in the comment.

melos.yaml
```diff
@@ -26,6 +26,15 @@ scripts:
...
 
+  upload:coverage:ios: >
+    ./build-support/codecov.sh -F ios-unit-tests
+
+  upload:coverage:flutter: >
+    ./build-support/codecov.sh -F flutter-unit-tests
+
+  upload:coverage:android: >
+    ./build-support/codecov.sh -F android-unit-tests
+
```
haverchuck, line 37: +1

haverchuck (approved):

PR #257, opened 2020-12-16
chore: categories use selectors to invoke plugins

jamesonwilliams:
Updates the category definitions to use a `_select` function. The `_select`
function will obtain the first available plugin, and make it available
to a closure. The category may then invoke a desired behavior on the
plugin within that closure.

Adds coverage over all functions that use `_select`. This introduces
three new unit test files for the Storage, DataStore, and Auth categories.

In order to get passing, tests, several other changes are made:
1. Refer to local packages by their paths;
2. Add equals, toString, and hashCode functions for some result types.

### DIFF
--- a/packages/amplify_core_plugin_interface/lib/amplify_auth_category.dart
+++ b/packages/amplify_core_plugin_interface/lib/amplify_auth_category.dart
@@ -5,7 +5,6 @@ part of amplify_core_plugin_interface;
 /// be registered and configured and then subsequent API calls will be forwarded
 /// to those plugins.
 class AuthCategory {
-  final errorMsg = "Auth plugin not added correctly";
   const AuthCategory();
   static List<AuthPluginInterface> plugins = [];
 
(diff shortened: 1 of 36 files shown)


packages/amplify_core_plugin_interface/lib/amplify_datastore_category.dart
```diff
@@ -34,42 +33,48 @@ class DataStoreCategory {
...
   }
 
   Future<void> save<T extends Model>(T model) {
-    return plugins.length == 1 ? plugins[0].save(model) : throw (errorMsg);
+    return _select((it) => it.save(model));
   }
 
   Stream<SubscriptionEvent<T>> observe<T extends Model>(
       ModelType<T> modelType) {
+    // TODO: Howto use _stream()? It will return a Future<Stream<...>>,
```
jamesonwilliams, line 58:
```suggestion
    // TODO: Howto use _select()? It will return a Future<Stream<...>>,
```

Amplifiyer: You can await on it? `return await _select((it) => it.observe());`

packages/amplify_core_plugin_interface/lib/amplify_auth_category.dart
```diff
@@ -14,64 +13,69 @@ class AuthCategory {
...
+    return _select((it) => it.getCurrentUser(request: request));
   }
 
   Future<AuthSession> fetchAuthSession({AuthSessionOptions options}) {
     var request = AuthSessionRequest(options: options);
-    return plugins.length == 1 ? plugins[0].fetchAuthSession(request: request) : throw(errorMsg);
+    return _select((it) => it.fetchAuthSession(request: request));
+  }
+
+  Future<R> _select<R>(FutureOr<R> onValue(AuthPluginInterface plugin)) {
```
Amplifiyer, line 76: Is the argument `AuthPluginInterface plugin` in the callback function only for type inference?

Amplifiyer:
Thanks @jamesonwilliams for cleaning this up. Can you accompany this with some simple tests on  `_select()`?

packages/amplify_core_plugin_interface/lib/amplify_datastore_category.dart
jamesonwilliams: I thought so! VS Code told me I'd need to put `async` on the method signature, to become:
```dart
async Stream<SubscriptionEvent<T>> observe<T extends Model>(ModelType<T> modelType) {
      ...
}
```
Would that be acceptable?

packages/amplify_core_plugin_interface/lib/amplify_auth_category.dart
jamesonwilliams: I believe so, yes. When you do the work to invert the dependency on `core`, and so will have a `Plugin`, this `_select` could become a base method in `Plugin`.

jamesonwilliams:
> Thanks @jamesonwilliams for cleaning this up. Can you accompany this with some simple tests on `_select()`?

Yea! Will follow up with these.

packages/amplify_core_plugin_interface/lib/amplify_auth_category.dart
Amplifiyer: sounds good! I'm currently working on the refactor and it's a big dependency mess right now 😞

PR #259, opened 2020-12-17
Implement the GraphQL API Query and Mutate functionality

Ashish-Nanda:
*Description of changes:*

This PR implements the foundation of the Amplify Flutter API category along with the GraphQL API Query and Mutate functionality. The main aspects of this PR include:
- The API Plugin Interface
- The API category Plugin 
- GraphQL API Query implementation
- GraphQL API Mutate implementation
- iOS and Android unit tests
- Dart/Flutter unit tests


packages/amplify_api/lib/method_channel_api.dart
```diff
@@ -0,0 +1,86 @@
...
+import 'package:flutter/foundation.dart';
+import 'package:amplify_api_plugin_interface/amplify_api_plugin_interface.dart';
+
+import 'amplify_api.dart';
+
+const MethodChannel _channel = MethodChannel('com.amazonaws.amplify/api');
+
+class AmplifyAPIMethodChannel extends AmplifyAPI {
+  @override
+  GraphQLOperation<T> query<T>({@required GraphQLRequest<T> request}) {
```
Amplifiyer, line 26: Did we have an API review for this?

packages/amplify_api/lib/method_channel_api.dart
```diff
@@ -0,0 +1,86 @@
...
+      Future<GraphQLResponse<T>> response =
+          _getMethodChannelResponse(methodName: 'query', request: request);
+
+      GraphQLOperation<T> result =
+          GraphQLOperation<T>(cancel: () {}, response: response);
+
+      return result;
+    } on ApiError catch (e) {
+      throw e;
+    }
```
Amplifiyer, line 37: Is this not a no op?

packages/amplify_api/pubspec.yaml
```diff
@@ -0,0 +1,68 @@
+name: amplify_api
+description: A new flutter plugin project.
+version: 0.0.1
+author:
+homepage:
```
Amplifiyer, line 5: Should fill these appropriately.

packages/amplify_api/pubspec.yaml
```diff
@@ -0,0 +1,68 @@
...
+  #       - asset: fonts/Schyler-Italic.ttf
+  #         style: italic
+  #   - family: Trajan Pro
+  #     fonts:
+  #       - asset: fonts/TrajanPro.ttf
+  #       - asset: fonts/TrajanPro_Bold.ttf
+  #         weight: 700
+  #
+  # For details regarding fonts in packages, see
+  # https://flutter.dev/custom-fonts/#from-packages
```
Amplifiyer, line 68: can be removed.

packages/amplify_api_plugin_interface/lib/src/GraphQLOperation.dart
```diff
@@ -0,0 +1,31 @@
...
+  Function _cancel;
+
+  GraphQLOperation({@required this.response, @required Function cancel}) {
+    _cancel = cancel;
+  }
+
+  Future<GraphQLResponse<T>> response;
+
+  void cancel() {
+    _cancel();
```
Amplifiyer, line 29: Move the member fields together at the top

packages/amplify_api_plugin_interface/pubspec.yaml
```diff
@@ -0,0 +1,19 @@
+name: amplify_api_plugin_interface
+description: A new Flutter package project.
+version: 0.0.1
+author:
+homepage:
```
Amplifiyer, line 5: Should be updated appropriately.

Amplifiyer (commented):
Is there anything we need to do to support pagination for list type query?

packages/amplify_api/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+## 0.0.1
+
+* TODO: Describe initial release.
```
haverchuck, line 3: Don't forget to update this.

packages/amplify_api/README.md
```diff
@@ -0,0 +1,7 @@
+# amplify_api
+
+The Amplify Flutter API category plugin.
```
haverchuck, line 3: Please don't forget to update.

packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/AmplifyApiPlugin.kt
```diff
@@ -0,0 +1,202 @@
...
+          createFlutterError(
+                  flutterResult,
+                  FlutterApiErrorMessage.AMPLIFY_API_QUERY_FAILED.toString(),
+                  createErrorMap(it))
+        }
+    )
+  }
+
+  @VisibleForTesting
+  fun mutate(flutterResult: Result, request: Map<String, Any>) {
```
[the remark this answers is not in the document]
haverchuck: I think we do similar things in other plugins' platform code.

packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/AmplifyApiPlugin.kt
```diff
@@ -0,0 +1,202 @@
...
+    this.mainActivity = null
+  }
+
+  override fun onDetachedFromActivityForConfigChanges() {
+    onDetachedFromActivity()
+  }
+
+  override fun onReattachedToActivityForConfigChanges(binding: ActivityPluginBinding) {
+    onAttachedToActivity(binding)
+  }
```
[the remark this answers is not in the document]
haverchuck: I believe we were using ActivityAware in other plugins (Auth) to setup an event channel for Hub, but I don't think it's necessary or really the right way to go (I'm working on changing this in Auth).  Here, it probably depends on how exactly you'll be implementing subscribe.  

TLDR: It depends on `subscribe` implementation.

packages/amplify_api/ios/Classes/SwiftAmplifyApiPlugin.swift
```diff
@@ -0,0 +1,124 @@
...
+        case "query":
+            query(flutterResult: result, request: arguments)
+        case "mutate":
+            mutate(flutterResult: result, request: arguments)
+        default:
+            result(FlutterMethodNotImplemented)
+        }
+    }
+    
+    func query(flutterResult: @escaping FlutterResult, request: [String: Any]) {
```
haverchuck, line 46: Following the 'bridge' pattern we've been using may make unit testing these easier.

packages/amplify_api/pubspec.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+environment:
+  sdk: ">=2.7.0 <3.0.0"
+  flutter: ">=1.20.0 <2.0.0"
+
+dependencies:
+  flutter:
+    sdk: flutter
+  amplify_api_plugin_interface:
+    path: ../amplify_api_plugin_interface
```
haverchuck, line 15: Local paths not needed w/ melos.

packages/amplify_api_plugin_interface/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+## [0.0.1] - TODO: Add release date.
+
+* TODO: Describe initial release.
```
haverchuck, line 3: Must remember to update upon release.

packages/amplify_api_plugin_interface/lib/src/GraphQLOperation.dart
haverchuck: I'm slightly hesitant about this type. I think it's possible that customers could see `GraphQLOperation` and expect something more [akin to this.](https://github.com/zino-app/graphql-flutter/blob/a3e4401c09aa36f11a2f341f759a18702b32516d/packages/graphql/lib/src/link/operation.dart). Perhaps you could share type that Kyle is using to support cancel (I'm going off memory, so that might be off base).

packages/amplify_api_plugin_interface/lib/src/GraphQLResponse.dart
```diff
@@ -0,0 +1,24 @@
...
+
+import 'package:flutter/foundation.dart';
+import './types.dart';
+
+class GraphQLResponse<T> {
+  T data;
+  List<GraphQLResponseError> errors = [];
+
+  GraphQLResponse({@required this.data, this.errors});
+}
```
haverchuck, line 24: We don't need to support GraphQL [extensions](https://dgraph.io/docs/graphql/api/requests/#extensions-field)?

packages/amplify_core/pubspec.yaml
```diff
@@ -15,6 +15,8 @@ dependencies:
   amplify_analytics_plugin_interface: ^0.0.1-dev.4
   amplify_auth_plugin_interface: ^0.0.1-dev.4
   amplify_datastore_plugin_interface: ^0.0.1-dev.4
+  amplify_api_plugin_interface:
+    path: ../amplify_api_plugin_interface #TODO: Remove this before release
```
haverchuck, line 19: local paths not needed.

packages/amplify_core_plugin_interface/pubspec.yaml
```diff
@@ -14,6 +14,8 @@ dependencies:
   amplify_auth_plugin_interface: ^0.0.1-dev.4
   amplify_analytics_plugin_interface: ^0.0.1-dev.4
   amplify_datastore_plugin_interface: ^0.0.1-dev.4
+  amplify_api_plugin_interface:
+    path: ../amplify_api_plugin_interface #TODO: Remove this before release
```
haverchuck, line 18: Not needed.

haverchuck (changes requested):
Overall it's looking good - left some comments.

packages/amplify_api/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.2-all.zip
```
richardmcclellan, line 5: The latest version is 6.6.1.  Any reason not to use it?

```suggestion
distributionUrl=https\://services.gradle.org/distributions/gradle-6.6.1-all.zip
```

packages/amplify_api/android/build.gradle
```diff
@@ -0,0 +1,45 @@
+group 'com.amazonaws.amplify.amplify_api'
+version '1.0-SNAPSHOT'
+
+buildscript {
+    ext.kotlin_version = '1.3.50'
```
richardmcclellan, line 5: Should you use the latest version here?  
```suggestion
    ext.kotlin_version = '1.4.21'
```

packages/amplify_api/android/build.gradle
```diff
@@ -0,0 +1,45 @@
...
+    repositories {
+        google()
+        jcenter()
+    }
+
+    dependencies {
+        classpath 'com.android.tools.build:gradle:3.5.0'
```
richardmcclellan, line 12: Should you use the latest version here? 
```suggestion
        classpath 'com.android.tools.build:gradle:4.1.1'
```

packages/amplify_api_plugin_interface/lib/src/GraphQLResponse.dart
richardmcclellan: For reference, we don't support extensions on a response for JS, iOS or Android today, and there hasn't been a use case for it yet.

packages/amplify_api/android/build.gradle
```diff
@@ -0,0 +1,45 @@
...
+    }
+    lintOptions {
+        disable 'InvalidPackage'
+    }
+}
+
+dependencies {
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+    implementation "com.amplifyframework:aws-api:1.6.0"
+    implementation "com.amplifyframework:aws-api-appsync:1.6.0"
```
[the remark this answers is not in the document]
Ashish-Nanda: > Is there anything we need to do to support pagination for list type query?

Not for these APIs where the GraphQL document and variables need to be provided, since the customer explicitly decides what they want in the selection set, such as the pagination token.
In the future when we support model based APIs, we will need to handle it.

packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/AmplifyApiPlugin.kt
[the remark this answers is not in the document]
Ashish-Nanda: Removed it as it wasn't needed.

packages/amplify_api/ios/Classes/FlutterApiResponseUtils.swift
```diff
@@ -0,0 +1,47 @@
...
+            ]
+            print("Received a partially successful GraphQL response: \(result)")
+            flutterResult(result)
+        case .transformationError(let rawResponse, let error):
+            print("Received a partially successful GraphQL response with a transformation error: \(error)")
+            FlutterApiErrorUtils.handleAPIError(flutterResult: flutterResult, error: error, msg: failureMessage, rawResponse: rawResponse)
+        case .unknown(let errorDescription, let recoverySuggestion, _):
+            print("An unknown error occured: \(errorDescription)")
+            let errorMap = FlutterApiErrorUtils.createErrorMap(localizedError: errorDescription, recoverySuggestion: recoverySuggestion)
+            FlutterApiErrorUtils.createFlutterError(flutterResult: flutterResult, msg: failureMessage, errorMap: errorMap)
```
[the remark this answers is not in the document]
Ashish-Nanda: As discussed offline:
In the `transformationError` case the error type in iOS is also an APIError, and in the `unknown` case it is also a failure due to an unknown Error having occurred. For eg. if the response failed to deserialize correctly, it results in the `transformationError` case and throws and APIError.
Its not a GraphQLResponse error (from the server) like the partial success case above, and we are treating it as an APIError/failure as it should be.

packages/amplify_api/ios/Classes/SwiftAmplifyApiPlugin.swift
Ashish-Nanda: Done

packages/amplify_api/ios/amplify_api.podspec
```diff
@@ -0,0 +1,25 @@
...
+  s.name             = 'amplify_api'
+  s.version          = '0.0.1'
+  s.summary          = 'A new flutter plugin project.'
+  s.description      = <<-DESC
+A new flutter plugin project.
+                       DESC
+  s.homepage         = 'http://example.com'
+  s.license          = { :file => '../LICENSE' }
+  s.author           = { 'Your Company' => 'email@example.com' }
+  s.source           = { :path => '.' }
```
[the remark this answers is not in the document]
Ashish-Nanda: Updated

packages/amplify_api/lib/method_channel_api.dart
Ashish-Nanda: Yes, the interface and implementation is aligned with those approved in the review.

packages/amplify_api/lib/method_channel_api.dart
Ashish-Nanda: Removed

packages/amplify_api/pubspec.yaml
Ashish-Nanda: Done

packages/amplify_api/pubspec.yaml
Ashish-Nanda: Will remove before publishing.

packages/amplify_api/pubspec.yaml
Ashish-Nanda: Done

packages/amplify_api_plugin_interface/CHANGELOG.md
Ashish-Nanda: Updated. Will add the date at the time of the release

packages/amplify_api_plugin_interface/lib/src/GraphQLOperation.dart
Ashish-Nanda: Updated.
Regarding the type name, it is inline with what we decided in the API review, and is also similar to what customers selected as their preferred DX in the Rest API RFC.

packages/amplify_api_plugin_interface/lib/src/GraphQLResponse.dart
Ashish-Nanda: As Richard mentioned, its not supported by Amplify libraries at this time.

packages/amplify_api_plugin_interface/pubspec.yaml
Ashish-Nanda: Updated

packages/amplify_core/pubspec.yaml
Ashish-Nanda: Will remove this before release

packages/amplify_core_plugin_interface/pubspec.yaml
Ashish-Nanda: Will remove this before release

Ashish-Nanda (commented):
Addressed PR comments along with adding some more unit tests and other improvements

haverchuck (approved):

Amplifiyer (commented):
Looks good, minor comments and questions.

iOS tests are failing, please fix them before pushing to master.

packages/amplify_api/ios/Classes/FlutterApiErrorUtils.swift
```diff
@@ -0,0 +1,51 @@
...
+class FlutterApiErrorUtils {
+    
+    static func handleAPIError(flutterResult: @escaping FlutterResult, error: APIError, msg: String, rawResponse: String? = nil) {
+        var errorMap =
+            createErrorMap(localizedError: error.errorDescription, recoverySuggestion: error.recoverySuggestion)
+        if(rawResponse != nil) {
+            errorMap["rawResponse"] = rawResponse
+        }
+        createFlutterError(flutterResult:flutterResult, msg: msg, errorMap: errorMap)
+    }
```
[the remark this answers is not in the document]
Ashish-Nanda: This is a field in the details map, so it will present in the ApiError.details in Dart as well when this error occurs.
The `transformationError` case is explicitly handled in iOS, when the response from the server that could not be transformed into the specified response type. The `RawGraphQLResponse` contains the entire response from the service, and so it makes sense to add it to the details. 
In the future if Android handles this case explicitly and provides the `RawGraphQLResponse`, we can add it to the details for Android as well. Also deserialization and transformation is a library level bug, so it may occur in one platform and not the other for a given API call, but I do prefer the way Amplify iOS explicitly handles it and provides the raw response.

packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/FlutterApiErrorUtils.kt
```diff
@@ -0,0 +1,54 @@
...
+
+package com.amazonaws.amplify.amplify_api
+
+import android.os.Handler
+import android.os.Looper
+import androidx.annotation.NonNull
+import com.amplifyframework.api.ApiException
+import io.flutter.plugin.common.MethodChannel
+
+class FlutterApiErrorUtils {
```
jamesonwilliams, line 24: Avoid `Utils` in all names.

This looks like a factory, though, since it builds errors? `FlutterApiErrorFactory`?

packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/FlutterApiRequestUtils.kt
```diff
@@ -0,0 +1,46 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazonaws.amplify.amplify_api
+
+import com.amplifyframework.AmplifyException
+
+class FlutterApiRequestUtils {
```
jamesonwilliams, line 20: Same here, avoid `Utils`. Is this a `RequestFactory`?

packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/FlutterGraphQLApiModule.kt
```diff
@@ -0,0 +1,111 @@
...
+import android.os.Handler
+import android.os.Looper
+import com.amazonaws.amplify.amplify_api.types.FlutterApiErrorMessage
+import com.amplifyframework.api.aws.GsonVariablesSerializer
+import com.amplifyframework.api.graphql.SimpleGraphQLRequest
+import com.amplifyframework.core.Amplify
+import io.flutter.plugin.common.MethodChannel
+
+class FlutterGraphQLApiModule {
+    companion object {
```
jamesonwilliams, line 27: Seems to be a pattern in this PR if using static classes to warehouse some code.

In this one, look at the signatures of the methods. They seem to all need a method channel. Maybe you should be passing that in as a dependency, and creating an instance of this class.

What does `Module` mean?

Alternately, why don't you split this out into two things, a `FlutterGraphQlQuery` and a `FlutterGraphQlMutation`. You can make instances of them, and passa method channel as a dependency. They'll have better names.

They're looking like wrappers around the native versions of those calls, right?

packages/amplify_api/android/src/test/kotlin/com/amazonaws/amplify/amplify_api/GraphQLApiUnitTests.kt
```diff
@@ -0,0 +1,299 @@
...
+class GraphQLApiUnitTests {
+    lateinit var flutterPlugin: AmplifyApiPlugin
+    private var mockApi = mock(ApiCategory::class.java)
+    private val mockResult: MethodChannel.Result = mock(MethodChannel.Result::class.java)
+    private val mockGraphQLOperation = mock(GraphQLOperation::class.java)
+
+    @Before
+    fun setup() {
+        flutterPlugin = AmplifyApiPlugin()
+        setFinalStatic(Amplify::class.java.getDeclaredField("API"), mockApi)
```
jamesonwilliams, line 49: #270 and #272 show how to avoid this. Mock the behavior of the native plugin, instead. Leave the category alone.

You may need to open up a second constructor in your `AmplifyApiPlugin`. It should call onto `ApiCategoryBehavior`, instead of `Amplfy.API` directly.

packages/amplify_api/android/src/test/kotlin/com/amazonaws/amplify/amplify_api/GraphQLApiUnitTests.kt
```diff
@@ -0,0 +1,299 @@
...
+                createdAt
+                }
+            }
+        }"""
+        testRequest["variables"] = HashMap<String, Any>();
+
+        val graphQLResponse: GraphQLResponse<String> = GraphQLResponse("result", null)
+
+        doAnswer { invocation ->
+            Assert.assertEquals(
```
jamesonwilliams, line 69: Instead of asserting in your invocation lambda, you should:
1. Tighten the `when` clause, so that the answer only fires when this condition is met
2. Use a `verify` call towards the end of the test method, that captures the input value using an `ArgumentCaptor`.

packages/amplify_api/android/src/test/kotlin/com/amazonaws/amplify/amplify_api/GraphQLApiUnitTests.kt
```diff
@@ -0,0 +1,299 @@
...
+                any<GraphQLRequest<String>>(),
+                any(),
+                any())
+
+        flutterPlugin.onMethodCall(
+                MethodCall("query", testRequest),
+                mockResult
+        )
+
+        verify(mockResult, times(1)).success(mapOf(
```
jamesonwilliams, line 92: Drop `times(1)` throughout the PR. `verify(foo)` is identically equal to `verify(foo, times(1))`. The latter would only be useful if you have a test that tests invocation counts other than 1 sometimes.

packages/amplify_api/android/src/test/kotlin/com/amazonaws/amplify/amplify_api/GraphQLApiUnitTests.kt
```diff
@@ -0,0 +1,299 @@
...
+                id
+                name
+                createdAt
+                }
+            }
+        }"""
+        testRequest["variables"] = HashMap<String, Any>()
+
+        doAnswer { invocation ->
+            Assert.assertEquals(
```
jamesonwilliams, line 132: ditto

packages/amplify_api/android/src/test/kotlin/com/amazonaws/amplify/amplify_api/GraphQLApiUnitTests.kt
```diff
@@ -0,0 +1,299 @@
...
+                                mapOf(
+                                        "platform" to "Android",
+                                        "localizedErrorMessage" to "AmplifyException",
+                                        "recoverySuggestion" to ApiException.REPORT_BUG_TO_AWS_SUGGESTION,
+                                        "errorString" to apiException.toString()))
+        )
+
+    }
+
+    private fun setFinalStatic(field: Field, newValue: Any?) {
```
jamesonwilliams, line 292: Once you update your flutter plugin to call `ApiCategoryBehavior` instead of `Amplify.API` directly, you will no longer need to do this. You can just use a `mock(ApiPlugin)` and then arrange behaviors on it. In your test, you'll pass the mock plugin down into your Flutter wrapper plugin.

packages/amplify_api/example/android/app/build.gradle
```diff
@@ -0,0 +1,63 @@
...
+if (flutterVersionName == null) {
+    flutterVersionName = '1.0'
+}
+
+apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
+apply from: "$flutterRoot/packages/flutter_tools/gradle/flutter.gradle"
+
+android {
+    compileSdkVersion 28
```
jamesonwilliams, line 29: As @richardmcclellan points out, we should go through the entire code base and update all these out-of-date Android version. This one should say 30. The dependency version can mostly be updated, too. The Gradle and Android Gradle Plugin version can all be updated. Probably another PR, I guess.

packages/amplify_api/example/lib/main.dart
```diff
@@ -0,0 +1,132 @@
...
+
+  Future<void> initPlatformState() async {
+    AmplifyAPI apiPlugin = AmplifyAPI();
+
+    await amplify.addPlugin(apiPlugins: [apiPlugin]);
+
+    // Configure
+    await amplify.configure(amplifyconfig);
+    setState(() {
+      _isAmplifyConfigured = true;
```
jamesonwilliams, line 50: Seems like, if you're going to keep this state boolean, might as well do it in the façade so that people can ask like `Amplify.isConfigured()`.

packages/amplify_api/example/lib/main.dart
```diff
@@ -0,0 +1,132 @@
...
+        items {
+          id
+          name
+          createdAt
+        }
+      }
+    }''';
+
+    var operation = await Amplify.API
+        .query<String>(request: GraphQLRequest(document: graphQLDocument));
```
jamesonwilliams, line 66: I didn't ask this in the design review, but last chance before it's permanent -- there's no way to omit this `<String>` syntax, right?

packages/amplify_api/ios/Classes/AmplifyApiPlugin.m
```diff
@@ -0,0 +1,15 @@
...
+#if __has_include(<amplify_api/amplify_api-Swift.h>)
+#import <amplify_api/amplify_api-Swift.h>
+#else
+// Support project import fallback if the generated compatibility header
+// is not copied when this plugin is created as a library.
+// https://forums.swift.org/t/swift-static-libraries-dont-copy-generated-objective-c-header/19816
+#import "amplify_api-Swift.h"
+#endif
+
+@implementation AmplifyApiPlugin
```
jamesonwilliams, line 11: Whoa, what's up with the Objective C?

packages/amplify_api/ios/Classes/ApiBridge.swift
```diff
@@ -0,0 +1,55 @@
...
+ * permissions and limitations under the License.
+ */
+
+import Amplify
+import AmplifyPlugins
+import Flutter
+
+class ApiBridge {
+    
+    func get(request: RESTRequest, listener: ((AmplifyOperation<RESTOperationRequest, Data, APIError>.OperationResult) -> Void)?) -> RESTOperation?{
```
jamesonwilliams, line 22: You could do a typealias for this really long operation type (and, for that matter, its result type, as well.)

packages/amplify_api/ios/Classes/ApiBridge.swift
```diff
@@ -0,0 +1,55 @@
...
+    func put(request: RESTRequest, listener: ((AmplifyOperation<RESTOperationRequest, Data, APIError>.OperationResult) -> Void)?) -> RESTOperation?{
+        return Amplify.API.put(request: request, listener: listener)
+    }
+
+    func post(request: RESTRequest, listener: ((AmplifyOperation<RESTOperationRequest, Data, APIError>.OperationResult) -> Void)?) -> RESTOperation?{
+        return Amplify.API.post(request: request, listener: listener)
+    }
+    
+    func delete(request: RESTRequest, listener: ((AmplifyOperation<RESTOperationRequest, Data, APIError>.OperationResult) -> Void)?) -> RESTOperation?{
+        return Amplify.API.delete(request: request, listener: listener)
```
jamesonwilliams, line 35: Instead of `Amplify.API`, call through our abstraction, the category behavior, `ApiCategoryBehavior`. Same comment applies to both iOS & Android impls.

packages/amplify_api/ios/Classes/FlutterApiResponseUtils.swift
```diff
@@ -0,0 +1,47 @@
...
+import Amplify
+import AmplifyPlugins
+
+class FlutterApiResponseUtils {
+    static func handleGraphQLErrorResponse(flutterResult: @escaping FlutterResult, errorResponse: GraphQLResponseError<String>, failureMessage: String) {
+        switch(errorResponse) {
+        case .error(let errorList):
+            let result: [String: Any?] = [
+                "data": nil,
+                "errors": errorList.map { $0.message }
```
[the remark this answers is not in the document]
Amplifiyer: Please create a story for this so we don't loose track of it.

packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/AmplifyApiPlugin.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    private lateinit var channel: MethodChannel
+    private lateinit var context: Context
+    private val LOG = Amplify.Logging.forNamespace("amplify:flutter:api")
+
+    override fun onAttachedToEngine(@NonNull flutterPluginBinding: FlutterPlugin.FlutterPluginBinding) {
+        channel = MethodChannel(flutterPluginBinding.binaryMessenger, "com.amazonaws.amplify/api")
+        channel.setMethodCallHandler(this)
+        context = flutterPluginBinding.applicationContext
+        Amplify.addPlugin(AWSApiPlugin())
```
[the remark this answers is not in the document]
Amplifiyer: Please create a story for this so we don't lose track of it. do we know the exposure of not adding this? cc: @jamesonwilliams

packages/amplify_api/ios/Classes/ApiBridge.swift
```diff
@@ -0,0 +1,55 @@
...
+                                     listener: GraphQLOperation<R>.ResultListener?) -> GraphQLOperation<R> {
+        return Amplify.API.mutate(request: request, listener: listener)
+    }
+
+    func subscribe<R>(request: GraphQLRequest<R>,
+                             valueListener: GraphQLSubscriptionOperation<R>.InProcessListener?,
+                             completionListener: GraphQLSubscriptionOperation<R>.ResultListener?)
+        -> GraphQLSubscriptionOperation<R> {
+        return Amplify.API.subscribe(request: request, valueListener: valueListener, completionListener: completionListener)
+    }
```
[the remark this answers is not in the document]
Amplifiyer: hmm, should we do this for Kotlin as well then to keep similar implementation across platforms? Right now in Kotlin these are combined in one module `FlutterGraphQLApiModule.kt` and the whole file doesn't get tested since it's mocked.

packages/amplify_api/ios/Classes/FlutterApiErrorUtils.swift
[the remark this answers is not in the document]
Amplifiyer: Gotcha, so for this particular case, the `transformationError` in either platform results in exactly the same error in Dart?

packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/AmplifyApiPlugin.kt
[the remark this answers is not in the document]
richardmcclellan: Allowed custom auth providers is required when customers want to use their own OIDC provider.  It's probably not a P0 feature, but something that should be a fast follow.

Amplifiyer (approved):
Looks good. Approving the PR, some comments that I left can be addressed in follow up PRs, please create stories for them.

packages/amplify_api/lib/method_channel_api.dart
```diff
@@ -0,0 +1,80 @@
...
+  ApiError _formatError(PlatformException e) {
+    return ApiError(code: e.code, message: e.message, details: e.details);
+  }
+
+  List<GraphQLResponseError> _deserializeGraphQLResponseErrors(
+      Map<String, dynamic> result) {
+    //TODO: Deserialize errors list into List<GraphQLResponseError>
+    return [];
+  }
+}
```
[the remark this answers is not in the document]
Ashish-Nanda: Done

packages/amplify_api/android/src/test/kotlin/com/amazonaws/amplify/amplify_api/GraphQLApiUnitTests.kt
Ashish-Nanda: Good point. Updated.

packages/amplify_api/android/src/test/kotlin/com/amazonaws/amplify/amplify_api/GraphQLApiUnitTests.kt
Ashish-Nanda: Nice! Will address as part of the test refactor PR.

packages/amplify_api/example/android/app/build.gradle
Ashish-Nanda: Agree. Will create a story to track this.

packages/amplify_api/example/lib/main.dart
Ashish-Nanda: Makes sense. We can consider adding this change in the current facade PR.

packages/amplify_api/example/lib/main.dart
Ashish-Nanda: It doesnt need to be provided explicitly as long as it can be inferred from somewhere, for example the `GraphQLRequest` thats passed in. Otherwise it will be dynamic. 
The documentation [here]( https://github.com/dart-lang/sdk/blob/master/pkg/dev_compiler/doc/GENERIC_METHODS.md#calling-generic-methods) provides more details.

packages/amplify_api/ios/Classes/AmplifyApiPlugin.m
Ashish-Nanda: This file is auto generated by Flutter for the plugin 😄

packages/amplify_api/ios/Classes/ApiBridge.swift
Ashish-Nanda: Good point. It will be included in the existing PR to support REST.

packages/amplify_api/ios/Classes/ApiBridge.swift
Ashish-Nanda: Will do this as part of the test refactor PR as its related to that.

Issue #274, opened 2020-12-23
Upcoming breaking changes for amplify flutter repository.


Amplifiyer:
Hi all,

We have a few breaking changes coming up in our next release as detailed below. Please upgrade your apps appropriately when you upgrade amplify_flutter.

1. **[Released in 0.0.2-dev.1]**  Package amplify_core will be renamed to amplify_flutter and amplify_core will only contain base types and core utils.
For more details, see this RFC https://github.com/aws-amplify/amplify-flutter/issues/263 and PR https://github.com/aws-amplify/amplify-flutter/issues/273.

The breaking change will be the path to importing `Amplify`

```diff
- import 'package:amplify_core/amplify_core.dart';
+ import 'package:amplify_flutter/amplify.dart';
```

2. **[Released in 0.0.2-dev.1]**  Top level Amplify class is now a static singleton, so you don't need to instantiate it before calling `configure()` or `addPlugin()`
```diff
- Amplify amplify = new Amplify();
- amplify.configure(amplifyconfig);
+ Amplify.configure(amplifyconfig);
```

3. **[Released in 0.0.2-dev.1]** `addPlugin` now takes a single plugin instance without the need for named parameters. You can alternatively use a new `addPlugins([AmplifyPlugins])` to add all the plugins at the same time.
```diff
AmplifyAuthCognito authPlugin = AmplifyAuthCognito();
AmplifyStorageS3 storagePlugin = AmplifyStorageS3();
AmplifyAnalyticsPinpoint analyticsPlugin = AmplifyAnalyticsPinpoint();

- amplify.addPlugin(authPlugins: [authPlugin], 
-        storagePlugins: [storagePlugin], 
-        analyticsPlugins: [analyticsPlugin]);
+ Amplify.addPlugin(authPlugin);
+ Amplify.addPlugin(storagePlugin);
+ Amplify.addPlugin(analyticsPlugin);

// or alternatively Amplify.addPlugins([authPlugin, storagePlugin, analyticsPlugin]);
```

4. **[Released in 0.0.2-dev.1]**  Hub Events will be exposed as instantiated Dart types with typed payload data, instead of simple Maps.  They will also be accessed with a new API that allow for subscriptions which listen on multiple channels simultaneously:

```dart
hubSubscription = Amplify.Hub.listen([HubChannel.DataStore], (msg) {
    /// do something with data
});
```

[...]

PR #275, opened 2020-12-25
Add a new amplify_core package for base types and utilities

Amplifiyer:
*Issue #, if available:* https://github.com/aws-amplify/amplify-flutter/issues/274

*Description of changes:*
1. Top level Amplify class is now a static singleton, so custoers don't need to instantiate it before calling `configure()` or `addPlugin()`. DX change:
```diff
- Amplify amplify = new Amplify();
- amplify.configure(amplifyconfig);
+ Amplify.configure(amplifyconfig);
```

2. `addPlugin` now takes a single plugin instance without the need for named parameters. Customers can alternatively use a new `addPlugins([AmplifyPlugins])` to add all the plugins at the same time. DX change:
```diff
AmplifyAuthCognito authPlugin = AmplifyAuthCognito();
AmplifyStorageS3 storagePlugin = AmplifyStorageS3();
AmplifyAnalyticsPinpoint analyticsPlugin = AmplifyAnalyticsPinpoint();

- amplify.addPlugin(authPlugins: [authPlugin], 
-        storagePlugins: [storagePlugin], 
-        analyticsPlugins: [analyticsPlugin]);
+ Amplify.addPlugin(authPlugin);
+ Amplify.addPlugin(storagePlugin);
+ Amplify.addPlugin(analyticsPlugin);

// or alternatively Amplify.addPlugins([authPlugin, storagePlugin, analyticsPlugin]);
```


packages/amplify_analytics_plugin_interface/pubspec.yaml
```diff
@@ -9,8 +9,9 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  plugin_platform_interface: ^1.0.1
   meta: ^1.1.8
+  amplify_core:
+    path: ../amplify_core
```
jamesonwilliams, line 14: cc: @haverchuck. I'm confused about when to use local pathing vs. not. If you use a version number, does melos publish to and read from some local development store, using those numbers? How to I publish to that development store, to prevent the need for the paths?

packages/amplify_core/lib/types/plugin/amplify_base_plugin_interface.dart
```diff
@@ -0,0 +1,23 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import 'package:meta/meta.dart';
+import 'package:plugin_platform_interface/plugin_platform_interface.dart';
+
+/// Interface for all amplify plugins
+abstract class AmplifyBasePluginInterface extends PlatformInterface {
```
jamesonwilliams, line 20: Naming-wise, I'd suggest to drop "Base", both from the class name and the file name.

```suggestion
abstract class AmplifyPluginInterface extends PlatformInterface {
```

For reference, both [iOS](https://github.com/aws-amplify/amplify-ios/blob/main/Amplify/Core/Plugin/Plugin.swift#L11) and [Android](https://github.com/aws-amplify/amplify-android/blob/main/core/src/main/java/com/amplifyframework/core/plugin/Plugin.java#L47) call this just `Plugin`.

If my reading is correct, Amplify JS doesn't actually have any common base type for plugins? A "pluggable" is just something that can be added to an individual category, by convention?

packages/amplify_storage_s3/lib/amplify_storage_s3.dart
```diff
@@ -35,7 +35,8 @@ class AmplifyStorageS3 extends StoragePluginInterface {
   static StoragePluginInterface get instance => _instance;
 
   static set instance(StoragePluginInterface instance) {
-    PlatformInterface.verifyToken(instance, _token);
+    PlatformInterface.verifyToken(
+        instance, _token);
```
jamesonwilliams, line 39:
```suggestion
    PlatformInterface.verifyToken(instance, _token);
```

jamesonwilliams (approved):
Looks great!

packages/amplify_analytics_plugin_interface/pubspec.yaml
Amplifiyer: When using `melos` we don't need local paths as running `melos bootstrap` in the root directory will setup proper  local symlinks for the dependencies and won't look for pub.get for them.

However the gotcha is that you won't be able to run `flutter pub get` in the individual packages since that will override the melos symlinks and will start using the `pub.get` versions (what you observed @jamesonwilliams  that local changes were not being picked up)

packages/amplify_core/lib/types/plugin/amplify_base_plugin_interface.dart
Amplifiyer: Refactoring in the branch.

packages/amplify_analytics_plugin_interface/pubspec.yaml
haverchuck: @Amplifiyer I guess I can see the utility of being able to run pub get on individual packages during development, but why commit this file with a local path?

packages/amplify_auth_plugin_interface/lib/amplify_auth_plugin_interface.dart
```diff
@@ -18,11 +18,11 @@ library amplify_auth_plugin_interface;
 import 'dart:async';
 import 'package:flutter/foundation.dart';
 import 'package:meta/meta.dart';
-import 'package:plugin_platform_interface/plugin_platform_interface.dart';
+import 'package:amplify_core/types/index.dart';
```
haverchuck, line 21: happy to see this!

packages/amplify_analytics_plugin_interface/pubspec.yaml
Amplifiyer: unfortunately this is required to run iOS unit tests in the pipeline until that package is available from pub.get (or even after that) since `flutter pub get` is necessary to be run before running iOS tests and running that basically overrides everything melos did and so iOS tests then run on the published versions and not the local dependencies.

This is the same issue we ran into amplify-js where we resolved it by using local publishing via verdaccio to use local package dependencies. We may need to find a solution for iOS tests here to possibly run without requiring `flutter pub get` (it's currently required to create/update flutter's Podfile)

packages/amplify_flutter/lib/amplify.dart
```diff
@@ -27,51 +30,50 @@ import 'categories/amplify_categories.dart';
...
-  static const AuthCategory Auth = const AuthCategory();
-  static const AnalyticsCategory Analytics = const AnalyticsCategory();
-  static const StorageCategory Storage = const StorageCategory();
-  static const DataStoreCategory DataStore = const DataStoreCategory();
+/// Top level singleton Amplify object.
+final AmplifyClass Amplify = new AmplifyClass._private();
+
+class AmplifyClass extends PlatformInterface {
+  AmplifyClass._private();
+
```
haverchuck, line 38: This `_private()` method is a factory constructor to ensure a singleton?

haverchuck (commented):
Left a couple of questions.  I'm still not sure why we are committing files with local paths.  Otherwise, LGTM.

packages/amplify_analytics_plugin_interface/pubspec.yaml
haverchuck: Ah, I see.  So this is a temporary pipeline workaround for a new package?

haverchuck (approved):
LGTM

packages/amplify_flutter/lib/amplify.dart
Amplifiyer: yes, a named constructor starting with a `_` is private to the library (like any other identifier in dart) and so cannot be initialized by the apps. This ensure that only one instance exists for this class which is accessed by the static `Amplify`.

PR #282, opened 2020-12-31
fix(amplify_api): Remove cancelled operations iOS

fjnoyp:
Quick fix to get operations to be removed on failed rest operations in iOS 

### DIFF
--- a/packages/amplify_api/ios/Classes/FlutterRestApi.swift
+++ b/packages/amplify_api/ios/Classes/FlutterRestApi.swift
@@ -34,8 +34,14 @@ public class FlutterRestApi {
             let restOperation = function(restRequest) { result in
                 switch result {
                     case .success(let data):
+                        if(!cancelToken.isEmpty){
+                            OperationsManager.removeOperation(cancelToken: cancelToken)
+                        }
                         self.prepareRestResponseResult(flutterResult: flutterResult, data: data, cancelToken: cancelToken)
                     case .failure(let apiError):
+                        if(!cancelToken.isEmpty){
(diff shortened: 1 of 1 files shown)


Ashish-Nanda (approved):
LGTM

packages/amplify_api/ios/Classes/FlutterRestApi.swift
```diff
@@ -57,10 +63,6 @@ public class FlutterRestApi {
     }
 
     private static func prepareRestResponseResult(flutterResult: @escaping FlutterResult, data: Data, cancelToken: String = ""){
```
Amplifiyer, line 65:
```suggestion
    private static func prepareRestResponseResult(flutterResult: @escaping FlutterResult, data: Data){
```

packages/amplify_api/ios/Classes/FlutterRestApi.swift
```diff
@@ -34,8 +34,14 @@ public class FlutterRestApi {
             let restOperation = function(restRequest) { result in
                 switch result {
                     case .success(let data):
+                        if(!cancelToken.isEmpty){
+                            OperationsManager.removeOperation(cancelToken: cancelToken)
+                        }
                         self.prepareRestResponseResult(flutterResult: flutterResult, data: data, cancelToken: cancelToken)
                     case .failure(let apiError):
```
Amplifiyer, line 41:
```suggestion
                        }
                        self.prepareRestResponseResult(flutterResult: flutterResult, data: data)
                    case .failure(let apiError):
```

Issue #285, opened 2021-01-01
[android] cannot build in 0.0.1-dev.5


WingCH:
android is normal in `^0.0.1-dev.4`, but `^0.0.1-dev.5` have error

```yaml
amplify_core: ^0.0.1-dev.5
amplify_auth_cognito: ^0.0.1-dev.5
amplify_storage_s3: ^0.0.1-dev.5
```

```json
Plugin project :connectivity_macos not found. Please update settings.gradle.

FAILURE: Build completed with 2 failures.

1: Task failed with an exception.
-----------
* Where:
Build file '/Users/wingch/.pub-cache/hosted/pub.dartlang.org/amplify_auth_cognito-0.0.1-dev.5/android/build.gradle' line: 27

* What went wrong:
A problem occurred evaluating project ':amplify_auth_cognito'.
> Could not read script '/Users/wingch/.pub-cache/hosted/build-support/coverage.gradle' as it does not exist.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
==============================================================================

2: Task failed with an exception.
-----------
* What went wrong:
A problem occurred configuring project ':amplify_auth_cognito'.
> compileSdkVersion is not specified. Please add it to build.gradle

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
==============================================================================

* Get more help at https://help.gradle.org

BUILD FAILED in 934ms
Exception: Gradle task assembleDebug failed with exit code 1
```

```
[✓] Flutter (Channel stable, 1.22.5, on macOS 11.1 20C69 darwin-x64, locale zh-Hant-HK)
    • Flutter version 1.22.5 at /usr/local/bin/flutter
    • Framework revision 7891006299 (3 weeks ago), 2020-12-10 11:54:40 -0800
    • Engine revision ae90085a84
    • Dart version 2.10.4

 
[✓] Android toolchain - develop for Android devices (Android SDK version 29.0.2)
    • Android SDK at /Users/wingch/Library/Android/sdk
    • Platform android-30, build-tools 29.0.2
    • ANDROID_HOME = /Users/wingch/Library/Android/sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b3-6915495)
[...]

PR #293, opened 2021-01-04
fix relative path for coverage.gradle to be local to the package

Amplifiyer:
*Issue #, if available:* #285

*Description of changes:* Change the relative path of `coverage.gradle` to be local to the flutter package instead of reading from root which is not published to pub.dev.

Testing with unpub before releasing to make sure it works.


haverchuck:
#285

jamesonwilliams (approved):
Ideally, I think we would have a setup more like:
1. A single `coverage.gradle` source document
2. Some build time logic to copy it to each module as a build artifact
3. Modules get published with the programmatically copied file

However, if the build is currently not functioning, it would be better to go with your current fix, than to spend an afternoon playing around with the build.

haverchuck (approved):
Include changelogs and version bumps for immediate release. Otherwise, LGTM.

Ashish-Nanda (approved):

PR #307, opened 2021-01-13
feat(temporal): Add support for temporal types

fjnoyp:
Follow AppSync Temporal Spec: https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html

*Issue #, if available:*

*Description of changes:*


packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_time.dart
```diff
@@ -0,0 +1,128 @@
...
+    // Extract Time
+    var match = regExp.matchAsPrefix(regexString);
+
+    int hours = int.parse(match.group(1));
+    int minutes = int.parse(match.group(2));
+    int seconds = Temporal.getIntOr0(match.group(4));
+    _nanoseconds = Temporal.getIntOr0(match.group(6));
+    //int milliseconds =
+
+    _dateTime = DateTime.utc(1970, 1, 1, hours, minutes, seconds);
```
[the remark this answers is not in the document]
fjnoyp: The reason is to be consistent with Android's implementation which sets it to 1970/01/01

This can impact developers as the method "toDateTime" returns a Dart DateTime which needs to have 1970/01/01 (Android has something similar in their toDate method which auto sets to 1970/01/01)

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_datetime.dart
```diff
@@ -0,0 +1,112 @@
...
+        (dateTime.millisecond * 1000000) + (dateTime.microsecond * 1000);
+
+    if (offset.inDays > 0) {
+      throw new Exception("Cannot have an offset in days (hh:mm:ss)");
+    }
+
+    _offset = offset;
+  }
+
+  // DateTime Z is required
```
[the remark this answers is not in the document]
fjnoyp: For AWSDateTime specs it says: 
`The time zone offset is compulsory for this scalar. The time zone offset must either be Z (representing the UTC time zone) or be in the format ±hh:mm:ss. 
`
https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_date.dart
```diff
@@ -0,0 +1,111 @@
...
+    // TODO: enforce month 1-12
+    RegExp regExp = new RegExp(
+        r'^([0-9]{4}-[0-1][0-9]-[0-3][0-9])((z|Z)|((\+|-)[0-2][0-9]:[0-5][0-9](:[0-5][0-9])?))?',
+        caseSensitive: false,
+        multiLine: false);
+
+    // Validate
+    String regexString = regExp.stringMatch(iso8601String);
+    if (regexString != iso8601String) {
+      throw new Exception("invalid string input");
```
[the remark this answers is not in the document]
fjnoyp: If the user does not pass something in the form YYYY-MM-DD, an exception will be thrown.  So 030303 would not be recognized and consequently an exception would be thrown.

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_date.dart
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  Duration getOffset() {
+    return _offset;
+  }
+
+  // return DateTime with offset added
+  //Date is set as January 1, 1970.
+  DateTime toDateTime() {
+    return _offset == null ? _dateTime : _dateTime.subtract(_offset);
```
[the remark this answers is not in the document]
fjnoyp: Yeah that sounds about right.

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_date.dart
```diff
@@ -0,0 +1,111 @@
...
+      _offset = Temporal.stringToOffset(match.group(2));
+    else if (iso8601String.toLowerCase().contains("z")) _hasZ = true;
+  }
+
+  Duration getOffset() {
+    return _offset;
+  }
+
+  // return DateTime with offset added
+  //Date is set as January 1, 1970.
```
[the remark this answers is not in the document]
fjnoyp: yes thanks I'll remove that

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_datetime.dart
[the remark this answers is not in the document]
lawmicha: I feel like there's something to be said about hardcoding AppSync type specific requirements at the Temporal.DateTime level, like is this AppSync.AWSDateTime or Temporal.DateTime? I'm not sure what the solution is since even if we relax the validation because iso8601 spec allows it, it would then fail at the API call. perhaps it's up to the API plugin to validate the iso8601String of Temporal.DateTime's. This would still be troublesome since developers are actually calling DataStore and so SQL wouldn't reject it, then only in the sync engine would it be rejected.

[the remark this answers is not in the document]
fjnoyp: Good point, though isn't the entire purpose of Temporal.DateTime to be stored to AppSync?

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_date.dart
```diff
@@ -0,0 +1,111 @@
...
+    var match = regExp.matchAsPrefix(regexString);
+
+    // Parse cannot take a YYYY-MM-DD as UTC!
+    DateTime dateTime = DateTime.parse(match.group(1));
+    _dateTime = DateTime.utc(dateTime.year, dateTime.month, dateTime.day);
+
+    // Extract Offset
+    if (match.group(2) != null && match.group(2).isNotEmpty)
+      _offset = Temporal.stringToOffset(match.group(2));
+    else if (iso8601String.toLowerCase().contains("z")) _hasZ = true;
```
richardmcclellan, line 96: If the `iso8601String` does contain `Z`, you should set the duration to 0 seconds as well.  That way, calling `getOffset()` will correctly return a Duration of 0 seconds, instead of null.

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_date.dart
```diff
@@ -0,0 +1,111 @@
...
+    // Extract date portion of DateTime.ISO8601String
+    buffer.write(_dateTime.toIso8601String().substring(0, 10));
+
+    // Duration.toString returns string of form -9:30:00.000000 / 9:30:00.000000
+    // But we need -09:30 / +09:30
+    if (_offset != null) {
+      buffer.write(Temporal.durationToString(_offset));
+    } else if (_hasZ) {
+      buffer.write("Z");
+    }
```
richardmcclellan, line 122: Assuming you make my suggested change to set duration to 0 seconds when "Z" is present, this logic should change to:
```suggestion
    if (_offset != null) {
      if(_hasZ) {
        buffer.write("Z");
      } else {
        buffer.write(Temporal.durationToString(_offset));
      }
    }
```

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_time.dart
```diff
@@ -0,0 +1,151 @@
...
+
+import 'temporal.dart';
+
+/// Represents a valid extended ISO-8601 Time string, with an optional timezone offset.
+/// hh:mm:ss.sss±hh:mm:ss
+/// hh:mm:ss.sss
+/// https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html#appsync-defined-scalars
+class TemporalTime {
+  DateTime _dateTime;
+  int _nanoseconds; // DateTime only stores millisecond and microsecond
```
richardmcclellan, line 24: It's unfortunate that a Dart `DateTime` can't store nanoseconds :unamused:

From what I can tell, it does look like you are correctly saving the milliseconds, microseconds, and nanoseconds within this member variable though.

One minor suggestion though - I think it might simplify some of the conversion logic if you store the milliseconds and microseconds as a part of `_dateTime`, and then just use `_nanoseconds` for just the nanoseconds components (a value 0...999).  Then, the `toDateTime` method can be simplified to a simple getter, which just returns `_dateTime`.

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_time.dart
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  /// Return offset
+  Duration getOffset() {
+    return _offset;
+  }
+
+  /// Return DateTime with offset added
+  /// Date fields are set as January 1, 1970.
+  DateTime toDateTime() {
```
richardmcclellan, line 109: Perhaps call this `getDateTime()` instead?  This method is not _converting_ the `TemporalDateTime`, it is simply returning the "instant" component of it.

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_datetime.dart
```diff
@@ -0,0 +1,112 @@
...
+      _offset = Temporal.stringToOffset(match.group(7));
+  }
+
+  Duration getOffset() {
+    return _offset;
+  }
+
+  DateTime toDateTime() {
+    DateTime toReturn =
+        _offset == null ? _dateTime : _dateTime.subtract(_offset);
```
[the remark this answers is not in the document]
fjnoyp: Thanks, I have updated all Temporal types toDateTime to just return the underlying DateTime.

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_timestamp.dart
```diff
@@ -0,0 +1,29 @@
+import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart';
+
+///
+class TemporalTimestamp {
+  int _secondsSinceEpoch;
+
+  ///
+  TemporalTimestamp(DateTime date) {
+    var ms = (new DateTime.now()).millisecondsSinceEpoch;
```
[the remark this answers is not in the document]
fjnoyp: Thanks, great catch!

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_timestamp.dart
```diff
@@ -0,0 +1,29 @@
...
+    _secondsSinceEpoch = (ms / 1000).round();
+  }
+
+  ///
+  TemporalTimestamp.fromSeconds(int secondsSinceEpoch) {
+    _secondsSinceEpoch = secondsSinceEpoch;
+  }
+
+  int toInt() {
```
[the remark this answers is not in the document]
fjnoyp: changed to `toSeconds()`

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_date.dart
```diff
@@ -0,0 +1,111 @@
+import 'package:date_time_format/date_time_format.dart';
+
+import 'temporal.dart';
+
+///
+class TemporalDate {
+  DateTime _dateTime;
+  Duration _offset;
+  bool _hasZ =
```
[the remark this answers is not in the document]
fjnoyp: Thanks for the explanation there.  If the distinction isn't important then I think removing the `hasZ` boolean is preferable to keep our code cleaner while not affecting the customer experience.

Amplifiyer (commented):
There are some new test failures that need to be looked at.

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_date.dart
fjnoyp: Thanks!  Updated TemporalDate, TemporalDateTime, and TemporalTime to do Duration of 0 seconds if z is present.

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_date.dart
fjnoyp: Changed code to: 
`    if (_offset != null) {
      if (_offset.inSeconds == 0) {
        buffer.write("Z");
      } else {
        buffer.write(Temporal.durationToString(_offset));
      }
    }`

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_time.dart
fjnoyp: Changed code based on your minor suggestion though the changes require more complexity in the fromString and format methods in the end.

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_time.dart
fjnoyp: good call, changed the method name.

fjnoyp: Though could that confuse users if they thought they can call getDateTime and then modify the underlying TemporalDate/TemporalDatetime object as well?

fjnoyp: The way Dart DateTime works is if you call "add" or "subtract" a duration, a new DateTime object is returned instead of modifying the existing one.

fjnoyp: Do you think there should be a setDateTime method as well?

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal.dart
```diff
@@ -0,0 +1,59 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+// ignore_for_file: public_member_api_docs
```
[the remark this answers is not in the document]
fjnoyp: I removed the line : // ignore_for_file: public_member_api_docs

This file is a static helper file for the other Temporal classes and shouldn't be public ideally.

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_time.dart
richardmcclellan: > Though could that confuse users if they thought they can call getDateTime and then modify the underlying TemporalDate/TemporalDatetime object as well?

As a general rule, all objects that we return to customers, including these Temporal objects, should be immutable.    To achieve that, you'd actually want to return a copy of `_dateTime` in this method, rather than the _dateTime instance itself.

> The way Dart DateTime works is if you call "add" or "subtract" a duration, a new DateTime object is returned instead of modifying the existing one.

I do understand that - my point was that you shouldn't ever be adding/subtracting the duration from the instant (a.k.a. Dart DateTime), ever.  The instant, combined with the offset, can be used by the customer to compute hour, minute, second for display on a UI.  

> Do you think there should be a setDateTime method as well?

No, like mentioned above, all objects that we return to customers should be immutable.

fjnoyp: Sounds good, that works perfectly then.  The Dart DateTime class is inherently immutable as it has no methods that actually allow you to modify the underlying instance - thus I think no action is needed here then.  Returning the _dateTime instance variable is already returning an immutable.

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal.dart
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
haverchuck, line 2: 2021

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_date.dart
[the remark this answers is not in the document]
haverchuck: We want this to be a vanilla exception and not a datastore exception?

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_datetime.dart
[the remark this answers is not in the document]
haverchuck: Could this type also be relevant for the API category, not just Datastore? If so we could move it to amplify_core?

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_datetime.dart
```diff
@@ -0,0 +1,157 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import 'temporal.dart';
+
+/// Represents a valid extended ISO-8601 DateTime string.  The time zone offset is compulsory.
+/// YYYY-MM-DDThh:mm:ss.sssZ  (ISO_OFFSET_DATE_TIME)
+/// https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html#appsync-defined-scalars
+class TemporalDateTime {
```
haverchuck, line 21: If these types can be used across categories (say, in API.Graphql) then perhaps they should live in amplify_core? What do others think?

haverchuck (changes requested):
Left a couple of comments/questions inline.

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_datetime.dart
richardmcclellan: Ideally, `amplify_core` should be kept small, since it is a dependency of all the plugins.   Moving the `Temporal` types and other API concepts to `amplify_core` will likely result in bloat over time.  In `amplify-android`, `Temporal.java` lives in a module called `aws-api-appsync`, which is a module containing all the common code needed by `aws-api` and `aws-datastore`.  Could a similar pattern work for Flutter?

richardmcclellan (approved):
Looks good to me overall!  Just left one comment regarding whether these types should be put in the DataStore package vs somewhere else.

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_date.dart
```diff
@@ -0,0 +1,131 @@
...
+
+  /// Return offset
+  Duration getOffset() {
+    return _offset;
+  }
+
+  /// Return DateTime
+  DateTime getDateTime() {
+    return _dateTime;
+  }
```
Amplifiyer, line 105: We should rename this as `getDateTimeInUtc()` since we are manipulating what customers provided and converting the provided datetime to UTC, it's not really the same what they provided.

```dart
var myDate = DateTime.now()
var temporalDate = TemporalDate(myDate);

print(myDate == temporalDate.getDateTime()); // returns false;
```

Ref: https://api.dart.dev/stable/2.10.5/dart-core/DateTime/operator_equals.html

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_datetime.dart
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import 'temporal.dart';
+
+/// Represents a valid extended ISO-8601 DateTime string.  The time zone offset is compulsory.
```
Amplifiyer, line 18: >The time zone offset is compulsory.

Why is this a requirement? DartTime doesn't have this.

Amplifiyer (commented):
Looks good, some minor comments and improvements

packages/amplify_datastore/ios/Classes/types/temporal/FlutterTemporal.swift
```diff
@@ -0,0 +1,48 @@
...
+        return FlutterTemporal(Foundation.Date())
+    }
+    
+    // Not needed
+    var foundationDate: Date
+    
+    // Not needed
+    init(_ date: Date) {
+        self.iso8601 = ""
+        self.foundationDate = Date()
```
richardmcclellan, line 32:
```suggestion
        self.foundationDate = date
```

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_datetime.dart
fjnoyp: Moving to a separate `aws-api-appsync` module makes sense to me.  This will change our imports so it will change the codegen models as well.  
@haverchuck @Amplifiyer @Ashish-Nanda what are your thoughts on making that change?

Amplifiyer: Can't we just re-export it from here? If yes, We can hold off on doing the refactor until API category needs this and move it then to what makes most sense at that time. If it is going to be a breaking change, then let's figure it out right now.

packages/amplify_datastore/ios/Classes/types/temporal/FlutterTemporal.swift
```diff
@@ -0,0 +1,48 @@
...
+    }
```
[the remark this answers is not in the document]
fjnoyp: This constructor is not needed, only another type of constructor is.  However in order to conform to TemporalSpec, we need to implement the other constructor to get compiling code.

packages/amplify_datastore/ios/Classes/types/temporal/FlutterTemporal.swift
```diff
@@ -0,0 +1,48 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import Amplify
+
+import Foundation
+
+struct FlutterTemporal: TemporalSpec {
+    // Not needed
```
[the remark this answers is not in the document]
fjnoyp: Done

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_datetime.dart
fjnoyp: https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html#appsync-defined-scalars

> The time zone offset is compulsory for this scalar.

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_datetime.dart
fjnoyp: @Amplifiyer 

I believe it will be a breaking change because the dart file imports will change for the codegen models: 

will change to whatever new package it's in

Amplifiyer: If the new package gets exported from `amplify_datastore_plugin_interface.dart` then nothing will need to change, right?

Amplifiyer (approved):
Great work! 🚀

haverchuck (approved):

PR #314, opened 2021-01-20
chore(amplify_flutter): Error handling refactor

Amplifiyer:
*Issue #, if available:*

*Description of changes:*


packages/amplify_core/android/src/main/kotlin/com/amazonaws/amplify/amplify_core/exception/ExceptionMessages.kt
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
haverchuck, line 2: Copyright 2021?

packages/amplify_core/android/src/main/kotlin/com/amazonaws/amplify/amplify_core/exception/ExceptionMessages.kt
```diff
@@ -0,0 +1,42 @@
...
+
+class ExceptionMessages {
+    companion object {
+
+        @JvmStatic
+        val createGithubIssueString: String =
+                """
+        Please take a look at
+        https://github.com/aws-amplify/amplify-flutter/issues to see if there are any existing issues that
+        match your scenario, and file an issue with the details of the bug if there isn't.
```
haverchuck, line 26: Please take a look at https://github.com/aws-amplify/amplify-flutter/issues to see if there are any existing issues that match your scenario, and feel free to comment on any exiting issues you find. If your scenario is not described in an existing issue, please file create one with a detailed desription of the bug.

packages/amplify_core/example/.metadata
```diff
@@ -0,0 +1,10 @@
+# This file tracks properties of this Flutter project.
```
haverchuck, line 1: Do we have a concrete reason for including an example app for core?  iOS unit tests?

packages/amplify_core/example/README.md
```diff
@@ -0,0 +1,16 @@
+# amplify_core_example
+
+Demonstrates how to use the amplify_core plugin.
+
+## Getting Started
+
+This project is a starting point for a Flutter application.
```
haverchuck, line 7: We can make this match the other examples, or remove ## Getting Started entirely.

packages/amplify_core/example/ios/Podfile
```diff
@@ -0,0 +1,41 @@
+# Uncomment this line to define a global platform for your project
+# platform :ios, '9.0'
```
haverchuck, line 2: If we want this to run I think we might need to uncomment this line and set to `:ios, 11.0`

packages/amplify_core/ios/Classes/Support/ErrorMessages.swift
```diff
@@ -0,0 +1,25 @@
...
+ * permissions and limitations under the License.
+ */
+
+/// Commonly used cross-category error messages.
+public struct ErrorMessages {
+    public static let defaultFallbackErrorMessage: String =
+        """
+        Translating platform exception failed. Please take a look at \
+        https://github.com/aws-amplify/amplify-flutter/issues to see if there are any existing issues that \
+        match your scenario, and file an issue with the details of the bug if there isn't.
```
haverchuck, line 22: See matching suggestion for Kotlin file.

haverchuck: Just to confirm - we don't need `missingRecoverySuggestion` or `missingExceptionMessage`?

packages/amplify_core/ios/amplify_core.podspec
```diff
@@ -0,0 +1,23 @@
...
+  s.description      = <<-DESC
+A base package shared across Amplify Flutter library.
+                       DESC
+  s.homepage = 'https://github.com/aws-amplify/amplify-flutter'
+  s.license = 'Apache License, Version 2.0'
+  s.author = { 'Amazon Web Services' => 'amazonwebservices' }
+  s.source = { :git => 'https://github.com/aws-amplify/amplify-flutter.git' }
+  s.source_files = 'Classes/**/*'
+  s.dependency 'Flutter'
+  s.platform = :ios, '9.0'
```
haverchuck, line 18: Aren't we using a minimum of 11.0?

haverchuck (approved):
Made a couple of minor suggestions, mostly not code-related. Otherwise looks great! I'm approving assuming you'll look at the suggestions and either make them or explicitly decline.

packages/amplify_core/android/src/main/kotlin/com/amazonaws/amplify/amplify_core/exception/ExceptionMessages.kt
Amplifiyer: I matched it with https://github.com/aws-amplify/amplify-ios/blob/e2f7f64e2a501f3b2807481ca921e71710c82b22/Amplify/Core/Support/AmplifyErrorMessages.swift#L14 to avoid a lot of different message strings

packages/amplify_core/example/.metadata
Amplifiyer: As discussed offline, it is needed for iOS unit tests.

packages/amplify_core/example/README.md
Amplifiyer: Removed the getting started section.

packages/amplify_core/example/ios/Podfile
Amplifiyer: This example is unused.

packages/amplify_core/ios/amplify_core.podspec
Amplifiyer: in your app, Cocoapods (?) will resolve to the max of the min dependency. This merely says that this package doesn't need anything more than 9.0 so other packages at 9.0 can use it as well.

packages/amplify_core/ios/Classes/Support/ErrorMessages.swift
Amplifiyer: Good point, will add.

packages/amplify_core/android/src/main/kotlin/com/amazonaws/amplify/amplify_core/exception/ExceptionMessages.kt
Amplifiyer: Done

fjnoyp:
Why do we have "ErrorUtil.swift" on iOS but "ExceptionUtil.kt" on Android.  Furthermore within ExceptionUtil we have a method called "createSerializedError".

Can we consistently use either Error or Exception between both iOS and Android?

Amplifiyer:
> Why do we have "ErrorUtil.swift" on iOS but "ExceptionUtil.kt" on Android. 

This was so I keep consistent with the platforms (android call them Exception and iOS call them Error)

> Furthermore within ExceptionUtil we have a method called "createSerializedError".

This is because I couldn't figure out a way to centralize/generalize the error serialization to put this method in a common util in iOS.

Evidently, to be consistent with my first answer, it should be called `createSerializedException` in android.

fjnoyp (commented):
Looks good, just that question about naming between iOS / Android

.circleci/config.yml
```diff
@@ -122,6 +123,7 @@ jobs:
           name: Build example APKs
           command: |
             melos run build:examples:android
+          no_output_timeout: 20m
```
haverchuck, line 126: Nice. Just out of curiousity - any particular reason we arrived at this number?

fjnoyp (approved):

.circleci/config.yml
Amplifiyer: No particular reason. Some of the android tasks are taking upwards of 10 mins which kills the job and doesn't even tell us which one. This will at least complete the job and in the logs we can tell which ones are taking longer to troubleshoot them separately.

haverchuck (approved):

PR #353, opened 2021-02-05
Enable Sign In With Apple for IOS(Native flow) and Android(Web flow) devices 

omchiii:
*Issue #, if available:*

Unable to Sign in/Sign up with Apple.

*Description of changes:*

1. Added a new case for AuthProvider enum.
2. Handled the case in swift.
3. Added "apple.com" as an allowed provider.

I was able to successfully login using Apple iD. I just followed these steps : https://docs.amplify.aws/lib/auth/social_signin_web_ui/q/platform/ios#setup-your-auth-provider 


omchiii:
@Amplifiyer can you review please?

Amplifiyer:
@omchiii We are currently testing this change, will merge this post testing. CC: @haverchuck

omchiii:
@Amplifiyer please check out my latest commit, which enables Sign in with Apple for Android as well. What is great about this is when I used an android device it uses the web auth flow for Apple sign in but on a IOS device it uses the native one. And they both work great.

haverchuck:
@omchiii Thanks for posting this! I am in the process of testing, but I am fairly confident this should work.

talamaska:
I'm just curious, what's the value of adding apple signin in android?

omchiii:
> I'm just curious, what's the value of adding apple signin in android?

You would be able to sign in through your Apple account. Not the native way with the prompt asking your phone password or face id, but the web way. Just like Google sign in. This feature is available in the Native Android Amplify library, so it would not hurt having it here as well.

omchiii:
@haverchuck Any updates?

haverchuck:
@omchiii I have pulled this down, tested, and rebased. Works great!  I encoutered some issues trying to push to your branch after rebasing (and making another necessary change for hosted ui), so I am pushing this branch to my fork.  We will list you as a co-author on that commit.

Issue #157, opened 2020-10-17
PlatformException - Cognito


arealguru:
**Bug Description**
I'm trying to set up just the `AmplifyAuthCognito()` plugin, but I'm having difficulty as a PlatformException (most likely a bad `amplifyconfiguration.dart` was keeping me from getting anywhere. I looked at issues #100 and #153 (#100 was much closer to my situation), and I wasn't able to replicate the fixes from those configurations in mine.

**To Reproduce**
I used the Amplify CLI and the tutorial and I currently have Auth set up; `amplify status` outputs 

```
| Category | Resource name     | Operation | Provider plugin   |
| -------- | ----------------- | --------- | ----------------- |
| Auth     | <value> | No Change | awscloudformation |
```

This is what my config looks like within `amplifyconfiguration.dart`.

```
const amplifyconfig = ''' {
    "UserAgent": "aws-amplify-cli/2.0",
    "Version": "1.0",
    "auth": {
        "plugins": {
            "awsCognitoAuthPlugin": {
                "UserAgent": "aws-amplify-cli/0.1.0",
                "Version": "0.1.0",
                "IdentityManager": {
                    "Default": {}
                },
                "CredentialsProvider": {
                    "CognitoIdentity": {
                        "Default": {
                            "PoolId": "<value>",
                            "Region": "us-east-2"
                        }
                    }
                },
                "CognitoUserPool": {
                    "Default": {
                        "PoolId": "us-east-2_<value>",
                        "AppClientId": "<value>",
                        "AppClientSecret": "<value>",
                        "Region": "us-east-2"
                    }
                },
                "Auth": {
                    "Default": {
                        "authenticationFlowType": "USER_SRP_AUTH"
                    }
                }
            }
        }
    }
}''';
```

I'm then running the following test code:

```
class _MyHomePageState extends State<MyHomePage> {
  bool _amplifyConfigured = false;
  Amplify amplifyInstance = Amplify();

  @override
  initState() {
    super.initState(); 
    _configureAmplify();
  }

  void _configureAmplify() async {
    AmplifyAuthCognito authPlugin = AmplifyAuthCognito();
    amplifyInstance.addPlugin(
      authPlugins: [authPlugin]
    );

    // Once Plugins are added, configure Amplify
    await amplifyInstance.configure(amplifyconfig);
    try {
      setState(() {
        _amplifyConfigured = true;
      });
    } catch (e) {
      print(e);
    }
  }

  @override
[...]

Issue #153, opened 2020-10-13
PlatformException(AmplifyException, Failed to Configure Amplify, The operation couldn’t be completed.


cvalab:
PlatformException(AmplifyException, Failed to Configure Amplify, The operation couldn’t be completed. (Amplify.PluginError error 2.), null)
[VERBOSE-2:profiler_metrics_ios.mm(184)] Error retrieving thread information: (ipc/send) invalid destination port
Fatal error: Analytics category is not configured. Call Amplify.configure() before using any methods on the category.: file /Users/av/StudioProjects/dietian/flutter-customer-application/flutter-customer-application/ios/Pods/Amplify/Amplify/Categories/Analytics/AnalyticsCategory.swift, line 18
Lost connection to device.

[✓] Flutter (Channel beta, 1.22.0-12.2.pre, on Mac OS X 10.15.3 19D76, locale ru-UA)

  Future<bool> initAwsAmplify(String amplifyconfig) async {
    final AmplifyAuthCognito auth = AmplifyAuthCognito();
    //AmplifyStorageS3 storage = AmplifyStorageS3();
    final AmplifyAnalyticsPinpoint analytics = AmplifyAnalyticsPinpoint();

    await amplify.addPlugin(authPlugins: [auth],
        //storagePlugins: [storage],
        analyticsPlugins: [analytics]);
    // Initialize AmplifyFlutter
    await amplify.configure(amplifyconfig);
    return true;
  }

  amplify_core:
    git:
      url: https://github.com/aws-amplify/amplify-flutter.git
      #ref: fix-auth-cognito-podfile #master
      ref: master
      path: packages/amplify_core/

  amplify_auth_cognito:
    git:
      url: https://github.com/aws-amplify/amplify-flutter.git
      #ref: fix-auth-cognito-podfile #master
      ref: master
      path: packages/amplify_auth_cognito

  amplify_analytics_pinpoint:
    git:
      url: https://github.com/aws-amplify/amplify-flutter.git
      #ref: fix-auth-cognito-podfile #master
      ref: master
      path: packages/amplify_analytics_pinpoint

Cognito working fine.

cvalab:
<img width="802" alt="Снимок экрана 2020-10-13 в 07 46 16" src="https://user-images.githubusercontent.com/68106574/95816712-6b00ce80-0d28-11eb-8fb9-8c5b29ce8c9c.png">

haverchuck:
[...]

PR #355, opened 2021-02-05
fix(amplify_flutter): Provide better error messages when configure() fails due to plugin's configuration not present

Amplifiyer:
fix(amplify_flutter): Provide better error messages when configure() fails due to plugin's configuration not present

This is only a temporary patch to display correct error messages. A solution that doesn't require customers to maintain strict pubspec dependencies will be coming later.

issues: #157 , #153


fjnoyp (approved):

haverchuck (approved):
LGTM

Issue #305, opened 2021-01-10
SignUp/SignIn Fails in Release build apk: com.amazonaws.amplify/auth_cognito: Failed to handle method call g.i: null cannot be cast to non-null type java.lang.reflect.Method


uttusharma:
**Bug**
Amplify fails to signup/sign-in on release build apk. 

**To Reproduce**
1. Configure Amplify like this
```
AmplifyAuthCognito auth = new AmplifyAuthCognito();
    amplify.addPlugin(authPlugins: [auth]);
    try {
      await amplify.configure(amplifyconfig);
      print('Successfully configured Amplify 🎉');
    } catch (e) {
      print('Could not configure Amplify ☠️');
      print(e);
    }
```
2. Code to execute 
```
final SignUpResult res = await Amplify.Auth.signUp(
            username: event.email,
            password: event.pass,
            options: CognitoSignUpOptions(userAttributes: userAttributes));
 ```

3. Now run ```flutter build apk --release``` or Generate Signed Release build apk.  **Note:** Don't run ``` flutter run --release``` as it works fine with that.
4. Now run the release apk on the device and in the adb logcat you'll see 
```
 2021-01-10 23:51:24.645 29270-29270/?  E/MethodChannel#com.amazonaws.amplify/auth_cognito: Failed to handle method call
    g.i: null cannot be cast to non-null type java.lang.reflect.Method
        at com.amazonaws.amplify.amplify_auth_cognito.types.FlutterSignUpRequest.b(Unknown Source:203)
        at com.amazonaws.amplify.amplify_auth_cognito.types.FlutterSignUpRequest.<init>(Unknown Source:90)
        at com.amazonaws.amplify.amplify_auth_cognito.AuthCognito.w(Unknown Source:10)
        at com.amazonaws.amplify.amplify_auth_cognito.AuthCognito.h(Unknown Source:150)
        at f.a.d.a.j$a.a(Unknown Source:17)
        at io.flutter.embedding.engine.e.b.e(Unknown Source:57)
        at io.flutter.embedding.engine.FlutterJNI.handlePlatformMessage(Unknown Source:4)
        at android.os.MessageQueue.nativePollOnce(Native Method)
        at android.os.MessageQueue.next(MessageQueue.java:325)
        at android.os.Looper.loop(Looper.java:170)
        at android.app.ActivityThread.main(ActivityThread.java:7055)
        at java.lang.reflect.Method.invoke(Native Method)
[...]

PR #362, opened 2021-02-08
fix(amplify_auth_cognito): force user attribute to be string

haverchuck:
*Description of changes:*
Changes interface to only allows Strings for user attribute on signup. This accomodates amplify-android signUp API, and avoids the need to either expand our API surface or do extensive parsing on platform side. Custom attributes that are 'Numbers' will still be created as such, and will still be parsed as such when customers request them.

### DIFF
--- a/packages/amplify_auth_plugin_interface/lib/src/SignUp/SignUpOptions.dart
+++ b/packages/amplify_auth_plugin_interface/lib/src/SignUp/SignUpOptions.dart
@@ -16,7 +16,7 @@
 import 'package:flutter/foundation.dart';
 
 class SignUpOptions {
-  final Map<String, dynamic> userAttributes;
+  final Map<String, String> userAttributes;
   const SignUpOptions({@required this.userAttributes});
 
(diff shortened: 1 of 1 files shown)


Amplifiyer:
So customers need to do parsing for numbers?

haverchuck:
@Amplifiyer Confirmed that this matches [amplify-android](https://github.com/aws-amplify/amplify-android/blob/6699348354feda0c9749b3c0155a51626ce6defd/core/src/main/java/com/amplifyframework/auth/AuthUserAttribute.java#L32) and [amplify-ios](https://github.com/aws-amplify/amplify-ios/blob/e2f7f64e2a501f3b2807481ca921e71710c82b22/Amplify/Categories/Auth/Models/AuthUserAttribute.swift#L12).

Amplifiyer (approved):

fjnoyp (approved):

haverchuck:
This PR fixes: https://github.com/aws-amplify/amplify-flutter/issues/305

PR #366, opened 2021-02-09
fix(amplify_auth_cognito): accounts for zoneinfo camelcase in Android

haverchuck:
*Description of changes:*
Accounts for amplify-android using camel case version of 'zoneinfo'.

### DIFF
--- a/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterSignUpRequest.kt
+++ b/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterSignUpRequest.kt
@@ -46,7 +46,11 @@ data class FlutterSignUpRequest(val map: HashMap<String, *>) {
                 }
                 authUserAttributes.add(AuthUserAttribute(AuthUserAttributeKey.custom(keyCopy), value))
             } else {
-                var t: Method = attributeMethods.asIterable().find { it.name.equals(convertSnakeToCamel(key)) } as Method;
+                // special case for amplify-android camelCase zoneinfo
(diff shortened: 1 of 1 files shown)


packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterSignUpRequest.kt
```diff
@@ -46,7 +46,11 @@ data class FlutterSignUpRequest(val map: HashMap<String, *>) {
                 }
                 authUserAttributes.add(AuthUserAttribute(AuthUserAttributeKey.custom(keyCopy), value))
             } else {
-                var t: Method = attributeMethods.asIterable().find { it.name.equals(convertSnakeToCamel(key)) } as Method;
+                // special case for amplify-android camelCase zoneinfo
+                if (key == "zoneinfo") {
+                    keyCopy = "zoneInfo"
+                }
+                var t: Method = attributeMethods.asIterable().find { it.name.equals(convertSnakeToCamel(keyCopy)) } as Method;
```
Amplifiyer, line 53: This went from finding for any key in all of userAttributes, to finding `zoneInfo` or `key` which are the only two possible values in this else block. Is this accurate? Maybe it needs unit test

Jordan-Nelson:
@haverchuck I think we can close this. This should not be an issue anymore since it is mapped to the static function
https://github.com/aws-amplify/amplify-flutter/blob/4337cd19fa62b57980e44966885ccd4fff9aa765/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/utils/UserAttributeDeserialization.kt#L44

Issue #327, opened 2021-01-26
[amplify_auth_cognito][Android]Smart cast to 'String' is impossible, because 'req.provider' is a complex expression


WingCH:
When i update to `^0.0.2-dev.1`, has the following error, only in android

```
e: /Users/wingch/.pub-cache/hosted/pub.dartlang.org/amplify_auth_cognito-0.0.2-dev.1/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt: (397, 56): Smart cast to 'String' is impossible, because 'req.provider' is a complex expression

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':amplify_auth_cognito:compileDebugKotlin'.
> Compilation error. See log for more details

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 3s
```
```
amplify_flutter: ^0.0.2-dev.1
amplify_auth_cognito: ^0.0.2-dev.1
amplify_storage_s3: ^0.0.2-dev.1
```

```
[✓] Flutter (Channel stable, 1.22.6, on macOS 11.1 20C69 darwin-x64, locale zh-Hant-HK)
    • Flutter version 1.22.6 at /Users/wingch/fvm/versions/stable
    • Framework revision 9b2d32b605 (3 days ago), 2021-01-22 14:36:39 -0800
    • Engine revision 2f0af37152
    • Dart version 2.10.5

[✓] Android toolchain - develop for Android devices (Android SDK version 29.0.3)
    • Android SDK at /Users/wingch/Library/Android/sdk
    • Platform android-30, build-tools 29.0.3
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b3-6915495)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 12.3)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 12.3, Build version 12C33
    • CocoaPods version 1.9.3

[!] Android Studio (version 4.1)
    • Android Studio at /Applications/Android Studio.app/Contents
    ✗ Flutter plugin not installed; this adds Flutter specific functionality.
[...]

PR #381, opened 2021-02-12
fix(amplify_datastore): fix compile issue with newer gradle plugin

Amplifiyer:
*Issue #, if available:* #327

*Description of changes:*

### DIFF
--- a/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
+++ b/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
@@ -421,7 +421,7 @@ public class AuthCognito : FlutterPlugin, ActivityAware, MethodCallHandler, Plug
         } else {
           mainActivity.let {
             Amplify.Auth.signInWithSocialWebUI(
-                    AuthProvider::class.java.getMethod(req.provider).invoke(null) as AuthProvider,
+                    AuthProvider::class.java.getMethod(req.provider!!).invoke(null) as AuthProvider,
                     it,
                     { result ->
                       if (!resultSubmitted) {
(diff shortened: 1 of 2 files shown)


packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
```diff
@@ -421,7 +421,7 @@ public class AuthCognito : FlutterPlugin, ActivityAware, MethodCallHandler, Plug
         } else {
           mainActivity.let {
             Amplify.Auth.signInWithSocialWebUI(
-                    AuthProvider::class.java.getMethod(req.provider).invoke(null) as AuthProvider,
+                    AuthProvider::class.java.getMethod(req.provider!!).invoke(null) as AuthProvider,
```
jamesonwilliams, line 424: Huh! This field is [declared](https://github.com/aws-amplify/amplify-flutter/blob/c86f423fa1a72c805968bb395436bf9c9f10c379/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterSignInWithWebUIRequest.kt#L21):
```kotlin
 val provider: String? = map?.get("authProvider") as String?;
```
What are the implications of `provider` being `null`, there? Will it ever reach this `!!`?

Amplifiyer: No, the provider cannot be null! It's the else branch of `if(req.provider == null)`. Really weird that it's a compile issue while targeting newer sdk level(29).

haverchuck (approved):

Ashish-Nanda (approved):

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
jamesonwilliams: Oh whoops, I missed the `else`, sorry. :-D Proceed!

jamesonwilliams (approved):

PR #382, opened 2021-02-12
fix(amplify_auth_cognito): globalSignout fix

haverchuck:
*Description of changes:*
Fixes globalsignout on Android.


packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthCognitoPluginTest.kt
```diff
@@ -205,6 +208,23 @@ class AmplifyAuthCognitoPluginTest {
...
+                    "globalSignOut" to true
+            )
+        )
+
+        // Act
+        val request = FlutterSignOutRequest(data)
+
+        // Assert
+        assert(request.signOutOptions is AuthSignOutOptions)
+        assert(request.signOutOptions.isGlobalSignOut)
```
jamesonwilliams, line 225: These look like the Kotlin stdlib [`assert`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/assert.html), instead of a jUnit `assertFoo`-style function. Was this your intention? Does the code base currently use `assert`?

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterSignOutRequest.kt
```diff
@@ -18,12 +18,12 @@ package com.amazonaws.amplify.amplify_auth_cognito.types
...
 data class FlutterSignOutRequest(val map: HashMap<String, *>) {
-  var signOutOptions: AuthSignOutOptions = setOptions()
   val options: HashMap<String, *>? = map["options"] as HashMap<String, *>?;
+  var signOutOptions: AuthSignOutOptions = setOptions()
 
-  fun setOptions(): AuthSignOutOptions {
+  private fun setOptions(): AuthSignOutOptions {
     var globalSignOut: Boolean = false;
-    if (options != null && options.get("globalSignOut") != null) {
+    if (options?.get("globalSignOut") != null) {
```
Amplifiyer, line 26: what's the fix here?

jamesonwilliams (approved):
Question about `assert`, otherwise looks good.

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterSignOutRequest.kt
Amplifiyer: Nvmd, the fix is elsewhere.

PR #383, opened 2021-02-12
fix(amplify_analytics): fix analyticsProperties

fjnoyp:
*Issue #, if available:*

*Description of changes:*

Fix sending of AnalyticsProperties in iOS 

Before iOS was converting Dart maps representing AnalyticsProperties to blank maps every time. 

### DIFF
--- a/packages/amplify_analytics_pinpoint/android/src/main/kotlin/com/amazonaws/amplify/amplify_analytics_pinpoint/AmplifyAnalyticsBuilder.kt
+++ b/packages/amplify_analytics_pinpoint/android/src/main/kotlin/com/amazonaws/amplify/amplify_analytics_pinpoint/AmplifyAnalyticsBuilder.kt
@@ -98,9 +98,12 @@ class AmplifyAnalyticsBuilder {
                         val locationMap = item.value as HashMap<String, String>
                         userProfileBuilder.location(createUserLocation(locationMap))
                     }
-                    "properties" -> {
+                    "propertiesMap" -> {
                         val propertiesMap = item.value as HashMap<String, Any>
                         userProfileBuilder.customProperties(createAnalyticsProperties(propertiesMap))
+                    }
+                    "propertiesTypesMap" -> {
+                        // Ignore, we can infer types just using propertiesMap
                     }
                     // This case should not be possible as UserProfile is typed on Dart side
                     else -> {
--- a/packages/amplify_analytics_pinpoint/ios/Classes/AmplifyAnalyticsBuilder.swift
+++ b/packages/amplify_analytics_pinpoint/ios/Classes/AmplifyAnalyticsBuilder.swift
@@ -18,8 +18,27 @@ import AmplifyPlugins
 
 public class AmplifyAnalyticsBuilder {
 
-    public static func createAnalyticsProperties(propertiesMap : Dictionary<String, Any>) -> AnalyticsProperties{
-        return propertiesMap as? AnalyticsProperties ?? [:]
+    public static func createAnalyticsProperties(map : [String: Any]) -> AnalyticsProperties{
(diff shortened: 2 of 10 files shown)


packages/amplify_analytics_pinpoint/android/src/main/kotlin/com/amazonaws/amplify/amplify_analytics_pinpoint/AmplifyAnalyticsBuilder.kt
```diff
@@ -98,10 +98,13 @@ class AmplifyAnalyticsBuilder {
...
                         userProfileBuilder.location(createUserLocation(locationMap))
                     }
-                    "properties" -> {
+                    "propertiesMap" -> {
                         val propertiesMap = item.value as HashMap<String, Any>
                         userProfileBuilder.customProperties(createAnalyticsProperties(propertiesMap))
                     }
+                    "propertiesTypesMap" -> {
+                        // Ignore, we can infer types just using propertiesMap
+                    }
```
Amplifiyer, line 107: probably a good idea to do it anyways? I've seen in DataStore `1` being translated to `1.0` after traveling through method channel serialization. If that didn't happen in your testing, then we can keep it as is.

packages/amplify_analytics_pinpoint/ios/Classes/AmplifyAnalyticsBuilder.swift
```diff
@@ -18,8 +18,27 @@ import AmplifyPlugins
...
+                    analyticsProperties[key] = propertiesMap[key] as! String
+                case "INT":
+                    analyticsProperties[key] = propertiesMap[key] as! Int
+                case "DOUBLE":
+                    analyticsProperties[key] = propertiesMap[key] as! Double
+                case "BOOL":
+                    analyticsProperties[key] = propertiesMap[key] as! Bool
+                default:
+                    print("Unknown type for AnalyticsProperties")
+            }
```
Amplifiyer, line 39: If we can avoid force unwrap, we should do that

packages/amplify_analytics_pinpoint/android/src/main/kotlin/com/amazonaws/amplify/amplify_analytics_pinpoint/AmplifyAnalyticsBuilder.kt
fjnoyp: hmm what are you referring to?  We ignore this case as it's already handled for the propertiesMap case.  

I've verified that for Android a double property sent as 10.0 and an int property sent as 10 are saved as their proper types.

fjnoyp: iOS was a bit trickier hence the extra data structure propertiesTypesMap being sent from Dart

Amplifiyer: I was referring to inferring the types in android. There were some issues in DataStore when `int` were being converted to `double` by GSON library. If that's not the case here, then we can ignore it.

packages/amplify_analytics_pinpoint/ios/Classes/AmplifyAnalyticsBuilder.swift
fjnoyp: propertiesTypesMap needs to be force unwrapped as it's used for the foreach.  But I can use ? for String/Int/Double/Bool casts.

Amplifiyer (approved):

Issue #367, opened 2021-02-09
Documentation all APIs and exceptions


parikh20:
Is there a place where all of the exception names and details are cleanly laid out for in app handling? Having to manually check for each possible type just to get the name of the exception is getting quite exhausting.

Amplifiyer:
@parikh20, we are completely rewriting the error handling such that the errors received from the APIs will be consistent and easy to inspect. We will update our documentation after the release.

Jordan-Nelson:
Hi @parikh20 - The documentation on pub has a section for Exceptions. For example, you can reference the "Exceptions / Errors" section [here](https://pub.dev/documentation/amplify_flutter/latest/amplify_flutter/amplify_flutter-library.html) for amplify_flutter or [here](https://pub.dev/documentation/amplify_auth_cognito/latest/amplify_auth_cognito/amplify_auth_cognito-library.html) for amplify_auth_cognito.

I know this request is quite old and those Exceptions may not have been documented when you opened the issue. If there is something more you are looking for, please let me know.

Jordan-Nelson:
Closing this issue out as completed. Please let us know if the links shared above are not what you are looking for.

PR #397, opened 2021-02-18
docs: Adds documentation for public classes of auth category

haverchuck:
*Issue*
#367 

*Description of changes:*
Adds documentation for Auth category + Cognito plugin.


fjnoyp (approved):

packages/amplify_auth_cognito/lib/amplify_auth_cognito.dart
```diff
@@ -34,7 +35,7 @@ class AmplifyAuthCognito extends AuthPluginInterface {
   static AmplifyAuthCognito _instance = AmplifyAuthCognitoMethodChannel();
   static AuthStreamController streamWrapper = AuthStreamController();
 
-
+  /// Sets the instace of the AmplifyAuthCognito plugin to use a defined MethodChannel
```
fjnoyp, line 38: typo instance

packages/amplify_auth_cognito/lib/src/CognitoPasswords/CognitoResetPasswordResult.dart
```diff
@@ -15,6 +15,8 @@
 
 import 'package:amplify_auth_plugin_interface/amplify_auth_plugin_interface.dart';
 
+/// Cognito extension of ResetPasswordResult
 class CognitoResetPasswordResult extends ResetPasswordResult {
+  /// Default constructor for CognitoResetPasswordResult
   CognitoResetPasswordResult({nextStep, isPasswordReset}) : super(isPasswordReset: isPasswordReset, nextStep: nextStep);
```
Amplifiyer, line 21: I don't think there is an `@inherit` annotation to automatically copy the super class's doc. If not, we should just copy paste it here. When customers are using `CognitoResetPasswordResult`, they won't get any help what that is until they reach `ResetPasswordResult` e.g. in IDE.

packages/amplify_auth_cognito/lib/src/CognitoSession/AWSCognitoUserPoolTokens.dart
```diff
@@ -18,10 +18,21 @@ import 'dart:io' show Platform;
 import 'package:amplify_auth_plugin_interface/amplify_auth_plugin_interface.dart';
 import 'package:flutter/material.dart';
 
+/// Class to wrap AWS Cognito UserPool tokens
```
Amplifiyer, line 21: One way we can improve docs is instead of _explaining_ what the class/variable/method is from the library architecture perspective, talk about what it is from customer's use case perspective or how to use it or to document the behavior.

E,g. here the doc could be something like

```suggestion
import 'package:amplify_auth_plugin_interface/amplify_auth_plugin_interface.dart';
import 'package:flutter/material.dart';

/// Holds cognito user pool specific auth tokens. 
///
/// Returned from `Auth.fetchAuthSession()` API in the `CognitoAuthSession` object.
/// Todo: Use it to... or Use it as ... etc. if applicable.
```

packages/amplify_flutter/lib/categories/amplify_auth_category.dart
```diff
@@ -115,27 +183,47 @@ class AuthCategory {
...
+  /// Other AuthException subclasses are possible as a result of network issues,
+  /// misconfiguration, internal service errors or other less common reasons.
   Future<List<AuthUserAttribute>> fetchUserAttributes({AuthUserAttributeOptions options}) {
     var request = AuthUserAttributeRequest(options: options);
     return plugins.length == 1
         ? plugins[0].fetchUserAttributes(request: request)
         : throw _pluginNotAddedException("Auth");
   }
 
+  /// Returns [AuthSession] or a plugin-specific subclass.
```
Amplifiyer, line 210: How do we best convey that customers **need** to cast it to a plugin specific class else it won't work? I think it's important since I've seen users struggling to figure out how to access the tokens. Same goes for other APIs as well where the cast is mandatory.

packages/amplify_flutter/lib/categories/amplify_auth_category.dart
```diff
@@ -115,27 +183,47 @@ class AuthCategory {
...
+  ///
+  /// Throws an [AuthException], or possibly a subclass such as [SessionExpiredException] or [SignedOutException].
+  /// Other AuthException subclasses are possible as a result of network issues,
+  /// misconfiguration, internal service errors or other less common reasons.
   Future<AuthSession> fetchAuthSession({AuthSessionOptions options}) {
```
Amplifiyer, line 215: For all the APIs where we know `options` are used for some specific use cases, we should add them in their respective docs. `Options` in general is the most ambiguous parameter we have and is not really easily understood where and why one should use one.

Amplifiyer (commented):
It's a great start, thanks for volunteering to pick this up. I think one way we can improve docs here is instead of explaining what the class/variable/method is from the library's architecture perspective, talk about what it is from customer's use case perspective or how to use it or use docs to document the API behavior. These docs should be most useful from end user's perspective.

packages/amplify_auth_cognito/lib/amplify_auth_cognito.dart
haverchuck: fixed.

packages/amplify_auth_cognito/lib/amplify_auth_cognito.dart
```diff
@@ -25,6 +25,7 @@ import './amplify_auth_cognito_stream_controller.dart';
 export './src/types.dart';
 export 'package:amplify_auth_plugin_interface/src/types.dart';
 
+/// Constructs an AmplifyAuthCognito plugin
```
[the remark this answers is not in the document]
haverchuck: Made this change.

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
```diff
@@ -19,7 +19,6 @@ import 'package:amplify_auth_plugin_interface/amplify_auth_plugin_interface.dart
 import 'amplify_auth_cognito.dart';
 import 'amplify_auth_error_handling.dart';
 
-// ignore_for_file: public_member_api_docs
 const MethodChannel _channel = MethodChannel('com.amazonaws.amplify/auth_cognito');
 
 /// An implementation of [AmplifyPlatform] that uses method channels.
```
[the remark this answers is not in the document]
haverchuck: Fixed.

packages/amplify_auth_cognito/lib/amplify_auth_cognito_stream_controller.dart
```diff
@@ -17,10 +17,14 @@ import 'dart:async';
...
 import 'src/CognitoHubEvents/AuthHubEvent.dart';
 
+/// An event channel for receiving HUB events over a PlatformChannel.
 EventChannel channel = const EventChannel("com.amazonaws.amplify/auth_cognito_events");
+/// A subscription established for the event stream.
 StreamSubscription eventStream;
 
+/// Provides a StreamController for HUB
 class AuthStreamController {
+  /// The getter for the _authStreamController
```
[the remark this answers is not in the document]
haverchuck: The AuthStreamController class is used externally by amplify_flutter, so I don't believe it can be private - but I have noted in the docs that it is 'for internal usage'.  The channel and eventStream I've marked as private.

packages/amplify_auth_cognito/lib/src/CognitoHubEvents/AuthHubEvent.dart
```diff
@@ -16,6 +16,8 @@
 import 'package:amplify_core/types/hub/HubEvent.dart';
 import 'package:amplify_core/types/hub/HubEventPayload.dart';
 
+/// The Auth HubEvent subclass
 class AuthHubEvent extends HubEvent {
+    /// Default constructor for AuthHubEvent
     AuthHubEvent(String eventName, {HubEventPayload payload}) : super(eventName, payload: payload);
```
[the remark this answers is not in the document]
haverchuck: I'll be adding `// ignore: public_member_api_docs` for these constructors.

packages/amplify_auth_cognito/lib/src/CognitoPasswords/CognitoResetPasswordResult.dart
haverchuck: For this particular example, I'm doing this:
```
/// Wraps the result of a Cognito reset password operation,
/// which will include a [isPasswordReset] flag and a [nextStep] detailing the 
/// next step in the password reset process (if any). 
```
This follows the do's and dont's of flutter documentation, which suggests description of attributes and/or parameters with square brackets rather than "@param" style documentation (I'm assuming this preference has something to do with the lack of overloads and a decreased need to itemize parameters.

packages/amplify_auth_cognito/lib/src/CognitoSession/AWSCognitoUserPoolTokens.dart
haverchuck: Added this on my end.

haverchuck: Also I think this is a good opportunity to explain how our fetchAuthSession API works in `CognitoAuthSession.dart`:
```
/// A Cognito extension of AuthSession
/// 
/// Instantiated when fetchAuthSession is called with [getAWSCredentials] set to true.
/// If a Cognito Identity Pool is configured with unauthenticated access enabled, the 
/// [credentials] and [identityId] will be populated if fetchAuthSession is called
///  prior to user signin. 
/// If fetchAuthSession is called and an Identity Pool is not 
/// configured or is not configured with unauthenticated access enabled, then a
/// [SignedOutException] will be thrown.
```

packages/amplify_flutter/lib/categories/amplify_auth_category.dart
haverchuck: It's tricky because I don't want to go into too much plugin-specific detail here since this documentation should broadly define the category behavior.

packages/amplify_flutter/lib/categories/amplify_auth_category.dart
haverchuck: Agree with this.  I've added documentation for options in the SignUp API, and will add more once I fix the validationData/clientMetadata and get it into our official documentation (also once we figure out global signout).

viveky259259:
@haverchuck please resolve these conflicts, so it can merged into master

Issue #422, opened 2021-03-03
DataStore query fail on iOS


hyobbb:
**Describe the bug**
`Query failed: DataStoreException(message: The operation couldn’t be completed. (SQLite.Result error 0.), recoverySuggestion: The operation couldn’t be completed. (SQLite.Result error 0.), underlyingException: The operation couldn’t be completed. (SQLite.Result error 0.))`

query relational model causes error (query comments by postId for example)
on Android it doesn't throw error.

Also, HotRestart on iOS crash app because of AmplifyAlreadyConfiguredException although it should be caught by try&catch.

```
[VERBOSE-2:ui_dart_state.cc(186)] Unhandled Exception: AmplifyException(message: An unknown exception has happened.           Please take a look at 
        https://github.com/aws-amplify/amplify-flutter/issues to see if there are any existing issues that 
        match your scenario, and file an issue with the details of the bug if there isn't.
  , recoverySuggestion: We currently don't have a recovery suggestion for this exception., underlyingException: PlatformException(AmplifyAlreadyConfigured, Translating platform exception failed. Please take a look at https://github.com/aws-amplify/amplify-flutter/issues to see if there are any existing issues that match your scenario, and file an issue with the details of the bug if there isn't., {recoverySuggestion: Remove the duplicate call to `Amplify.configure()`, underlyingException: nil, message: Amplify has already been configured.}, null))
#0      AmplifyClass.configure (package:amplify_flutter/amplify.dart:187:9)
<asynchronous suspension>
#1      M<…>
```

fjnoyp:
Hi @hyobbb your issue sounds similar to #415 

Have you changed or modified your schema at all?

hyobbb:
@fjnoyp Hi. Yes I did change schema like below

```
enum TagStatus {
  ACTIVE
  INACTIVE
}

type Tag @model {
  id: ID!
  text: String!
  rating: Int!
  status: TagStatus!
  # New field with @connection
  videos: [Video] @connection(keyName: "byTag", fields: ["id"])
}

# New model
type Video @model
@key(name: "byTag", fields: ["tagID", "contentUrl"]) {
  id: ID!
  tagID: ID!
  tag: Tag! @connection(fields: ["tagID"])
  contentUrl: String!
  thumbUrl : String!
}
```

Also, will it support new null safety feature of flutter 2.0 soon?

fjnoyp:
[...]

Issue #415, opened 2021-03-01
After changing the schema, deserializeException occurs in iOS.


Arosf:
**Describe the bug**
I added a name to User as follows, generated code, did an amplify push, and built the iOS app, but an exception occurred. 
The android app did not throw any exceptions. 

```graphql
type User @model {
  id: ID!
  name: String <-- new
}
```

An exception was thrown when the following code was executed. 

```dart
final user = await datastore.query(
  Amp.User.classType,
  where: Amp.User.ID.eq(userId),
);
```

The error message displayed is as follows. 

```
DataStoreException (DataStoreException(message: The operation couldn’t be completed. (SQLite.Result error 0.), recoverySuggestion: The operation couldn’t be completed. (SQLite.Result error 0.), underlyingException: The operation couldn’t be completed. (SQLite.Result error 0.)))
```

Stack trace when an exception is thrown.

```
#0      AmplifyDataStoreMethodChannel.query
package:amplify_datastore/method_channel_datastore.dart:73
<asynchronous suspension>
#1      AmplifyDataStore.query
package:amplify_datastore/amplify_datastore.dart:71
#2      DataStoreCategory.query
package:amplify_flutter/categories/amplify_datastore_category.dart:52
#3      UserRepository.fetch
package:flutter_template/…/user/user_repository.dart:14
<asynchronous suspension>
#4      _UserPageState._checkAndRegistUser
package:flutter_template/main.dart:262
#5      _UserPageState.initState

```

**Platform**
Amplify Flutter current supports iOS and Android.  This issue is reproducible in (check all that apply):
[] Android
[x] iOS

**Smartphone (please complete the following information):**
 - Device: [e.g. iPhone12 Pro Max (ios simulator)]
 - OS: [e.g. iOS14.4]
 - Version [flutter 1.22.6]

haverchuck:
@Arosf Which version of amplify_datastore are you using?

fjnoyp:
Thanks @Arosf we are able to reproduce this issue on our side and are looking into it.

Arosf:
@haverchuck amplify_datastore version is 0.1.0

fjnoyp:
Hi @Arosf the issue you encountered comes from the underlying Amplify iOS Library behavior.  I have been in communication with their devs and created an issue on their Amplify iOS repo linked here: https://github.com/aws-amplify/amplify-ios/issues/1083.  

[...]

PR #439, opened 2021-03-11
fix(amplify_datastore): fix ios schema change bug

fjnoyp:
Send ModelProvider version to Datastore iOS plugin to fix bug where changing a schema causes ios to crash afterwards.

https://github.com/aws-amplify/amplify-flutter/issues/415
https://github.com/aws-amplify/amplify-flutter/issues/422

*Issue #, if available:*

*Description of changes:*


Amplifiyer:
This is great! Would love to see iOS unit tests for this.

HuiSF:
Hi @fjnoyp it looks like we would need to merge the PR to fix the root cause. Could you please update the PR and we will review to merge. Thanks!

HuiSF (approved):
Thanks for the fix.

fjnoyp:
Hey @HuiSF thanks for bringing this PR back to the front.  I'm testing it again now.  Encountered a weird issue where changing the model schema would clear the models.  I'm looking into that now before pushing this PR forward.

HuiSF:
Thanks for the heads up @fjnoyp .
Clearing models? Do you mean it clears the local DB? (This is expected behavior when model provider version changes)

fjnoyp:
Thanks for the quick response.  Yes, okay that's good to know then.  Should be good to go then. @HuiSF

fjnoyp:
I'll get another approval so we can move forward here.

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
```diff
@@ -81,10 +81,19 @@ public class SwiftAmplifyDataStorePlugin: NSObject, FlutterPlugin {
...
-            return //TODO
+
+        guard let modelSchemaList = args["modelSchemas"] as? [[String: Any]],
+              let modelProviderVersion = args["modelProviderVersion"] as? String else {
+
+            FlutterDataStoreErrorHandler.handleDataStoreError(
+                error:
+                    DataStoreError.decodingError(
+                        "Received invalid request from Dart, modelSchemas and/or modelProviderVersion are not available. Request: " + args.description,
+                        "Check the values that are being passed from Dart."
```
cshfang, line 92: Nit: Since this is a Flutter handler I wonder if we should just reference Flutter directly rather than Dart

cshfang (approved):
Just a nit comment but lgtm otherwise

Issue #434, opened 2021-03-08
Firebase-Amplify dependency conflicts


exampaqs:
And because exampaqs depends on amplify_storage_s3 ^0.1.0 which depends on plugin_platform_interface ^1.0.1, firebase_auth
  ^1.0.0-1.0.nullsafety.0 requires firebase_core >=0.8.0-1.0.nullsafety.2 <0.8.0-2.0.nullsafety.0.
So, because exampaqs depends on both firebase_auth ^1.0.0-1.0.nullsafety.0 and firebase_core ^1.0.0-1.0.nullsafety.0, version solving
  failed.

because i upgraded firebase version and other dependencies in my pubspec.yaml file, i get the error above in my console because my project also s3 bucket and cognito-auth heavily. And when i comment out all amplify packages, the code works, so I feel there is a need for amplify upgrade asap.

Firebase depends on firebase_core_platform_interface 4.0.0 while amplify depends on  plugin_platform_interface ^1.0.1

fjnoyp:
Hi @exampaqs thanks for sharing your issue here. 

Could you provide your full pubspec.yaml so we can investigate all your dependencies?

exampaqs:
> Hi @exampaqs thanks for sharing your issue here.
> 
> Could you provide your full pubspec.yaml so we can investigate all your dependencies?


---------------------------------------.yaml file---------------------------------------------------
version: 1.0.0+1

environment:
  sdk: ">=2.10.0 <3.0.0"

dependencies:
  flutter:
    sdk: flutter
  # amplify_analytics_pinpoint: ^0.1.0
  # amplify_auth_cognito: ^0.1.0
  # amplify_flutter: ^0.1.0
  amplify_storage_s3: ^0.1.0
  # cached_network_image: ^2.5.0
  cloud_firestore: ^1.0.0-1.0.nullsafety.0
  connectivity: ^3.0.2
  countdown: ^0.1.0
  expandable: ^4.1.4
  firebase_auth: ^1.0.0-1.0.nullsafety.0
  firebase_core: ^1.0.0-1.0.nullsafety.0
  flutter_bloc: ^7.0.0-nullsafety.0
  flutter_countdown_timer: ^4.0.0
  flutter_html: ^2.0.0-nullsafety.0
  flutter_launcher_icons: ^0.9.0
  flutter_svg: ^0.21.0-nullsafety.0
  hexcolor: ^2.0.0
  http: ^0.13.0-nullsafety.0
[...]

PR #448, opened 2021-03-16
fix(amplify): upgrade uuid

fjnoyp:
*Issue #, if available:* #434

*Description of changes:*

Upgrade UUID dependency and remove conflicting flutter driver code 


packages/amplify_auth_cognito/example/test_driver/app.dart
```diff
@@ -9,3 +10,4 @@ void main() {
   // any widget you are interested in testing.
   app.main();
 }
+ */
```
Amplifiyer, line 13: Do we need this file at all?

fjnoyp: @haverchuck what are your thoughts - can I delete these?

fjnoyp: Confirmed with Dustin that we can delete these.

haverchuck (approved):

packages/amplify_datastore/pubspec.yaml
```diff
@@ -14,7 +14,7 @@ dependencies:
   amplify_core: 0.1.0
   plugin_platform_interface: ^1.0.1
   date_time_format: ^1.1.0
-  uuid: ^2.2.2
```
Ashish-Nanda, line 17: Is the `uuid` dependency needed in the `amplify_datastore` package when its already a dependency in the `amplify_datastore_plugin_interface` package?

fjnoyp: Thanks for the comment.  It’s a good point, the datastore package doesn’t need uuid.  I just tested and validated removing it won’t cause any problems.  I’ll remove the package.

Ashish-Nanda (approved):
Looks good 🚀

Issue #140, opened 2020-10-08
S3 download failure crashes app


JamesMcIntosh:
**Describe the bug**
When a download operation is failing the App is crashing because the a message reply has already been sent.

**To Reproduce**

```
final String path = Directory.systemTemp.createTempSync().path + "/tmpfile";
final DownloadFileOptions options = S3DownloadFileOptions(accessLevel: StorageAccessLevel.protected, targetIdentityId: "ABC123");
    final DownloadFileResult result = await Amplify.Storage.downloadFile(key: "cannot_access.key", local: File(path), options: options);
```

```
Shutting down VM
FATAL EXCEPTION: main
Process: com.myovolt.myovolt, PID: 22828
java.lang.IllegalStateException: Reply already submitted
	at io.flutter.embedding.engine.dart.DartMessenger$Reply.reply(DartMessenger.java:139)
	at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler$1.error(MethodChannel.java:240)
	at com.amazonaws.amplify.amplify_storage_s3.AmplifyStorageOperations$StorageOperations$prepareError$1.run(AmplifyStorageOperations.kt:208)
	at android.os.Handler.handleCallback(Handler.java:914)
	...
```

To see this in action put attach the Android debugger with a breakpoint in error closure for
[downloadFile()](https://github.com/aws-amplify/amplify-flutter/blob/5a5cef48784884c6fc57d48675af69c0566e9cab/packages/amplify_storage_s3/android/src/main/kotlin/com/amazonaws/amplify/amplify_storage_s3/AmplifyStorageOperations.kt#L136) in `AmplifyStorageOperations.kt`

The first time prepareError is called it is with the S3 not found exception.
```
accept:136, AmplifyStorageOperations$StorageOperations$downloadFile$2 (com.amazonaws.amplify.amplify_storage_s3)
accept:30, AmplifyStorageOperations$StorageOperations$downloadFile$2 (com.amazonaws.amplify.amplify_storage_s3)
onError:195, AWSS3StorageDownloadFileOperation$DownloadTransferListener (com.amplifyframework.storage.s3.operation)
run:288, TransferStatusUpdater$4 (com.amazonaws.mobileconnectors.s3.transferutility)
handleCallback:914, Handler (android.os)
```

Second time it is called because of the "Failed" state change listened to by the 
[...]

PR #456, opened 2021-03-19
Send only one error response when S3 download fails

JamesMcIntosh:
*Issue #140

*Description of changes:*

Add check for double submission of error into event channel.


Ashish-Nanda (approved):
Approved!
Thanks for your contribution @JamesMcIntosh

Issue #435, opened 2021-03-09
Null-Safety Support


raywhiteside:
I assume at some point null safety will be supported. I'd like to track the status and interest here.

fjnoyp:
Hi @raywhiteside thanks for creating this issue.  We don't have any set dates for delivering this, but are actively looking into it.  We can update you here.

cto-leaps:
I'm pretty sure a lot of people are looking to move to null-safety and not having all dependencies supporting it will be an issue soon. Do you guys have any idea of an ETA for this?

24601:
I would suggest a heavy priority be put on this. Frankly, a brand-new (as opposed to new rev with old legacy code) project/library like Amplify Flutter NOT supporting this amazes me (to the point of it almost being negligent, given null safety was no surprise, been in plan for quite some time in Flutter, almost as if AWS didn't hire or engage proper Flutter architects to plan this).

Especially considering that null-safe packages are compatible for those not using it, but you're locking out those who are or at least being the delaying dependency for those who otherwise could move (e.g. people not using null-safety could still use amplify-flutter, but people using null safety cannot use amplify-flutter).

Ashish-Nanda:
@24601 I understand your frustration, but as you can see above https://github.com/aws-amplify/amplify-flutter/issues/435#issuecomment-795408009 it is clear that we are actively working on the null safety migrations and acknowledge its importance. We cannot disclose our internal roadmap or dates, **but this is one of our top priorities at this time**. 

Also this is not a surprise for us. The project is new, but was started much before migrations of a lot of community packages began. It was planned to have the migration for all packages (we have 12) after GA, and at the same time to ensure they work together as many are interdependent. Please remember null safety migrations are still happening actively across the Dart/Flutter ecosystem with [775 packages migrating just recently](https://twitter.com/dart_lang/status/1370419316324245509?s=20), and different projects will take take time based on size, complexity, dependencies, etc. 

While we encourage feedback we request that it is constructive, respectful and adheres to our [community guidelines](https://aws.github.io/code-of-conduct). I assure you we understand the importance of the null safety migration and are actively working on it. We will update this thread as soon as we have a preview version available for customers to try.

meggamind:
@Ashish-Nanda Do you need help with this? I would like to make a PR if ya'll need help

24601:
[...]

PR #481, opened 2021-03-31
Null safety core

fjnoyp:
*Description of changes:*

Null Safety upgrade for (to min Dart 2.12): 
Core
Flutter
All Plugin Interface Packages 

*Issues*
#435 

Please note you won't be able to run a test app with these changes as the actual plugins have not been upgraded to NS in this PR. 


fjnoyp:
Please note that NS packages cannot rely on non NS packages for running unit tests. 

Furthermore, the android and ios unit tests will fail because the Pubspec dependency upgrades have not been performed on the actual plugins. 

We can only rely on unit tests once all of the sub PRs have been merged into the main null-safety-master PR.

packages/amplify_api_plugin_interface/lib/src/RestAPI/RestOptions.dart
```diff
@@ -17,22 +17,19 @@ import 'package:flutter/foundation.dart';
...
-  Map<String, String> queryParameters;
-  Map<String, String> headers;
+  Uint8List? body;
+  Map<String, String>? queryParameters;
+  Map<String, String>? headers;
 
   RestOptions({
     this.apiName,
-    @required this.path,
+    required this.path,
```
haverchuck, line 28: So in general with these changes, we are moving issues with missing parameters from runtime to compile time? Is that correct? If so, I don't think this necessarily amounts to a 'breaking' change, but there may be some developers who find their apps suddenly don't compile, right?

packages/amplify_auth_cognito/example/pubspec.yaml
```diff
@@ -7,6 +7,7 @@ publish_to: 'none' # Remove this line if you wish to publish to pub.dev
 
 environment:
   sdk: ">=2.7.0 <3.0.0"
+  flutter: ">=1.17.0 <2.0.0"
```
haverchuck, line 10: Minor nit: If we are adding this now, let's make it 1.20.0 instead of 1.17.0 in order to match our docs (for this example app, anyway).

packages/amplify_auth_plugin_interface/lib/src/Session/AuthSession.dart
```diff
@@ -16,10 +16,9 @@
 import 'package:amplify_auth_plugin_interface/amplify_auth_plugin_interface.dart';
 import 'package:flutter/foundation.dart';
 
-
 class AuthSession {
-  bool isSignedIn;
-  AuthSession({@required this.isSignedIn}) {
+  bool? isSignedIn;
+  AuthSession({this.isSignedIn}) {
```
haverchuck, line 21: Why is this becoming nullable?

packages/amplify_auth_plugin_interface/lib/src/SignIn/AuthNextSignInStep.dart
```diff
@@ -18,5 +18,5 @@ import '../types/AuthNextStep.dart';
 
 class AuthNextSignInStep extends AuthNextStep {
   String signInStep;
-  AuthNextSignInStep({additionalInfo, @required codeDeliveryDetails, @required this.signInStep}) : super(additionalInfo: additionalInfo, codeDeliveryDetails: codeDeliveryDetails);
+  AuthNextSignInStep({required additionalInfo, required codeDeliveryDetails, required this.signInStep}) : super(additionalInfo: additionalInfo, codeDeliveryDetails: codeDeliveryDetails);
```
haverchuck, line 21: Why is it necessary to make `additionalInfo` required?

packages/amplify_auth_plugin_interface/lib/src/SignUp/AuthNextSignUpStep.dart
```diff
@@ -18,5 +18,5 @@ import '../types/AuthNextStep.dart';
 
 class AuthNextSignUpStep extends AuthNextStep {
   String signUpStep;
-  AuthNextSignUpStep({additionalInfo, @required codeDeliveryDetails, @required this.signUpStep}) : super(additionalInfo: additionalInfo, codeDeliveryDetails: codeDeliveryDetails);
+  AuthNextSignUpStep({required additionalInfo, required codeDeliveryDetails, required this.signUpStep}) : super(additionalInfo: additionalInfo, codeDeliveryDetails: codeDeliveryDetails);
```
haverchuck, line 21: Necessary to make `additionalInfo` required?

packages/amplify_auth_plugin_interface/lib/src/SignUp/SignUpRequest.dart
```diff
@@ -21,14 +21,13 @@ class SignUpRequest {
...
+  SignUpRequest(
+      {required this.username, required this.password, required this.options});
   Map<String, dynamic> serializeAsMap() {
     final Map<String, dynamic> pendingRequest = <String, dynamic>{};
     pendingRequest['username'] = username;
     pendingRequest['password'] = password;
-    if (options != null) {
-      pendingRequest['options'] = options.serializeAsMap();
-    }
+    pendingRequest['options'] = options.serializeAsMap();
```
haverchuck, line 30: Nice catch removing this `if` check.

packages/amplify_datastore_plugin_interface/lib/amplify_datastore_plugin_interface.dart
```diff
@@ -25,7 +25,6 @@ import 'src/types/models/model.dart';
 import 'src/types/query/query_field.dart';
 import 'src/types/models/subscription_event.dart';
 
-export 'src/types/models/flutter_serialized_model.dart';
```
haverchuck, line 28: Why are we deleting this file?

packages/amplify_datastore_plugin_interface/lib/src/types/models/model_association.dart
```diff
@@ -20,22 +20,22 @@ enum ModelAssociationEnum { HasMany, HasOne, BelongsTo }
...
     this.associatedName,
     this.associatedType,
   });
 
   ModelAssociation copyWith({
-    String name,
-    String targetName,
-    String associatedName,
-    String associatedType,
+    ModelAssociationEnum? associationType,
```
haverchuck, line 35: Is this an existing bug being fixed here?

packages/amplify_datastore_plugin_interface/lib/src/types/models/model_field.dart
```diff
@@ -24,37 +25,37 @@ import 'model_field_type.dart';
...
 
   ModelField copyWith({
-    String name,
-    String type,
-    bool isRequired,
-    bool isArray,
-    ModelAssociation association,
-    List<AuthRule> authRules,
+    String? name,
+    ModelFieldType? type,
```
haverchuck, line 54: Preexisting bug?

packages/amplify_datastore_plugin_interface/lib/src/types/models/subscription_event.dart
```diff
@@ -24,20 +24,19 @@ class SubscriptionEvent<T extends Model> {
   final ModelType<T> modelType;
   final EventType eventType;
 
-  const SubscriptionEvent({this.item, this.modelType, this.eventType});
+  const SubscriptionEvent(
+      {required this.item, required this.modelType, required this.eventType});
```
haverchuck, line 28: Is this also a pre-existing bug? If not, why are we switching these three params from optional to required?

packages/amplify_datastore_plugin_interface/lib/src/types/query/query_sort.dart
```diff
@@ -30,7 +30,7 @@ class QuerySortBy {
   final QuerySortOrder order;
   final String field;
 
-  const QuerySortBy({this.order, this.field});
+  const QuerySortBy({required this.order, required this.field});
```
haverchuck, line 33: Why are we going from optiona -> required?

packages/amplify_flutter/lib/amplify.dart
```diff
@@ -37,13 +37,12 @@ export 'package:amplify_core/types/index.dart';
...
 
 /// Top level singleton Amplify object.
-final AmplifyClass Amplify = new AmplifyClass._private();
+final AmplifyClass Amplify = new AmplifyClass();
 
 /// This is a private class and customers are not expected to
 /// instantiate an object of this class. Please use top level
 /// `Amplify` singleton object for making calls to methods of this class.
 class AmplifyClass extends PlatformInterface {
-  AmplifyClass._private();
```
haverchuck, line 46: What is the reason for this change?

haverchuck (commented):
Has this been definitetively tested on Flutter 1.20.0 -> ^2.0.0?

packages/amplify_analytics_plugin_interface/pubspec.yaml
```diff
@@ -4,7 +4,7 @@ version: 0.1.1
 homepage: https://github.com/aws-amplify/amplify-flutter/tree/master/packages/amplify_analytics_plugin_interface
 
 environment:
-  sdk: ">=2.7.0 <3.0.0"
+  sdk: '>=2.12.0 <3.0.0'
```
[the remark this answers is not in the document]
fjnoyp: This changes the minimum Dart SDK version from 2.7.0 t0 2.12.0. This is necessary to update to Null Safety

packages/amplify_flutter/lib/amplify.dart
fjnoyp: That line causes a compile issue because PlatformInterface doesn't have an equivalent parameterless constructor.

packages/amplify_analytics_plugin_interface/lib/amplify_analytics_plugin_interface.dart
```diff
@@ -25,13 +25,13 @@ export 'src/types.dart';
 
 abstract class AnalyticsPluginInterface extends AmplifyPluginInterface {
   /// Constructs a AmplifyPlatform.
-  AnalyticsPluginInterface({@required Object token}) : super(token: token);
+  AnalyticsPluginInterface({required Object token}) : super(token: token);
```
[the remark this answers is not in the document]
fjnoyp: Good call, yes that import is not necessary anymore.  Removed it from the files.

packages/amplify_api_plugin_interface/lib/src/RestAPI/RestOptions.dart
fjnoyp: Yes that's correct.

packages/amplify_auth_cognito/example/pubspec.yaml
fjnoyp: hmm that line shouldn't be necessary.  I think it would be better if I just removed it - what do you think?

My changes should primarily be about the dart SDK version moving from >=2.7.0 to >=2.12.0

packages/amplify_auth_plugin_interface/lib/src/Session/AuthSession.dart
fjnoyp: Good question ... it shouldn't be.  However when I looked for usages of this class I don't see it anywhere - is this in use still?

packages/amplify_auth_plugin_interface/lib/src/SignIn/AuthNextSignInStep.dart
fjnoyp: You're right it's not required as AuthNextStep provides a default value.

packages/amplify_auth_plugin_interface/lib/src/SignUp/AuthNextSignUpStep.dart
fjnoyp: You're right it's not required as AuthNextStep provides a default value. Made it optional.

packages/amplify_datastore_plugin_interface/lib/amplify_datastore_plugin_interface.dart
fjnoyp: We don't use that file

packages/amplify_datastore_plugin_interface/lib/src/types/models/model_association.dart
fjnoyp: What are you referring to here?  The types you be nullable: 

  ModelAssociation copyWith({
    ModelAssociationEnum? associationType,
    String? targetName,
    String? associatedName,
    String? associatedType,
  }) {

packages/amplify_datastore_plugin_interface/lib/src/types/models/model_field.dart
fjnoyp: Yes this is a fix for a bug!

packages/amplify_datastore_plugin_interface/lib/src/types/models/subscription_event.dart
fjnoyp: Based on how this class was used it appeared that the fields were necessary. 

Looking at the usage of that constructor you can see only the fromMap constructor calls it and passes all fields. 

`  const SubscriptionEvent(
      {required this.item, required this.modelType, required this.eventType});

  factory SubscriptionEvent.fromMap(
      Map<dynamic, dynamic> map, ModelType<T> modelType) {
    Map<String, dynamic> serializedItem =
        Map<String, dynamic>.from(map["item"]);

    return SubscriptionEvent(
        item: modelType.fromJson(
            Map<String, dynamic>.from(serializedItem["serializedData"])),
        eventType: EventType.values
            .firstWhere((e) => describeEnum(e) == map["eventType"] as String?),
        modelType: modelType);
  }
}
`

packages/amplify_datastore_plugin_interface/lib/src/types/query/query_sort.dart
fjnoyp: Same as the above reason, all usages of this constructor method pass the fields: 

```
  // Sorting (Discussed later in this doc)
  QuerySortBy ascending() {
    return new QuerySortBy(field: fieldName, order: QuerySortOrder.ascending);
  }

  QuerySortBy descending() {
    return new QuerySortBy(field: fieldName, order: QuerySortOrder.descending);
  }
```

packages/amplify_auth_plugin_interface/lib/src/Session/AuthSession.dart
haverchuck: I believe it is.

packages/amplify_datastore_plugin_interface/lib/src/types/query/query_sort.dart
haverchuck: This won't break any customers?

packages/amplify_flutter/lib/amplify.dart
haverchuck: @Amplifiyer  what do you think of this?

packages/amplify_api_plugin_interface/lib/src/RestAPI/RestOptions.dart
haverchuck: We should discuss this as a team.

packages/amplify_flutter/lib/amplify.dart
Amplifiyer: I'm not sure why it is causing the compile issue since we are not calling super's constructor. a named `_` constructor help us to keep this a singleton and only initialized by us. Making this constructor public will open a huge can of worms.

packages/amplify_api_plugin_interface/lib/src/RestAPI/RestOptions.dart
fjnoyp: Confirmed with @Ashish-Nanda that we'll be doing a major version bump for this Null Safety work.

packages/amplify_auth_plugin_interface/lib/src/Session/AuthSession.dart
fjnoyp: Sounds good, switched to required

packages/amplify_datastore_plugin_interface/lib/src/types/query/query_sort.dart
fjnoyp: It will cause compile errors for customers not passing those fields.  However, aren't these fields require in the native side as well.  For example here are the constructors for Android: 

```
    public QuerySortBy(@NonNull String field, @NonNull QuerySortOrder sortOrder) {
        this(null, field, sortOrder);
    }

    /**
     * Constructor for {@code QuerySortBy}.
     *
     * @param modelName name of the model being sorted.
     * @param field name of field to sort by.
     * @param sortOrder order to sort by, either ASCENDING or DESCENDING.
     */
    public QuerySortBy(@Nullable String modelName, @NonNull String field, @NonNull QuerySortOrder sortOrder) {
        this.modelName = modelName;
        this.field = Objects.requireNonNull(field);
        this.sortOrder = Objects.requireNonNull(sortOrder);
    }
```

packages/amplify_flutter/lib/amplify.dart
fjnoyp: This is the compile error we get if the AmplifyClass._private(); constructor is present: 

> The superclass 'PlatformInterface' doesn't have a zero argument constructor.

I did not add a public constructor.  In the existing code we have a constructor: 

`  AmplifyClass() : super(token: _token);
`

That was added on 1/5/21 in the PR refactor amplify_core to amplify_flutter

Amplifiyer: I see that's just left over from a previous implementation (since the constructor is all tucked in the middle). You can still use the same thing with private constructor and pass an argument to super. 

```dart
class A {
  int superParameter;
  A(int p) {
    this.superParameter = p;
  }
  void printVal() {
    print(superParameter);
  }
}

class B extends A {
  B._private(): super(1);
}

var b = B._private();

void main() {
    b.printVal();
}
```

fjnoyp: Alright switched to private constructor

final AmplifyClass Amplify = new AmplifyClass._private();

AmplifyClass._private() : super(token: _token);

packages/amplify_api_plugin_interface/lib/src/GraphQL/GraphQLOperation.dart
```diff
@@ -13,14 +13,13 @@
  * permissions and limitations under the License.
  */
 
-import 'package:flutter/foundation.dart';
 import '../types.dart';
 
 class GraphQLOperation<T> {
-  Function _cancel;
+  late Function _cancel;
```
Amplifiyer, line 19: Is the usage of late correct here since it's being initialized in one and only constructor? I guess it's not really a lazy initialized var?

packages/amplify_api_plugin_interface/lib/src/GraphQL/GraphQLResponse.dart
```diff
@@ -13,12 +13,11 @@
  * permissions and limitations under the License.
  */
 
-import 'package:flutter/foundation.dart';
 import '../types.dart';
 
 class GraphQLResponse<T> {
   T data;
-  List<GraphQLResponseError> errors = [];
+  List<GraphQLResponseError>? errors = [];
```
Amplifiyer, line 20: Is this accurate? Didn't we already initialize it to be non-null. Are we saying that the caller can initialize it to null and that's expected?

packages/amplify_auth_plugin_interface/lib/src/Password/ResetPasswordRequest.dart
```diff
@@ -27,7 +27,7 @@ class ResetPasswordRequest {
       pendingRequest['username'] = username;
     }
     if (options != null) {
-      pendingRequest['options'] = options.serializeAsMap();
+      pendingRequest['options'] = options!.serializeAsMap();
     }
```
Amplifiyer, line 31: Not necessary I guess? You would have gotten a warning from dart analyzer https://dart.dev/null-safety/understanding-null-safety#unnecessary-code-warnings

You can run the dart analysis and remove similar code.

packages/amplify_auth_plugin_interface/lib/src/Password/ResetPasswordStep.dart
```diff
@@ -13,10 +13,10 @@
...
  */
 
-import 'package:flutter/foundation.dart';
+
 import '../types/AuthNextStep.dart';
 
 class ResetPasswordStep extends AuthNextStep {
   String updateStep;
-  ResetPasswordStep({additionalInfo, @required codeDeliveryDetails, @required this.updateStep}) : super(additionalInfo: additionalInfo, codeDeliveryDetails: codeDeliveryDetails);
+  ResetPasswordStep({required additionalInfo, required codeDeliveryDetails, required this.updateStep}) : super(additionalInfo: additionalInfo, codeDeliveryDetails: codeDeliveryDetails);
```
Amplifiyer, line 21: Is this intended? `additionalInfo` was not required earlier?

packages/amplify_auth_plugin_interface/lib/src/Password/UpdatePasswordRequest.dart
```diff
@@ -13,21 +13,21 @@
...
-  UpdatePasswordRequest({@required this.newPassword, @required this.oldPassword, this.options});
+  UpdatePasswordRequest({required this.newPassword, required this.oldPassword, this.options});
   Map<String, dynamic> serializeAsMap() {
     final Map<String, dynamic> pendingRequest = <String, dynamic>{};
     pendingRequest['newPassword'] = newPassword;
     pendingRequest['oldPassword'] = oldPassword;
     if (options != null) {
-      pendingRequest['options'] = options.serializeAsMap();
+      pendingRequest['options'] = options!.serializeAsMap();
     }
```
Amplifiyer, line 31: Same, if options is guaranteed to be non-null, `!` shouldn't be required.

packages/amplify_auth_plugin_interface/lib/src/Session/AuthUser.dart
```diff
@@ -13,11 +13,11 @@
  * permissions and limitations under the License.
  */
 
-import 'package:flutter/foundation.dart';
+
```
Amplifiyer, line 18: too many new lines.

packages/amplify_auth_plugin_interface/lib/src/SignIn/SignInWithWebUIRequest.dart
```diff
@@ -18,12 +18,12 @@ import 'package:flutter/foundation.dart';
...
 class SignInWithWebUIRequest {
-  AuthProvider provider;
+  AuthProvider? provider;
   SignInWithWebUIRequest({this.provider});
   Map<String, dynamic> serializeAsMap() {
     final Map<String, dynamic> pendingRequest = <String, dynamic>{};
     if (this.provider != null) {
-      pendingRequest["authProvider"] = describeEnum(provider);
+      pendingRequest["authProvider"] = describeEnum(provider!);
     }
```
Amplifiyer, line 27: what is the right usage here `this.provider` or `provider`. We should stick to the one type of usage.

packages/amplify_auth_plugin_interface/lib/src/SignUp/ResendSignUpCodeRequest.dart
```diff
@@ -13,14 +13,12 @@
...
 
-import 'package:flutter/foundation.dart';
 
-import './SignUpOptions.dart';
 
 class ResendSignUpCodeRequest {
   String username;
 
-  ResendSignUpCodeRequest({@required this.username,});
+  ResendSignUpCodeRequest({required this.username,});
```
Amplifiyer, line 21: probably don't need a `,` in the end. I know not your code but just a good cleanup

packages/amplify_auth_plugin_interface/lib/src/types/AuthNextStep.dart
```diff
@@ -17,9 +17,9 @@ import 'package:flutter/foundation.dart';
 import 'AuthCodeDeliveryDetails.dart';
 
 class AuthNextStep {
-  Map<dynamic, dynamic> additionalInfo;
-  AuthCodeDeliveryDetails codeDeliveryDetails;
-  AuthNextStep({@required codeDeliveryDetails, additionalInfo = const {}}) {
+  late Map<dynamic, dynamic> additionalInfo;
+  late AuthCodeDeliveryDetails codeDeliveryDetails;
```
Amplifiyer, line 21: Some context on why late is required here will be helpful.

packages/amplify_datastore_plugin_interface/lib/src/types/models/model_field.dart
```diff
@@ -69,18 +69,16 @@ class ModelField {
   Map<String, dynamic> toMap() {
     Map<String, dynamic> map = {
       'name': name,
-      'type': type.toMap(),
+      'type': type!.toMap(),
```
Amplifiyer, line 72: Shouldn't type be declared `required` since it's required 100% of the time?

packages/amplify_datastore_plugin_interface/lib/src/types/models/model_field_definition.dart
```diff
@@ -101,18 +100,18 @@ class ModelFieldDefinition {
...
-    String associatedName;
-    String associatedType;
+    String? associatedName;
+    String? associatedType;
 
     associatedName = associatedKey?.fieldName;
     associatedType = associatedKey?.fieldType != null
-        ? associatedKey.fieldType.ofModelName
+        ? associatedKey!.fieldType!.ofModelName
         : ofModelName;
```
Amplifiyer, line 115: Not sure if we can do the following, but would be cool if we can.

```dart
associatedType = associatedKey?.fieldType?.ofModelName ?? ofModelName;
```

packages/amplify_datastore_plugin_interface/lib/src/types/models/model_schema_definition.dart
```diff
@@ -19,16 +19,12 @@ import 'model_field_definition.dart';
 import 'model_schema.dart';
 
 class ModelSchemaDefinition {
-  String name;
-  String pluralName;
-  List<AuthRule> authRules;
-  Map<String, ModelField> fields;
+  String? name;
```
Amplifiyer, line 22: name is probably required?

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal.dart
```diff
@@ -39,7 +39,7 @@ class Temporal {
         caseSensitive: false,
         multiLine: false);
 
-    var match = regExp.matchAsPrefix(offsetString);
+    var match = regExp.matchAsPrefix(offsetString)!;
```
Amplifiyer, line 42: I feel this is an "escape hatch" or a work around using null safety. Null safety very correctly told us that the value here is nullable and your code expects a non-nullable and we are forcing (or overridding) the null safety here.

In general, I think we should avoid use of force un-wrapping (i.e. `!` operator) as much as we can and fail gracefully when we detect a null.

packages/amplify_datastore_plugin_interface/test/testData/Blog.dart
```diff
@@ -13,18 +13,19 @@
...
+import 'ModelProvider.dart';
+
 /** This is an auto generated class representing the Blog type in your schema. */
 @immutable
 class Blog extends Model {
   static const classType = const BlogType();
   final String id;
   final String name;
-  final List<Post> posts;
+  final List<Post>? posts;
```
Amplifiyer, line 28: Does it mean we need codegen changes?

packages/amplify_datastore_plugin_interface/test/testData/Comment.dart
```diff
@@ -70,7 +70,7 @@ class Comment extends Model {
     return buffer.toString();
   }
 
-  Comment copyWith({@required String id, Post post, @required String content}) {
+  Comment copyWith({required String id, Post? post, required String content}) {
```
Amplifiyer, line 73: copyWith probably won't need `required`

packages/amplify_api_plugin_interface/lib/src/GraphQL/GraphQLOperation.dart
fjnoyp: Thanks great point, updated the method: 

GraphQLOperation({required this.response, required Function cancel})
      : _cancel = cancel;

packages/amplify_api_plugin_interface/lib/src/GraphQL/GraphQLResponse.dart
fjnoyp: Good point.  I think this should be switched to just nullable and no initialization.

@Ashish-Nanda perhaps you can provide some insight here on how this class is used.  Based on the usage I see, it appears this field can be nullable and doesn't need the blank list [] initialization?

packages/amplify_auth_plugin_interface/lib/src/Password/ResetPasswordRequest.dart
fjnoyp: Hmm I do not get warnings from the dart analyzer. 

Why would this be unnecessary? 

As options is a nullable type we need to make sure it's not null before getting its value.  If we just called options!.serializeAsMap() if options was null the program would crash.

packages/amplify_auth_plugin_interface/lib/src/Password/ResetPasswordStep.dart
fjnoyp: You're right it shouldn't need to be required especially because AuthNextStep superclass has a default value for additionalInfo. 

@haverchuck can you verify?

packages/amplify_auth_plugin_interface/lib/src/Password/UpdatePasswordRequest.dart
fjnoyp: options is not guaranteed to be non null.  options is a nullable type.

packages/amplify_auth_plugin_interface/lib/src/SignIn/SignInWithWebUIRequest.dart
fjnoyp: Either way is fine.  "this" is added to make it more explicit that we are using an instance variable vs. some local or method parameter variable.  

We have no set standard for this throughout all of our plugins.  If we want to stick to one type of usage we should probably come together as a team and decide on a common path forward before making the change to all of our plugins.

packages/amplify_auth_plugin_interface/lib/src/SignUp/ResendSignUpCodeRequest.dart
fjnoyp: Sounds good

packages/amplify_auth_plugin_interface/lib/src/types/AuthNextStep.dart
fjnoyp: Do you think it's necessary to add comments to the dart file to explain?  I'll write the explanation here for now.  If a type is non nullable, then the compiler is checking to see that it is initialized somewhere in the constructor or variable declaration, if it can't find that it will throw an error until late keyword is added. 

1) Even though we have a default value of const {} for additionalInfo, that does not satisfy the compiler.  If you remove the late modifier, the compiler will complain that the field isn't initialized.  

2) codeDeliveryDetails is only initialized within the constructor method and not within the constructor parameters or when it was declared.

packages/amplify_datastore_plugin_interface/lib/src/types/models/model_field.dart
fjnoyp: thanks updated it.  Change name, type, isRequired, and isArray to be non nullable.

packages/amplify_datastore_plugin_interface/lib/src/types/models/model_field_definition.dart
fjnoyp: that works!  added that in

packages/amplify_datastore_plugin_interface/lib/src/types/models/model_schema_definition.dart
fjnoyp: yes it is, updated that thanks

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal.dart
fjnoyp: Alright that makes sense, added a graceful case to return a blank duration if no match was found.  However, isn't the ! operator necessary when making any access on a nullable field? 

How would you write this code without using the ! operator? 

image

packages/amplify_datastore_plugin_interface/test/testData/Comment.dart
fjnoyp: thanks.  Let's ignore the dart datastore codegen models for now.  There is a bit more complexity going on there that we can address in a future meeting (scheduled for next Monday) and eventually within the dedicated Datastore NS PR.

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal.dart
fjnoyp: Do you think it would be better if we called match?.group everwhere we could?  And just propagated the nullable type as far down as possible?

packages/amplify_auth_plugin_interface/lib/src/Password/ResetPasswordRequest.dart
Amplifiyer: I'm talking about the `!` being unnecessary since we already established  on line 29 that options is non-nullable so force unwrapping shouldn't be required

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal.dart
Amplifiyer: hmm I think this is one of those places where the language (compiler) is not "smart enough" to know that the var is guaranteed to be non-nullable and so force unwrap is fine. 

But for cases like below

```diff
   static AmplifyException fromMap(Map<String, String> serializedException) {
-    return AmplifyException(serializedException["message"],
+    return AmplifyException(serializedException["message"]!,
```

We know that the `serializedException["message"]` should be non-null, and so we should check and throw an exception if it is and in the else use it without the force unwrap (like swift `guard`)

packages/amplify_auth_plugin_interface/lib/src/Password/ResetPasswordRequest.dart
fjnoyp: Unfortunately the compiler isn't smart enough.  Removing the ! will result in an error as it thinks options can still be null.

packages/amplify_datastore_plugin_interface/lib/src/types/query/query_sort.dart
haverchuck: Hmm ok, as long as this is the case for both native libraries (and we mark this as breaking change - we should specifically call out all breaking changes in the change log and possibly in a pinned issue).

packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal.dart
fjnoyp: Sounds good updated that example and combed through the code looking for similar changes to make.

packages/amplify_auth_plugin_interface/lib/src/Password/ResetPasswordStep.dart
haverchuck: Correct.

PR #483, opened 2021-03-31
Null safety analytics

fjnoyp:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify_analytics_pinpoint/example/pubspec.yaml
+++ b/packages/amplify_analytics_pinpoint/example/pubspec.yaml
@@ -6,7 +6,7 @@ description: Demonstrates how to use the amplify_analytics_pinpoint plugin.
 publish_to: 'none' # Remove this line if you wish to publish to pub.dev
 
 environment:
-  sdk: ">=2.7.0 <3.0.0"
+  sdk: ">=2.12.0 <3.0.0"
 
 dependencies:
   flutter:
--- a/packages/amplify_analytics_pinpoint/example/pubspec.yaml
+++ b/packages/amplify_analytics_pinpoint/example/pubspec.yaml
@@ -6,7 +6,7 @@ description: Demonstrates how to use the amplify_analytics_pinpoint plugin.
 publish_to: 'none' # Remove this line if you wish to publish to pub.dev
 
 environment:
-  sdk: ">=2.12.0 <3.0.0"
+  sdk: '>=2.12.0 <3.0.0'
 
 dependencies:
   flutter:
--- a/packages/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
+++ b/packages/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
@@ -37,7 +37,7 @@ class AmplifyAnalyticsPinpoint extends AnalyticsPluginInterface {
   /// class that extends [AnalyticsPluginInterface] when they register themselves.
   static set instance(AnalyticsPluginInterface instance) {
     PlatformInterface.verifyToken(instance, _token);
-    _instance = instance;
+    _instance = instance as AmplifyAnalyticsPinpoint;
   }
 
   // Public facing methods
@@ -46,7 +46,7 @@ class AmplifyAnalyticsPinpoint extends AnalyticsPluginInterface {
     return _instance.addPlugin();
   }
 
-  Future<void> recordEvent({@required AnalyticsEvent event}) async {
+  Future<void> recordEvent({required AnalyticsEvent event}) async {
     return _instance.recordEvent(event: event);
   }
 
@@ -55,13 +55,13 @@ class AmplifyAnalyticsPinpoint extends AnalyticsPluginInterface {
   }
 
   Future<void> registerGlobalProperties(
-      {@required AnalyticsProperties globalProperties}) async {
+      {required AnalyticsProperties globalProperties}) async {
     return _instance.registerGlobalProperties(
         globalProperties: globalProperties);
   }
 
   Future<void> unregisterGlobalProperties(
-      {@required List<String> propertyNames}) async {
+      {required List<String> propertyNames}) async {
     return _instance.unregisterGlobalProperties(propertyNames: propertyNames);
   }
 
@@ -74,8 +74,8 @@ class AmplifyAnalyticsPinpoint extends AnalyticsPluginInterface {
   }
 
   Future<void> identifyUser(
-      {@required String userId,
-      @required AnalyticsUserProfile userProfile}) async {
+      {required String userId,
+      required AnalyticsUserProfile userProfile}) async {
     return _instance.identifyUser(userId: userId, userProfile: userProfile);
   }
 }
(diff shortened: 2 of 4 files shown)


packages/amplify_analytics_pinpoint/example/pubspec.yaml
```diff
@@ -6,7 +6,7 @@ description: Demonstrates how to use the amplify_analytics_pinpoint plugin.
 publish_to: 'none' # Remove this line if you wish to publish to pub.dev
 
 environment:
-  sdk: ">=2.7.0 <3.0.0"
+  sdk: ">=2.12.0 <3.0.0"
```
cshfang, line 9: Nit: We should try to be consistent about using double vs single quotes. I *think* the convention is single based on Flutter docs around `pubspec.yaml`

packages/amplify_analytics_pinpoint/pubspec.yaml
```diff
@@ -4,7 +4,7 @@ version: 0.1.4
 homepage: https://github.com/aws-amplify/amplify-flutter/tree/master/packages/amplify_analytics_pinpoint
 
 environment:
-  sdk: ">=2.7.0 <3.0.0"
+  sdk: '>=2.12.0 <3.0.0'
   flutter: ">=1.20.0"
```
cshfang, line 8: Copied from comment left on #485

Nit: This might be a good opportunity to rein in the consistency between " vs ' usage in our pubspec. I think the preferred convention is ' like you have it here for multi-part constraints and not even having quotes at all for single-part constraints i.e.:
```yaml
flutter: >=1.10.0
```

packages/amplify_analytics_pinpoint/lib/method_channel_amplify.dart
```diff
@@ -74,7 +74,10 @@ class AmplifyAnalyticsPinpointMethodChannel extends AmplifyAnalyticsPinpoint {
   }
 
   @override
-  Future<void> unregisterGlobalProperties({List<String> propertyNames}) async {
+  Future<void> unregisterGlobalProperties({List<String>? propertyNames}) async {
+    if (propertyNames == null) {
+      propertyNames = <String>[];
+    }
```
cshfang, line 80: Think we can avoid this check by providing a default:
```dart
Future<void> unregisterGlobalProperties({List<String>? propertyNames = const []}) async {
```
We can also safely cast as non-null after that if necessary

packages/amplify_analytics_pinpoint/pubspec.yaml
fjnoyp: changed to ^1.10.0

packages/amplify_analytics_pinpoint/example/pubspec.yaml
fjnoyp: yes single is best.

packages/amplify_analytics_pinpoint/lib/method_channel_amplify.dart
fjnoyp: sounds good, updated the code.

cshfang (approved):
Lgtm 👍

packages/amplify_analytics_pinpoint/lib/method_channel_amplify.dart
```diff
@@ -74,7 +74,7 @@ class AmplifyAnalyticsPinpointMethodChannel extends AmplifyAnalyticsPinpoint {
   }
 
   @override
-  Future<void> unregisterGlobalProperties({List<String> propertyNames}) async {
+  Future<void> unregisterGlobalProperties({List<String>? propertyNames = const []}) async {
```
ragingsquirrel3, line 77: non-blocking question: Why not make this required as unregistering nothing would not be very useful? Is it to minimize upgrading/migrating for customers in the event they have already called this with nothing?

ragingsquirrel3 (approved):
LGTM. I had a small, non-blocking question.

packages/amplify_analytics_pinpoint/lib/method_channel_amplify.dart
fjnoyp: hey, if you pass nothing to unregisterGlobalProperties it will unregister all global properties!  This is the behavior of the native APIs that we emulate.

haverchuck (approved):

packages/amplify_analytics_pinpoint/lib/method_channel_amplify.dart
ragingsquirrel3: oh man! thanks for the important info.

packages/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
```diff
@@ -37,7 +37,7 @@ class AmplifyAnalyticsPinpoint extends AnalyticsPluginInterface {
   /// class that extends [AnalyticsPluginInterface] when they register themselves.
   static set instance(AnalyticsPluginInterface instance) {
     PlatformInterface.verifyToken(instance, _token);
-    _instance = instance;
+    _instance = instance as AmplifyAnalyticsPinpoint;
```
Jordan-Nelson, line 40: I left a similar comment on #485. Any reason not to update the setter/getter to be of type `AmplifyAnalyticsPinpoint` instead of casting?

```dart
  static AmplifyAnalyticsPinpoint get instance => _instance;

  static set instance(AmplifyAnalyticsPinpoint instance) {
    _instance = instance;
  }
```

packages/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
```diff
@@ -46,7 +46,7 @@ class AmplifyAnalyticsPinpoint extends AnalyticsPluginInterface {
     return _instance.addPlugin();
   }
 
-  Future<void> recordEvent({@required AnalyticsEvent event}) async {
```
Jordan-Nelson, line 49: `import 'package:flutter/cupertino.dart';` can be removed with the removal of `@required`

packages/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
fjnoyp: Good point, we shouldn't be casting the instance parameter of type AnalyticsPluginInterface to AmplifyAnalyticsPinpoint if we're going to be saving it as type AnalyticsPluginInterface anyways!

fjnoyp: `  static set instance(AnalyticsPluginInterface instance) {
    _instance = instance;
  }`

We should be saving as Interface type though to allow for others to provide their own interface implementation though.

packages/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
fjnoyp: thanks, removed

cshfang (approved):

PR #484, opened 2021-03-31
Null safety api

fjnoyp:
*Issue #, if available:*

*Description of changes:*


packages/amplify_api/example/ios/Runner.xcodeproj/project.xcworkspace/contents.xcworkspacedata
```diff
@@ -2,6 +2,6 @@
 <Workspace
    version = "1.0">
    <FileRef
-      location = "group:Runner.xcodeproj">
+      location = "self:">
```
cshfang, line 5: Could you help me understand this change?

packages/amplify_api/lib/method_channel_api.dart
```diff
@@ -107,7 +107,7 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
 
     StreamSubscription _subscription = filteredStream.listen((event) {
       if (event['type'] == 'DONE') {
-        onDone();
+        onDone!();
```
cshfang, line 110: Isn't `onDone` nullable? If it's not provided, this would cause an issue right? I think we could either initialize `onDone` with some kind of noop if one is not provided or we need to perform a null check here

packages/amplify_api/lib/method_channel_api.dart
```diff
@@ -143,15 +143,15 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
...
+    required GraphQLRequest<T> request,
   }) async {
     try {
       final Map<String, dynamic> result =
-          await _channel.invokeMapMethod<String, dynamic>(
+      (await (_channel.invokeMapMethod<String, dynamic>(
         methodName,
         request.serializeAsMap(),
-      );
+      )))!;
```
cshfang, line 154: Are we sure this is non-null?

packages/amplify_api/lib/method_channel_api.dart
```diff
@@ -206,8 +206,8 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
 
     // Attempt switch to proper async
     try {
-      final Map<String, dynamic> data = await _channel
-          .invokeMapMethod<String, dynamic>(methodName, inputsMap);
+      final Map<String, dynamic> data = (await (_channel
+          .invokeMapMethod<String, dynamic>(methodName, inputsMap)))!;
```
cshfang, line 210: Ditto here

packages/amplify_api/pubspec.yaml
```diff
@@ -4,7 +4,7 @@ version: 0.1.4
 homepage: https://github.com/aws-amplify/amplify-flutter/tree/master/packages/amplify_api
 
 environment:
-  sdk: ">=2.7.0 <3.0.0"
+  sdk: ">=2.12.0 <3.0.0"
   flutter: ">=1.20.0"
```
cshfang, line 8: Maybe we could switch to the
```yaml
flutter: ^1.20.0
```
for consistency?

packages/amplify_api/example/ios/Runner.xcodeproj/project.xcworkspace/contents.xcworkspacedata
fjnoyp: Not too sure either.  Change happens automatically.  Seems to fix some Xcode build issue.

Ever since updating to Flutter 2.0 whenever I open and run the project for iOS this change is automatically done ... For this time I forgot to discard these changes.  

Looked online and found this here: https://stackoverflow.com/questions/66729236/flutter-ios-build-issues saying that it fixes the archive step for xcode.  If that's true we should probably include this change to all plugins instead of just api in the future.

packages/amplify_api/lib/method_channel_api.dart
fjnoyp: ah good call, added null check

packages/amplify_api/lib/method_channel_api.dart
fjnoyp: thanks we can't be 100% sure so added the changes we did in the other plugins: 

```

      final Map<String, dynamic>? data = (await (_channel
          .invokeMapMethod<String, dynamic>(methodName, inputsMap)));
      if (data == null) throw nullMapFromMethodChannelException;
```

packages/amplify_api/lib/method_channel_api.dart
fjnoyp: same fix applied here

packages/amplify_api/pubspec.yaml
fjnoyp: sounds good, done

packages/amplify_api/example/lib/main.dart
```diff
@@ -16,6 +16,7 @@
 import 'package:amplify_api/amplify_api.dart';
 import 'package:amplify_api_example/graphql_api_view.dart';
 import 'package:amplify_api_example/rest_api_view.dart';
+import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
```
haverchuck, line 19: If the configuration is using an API_KEY would this still be needed

packages/amplify_api/example/pubspec.yaml
```diff
@@ -19,6 +19,8 @@ dependencies:
     # The example app is bundled with the plugin so we use a path dependency on
     # the parent directory to use the current plugin's version.
     path: ../
+  amplify_auth_cognito:
```
haverchuck, line 22: Same question as above

packages/amplify_api/lib/method_channel_api.dart
```diff
@@ -26,8 +26,11 @@ import 'amplify_api.dart';
 
 const MethodChannel _channel = MethodChannel('com.amazonaws.amplify/api');
 
+AmplifyException nullMapFromMethodChannelException =
```
haverchuck, line 29: Could you explain this change?

packages/amplify_api/example/ios/Runner.xcodeproj/project.xcworkspace/contents.xcworkspacedata
ragingsquirrel3: I have the same thing happen in my builds and agree it probably makes sense to include the change.

packages/amplify_api/lib/method_channel_api.dart
fjnoyp: The NS upgrade treats any information received from MethodChannels as nullable.

In our method_channel code we can either force cast this result, or throw an exception if the result is null. It's better to throw an exception instead of force cast, thus we throw "nullMapFromMethodChannelException" in case the returned result is actually null.

packages/amplify_api/example/ios/Runner.xcodeproj/project.xcworkspace/contents.xcworkspacedata
fjnoyp: ok we can make sure these changes are in the NS master branch.

packages/amplify_api/example/lib/main.dart
fjnoyp: Do you know how we might set that up?  I remember that even with setting "NO" for Restrict API Access that I 'd need to add the Auth plugin module still @haverchuck

packages/amplify_api/lib/amplify_api.dart
```diff
@@ -33,7 +33,7 @@ class AmplifyAPI extends APIPluginInterface {
 
   static set instance(APIPluginInterface instance) {
     PlatformInterface.verifyToken(instance, _token);
-    _instance = instance;
+    _instance = instance as AmplifyAPI;
```
Jordan-Nelson, line 36: Same as the other NS PRs, I'm not sure why the getters/setters are of type `APIPluginInterface` and not `AmplifyAPI`. I left more detailed comments on the other PRs.

packages/amplify_api/lib/amplify_api.dart
```diff
@@ -43,21 +43,21 @@ class AmplifyAPI extends APIPluginInterface {
 
   // ====== GraphQL =======
   @override
-  GraphQLOperation<T> query<T>({@required GraphQLRequest<T> request}) {
```
Jordan-Nelson, line 46: Same as the other NS PRs, `import 'package:flutter/foundation.dart';` is now used with `@required` removed.

packages/amplify_api/example/.gitignore
```diff
@@ -42,3 +42,22 @@ app.*.map.json
...
+build/
+dist/
+node_modules/
+aws-exports.js
+awsconfiguration.json
+amplifyconfiguration.json
+amplify-build-config.json
+amplify-gradle-config.json
+amplifytools.xcconfig
+.secret-*
```
ragingsquirrel3, line 63: nit: no new line at end of file.

packages/amplify_api/lib/amplify_api.dart
ragingsquirrel3: I had same question.

packages/amplify_api/example/lib/main.dart
ragingsquirrel3: I had to make a similar change to my version of this app to run it, so I'm also interested to see if there is a way around it. Absent some solution there, +1 this change.

packages/amplify_api/lib/method_channel_api.dart
```diff
@@ -143,16 +146,16 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
...
-      final Map<String, dynamic> result =
-          await _channel.invokeMapMethod<String, dynamic>(
+      final Map<String, dynamic>? result =
+          (await (_channel.invokeMapMethod<String, dynamic>(
         methodName,
         request.serializeAsMap(),
-      );
-
+      )));
+      if (result == null) throw nullMapFromMethodChannelException;
```
ragingsquirrel3, line 158: noob question: Why not just `throw AmplifyException(AmplifyExceptionMessages.nullReturnedFromMethodChannel);` here? Is it an optimization, convention, or is there some value to throwing same literal error every time?

ragingsquirrel3 (approved):
Other than Jordan's casting question and Dustin's question about running example app without cognito plugin, LGTM. Just a nit and a question.

haverchuck (commented):
@fjnoyp I still think it makes sense to remove the changes to the example app.  You don't actually need cognito depending on your configuration.  It might make sense to make this change in the future, but we should put some more thought into it - and in any case, it doesn't really have anything to do with null safety.

fyi - here's how to add an api_key for a project in the cli:
Screen Shot 2021-05-25 at 7 58 48 AM

packages/amplify_api/lib/amplify_api.dart
fjnoyp: changed instance type to be APIPlugin

packages/amplify_api/lib/amplify_api.dart
fjnoyp: removed unnecessary import

packages/amplify_api/example/.gitignore
fjnoyp: added new line

packages/amplify_api/lib/method_channel_api.dart
fjnoyp: We can do that too.  

I figured it would be cleaner in the code to have a dedicated type for that instead of repeating the code throughout.

packages/amplify_api/example/lib/main.dart
fjnoyp: After discussing with @haverchuck we will be removing the Auth addition changes to the API example app as it's out of scope for the NS upgrade.

packages/amplify_api/example/pubspec.yaml
fjnoyp: After discussing with @haverchuck we will be removing the Auth addition changes to the API example app as it's out of scope for the NS upgrade.

cshfang (approved):

packages/amplify_api/lib/method_channel_api.dart
ragingsquirrel3: no change needed, just curious

haverchuck (approved):

Ashish-Nanda (approved):
Looks good

PR #485, opened 2021-03-31
Null safety storage

fjnoyp:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify_storage_s3/example/.gitignore
+++ b/packages/amplify_storage_s3/example/.gitignore
@@ -41,3 +41,22 @@ app.*.map.json
 
 # Exceptions to above rules.
 !/packages/flutter_tools/test/data/dart_dependencies_test/**/.packages
+
+#amplify
+amplify/\#current-cloud-backend
+amplify/.config/local-*
+amplify/logs
+amplify/mock-data
+amplify/backend/amplify-meta.json
+amplify/backend/awscloudformation
+amplify/backend/.temp
+build/
+dist/
+node_modules/
+aws-exports.js
+awsconfiguration.json
+amplifyconfiguration.json
+amplify-build-config.json
+amplify-gradle-config.json
+amplifytools.xcconfig
+.secret-*
--- a/packages/amplify_storage_s3/example/.gitignore
+++ b/packages/amplify_storage_s3/example/.gitignore
@@ -60,3 +60,4 @@ amplify-build-config.json
 amplify-gradle-config.json
 amplifytools.xcconfig
 .secret-*
+
--- a/packages/amplify_storage_s3/lib/amplify_storage_s3.dart
+++ b/packages/amplify_storage_s3/lib/amplify_storage_s3.dart
@@ -36,7 +36,7 @@ class AmplifyStorageS3 extends StoragePluginInterface {
 
   static set instance(StoragePluginInterface instance) {
     PlatformInterface.verifyToken(instance, _token);
-    _instance = instance;
+    _instance = instance as AmplifyStorageS3;
   }
 
   @override
@@ -45,28 +45,28 @@ class AmplifyStorageS3 extends StoragePluginInterface {
   }
 
   @override
-  Future<UploadFileResult> uploadFile({@required UploadFileRequest request}) {
+  Future<UploadFileResult> uploadFile({required UploadFileRequest request}) {
     return _instance.uploadFile(request: request);
   }
 
   @override
-  Future<GetUrlResult> getUrl({@required GetUrlRequest request}) {
+  Future<GetUrlResult> getUrl({required GetUrlRequest request}) {
     return _instance.getUrl(request: request);
   }
 
   @override
-  Future<RemoveResult> remove({@required RemoveRequest request}) {
+  Future<RemoveResult> remove({required RemoveRequest request}) {
     return _instance.remove(request: request);
   }
 
   @override
-  Future<ListResult> list({@required ListRequest request}) {
+  Future<ListResult> list({required ListRequest request}) {
     return _instance.list(request: request);
   }
 
   @override
   Future<DownloadFileResult> downloadFile(
-      {@required DownloadFileRequest request}) {
+      {required DownloadFileRequest request}) {
     return _instance.downloadFile(request: request);
   }
 }
--- a/packages/amplify_storage_s3/lib/amplify_storage_s3.dart
+++ b/packages/amplify_storage_s3/lib/amplify_storage_s3.dart
@@ -18,7 +18,6 @@ library amplify_storage_plugin;
 import 'package:amplify_storage_plugin_interface/amplify_storage_plugin_interface.dart';
 import 'package:plugin_platform_interface/plugin_platform_interface.dart';
 import './method_channel_storage_s3.dart';
-import 'package:flutter/foundation.dart';
 
 export './src/types.dart';
 export 'package:amplify_storage_plugin_interface/src/types.dart';
@@ -29,14 +28,14 @@ class AmplifyStorageS3 extends StoragePluginInterface {
   /// Constructs a AmplifyStoragePlugin.
   AmplifyStorageS3() : super(token: _token);
 
-  static AmplifyStorageS3 _instance = AmplifyStorageS3MethodChannel();
+  static StoragePluginInterface _instance = AmplifyStorageS3MethodChannel();
 
   /// The default instance of [AmplifyStoragePlugin] to use.
   static StoragePluginInterface get instance => _instance;
 
   static set instance(StoragePluginInterface instance) {
     PlatformInterface.verifyToken(instance, _token);
-    _instance = instance as AmplifyStorageS3;
+    _instance = instance;
   }
 
   @override
--- a/packages/amplify_storage_s3/lib/amplify_storage_s3.dart
+++ b/packages/amplify_storage_s3/lib/amplify_storage_s3.dart
@@ -28,12 +28,12 @@ class AmplifyStorageS3 extends StoragePluginInterface {
   /// Constructs a AmplifyStoragePlugin.
   AmplifyStorageS3() : super(token: _token);
 
-  static StoragePluginInterface _instance = AmplifyStorageS3MethodChannel();
+  static AmplifyStorageS3 _instance = AmplifyStorageS3MethodChannel();
 
   /// The default instance of [AmplifyStoragePlugin] to use.
-  static StoragePluginInterface get instance => _instance;
+  static AmplifyStorageS3 get instance => _instance;
 
-  static set instance(StoragePluginInterface instance) {
+  static set instance(AmplifyStorageS3 instance) {
     PlatformInterface.verifyToken(instance, _token);
     _instance = instance;
   }
--- a/packages/amplify_storage_s3/lib/method_channel_storage_s3.dart
+++ b/packages/amplify_storage_s3/lib/method_channel_storage_s3.dart
@@ -26,6 +26,8 @@ import 'amplify_storage_s3.dart';
 const MethodChannel _channel =
     MethodChannel('com.amazonaws.amplify/storage_s3');
 
+const AmplifyException nullMapFromMethodChannelException = AmplifyException(AmplifyExceptionMessages.nullReturnedFromMethodChannel);
+
 /// An implementation of [AmplifyPlatform] that uses method channels.
 class AmplifyStorageS3MethodChannel extends AmplifyStorageS3 {
   @override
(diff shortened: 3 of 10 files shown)


packages/amplify_storage_s3/pubspec.yaml
```diff
@@ -4,7 +4,7 @@ version: 0.1.4
 homepage: https://github.com/aws-amplify/amplify-flutter/tree/master/packages/amplify_storage_s3
 
 environment:
-  sdk: ">=2.7.0 <3.0.0"
+  sdk: '>=2.12.0 <3.0.0'
   flutter: ">=1.10.0"
```
cshfang, line 8: Nit: This might be a good opportunity to rein in the consistency between `"` vs `'` usage in our pubspec. I think the preferred convention is `'` like you have it here for multi-part constraints and not even having quotes at all for single-part constraints i.e.:
```yaml
flutter: >=1.10.0
```

packages/amplify_storage_s3/lib/src/S3UploadFile/S3UploadFileOptions.dart
```diff
@@ -17,11 +17,11 @@ import 'package:flutter/foundation.dart';
...
 
 class S3UploadFileOptions extends UploadFileOptions {
-  String targetIdentityId;
+  String? targetIdentityId;
   S3UploadFileOptions(
       {StorageAccessLevel accessLevel = StorageAccessLevel.guest,
-      String contentType,
-      Map<String, String> metadata,
+      String? contentType,
+      Map<String, String>? metadata,
```
cshfang, line 24: Does the super need to be updated to be nullable?

packages/amplify_storage_s3/lib/src/S3List/S3ListOptions.dart
```diff
@@ -11,11 +11,12 @@ class S3ListOptions extends ListOptions {
...
-      'accessLevel': describeEnum(accessLevel),
-      'targetIdentityId': targetIdentityId
-    };
-    optionsMap.removeWhere((_, v) => v == null);
+    final Map<String, dynamic> optionsMap = <String, dynamic>{};
+    optionsMap["accessLevel"] = describeEnum(accessLevel);
+
+    if (targetIdentityId != null) {
+      optionsMap["targetIdentityId"] = targetIdentityId;
+    }
```
cshfang, line 19: Nit: We can still initialize this with accessLevel defined in the map literal right?
```dart
final Map<String, dynamic> optionsMap = <String, dynamic>{
  'accessLevel' = describeEnum(accessLevel)
};
```
I may have missed something in a different PR but I'm curious - why move away from the `removeWhere` pattern in just this one particular case?

packages/amplify_storage_s3/lib/method_channel_storage_s3.dart
```diff
@@ -46,13 +46,13 @@ class AmplifyStorageS3MethodChannel extends AmplifyStorageS3 {
...
-      {@required UploadFileRequest request}) async {
+      {required UploadFileRequest request}) async {
     try {
       final Map<String, dynamic> data =
-          await _channel.invokeMapMethod<String, dynamic>(
+      (await _channel.invokeMapMethod<String, dynamic>(
         'uploadFile',
         request.serializeAsMap(),
-      );
+      ))!;
```
cshfang, line 55: Are we certain this is non-null or should we actually be checking before the formatting step below? Ditto for the other functions updated in this file.

packages/amplify_storage_s3/example/.gitignore
```diff
@@ -41,3 +41,22 @@ app.*.map.json
...
+build/
+dist/
+node_modules/
+aws-exports.js
+awsconfiguration.json
+amplifyconfiguration.json
+amplify-build-config.json
+amplify-gradle-config.json
+amplifytools.xcconfig
+.secret-*
```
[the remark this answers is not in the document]
fjnoyp: I'm not aware of any enforcement we have for that.  Where in Github did you see that?

packages/amplify_storage_s3/pubspec.yaml
fjnoyp: Good idea.  

How did you get it to compile without ' ? as I get a compile error when I do that. 

image

packages/amplify_storage_s3/example/.gitignore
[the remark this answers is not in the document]
cshfang: image
Just when reviewing changed files. I know it's kind of been widely accepted convention to have it (due to relics which had issues with files not being newline terminated) and it's no big deal if we don't have it. I just noticed it because github pointed it out. It's _probably_ good practice but I don't feel strongly about it either way

packages/amplify_storage_s3/pubspec.yaml
cshfang: Oh I think you'd want to use caret syntax there:
```yaml
flutter: ^0.10.0
```

fjnoyp: Nice switched to that.

packages/amplify_storage_s3/lib/src/S3UploadFile/S3UploadFileOptions.dart
fjnoyp: Yes the UploadFileOptions super class already has nullable contentType and metadata: 

```
class UploadFileOptions extends StorageOptions {
  StorageAccessLevel accessLevel;
  String? contentType;
  Map<String, String>? metadata;

```

fjnoyp: We updated the Plugin Interfaces to be NS as part of the NS Core PR.  This was to get the Amplify Flutter Module to compile as it has dependencies on all the Plugin Interfaces.

packages/amplify_storage_s3/lib/src/S3List/S3ListOptions.dart
fjnoyp: Yes we can initialize in the map literal, I think this was a merge conflict override so I switched it back: 

` optionsMap["accessLevel"] = {describeEnum(accessLevel)};
`

cshfang: Any concerns with switching back to using
```dart
optionsMap.removeWhere((_, v) => v == null);
```
to stay consistent?

packages/amplify_storage_s3/pubspec.yaml
cshfang: Cool. My bad

packages/amplify_storage_s3/lib/method_channel_storage_s3.dart
fjnoyp: Yes that's a good point.  We should do what we did for Auth.  So get a nullable map and then check if it is null: 
```

final Map<String, dynamic>? data =
          (await _channel.invokeMapMethod<String, dynamic>(
        'downloadFile',
        request.serializeAsMap(),
      ));
      if (data == null) throw nullMapFromMethodChannelException;
```

packages/amplify_storage_s3/lib/src/S3List/S3ListOptions.dart
fjnoyp: Yeah good point, should probably do that.

packages/amplify_storage_s3/lib/src/S3List/S3ListOptions.dart
```diff
@@ -11,9 +11,9 @@ class S3ListOptions extends ListOptions {
 
   @override
   Map<String, dynamic> serializeAsMap() {
-    final Map<String, dynamic> optionsMap = {
-      'accessLevel': describeEnum(accessLevel),
-      'targetIdentityId': targetIdentityId
+    final Map<String, dynamic> optionsMap = <String, dynamic>{
+      "accessLevel" : describeEnum(accessLevel),
+      "targetIdentifyId" : targetIdentityId
```
cshfang, line 16: Nit: Why double quotes?

cshfang (approved):
Minor nit comment otherwise lgtm

packages/amplify_storage_s3/lib/src/S3List/S3ListOptions.dart
fjnoyp: thanks, I'll change to single quote

packages/amplify_storage_s3/lib/method_channel_storage_s3.dart
```diff
@@ -35,24 +37,25 @@ class AmplifyStorageS3MethodChannel extends AmplifyStorageS3 {
...
     try {
-      final Map<String, dynamic> data =
-          await _channel.invokeMapMethod<String, dynamic>(
+      final Map<String, dynamic>? data =
+          (await _channel.invokeMapMethod<String, dynamic>(
         'uploadFile',
         request.serializeAsMap(),
-      );
+      ));
+      if (data == null) throw nullMapFromMethodChannelException;
```
haverchuck, line 58: Could you please explain the need for `nullMapFromMethodChannelException`?

haverchuck (commented):
Left a question; mostly looks great!

packages/amplify_storage_s3/lib/method_channel_storage_s3.dart
fjnoyp: The NS upgrade treats any information received from MethodChannels as nullable.  

In our method_channel code we can either force cast this result, or throw an exception if the result is null.  It's better to throw an exception instead of force cast, thus we throw "nullMapFromMethodChannelException" in case the returned result is actually null.

packages/amplify_storage_s3/lib/amplify_storage_s3.dart
```diff
@@ -45,28 +45,28 @@ class AmplifyStorageS3 extends StoragePluginInterface {
   }
 
   @override
-  Future<UploadFileResult> uploadFile({@required UploadFileRequest request}) {
```
Jordan-Nelson, line 48: With `@required` removed, I think `import 'package:flutter/foundation.dart';` is now unused.

packages/amplify_storage_s3/lib/amplify_storage_s3.dart
```diff
@@ -36,7 +36,7 @@ class AmplifyStorageS3 extends StoragePluginInterface {
 
   static set instance(StoragePluginInterface instance) {
     PlatformInterface.verifyToken(instance, _token);
-    _instance = instance;
+    _instance = instance as AmplifyStorageS3;
```
Jordan-Nelson, line 39: Is there any reason the setter should accept type `StoragePluginInterface` instead of type `AmplifyStorageS3`? The instance should always be of type `AmplifyStorageS3` in this class, right?. If there is a scenario where that is not true, this cast is going to throw an exception. If it is true, I think we should just update the getter/setter.

```dart
  static AmplifyStorageS3 get instance => _instance;

  static set instance(AmplifyStorageS3 instance) {
    _instance = instance;
  }
```

packages/amplify_storage_s3/example/.gitignore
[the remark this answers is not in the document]
ragingsquirrel3: Small nit, but I support adding blank newlines at end of file simply for the fact that it helps avoid conflicts when adding new stuff below because the new content would not count as a modification to line 62.

packages/amplify_storage_s3/lib/amplify_storage_s3.dart
ragingsquirrel3: Thanks, same question.

packages/amplify_storage_s3/lib/src/S3List/S3ListOptions.dart
```diff
@@ -11,9 +11,9 @@ class S3ListOptions extends ListOptions {
 
   @override
   Map<String, dynamic> serializeAsMap() {
-    final Map<String, dynamic> optionsMap = {
-      'accessLevel': describeEnum(accessLevel),
-      'targetIdentityId': targetIdentityId
+    final Map<String, dynamic> optionsMap = <String, dynamic>{
+      'accessLevel' : describeEnum(accessLevel),
```
ragingsquirrel3, line 15: uber nit: Is the space before : needed?

ragingsquirrel3 (approved):
LGTM, just nits. Thanks!

packages/amplify_storage_s3/lib/amplify_storage_s3.dart
fjnoyp: thanks removed.

packages/amplify_storage_s3/lib/amplify_storage_s3.dart
fjnoyp: Yes you're right though I would say the instance should always be of interface type.  

We want to accept the interface type to allow for other implementations to be accepted.

packages/amplify_storage_s3/lib/src/S3List/S3ListOptions.dart
fjnoyp: good point, that spacing is inconsistent with the other files

haverchuck (approved):

packages/amplify_storage_s3/example/.gitignore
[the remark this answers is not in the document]
fjnoyp: okay sounds good, added a newline to the gitignore

cshfang (approved):

Ashish-Nanda (approved):
Looks good overall, just a minor comment.

PR #503, opened 2021-04-08
Add builder field into showDialog function.

emreakcan:
It seems like `showDialog` function is changed in new version of flutter, need to add builder field. It can't find `child` field.


haverchuck:
@emreakcan Thanks for submitting these fixes!  Do you know if they are backwards compatible with Flutter 1.x?

emreakcan:
@haverchuck 
Hello ! Yes I can confirm its backwards competible, tested on versions
`Stable channel 2.0.4` and `1.26.0` it was painful to downgrade !

emreakcan:
> Hello @emreakcan - Thanks for opening the PR. I think these changes were incorporated in another PR. I think this can be closed out. Would you agree?

Closed :)

PR #505, opened 2021-04-09
Change dynamic type to string, otherwise it throws an error.

emreakcan:
Throws [ERROR:flutter/lib/ui/ui_dart_state.cc(186)] Unhandled Exception, fixed it by changing dynamic to String.

### DIFF
--- a/example/lib/Views/SignUpView.dart
+++ b/example/lib/Views/SignUpView.dart
@@ -43,7 +43,7 @@ class _SignUpViewState extends State<SignUpView> {
       _signUpError = "";
     });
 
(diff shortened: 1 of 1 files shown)


Jordan-Nelson:
Hello @emreakcan - Thanks for opening the PR. I think this change was incorporated in another PR. I think this can be closed out. Would you agree?

emreakcan:
> Hello @emreakcan - Thanks for opening the PR. I think this change was incorporated in another PR. I think this can be closed out. Would you agree?

Yeah, closed. No worries, thank you :)

Issue #493, opened 2021-04-01
Unrecognized Auth Event


saltedpotatos:
**Describe the bug**
Listening to the Auth channel of the Hub subscription on a real device, the run terminal is littered with "Unrecognized Auth Event"

Debugging, it seems to trip over "Auth.fetchSessionAPI" and  "Auth.fetchUserAttributesAPI",

I do fetch credentials, as described [here](https://docs.amplify.aws/lib/auth/access_credentials/q/platform/flutter), and user attributes, as described [here](https://docs.amplify.aws/lib/auth/user-attributes/q/platform/flutter)

**To Reproduce**

pubspec:
```yaml
  amplify_flutter: ^0.1.1
  amplify_auth_cognito: ^0.1.1
```
Amplify Config:
```dart
 if (!Amplify.isConfigured) {
    try {
      await Amplify.addPlugin(AmplifyAuthCognito());
      await Amplify.configure(amplifyconfig);
    } catch (e) {
      print('Irritating amplify');
    }
  }
```

Hub subscription
```dart
 StreamSubscription monitorHubState() {
    return hubSubscription = Amplify.Hub.listen([HubChannel.Auth], (hubEvent) {
      switch (hubEvent.eventName) {
        case 'SIGNED_IN':
          {
            _log.fine('Hub says signed in');
            break;
          }
        case 'SIGNED_OUT':
          {
            _log.fine('Hub says signed out');
            break;
          }
        case 'SESSION_EXPIRED':
          {
            _log.fine('Hub says session expired');
            add(AuthEvent.signedOut());
            break;
          }
      }
    });
  }
```

**Platform**
Amplify Flutter current supports iOS and Android.  This issue is reproducible in (check all that apply):
[] Android
[x] iOS

Have not attempted to run on Android

**Smartphone (please complete the following information):**
 - Device: [e.g. iPhone6]  iPhone 11 Pro Max,
 - OS: [e.g. iOS8.1] 14.4.2,

**Additional context**
Does not seem to occur when using the simulator, but when running on a real device.
This is on Flutter 2.12 running with unsound null safety. I haven't tested against 2.10.

PR #517, opened 2021-04-15
fix: Remove logging of unhandled Hub auth events (#493)

cshfang:
*Issue:* #493

*Description of changes:* To maintain consistency across bridged platforms, select Auth events from Hub are explicitly handled. Previously, unhandled events were then just being either logged to console (iOS) or Amplify Logger (Android). This resulted in `"Unrecognized Auth Event"` frequently being logged - particularly in iOS - thus polluting logs. This change removes logging of unhandled events.


packages/amplify_auth_cognito/ios/Classes/AuthCognitoHubEventStreamHandler.swift
```diff
@@ -32,24 +32,22 @@ public class AuthCognitoHubEventStreamHandler: NSObject, FlutterStreamHandler {
...
             default:
-                print("Unrecognized Auth Event")
+                return // rather than polluting logs with unhandled platform events, simply do nothing
             }
         }
     }
 
-    func sendEvent(flutterEvent: [String : Any]) {
-        eventSink?(flutterEvent)
+    func sendEvent(eventName: String) {
```
haverchuck, line 48: Nice improvement.

haverchuck (approved):

haverchuck:
Should we include this in the next major version, as it *might* be breaking for some customers depending on how they are using logs?

cshfang:
> Should we include this in the next major version, as it _might_ be breaking for some customers depending on how they are using logs?

I'm skeptical that there would be a use for generic `Unrecognized Auth Event` in logs, especially since `amplify-ios` is the source of these (currently, `amplify-android` still only sends the 3 handled events) and we were only doing a `print` rather than logging to CW. So I don't think there can be any real impacts to customer logs.

haverchuck (approved):

Issue #162, opened 2020-10-24
Amplify Analytics - Backfill Missing Unit Tests 


fjnoyp:
Problem Statement: 
The current unit tests for Analytics do not fully cover all cases and platforms. 

1) Expand dart unit tests to test the serialization of input data types. Test using "expect" that these input data types are serialized correctly when being sent over the method channel.

2) Add iOS unit tests. 

These new tests should follow the implementation of unit tests for the Datastore Query API. 

This is a good first issue for someone looking to contribute to this repo.

PR #520, opened 2021-04-16
Analytics chore 162 tests

ragingsquirrel3:
- chore #162 

Updates and adds some unit tests for analytics, both in dart and iOS. For dart side, adds assertions for future completion, assertions for method channel serialization and some cases for enable/disable. On iOS unit tests, just adds trivial enable/disable cases.

### DIFF
--- a/packages/amplify_analytics_pinpoint/example/ios/unit_tests/AnalyticsUnitTests.swift
+++ b/packages/amplify_analytics_pinpoint/example/ios/unit_tests/AnalyticsUnitTests.swift
@@ -259,6 +259,38 @@ class AnalyticsUnitTests: XCTestCase {
             }
         )
     }
+    
+    func test_enable_returnsSuccess() throws {
+        class MockAnalyticsBridge: AnalyticsBridge {
+            override func enable(){}
+        }
+        
+        pluginUnderTest = SwiftAmplifyAnalyticsPinpointPlugin(bridge: MockAnalyticsBridge())
+        
+        pluginUnderTest.innerHandle(
+            method: "enable",
+            callArgs: [],
+            result: { (results) in
+                XCTAssertEqual(results as! Bool, true)
+            }
+        )
+    }
+    
+    func test_disable_returnsSuccess() throws {
+        class MockAnalyticsBridge: AnalyticsBridge {
+            override func disable(){}
+        }
+        
+        pluginUnderTest = SwiftAmplifyAnalyticsPinpointPlugin(bridge: MockAnalyticsBridge())
+        
+        pluginUnderTest.innerHandle(
+            method: "disable",
+            callArgs: [],
+            result: { (results) in
+                XCTAssertEqual(results as! Bool, true)
+            }
+        )
+    }
  
 
 }
--- a/packages/amplify_analytics_pinpoint/example/ios/unit_tests/AnalyticsUnitTests.swift
+++ b/packages/amplify_analytics_pinpoint/example/ios/unit_tests/AnalyticsUnitTests.swift
@@ -269,7 +269,7 @@ class AnalyticsUnitTests: XCTestCase {
         
         pluginUnderTest.innerHandle(
             method: "enable",
-            callArgs: [],
+            callArgs: nil,
             result: { (results) in
                 XCTAssertEqual(results as! Bool, true)
             }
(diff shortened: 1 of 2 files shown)


packages/amplify_analytics_pinpoint/test/amplify_analytics_pinpoint_test.dart
```diff
@@ -26,95 +26,235 @@ void main() {
...
+      'unregisterGlobalProperties correctly serializes input to method channel',
+      () async {
+    final propertiesToUnregister = ['name'];
+    analyticsChannel.setMockMethodCallHandler((MethodCall methodCall) async {
+      expect(methodCall.arguments, propertiesToUnregister);
+    });
+    expect(
+        analytics.unregisterGlobalProperties(
+            propertyNames: propertiesToUnregister),
+        completes);
```
fjnoyp, line 75: Nit: shouldn't this be indented so that lines 74 and 75 are aligned?

packages/amplify_analytics_pinpoint/test/amplify_analytics_pinpoint_test.dart
```diff
@@ -26,95 +26,235 @@ void main() {
...
+  });
+
+  test(
+      'recordEvent adds properties to event and correctly serializes input to method channel',
+      () async {
+    var event = AnalyticsEvent(eventName);
+    analyticsChannel.setMockMethodCallHandler((MethodCall methodCall) async {
+      expect(methodCall.arguments, {
+        'name': eventName,
+        'propertiesMap': {
```
fjnoyp, line 85: Not necessary, but if you have time perhaps we can move these maps to constants like you did for the event name and properties as well.

fjnoyp (commented):
Great refactor of the unit tests! Just some minor comments.

packages/amplify_analytics_pinpoint/example/ios/unit_tests/AnalyticsUnitTests.swift
```diff
@@ -259,6 +259,38 @@ class AnalyticsUnitTests: XCTestCase {
...
+    func test_enable_returnsSuccess() throws {
+        class MockAnalyticsBridge: AnalyticsBridge {
+            override func enable(){}
+        }
+        
+        pluginUnderTest = SwiftAmplifyAnalyticsPinpointPlugin(bridge: MockAnalyticsBridge())
+        
+        pluginUnderTest.innerHandle(
+            method: "enable",
+            callArgs: [],
```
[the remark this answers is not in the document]
ragingsquirrel3: Thanks for the explanation! Makes sense, addressed in last commit.

packages/amplify_analytics_pinpoint/example/ios/unit_tests/AnalyticsUnitTests.swift
```diff
@@ -259,6 +259,38 @@ class AnalyticsUnitTests: XCTestCase {
...
+    func test_disable_returnsSuccess() throws {
+        class MockAnalyticsBridge: AnalyticsBridge {
+            override func disable(){}
+        }
+        
+        pluginUnderTest = SwiftAmplifyAnalyticsPinpointPlugin(bridge: MockAnalyticsBridge())
+        
+        pluginUnderTest.innerHandle(
+            method: "disable",
+            callArgs: [],
```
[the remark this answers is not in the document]
ragingsquirrel3: also addressed in last commit

packages/amplify_analytics_pinpoint/test/amplify_analytics_pinpoint_test.dart
ragingsquirrel3: disclosure: not being acutely aware of dart indenting, I have been letting the vscode formatter handle this on save automatically and trusting it.

I thought your comment made sense, but when I tried to make that change, the formatter kept making it like this. I **think** this is correct now bc lines 73 and 75 are both arguments to `expect` so they are aligned. Line 74 is indented bc it's just an extension of line 73 that is too long to fit. Does that make sense?

packages/amplify_analytics_pinpoint/test/amplify_analytics_pinpoint_test.dart
ragingsquirrel3: Good call. I addressed in last commit and also tried to fully commit to the variables for others used in the file.

packages/amplify_analytics_pinpoint/test/amplify_analytics_pinpoint_test.dart
fjnoyp: good call, thanks for taking a look into that

fjnoyp (approved):

Issue #513, opened 2021-04-14
AuthException instead of specific exception


brunovsiqueira:
Hi there.

When cognito throws an LimitExceededException, for example when you call the resetPassword method, the result comes as a AuthException instance and not a LimitExceededException instance. I can see it is a LimitExceededException only in the "code" part of "underlyingException" attribute. It makes hard for us to catch the different exceptions properly.

Shouldn't Amplify be throwing a LimitExceededException in this case?

brunovsiqueira:
Using amplify 0.1.0.

haverchuck:
@brunovsiqueira This should be fixed in 0.1.3.

brunovsiqueira:
Thanks so much guys! I am going to test it and let you guys know.

brunovsiqueira:
It is working as expected now! Thank you all again!

PR #522, opened 2021-04-16
fix: handle LimitExceededException on Android

Jordan-Nelson:
*Issue #, if available:* #513

*Description of changes:*

- Add additional case for `LimitExceededException` from cognitoidentityprovider in AuthErrorHandler
- Add unit test for the new case

The issue can be reproduced by calling `Amplify.Auth.resetPassword()` multiple times on Android. The issue is not present on iOS for this exception.

Prior to this PR we can see the code will throw an `AuthException` instead of a `LimitExceededException`, but the details of the error message do indicate that it is a LimitExceededException.

Screen Shot 2021-04-16 at 4 47 01 PM

With the changes in this PR the same code will throw a `LimitExceededException`.

Screen Shot 2021-04-16 at 4 48 09 PM

### DIFF
--- a/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthErrorHandler.kt
+++ b/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthErrorHandler.kt
@@ -30,6 +30,7 @@ import com.amazonaws.services.cognitoidentityprovider.model.TooManyFailedAttempt
 import com.amazonaws.services.cognitoidentityprovider.model.TooManyRequestsException
 import com.amazonaws.services.cognitoidentityprovider.model.UnexpectedLambdaException
 import com.amazonaws.services.cognitoidentityprovider.model.UserLambdaValidationException
+import com.amazonaws.services.cognitoidentityprovider.model.LimitExceededException
 import com.amplifyframework.AmplifyException
 import com.amplifyframework.auth.AuthException
 import com.amplifyframework.core.Amplify
@@ -76,6 +77,7 @@ class AuthErrorHandler {
                     is UnexpectedLambdaException -> errorCode = "LambdaException"
                     is UserLambdaValidationException -> errorCode = "LambdaException"
                     is TooManyFailedAttemptsException -> errorCode = "FailedAttemptsLimitExceededException"
+                    is LimitExceededException -> errorCode = "LimitExceededException"
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthErrorHandler.kt
```diff
@@ -76,6 +77,7 @@ class AuthErrorHandler {
                     is UnexpectedLambdaException -> errorCode = "LambdaException"
                     is UserLambdaValidationException -> errorCode = "LambdaException"
                     is TooManyFailedAttemptsException -> errorCode = "FailedAttemptsLimitExceededException"
+                    is LimitExceededException -> errorCode = "LimitExceededException"
```
Jordan-Nelson, line 80: _Open Question_: Does this replaces the need for handling of `AuthException.LimitExceededException`

https://github.com/aws-amplify/amplify-flutter/blob/1ab9dc0b09b3e6dfa004d39beaae66f6688dd6d1/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthErrorHandler.kt#L59

Jordan-Nelson: After looking into this further, this is what I have found. 

The Android Amplify library can throw Cognito exceptions (exceptions from `com.amazonaws.services.cognitoidentityprovider`) and Amplify auth exceptions (exceptions from `com.amplifyframework.auth`).

The Android Amplify library will convert & map Cognito exceptions to specific Amplify auth exceptions for `signUp`, `confirmSignUp`, `resendSignUpCode`, `signIn`, and `confirmSignIn`. However, it will not convert & map the exception for any of the other methods (there are about a dozen other methods, for example: `fetchAuthSession`, `resetPassword`, `confirmResetPassword`, `updatePassword`, `fetchUserSessionAttributes`). It will instead throw a generic AuthException and nest the Cognito exception within the generic AuthException. This can be seen in [`AWSCognitoAuthPlugin.java`](https://github.com/aws-amplify/amplify-android/blob/main/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java).

We are currently handling a subset of exceptions from both libraries. We are first checking to see if the exception is one of 18 Amplify auth exceptions, and mapping it to a specific error code if it is. If it is not any of those 18 Amplify auth exceptions, then we are checking to see if one of 9 Cognito exceptions is nested within it. That can be seen in the code above.

We currently only have a mapping for 18 of 20 Amplify Auth exceptions and 9 of 37 Cognito exceptions. `LimitExceededException` (added above) is one of the 28 Cognito exceptions that we don't have a mapping for.

To map all possible exceptions to specific exceptions, either the Android Amplify library would need to consistently throw an Amplify exception (and we would need to add mappings for the missing 2 exceptions here) or we need to map all 20 Amplify exceptions and all 37 Cognito exceptions.

Jordan-Nelson: After discussing with @haverchuck, we decided to move forward with the fix above (pending a new unit test) and then add an item to the backlog to address this more holistically.

haverchuck (approved):

ragingsquirrel3 (approved):
LGTM, thanks!

Ashish-Nanda (approved):
Looks good. Great work on your first PR @Jordan-Nelson!

PR #528, opened 2021-04-20
chore: 0.1.3 release

haverchuck:
*Description of changes:*
Updates version numbers and Changelogs.


cshfang:
Would it be possible to include #517 in this release?

cshfang (approved):

packages/amplify_analytics_pinpoint/CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+## 0.1.3 (2021-04-21)
+
```
Jordan-Nelson, line 2: *nit:* Should we add something like, "dependency for amplify_core changed to 0.1.3"?

I would be confused if I saw a change log entry with no info in it at all.

Jordan-Nelson (approved):
I left a nit comment, but other than that lgtm. I think we should probably figure out a good way to handle change logs when only the dependencies have changed since it will be quite common as we are pinning exact version numbers.

packages/amplify_analytics_pinpoint/CHANGELOG.md
haverchuck: I agree.  I'm not going to do it here, since we haven't had a chance to discuss. We have a few of these entries already, so it's not a new problem.

Issue #487, opened 2021-03-31
How do we update custom attributes of a user in aws cognito user pool using amplify flutter


pavan613:
**Describe the bug**
How do we update custom attributes of a user in aws cognito user pool using amplify flutter. 
If this feature is not there in flutter at the moment when is it expected to come.
Is there any work around for this?

haverchuck:
@pavan613 You can currently create custom attributes for a new user with the signup api; the APIs to update cognito attributes (both custom and standard) are actively under development.

SalahAdDin:
@haverchuck It is related to the **API** `updateAttributes` function, isn't it?

Jordan-Nelson:
Hi @pavan613 - The APIs to update user attributes are now available in version 0.1.5. The [docs on user attributes](https://docs.amplify.aws/lib/auth/user-attributes/q/platform/flutter#update-user-attribute) have been updated to include the new APIs. I am going to close out this ticket as I believe those APIs are what you are looking for.

Issue #430, opened 2021-03-06
GetUserAttributeVerficationCode and VerifyUser method implementation. [proposed label] feature-request


vishalkhedeker:
Hi, i couldnt find getUserAttibuteVerificationCode and verifyAttribute methods in amplify_auth_cognito for verfiying more than one attribute.
Will this be added to amplify-flutter?
If so by when is it expected?

haverchuck:
@vishalkhedeker I don't have an exact date, but this is definitely on our roadmap!

hysuki:
@vishalkhedeker Until it is officially supported, it can be implemented through api and function with accessToken if necessary.

For example
**API**

```graphql
type Mutation {
    verifyPhoneNumber(access_token: String, username: String, phone_number: String): verifyPhoneNumberResponse @function(name: "verifyPhoneNumber-${env}")
    verifyPhoneNumberConfirm(access_token: String, code: String): verifyPhoneNumberResponse @function(name: "verifyPhoneNumberConfirm-${env}")
}
```

```dart
class VerifyPhoneNumber {
  var logger = Logger();

  String graphQLDocument = '''
  mutation VerifyPhoneNumber(\$access_token: String = "", \$phone_number: String = "", \$username: String = "") {
    verifyPhoneNumber(access_token: \$access_token, phone_number: \$phone_number, username: \$username) {
      status
    }
  }
  ''';

  final String accessToken;
  final String phoneNumber;
  final String username;

  VerifyPhoneNumber({
    @required this.accessToken,
    @required this.phoneNumber,
    @required this.username,
  })  : assert(accessToken != null),
        assert(phoneNumber != null),
        assert(username != null);

  Future<String> data() async {
    try {
      var operation = Amplify.API.mutate(request: GraphQLRequest<String>(document: graphQLDocument, variables: {
        'access_token': accessToken,
        'phone_number': phoneNumber,
        'username': username,
      }));

      var response = await operation.response;
      var data = response.data;

      logger.d(data);
      return data;
    } on ApiException catch (e) {
      logger.e(e.message);
      return Future.error(e.message);
    }
  }
}




class VerifyPhoneNumberConfirm {
  var logger = Logger();

  String graphQLDocument = '''
  mutation VerifyPhoneNumberConfirm(\$access_token: String = "", \$code: String = "") {
    verifyPhoneNumberConfirm(access_token: \$access_token, code: \$code) {
      status
    }
  }
  ''';

  final String accessToken;
[...]

Issue #193, opened 2020-11-16
Support for UpdateUserAttributes operation


jcblancomartinez:
Hi,

Support to invoke UpdateUserAttributes operation.

https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_UpdateUserAttributes.html

Thanks.

Regards.

jamesonwilliams:
This functionality will be added soon. This issue should be marked as a feature request.

The other platforms have an `updateUserAttributes(...)` method. There is [a placeholder for this functionality in the Amplify Flutter documentation](https://docs.amplify.aws/lib/auth/user-attributes/q/platform/flutter#update-user-attribute).

Jordan-Nelson:
Hi @jcblancomartinez - I just wanted to provide an update that `updateUserAttribute` was added in 0.1.5. Support for updating multiple user attributes at once has not yet been added though so I will keep this issue open. [Here are the docs](https://docs.amplify.aws/lib/auth/user-attributes/q/platform/flutter#update-user-attribute) on the functionality added in 0.1.5.

Jordan-Nelson:
Noting that there is currently a bug with `updateUserAttributes` on Android that will probably require a fix prior to this being implemented in Flutter. The success callback is called multiple times, which results in a fatal exception from the flutter engine due to the success callback on the method channel being invoked multiple times. Issue for reference: https://github.com/aws-amplify/amplify-android/issues/1263

haverchuck:
@jcblancomartinez This has been included in version 0.1.6! We will leave this issue open for a while for comment.

Jordan-Nelson:
Closing issue as the feature has been added as of amplify-flutter 0.1.6

PR #533, opened 2021-04-22
feat(auth): User Attribute Auth Methods

Jordan-Nelson:
*Issue #, if available:* #487, #554; related issues: #430, #193, #554

*Description of changes:*
- add `updateUserAttribute` method to Auth
- add `confirmUserAttribute` method to Auth
- add `resendUserAttributeConfirmationCode` method to Auth
- update sample app with new methods
- update android error mapping for new underlying exceptions


packages/amplify_auth_cognito/ios/Classes/FlutterSignUpRequest.swift
```diff
@@ -36,11 +35,8 @@ struct FlutterSignUpRequest {
     let rawAttributes: Dictionary<String, Any> = options["userAttributes"] as! Dictionary<String, String>
     var formattedAttributes: Array<AuthUserAttribute> = Array()
     for attr in rawAttributes {
-      if (standardAttributes.contains(attr.key)) {
-        formattedAttributes.append(AuthUserAttribute(.unknown(attr.key), value: attr.value as! String))
-      } else {
-        formattedAttributes.append(AuthUserAttribute(.custom(attr.key), value: attr.value as! String))
-      }
+        let stringValue: String = (attr.value is String) ? attr.value as! String : String(attr.value as! Int)
```
Jordan-Nelson, line 38: *Note:* This is outside of the scope of this PR, but we were not handling the scenario where this was an `Int`. I did not test it with the sign up flow, but with updateUserAttribute, it was crashing at first because I had handled it the same way it was handled here.

packages/amplify_auth_cognito/ios/Classes/FlutterUpdateUserAttributeResult.swift
```diff
@@ -0,0 +1,116 @@
...
+        "nextStep": [
+            "updateAttributeStep": self.updateAttributeStep,
+            "additionalInfo": self.additionalInfo,
+            "codeDeliveryDetails": self.codeDeliveryDetails
+        ]
+      ]
+    }
+}
+
+func isComplete(res: AmplifyOperation<AuthUpdateUserAttributeRequest, AuthUpdateAttributeResult, AuthError>.OperationResult) -> Bool {
```
Jordan-Nelson, line 45: *Note:* Several of these util methods are duplicated throughout a few classes. I followed the pattern from the other classes and duplicated them here, but I think they should be broken out.

packages/amplify_auth_cognito/ios/Classes/FlutterSignUpRequest.swift
Jordan-Nelson: I will attempt to test this in the sign up flow to confirm its currently a bug, but @haverchuck I would appreciate your eyes on this one.

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FormatUserAttribute.kt
```diff
@@ -0,0 +1,55 @@
...
+import java.lang.Error
+import java.lang.reflect.Method
+
+fun createAuthUserAttribute(key: String, value: String): AuthUserAttribute {
+    val authUserAttributeKey: AuthUserAttributeKey = createAuthUserAttributeKey(key);
+    return AuthUserAttribute(authUserAttributeKey, value)
+}
+
+fun createAuthUserAttributeKey(keyName: String): AuthUserAttributeKey {
+    when (keyName) {
```
Jordan-Nelson, line 30: *Note:* Previously we were using reflection to convert the string into an `AuthUserAttributeKey`. I changed it to an explicit map for a few reasons.
1. Consistency with iOS. iOS previously wasn't using an associated value, which was problematic, so I updated it to be a map.
2. Differences between platforms. The two platforms have some differences that were not being accounted for. For example, iOS uses "birthDate" whereas android uses "birthdate". An explicit map allows us to account for those differences.
3. Personal preference. I think being explicit here is a lot easier to understand.

packages/amplify_auth_cognito/ios/Classes/FlutterSignUpRequest.swift
Jordan-Nelson: Marking as resolved. After looking into this, the sign up flow doesn't allow users to pass in integer values. I updated the methods I added to match that.

ragingsquirrel3 (commented):
I did not get to review all the kotlin files, and not sure how much my noob feedback is worth right now, just left a few nits/comments, nothing substantial. Thanks for doing this, and I will look in more detail next week.

packages/amplify_flutter/lib/categories/amplify_auth_category.dart
```diff
@@ -166,4 +165,38 @@ class AuthCategory {
...
+  }) {
+    var request = UpdateUserAttributeRequest(
+        userAttributeKey: userAttributeKey, value: value);
+    return plugins.length == 1
+        ? plugins[0].updateUserAttribute(request: request)
+        : throw _pluginNotAddedException("Auth");
+  }
+
+  /// Confirms a user attribute update and returns a [ConfirmUserAttributeResult]
+  Future<ConfirmUserAttributeResult> confirmUserAttribute({
```
ragingsquirrel3, line 182: thanks for doc comment. I think a potential project would be to go around library and add these.

packages/amplify_auth_cognito/ios/Classes/FlutterConfirmUserAttributeRequest.swift
```diff
@@ -0,0 +1,40 @@
...
+    var confirmationCode: String
+    
+    init(dict: NSMutableDictionary) {
+        self.userAttributeKey = createAuthUserAttributeKey(keyName: dict["userAttributeKey"] as! String)
+        self.confirmationCode = dict["confirmationCode"] as! String
+    }
+    
+    static func validate(dict: NSMutableDictionary) throws {
+        let validationErrorMessage = "ConfirmUserAttribute Request malformed."
+        if (dict["userAttributeKey"] == nil) {
```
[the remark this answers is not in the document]
Jordan-Nelson: I had a similar thought initially. We are checking that the data from the method call is of type `NSMutableDictionary` on every request. I think that is sufficiently performing a null check.

https://github.com/aws-amplify/amplify-flutter/blob/6bc64a7b2bd4f22944b4147d6aa65a6d533663d4/packages/amplify_auth_cognito/ios/Classes/SwiftAuthCognito.swift#L55-L61

It is a little confusing because we *do* have a null check on Android.  I think the difference is that on Android we are checking that the dat is of type `HashMap<*, *>` (see below), which can be null, where `NSMutableDictionary` cannot.

https://github.com/aws-amplify/amplify-flutter/blob/6bc64a7b2bd4f22944b4147d6aa65a6d533663d4/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt#L120-L125

What I could not figure out, is why we are checking that the data is of type `HashMap<String, *>` on Android in the many of the validate methods. (example below). That seems unnecessary to me since we have already checked that it is of type `HashMap<String, *>`, and then cast it to `HashMap<String, Any>`. I intended to follow that in this PR for the sake on consistency, but I am now seeing that we are already inconsistently checking in Auth (roughly have of the validate methods check for it and half do not). I am going to look into it a bit more, but will plan to remove it from the validate methods I added in this PR unless I find a reason to keep it.

https://github.com/aws-amplify/amplify-flutter/blob/6bc64a7b2bd4f22944b4147d6aa65a6d533663d4/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterSignInRequest.kt#L42

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
```diff
@@ -359,4 +416,27 @@ class AmplifyAuthCognitoMethodChannel extends AmplifyAuthCognito {
   AuthSession _formatSessionResponse(Map<String, dynamic> res) {
     return CognitoAuthSession.init(sessionValues: res);
   }
+
+  UpdateUserAttributeResult _formatUpdateUserAttributeResponse(
+      Map<String, dynamic> res) {
+    return UpdateUserAttributeResult(
+        isUpdated: res["isUpdated"],
+        nextStep: AuthNextUpdateAttributeStep(
+            updateAttributeStep: res["nextStep"]["updateAttributeStep"],
```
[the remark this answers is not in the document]
Jordan-Nelson: As far as I know, we do not currently (at least not in Auth).

packages/amplify_auth_plugin_interface/lib/amplify_auth_plugin_interface.dart
```diff
@@ -91,4 +93,22 @@ abstract class AuthPluginInterface extends AmplifyPluginInterface {
   Future<SignInResult> signInWithWebUI({SignInWithWebUIRequest request}) {
     throw UnimplementedError('signInWithWebUI() has not been implemented.');
   }
+
+  Future<UpdateUserAttributeResult> updateUserAttribute(
```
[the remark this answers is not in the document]
Jordan-Nelson: Are you asking what the purpose of the plugin interface is

packages/amplify_flutter/lib/categories/amplify_auth_category.dart
Jordan-Nelson: Agreed. @haverchuck has [an open PR](https://github.com/aws-amplify/amplify-flutter/pull/397/files) that adds them to Auth but I think there are some open comments/questions.

packages/amplify_auth_plugin_interface/lib/amplify_auth_plugin_interface.dart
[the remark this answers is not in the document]
ragingsquirrel3: Well, mostly was just thrown off by seeing `UnimplementedError` but dk if there are some docs/explanations of plugin interface that add context I'm missing.

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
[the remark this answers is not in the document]
ragingsquirrel3: Sounds good, thanks.

packages/amplify_auth_cognito/ios/Classes/FlutterConfirmUserAttributeRequest.swift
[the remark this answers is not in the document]
ragingsquirrel3: ok thanks for explanation

packages/amplify_flutter/lib/categories/amplify_auth_category.dart
ragingsquirrel3: cool

packages/amplify_auth_plugin_interface/lib/amplify_auth_plugin_interface.dart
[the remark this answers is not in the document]
Jordan-Nelson: The docs on [federated plugins](https://flutter.dev/docs/development/packages-and-plugins/developing-packages#federated-plugins) might be what you are looking for. I believe the intention is to follow the standards set for federated plugins.

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterUpdateUserAttributeRequest.kt
```diff
@@ -0,0 +1,53 @@
...
+        fun validate(req : HashMap<String, *>?) {
+            if (req == null) {
+                throw InvalidRequestException(validationErrorMessage, ExceptionMessages.missingAttribute.format("request map"))
+            } else if (!req.containsKey("attribute") || req["attribute"] !is HashMap<*, *>) {
+                throw InvalidRequestException(validationErrorMessage, ExceptionMessages.missingAttribute.format( "attribute" ))
+            } else {
+                val attribute = req["attribute"] as HashMap<*, *>;
+                if (!attribute.containsKey("value")) {
+                    throw InvalidRequestException(validationErrorMessage, ExceptionMessages.missingAttribute.format( "value" ))
+                } else if (!attribute.containsKey("userAttributeKey")) {
```
ragingsquirrel3, line 45: Does it also need to validate the `attribute["userAttributeKey"] !is String` like it does for value?

packages/amplify_auth_cognito/example/lib/Widgets/ConfirmUserAttribute.dart
```diff
@@ -0,0 +1,103 @@
...
+        confirmationCode: _confirmationCodeController.text,
+      );
+      print(res);
+      _showSuccess('Attribute Update Confirmed');
+    } on AmplifyException catch (e) {
+      _showError(e.message);
+    }
+  }
+
+  void _resenCode() async {
```
ragingsquirrel3, line 46: nit: maybe "resendCode" ?

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterSignUpRequest.kt
```diff
@@ -24,32 +24,20 @@ import com.amplifyframework.AmplifyException
 import com.amplifyframework.auth.AuthUserAttribute
 import com.amplifyframework.auth.AuthUserAttributeKey
 import com.amplifyframework.auth.cognito.options.AWSCognitoAuthSignUpOptions
-import java.lang.reflect.Method
 
 data class FlutterSignUpRequest(val map: HashMap<String, *>) {
```
ragingsquirrel3, line 28: I wasn't sure why it was necessary to change sign up related files, I guess it made sense to do this as a small refactor based on utilities you added?

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterUpdateUserAttributeRequest.kt
Jordan-Nelson: In general we don't check the type of each attribute on the request (the exception being for anything we expect to be a HashMap, to avoid trying access a key/value on an object that isn't a HashMap). I think you could argue that we should though. I _think_ the reason that we generally don't check the type of everything is that the purpose of the validate methods is to fail fast when a mistake is made is serializing the object, opposed to being used as a way to enforce correct user input (that should, and is, done on the dart side with proper types).

The reason I _did_ add a type check for `value` is because Cognito lets customers define user attributes as a String or an Int. However, the iOS and Android SDKs expect a string regardless of what the type was set to in Cognito. I wanted to be very explicit that `value` should always be a String, and we should not be serializing it to an Int in any scenario.

I added a comment to note that iOS/Android SDKs expect strings regardless of the configuration in Cognito to at least make it clear why I have a type check on that specifically.

packages/amplify_auth_cognito/example/lib/Widgets/ConfirmUserAttribute.dart
Jordan-Nelson: 👍  Updated

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterSignUpRequest.kt
Jordan-Nelson: Yeah. At the very least I wanted to pull the list of standard attributes into a single location so that we are not updating it in two places.

Also, the current approach of using reflection has a bug. There is a standard attribute called "zoneinfo" (all lowercase according to how it is defined in cognito) but the Android SDK defines the method for creating that attribute key as `AuthUserAttributeKey.zoneInfo()` (treating zone and info and two words, camel cased). That means that currently there is no way to add the "zoneinfo" standard attribute with the flutter sdk. If a customer tries to add "zoneinfo" it will throw an error because of the difference in capitalization, but any other spelling ("zone_info", "zoneInfo", etc.) would be added as custom attribute.

Jordan-Nelson: This is the exception that is currently thrown when attempting to add "zoneinfo" today.

```
kotlin.TypeCastException: null cannot be cast to non-null type java.lang.reflect.Method
```

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterUpdateUserAttributeRequest.kt
ragingsquirrel3: cool, sounds good

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterSignUpRequest.kt
ragingsquirrel3: thanks for info

ragingsquirrel3 (approved):

BesavedTech:
Awesome, when will will this be up for the next release?

packages/amplify_auth_cognito/example/lib/Widgets/ConfirmUserAttribute.dart
```diff
@@ -0,0 +1,103 @@
...
+      _showInfo(
+          'Confirmation Code Sent via ${res.codeDeliveryDetails.deliveryMedium}');
+    } on AmplifyException catch (e) {
+      _showError(e.message);
+    }
+  }
+
+  @override
+  void initState() {
+    isNewAttribute = widget.userAttributeKey == null;
```
cshfang, line 61: Is this used anywhere?

Jordan-Nelson: It was not. Good catch. I removed it.

packages/amplify_auth_cognito/example/lib/Widgets/ViewUserAttributes.dart
```diff
@@ -0,0 +1,119 @@
...
+    }).catchError((error) {
+      var errorMessage =
+          'There was an error fetching the user attribute: $error';
+
+      if (isRefresh) {
+        _showError(errorMessage);
+      } else {
+        setState(() => _errorMessage = errorMessage);
+      }
+    }).whenComplete(() => setState(() => _isLoading = false));
```
haverchuck, line 45: I'm not quite sure this is working as intended.  When attempting to view attributes with a user w/ expired session, the loading widget hangs.

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
```diff
@@ -491,4 +546,25 @@ public class AuthCognito : FlutterPlugin, ActivityAware, MethodCallHandler, Plug
...
+  }
+
+  fun prepareConfirmUserAttributeResult(@NonNull flutterResult: Result) {
+    var parsedResult = mutableMapOf<String, Any>();
+    Handler (Looper.getMainLooper()).post {
+      flutterResult.success(parsedResult);
+    }
+  }
+
+  fun prepareResendUserAttributeConfirmationCodeResult(@NonNull flutterResult: Result, @NonNull result: AuthCodeDeliveryDetails) {
```
haverchuck, line 564: It's kind of odd that `resendSignUpCode` doesn't return `AuthCodeDeliveryDetails` as well, but I guess that's the case.

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FormatUserAttribute.kt
haverchuck: I am assuming you retested the signUp flow, correct?

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterUpdateUserAttributeResult.kt
```diff
@@ -0,0 +1,43 @@
...
+data class FlutterUpdateUserAttributeResult(private val raw: AuthUpdateAttributeResult) {
+    val isUpdated: Boolean = raw.isUpdated
+    val nextStep: Map<String, Any> = setNextStep();
+
+    private fun setNextStep(): Map<String, Any> {
+        return mapOf(
+            "updateAttributeStep" to raw.nextStep.updateAttributeStep.toString(),
+            "additionalInfo" to Gson().toJson(raw.nextStep.additionalInfo),
+            "codeDeliveryDetails" to mapOf(
+                "destination" to (raw.nextStep.codeDeliveryDetails?.destination ?: ""),
```
haverchuck, line 30: We are probably going to be changing this code a bit, as part of null safety.

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FormatUserAttribute.kt
```diff
@@ -0,0 +1,68 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazonaws.amplify.amplify_auth_cognito.types
```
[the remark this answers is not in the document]
haverchuck: We don't have a standard, unless it can be used across multiple plugins.

packages/amplify_auth_cognito/ios/Classes/FlutterResendUserAttributeConfirmationCodeResult.swift
```diff
@@ -0,0 +1,62 @@
...
+              deliveryMap["attributeName"] = "phone"
+            }
+            
+            if case let .sms(e) = codeDeliveryDetails.destination {
+              deliveryMap["destination"] = e! as String
+              deliveryMap["attributeName"] = "sms"
+              deliveryMap["deliveryMedium"] = "SMS"
+            }
+            
+            if case let .unknown(e) = codeDeliveryDetails.destination {
```
haverchuck, line 54: This is the exhausitve list for the swtich statement, right?  I believe that XCode will complain if it is not, but I wanted to make sure.

packages/amplify_auth_plugin_interface/lib/amplify_auth_plugin_interface.dart
```diff
@@ -80,7 +81,8 @@ abstract class AuthPluginInterface extends AmplifyPluginInterface {
     throw UnimplementedError('getCurrentUser() has not been implemented.');
   }
 
-  Future<List<AuthUserAttribute>> fetchUserAttributes({AuthUserAttributeRequest request}) {
+  Future<List<AuthUserAttribute>> fetchUserAttributes(
+      {AuthUserAttributeRequest request}) {
```
haverchuck, line 85: Why are we changing just the formatting for this one method?  Maybe leave this for now, and fix all of them later if the formatting is wrong?

packages/amplify_auth_plugin_interface/lib/amplify_auth_plugin_interface.dart
```diff
@@ -91,4 +93,23 @@ abstract class AuthPluginInterface extends AmplifyPluginInterface {
   Future<SignInResult> signInWithWebUI({SignInWithWebUIRequest request}) {
     throw UnimplementedError('signInWithWebUI() has not been implemented.');
   }
+
+  Future<UpdateUserAttributeResult> updateUserAttribute(
```
haverchuck, line 97: I think we should have consistent formatting on all methods in this interface file.

packages/amplify_auth_plugin_interface/lib/src/Session/AuthNextUpdateAttributeStep.dart
```diff
@@ -0,0 +1,40 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import 'package:amplify_auth_plugin_interface/amplify_auth_plugin_interface.dart';
+import 'package:flutter/foundation.dart';
+
+/// This wraps all the details around the next step in the user attribute update process.
+///
+/// It contains [updateAttributeStep] to denote the step and supplements it with
```
haverchuck, line 21: Good documentation.

packages/amplify_auth_cognito/example/lib/Widgets/ViewUserAttributes.dart
haverchuck: In Cognito you can view the app client, and then set the refresh token to expire in 60 minutes (and all other tokens in 5).  So... make this change in Cognito, signIn, and wait an hour.

packages/amplify_auth_plugin_interface/lib/amplify_auth_plugin_interface.dart
haverchuck: I think it would be preferable to revert this, but if you feel strongly its ok.

packages/amplify_auth_plugin_interface/lib/amplify_auth_plugin_interface.dart
Jordan-Nelson: I removed the trailing comma to be consistent with the file/project, but left the formatting from dartfmt on the newly added methods. I think we *should* be using dartfmt if we aren't.

haverchuck (approved):

PR #537, opened 2021-04-23
Bump version script

ragingsquirrel3:
When releasing a new version, manually changing the version number across lots of files is annoying. Here, I added a little command line utility to automate it. It changes pubspec files and adds a blank entry to CHANGELOG files, although adding contents there will still be manual because they differ from package to package. It can be run with `dart tool/bump_version.dart`.

There is also a change in runtime code to `configure` to read the version from pubspec instead of hardcoding there.

Screen Shot 2021-04-23 at 9 30 33 AM

### DIFF
--- a/packages/amplify_flutter/lib/amplify.dart
+++ b/packages/amplify_flutter/lib/amplify.dart
@@ -172,6 +172,10 @@ class AmplifyClass extends PlatformInterface {
           underlyingException: e.toString());
     }
     try {
+      // Read the version from local pubspec file.
+      File pubspecFile = new File('pubspec.yaml');
+      final String versionFromPubspec =
+          loadYaml(await pubspecFile.readAsString())['version'];
       var res = await AmplifyClass.instance
           ._configurePlatforms(_getVersion(), configuration);
       _isConfigured = res;
--- a/packages/amplify_flutter/lib/amplify.dart
+++ b/packages/amplify_flutter/lib/amplify.dart
@@ -172,10 +172,6 @@ class AmplifyClass extends PlatformInterface {
           underlyingException: e.toString());
     }
     try {
-      // Read the version from local pubspec file.
-      File pubspecFile = new File('pubspec.yaml');
-      final String versionFromPubspec =
-          loadYaml(await pubspecFile.readAsString())['version'];
       var res = await AmplifyClass.instance
           ._configurePlatforms(_getVersion(), configuration);
       _isConfigured = res;
--- a/tool/bump_version.dart
+++ b/tool/bump_version.dart
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import 'dart:io';
+
+void main() async {
+  await bumpVersionInAllPubspecFilesAndAddBlankEntryToChangelogs();
+}
+
+/// When executed, first prompts the user to enter old version, new version, and date.
+/// It then looks for instances of old version in pubspec files and updates them to new version.
+/// It also adds a blank entry to the top of CHANGELOG.md files.
+Future<void> bumpVersionInAllPubspecFilesAndAddBlankEntryToChangelogs() async {
+  Future<void> _addTextToTopOfFile(String filePath, String textToAdd) async {
+    final file = File(filePath);
+    var contents = await file.readAsString();
+    contents = '$textToAdd$contents';
+    await file.writeAsString(contents);
+  }
+
+  stdout.write('What is the old version number? \n');
+  String? oldVersion = stdin.readLineSync();
+
+  stdout.write('What is the new version number? \n');
+  String? newVersion = stdin.readLineSync();
+
+  stdout.write(
+      'What is the date for this release? Use international format yyyy-mm-dd e.g. 2020-07-29. \n');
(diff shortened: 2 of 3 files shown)


tool/bump_version.dart
```diff
@@ -0,0 +1,71 @@
...
+    patternsForAllPubspecs.forEach((pattern) async {
+      await Process.run('/bin/sh', [
+        '-c',
+        'sed -i "" "/amplify/s/$oldVersion/$newVersion/g" $pattern' // lines that have "amplify" and the old version, change old to new
+      ]);
+    });
+    // change the version: x at top of files
+    await Process.run('/bin/sh', [
+      '-c',
+      'sed -i "" "s/version: $oldVersion/version: $newVersion/g" $packagePubspecFilePattern' // instances of "version: <old>" to "version: <new>"
```
cshfang, line 59: Just wanna make sure there's no chance of collision with some other appearance of version? Afaik there isn't but figure you would know better at this time.

tool/bump_version.dart
```diff
@@ -0,0 +1,71 @@
...
+    await file.writeAsString(contents);
+  }
+
+  stdout.write('What is the old version number? \n');
+  String? oldVersion = stdin.readLineSync();
+
+  stdout.write('What is the new version number? \n');
+  String? newVersion = stdin.readLineSync();
+
+  stdout.write(
```
cshfang, line 39: Super minor concern over lack of validation for the user prompts but nothing it does is breaking and typos can just be fixed afterwards but it seems like an improvement we might consider at some point.

cshfang (approved):
Lgtm I think this is very helpful but I think we'll probably want some 👀  from the more tenured members.

packages/amplify_flutter/lib/amplify.dart
```diff
@@ -172,8 +170,12 @@ class AmplifyClass extends PlatformInterface {
           underlyingException: e.toString());
     }
     try {
+      // Read the version from local pubspec file.
+      File pubspecFile = new File('pubspec.yaml');
```
ragingsquirrel3, line 174: callout: dk if there is anything special to consider with the path, reading file, gets confusing and not totally aware of gotchas that might exist in flutter env.

tool/bump_version.dart
ragingsquirrel3: I think there is some room for improvement here to ensure it occurs at start of line.

As written, could match something like "plugin_ends_with_version: 1.0.0" if it happens to have same version as `oldVersion`, although I guess it's somewhat unlikely.

tool/bump_version.dart
ragingsquirrel3: I agree that would be a place to improve. It could also populate some of the input with a suggestion, like current date, or find old version from somewhere in the code. There are other CLI niceties that could be included, like outputting a summary of what was changed, but I thought this was good enough to start.

tool/bump_version.dart
```diff
@@ -0,0 +1,71 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import 'dart:io';
```
ragingsquirrel3, line 16: callout: dk if this would be better just as an sh script? I thought a dart project should have dart scripts, but it's basically sh called from dart.

packages/amplify_flutter/lib/amplify.dart
Jordan-Nelson: This is throwing an exception for me in the example projects in amplify_flutter and amplify_auth_cognito, on iOS and Android 🤔 

```
FileSystemException (FileSystemException: Cannot open file, path = 'pubspec.yaml' (OS Error: No such file or directory, errno = 2))
```

Jordan-Nelson: If this is working for everyone else, I am assuming something is messed up locally for me. @cshfang did this run fine for you?

Jordan-Nelson: Not related to that exception -

I see the advantage of reading it from pubspec.yaml, and I think that is probably better long term. But if there are any doubts about doing it that way right now, we could just go with the much less sophisticated solution of a find/replace just like all the other version numbers.

Amplify.dart
```dart
final _version = '0.1.4';

class AmplifyClass extends PlatformInterface {
  ...
}
```

bump_version.dart
```dart
await Process.run('/bin/sh', [
  '-c',
  'sed -i "" "s/final _version = \'$oldVersion\'/final _version = \'$newVersion\'/g" packages/amplify_flutter/lib/amplify.dart'
]);
```

cshfang: I was able to run the base example project but did not try any of the others

ragingsquirrel3: I did this in last commit with the small tweak that I did not add _version as top-level variable just to avoid changing this file here.

tool/bump_version.dart
```diff
@@ -0,0 +1,77 @@
...
+
+void main() async {
+  await bumpVersionInAllPubspecFilesAndAddBlankEntryToChangelogs();
+}
+
+/// When executed, first prompts the user to enter old version, new version, and date.
+/// It then looks for instances of old version in pubspec files and updates them to new version.
+/// It also adds a blank entry to the top of CHANGELOG.md files.
+Future<void> bumpVersionInAllPubspecFilesAndAddBlankEntryToChangelogs() async {
```
cshfang, line 25: This might be technically incomplete now that it also updates the top level `amplify.dart`. I think maybe we can make this name slightly less verbose and update the comment to something resembling:
```dart
/// When executed, first prompts the user to enter old version, new version, and date. It then
///   1. looks for instances of old version in pubspec files and updates them to new version.
///   2. replaces the version string in the top-level `amplify.dart` file.
///   3. adds a blank entry to the top of CHANGELOG.md files.
```
That just kind of lists out the changes this function makes?

ragingsquirrel3: good suggestion, I used this comment and changed to `_bumpVersionInFiles`.

cshfang (approved):
Lgtm let's try to make sure we give it a spin on the next release!

PR #540, opened 2021-04-24
chore: iOS optimistic pin

haverchuck:
*Description of changes:*
Uses optimistic version indicator to preclude only major version changes from amplify-ios.


cshfang (approved):

Ashish-Nanda (commented):
Shouldn't we pin to the exact version number similar to how we do it for Android?

Jordan-Nelson (approved):
LGTM

PR #545, opened 2021-04-27
fix(amplify_auth_cognito): Pass additionalInfo to AuthNextSignUpStep

### LINKED ISSUE #544: Public challenge parameters not returned from custom auth flow in iOS only
**Describe the bug**
When using a custom authentication flow in Cognito, the 'Create Auth Challenge' lambda trigger can return [public challenge parameters](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-create-auth-challenge.html#cognito-user-pools-lambda-trigger-syntax-create-auth-challenge-response) to the client in order to provide information about how to complete the challenge. For example, the trigger could return a generated Captcha URL or instructions on where an authentication code was sent to the user. This is essential information in order to implement many custom challenge use cases because the client can't complete authentication without that information.

On Android, amplify-flutter correctly returns the public challenge parameters in the `SignInResult` response (via `nextStep.additionalDetails`. However, on iOS `additionalDetails` is always null.

**To Reproduce**
1. Configure a 'Create Auth Challenge' lambda trigger in Cognito that returns public challenge parameters in the event response, e.g.:
```
event.response.publicChallengeParameters = {
  foo: 'bar'
};
```
2. Configure amplify-flutter to use the custom authentication flow:
```
...
          "Auth": {
            "Default": {
              "authenticationFlowType": "CUSTOM_AUTH"
            }
          }
...
```
3. Call signIn from the client: `final SignInResult response = Amplify.Auth.signIn(username, password);`
4. Examine the reponse, e.g.: `print(response.nextStep.additionalInfo);`

**Expected behavior**
Debug output: `flutter: {foo: bar}`

**Actual behavior (iOS only)**
`flutter: null`

**Platform**
Amplify Flutter current supports iOS and Android.  This issue is reproducible in (check all that apply):
[] Android
[x] iOS

**Smartphone (please complete the following information):**
 - Device: iPhone12 Simulator version 12.4
 - OS: iOS 14.4
 - amplify-flutter version: 0.1.4 (and all previous versions)

**Additional info**
I've traced down the issue and it spans both the amplify-flutter and amplify-ios libraries. I'll submit pull requests with the "fix" that makes the behavior consistent with android and reference this issue. Though, I'm not very familiar with the code base, so I don't know if my workaround will break any other use cases.

jcowley:
See details in issue #544

I'm submitting this mostly as an FYI since I was able to track down the issue(s), which span both amplify-flutter and amplify-ios. I'm not familiar with these code bases (or Swift for that matter), but I do know that these two changes resolve the issue. What I don't know is if these changes will break other use cases. So, let me know if there's another way of resolving this or additional changes needed.

There is a dependent fix in `amplify-ios` that is needed in order for this change to resolve the issue:
https://github.com/aws-amplify/amplify-ios/pull/1190

The problem here appears to be that `res["nextStep"]["additionalInfo"]` is never a String because it has already been json decoded into an object.

Also, there is one other place in the code (`_formatResetPasswordResponse()` in the same file) that may also need to be changed. However, that's not a use case that I use, so I didn't modify that function. (I can update the PR if needed.)

### DIFF
--- a/packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
+++ b/packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
@@ -316,9 +316,7 @@ class AmplifyAuthCognitoMethodChannel extends AmplifyAuthCognito {
         nextStep: AuthNextSignInStep(
             signInStep: res["nextStep"]["signInStep"],
             codeDeliveryDetails: res["nextStep"]["codeDeliveryDetails"],
-            additionalInfo: res["nextStep"]["additionalInfo"] is String
-                ? jsonDecode(res["nextStep"]["additionalInfo"])
-                : {}));
+            additionalInfo: res["nextStep"]["additionalInfo"] ?? {}));
   }
 
   UpdatePasswordResult _formatPasswordResponse(Map<String, dynamic> res) {
--- a/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterSignInResult.kt
+++ b/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterSignInResult.kt
@@ -27,7 +27,7 @@ data class FlutterSignInResult(private val raw: AuthSignInResult) {
 
     return mapOf(
       "signInStep" to raw.nextStep.signInStep.toString(),
-      "additionalInfo" to Gson().toJson(raw.nextStep.additionalInfo),
+      "additionalInfo" to (raw.nextStep.additionalInfo ?: emptyMap<String, String>()),
       "codeDeliveryDetails" to mapOf(
         "destination" to (raw.nextStep.codeDeliveryDetails?.destination ?: ""),
         "deliveryMedium" to (raw.nextStep.codeDeliveryDetails?.deliveryMedium?.name ?: ""),


packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
```diff
@@ -316,9 +314,7 @@ class AmplifyAuthCognitoMethodChannel extends AmplifyAuthCognito {
         nextStep: AuthNextSignInStep(
             signInStep: res["nextStep"]["signInStep"],
             codeDeliveryDetails: res["nextStep"]["codeDeliveryDetails"],
-            additionalInfo: res["nextStep"]["additionalInfo"] is String
-                ? jsonDecode(res["nextStep"]["additionalInfo"])
-                : {}));
+            additionalInfo: res["nextStep"]["additionalInfo"] ?? {}));
```
haverchuck, line 317: It looks like Android and iOS just use codeDeliveryDetails.  We can probably back this change out for now, and remove these unused attributes later.

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
```diff
@@ -299,9 +299,7 @@ class AmplifyAuthCognitoMethodChannel extends AmplifyAuthCognito {
         nextStep: AuthNextSignUpStep(
             signUpStep: res["nextStep"]["signUpStep"],
             codeDeliveryDetails: res["nextStep"]["codeDeliveryDetails"],
-            additionalInfo: res["nextStep"]["additionalInfo"] is String
-                ? jsonDecode(res["nextStep"]["additionalInfo"])
-                : {}));
+            additionalInfo: res["nextStep"]["additionalInfo"] ?? {}));
```
haverchuck, line 302: I think that we need to change the Android [FlutterSignInResult](https://github.com/aws-amplify/amplify-flutter/blob/6bc64a7b2bd4f22944b4147d6aa65a6d533663d4/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterSignInResult.kt#L30) class to accomodate this change.

From this (line 30):
```kotlin
 "additionalInfo" to Gson().toJson(raw.nextStep.additionalInfo),
```
to this:
```kotlin
"additionalInfo" to (raw.nextStep.additionalInfo ?: emptyMap()),
```
Have you found otherwise?

haverchuck (changes requested):
Thanks so much for submitting this fix! I've asked the ios team to look at your accompanying ios PR.  Please see my inline comments.

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
jcowley: `codeDeliveryDetails` would not work for custom auth flow, only with the out-of-the-box MFA cognito flow. The public challenge parameters can be anything, which can be unrelated to delivering a code (e.g. a Captcha flow).

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
jcowley: Let me take a look. If so, I'll update the PR.

jcowley: Confirmed that this did need to be updated for Android. Added your proposed fix and confirmed it works.

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
jcowley: @haverchuck: codeDeliveryDetails would not work for public challenge parameters since those can be any set of key/values returned, not just specific to code delivery. For example, you might be returning a Captcha url.

This change is definitely needed for the fix. I did back out the change that I made to `_formatSignUpResponse`, since that's not needed for the fix. Was that the line you meant to refer to?

haverchuck: I recalled that you had a change related to `additionalInfo` in the response from the resendSignUpCode method, but I might be mistaken (I'm not seeing it now).  These changes look ok to me.  I've asked our ios team to test your PR in the amplify-ios repo, and once they've released it I'll retest this PR a bit and then merge it.

haverchuck:
@jcowley It looks like this branch is out of date and needs to be updated, but since it's in your fork I cannot do it.  Would you mind updating it?

haverchuck (approved):

jcowley:
@haverchuck: rebased

PR #567, opened 2021-05-10
fix: Makes AuthCodeDelivery and AdditionalInfo fields nullable 

haverchuck:
*Description of changes:*
Makes AuthCodeDelivery and AdditionalInfo fields nullable on appropriate APIS

### DIFF
--- a/packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
+++ b/packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
@@ -293,7 +293,10 @@ class AmplifyAuthCognitoMethodChannel extends AmplifyAuthCognito {
 
   ResendSignUpCodeResult _formatResendSignUpResponse(
       Map<String, dynamic> res, method) {
-    return CognitoResendSignUpCodeResult(res["codeDeliveryDetails"]);
+    return CognitoResendSignUpCodeResult(AuthCodeDeliveryDetails(
+        attributeName: res["codeDeliveryDetails"]["attributeName"] ?? null,
+        deliveryMedium: res["codeDeliveryDetails"]["deliveryMedium"],
+        destination: res["codeDeliveryDetails"]["destination"]));
   }
 
   SignInResult _formatSignInResponse(Map<String, dynamic> res, String method) {
--- a/packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
+++ b/packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
@@ -22,14 +22,14 @@ extension IsEqual on SignInResult {
   bool isMostlyEqual(SignInResult comparator) {
     return comparator.isSignedIn == isSignedIn &&
         comparator.nextStep!.signInStep == nextStep!.signInStep &&
-        comparator.nextStep!.additionalInfo.length ==
-            nextStep!.additionalInfo.length &&
-        comparator.nextStep!.codeDeliveryDetails.destination ==
-            nextStep!.codeDeliveryDetails.destination &&
-        comparator.nextStep!.codeDeliveryDetails.attributeName ==
-            nextStep!.codeDeliveryDetails.attributeName &&
-        comparator.nextStep!.codeDeliveryDetails.deliveryMedium ==
-            nextStep!.codeDeliveryDetails.deliveryMedium;
+        comparator.nextStep!.additionalInfo?.length ==
+            nextStep!.additionalInfo?.length &&
+        comparator.nextStep!.codeDeliveryDetails!.destination ==
+            nextStep!.codeDeliveryDetails!.destination &&
+        comparator.nextStep!.codeDeliveryDetails!.attributeName ==
+            nextStep!.codeDeliveryDetails!.attributeName &&
+        comparator.nextStep!.codeDeliveryDetails!.deliveryMedium ==
+            nextStep!.codeDeliveryDetails!.deliveryMedium;
   }
 }
 
@@ -55,8 +55,8 @@ void main() {
           "nextStep": {
             "signInStep": "DONE",
             "codeDeliveryDetails": {
-              "deliveryMedium": "EMAIL",
               "attributeName": "email",
+              "deliveryMedium": "EMAIL",
               "destination": "test@test.test"
             }
           }
@@ -69,12 +69,11 @@ void main() {
     var expectation = CognitoSignInResult(
         isSignedIn: false,
         nextStep: AuthNextSignInStep(
-          additionalInfo: {},
-          codeDeliveryDetails: {
-            "deliveryMedium": "EMAIL",
-            "attributeName": "email",
-            "destination": "test@test.test"
-          },
+          additionalInfo: null,
+          codeDeliveryDetails: AuthCodeDeliveryDetails(
+              attributeName: "email",
+              deliveryMedium: "EMAIL",
+              destination: "test@test.test"),
           signInStep: "DONE",
         ));
     var res = await auth.confirmSignIn(
--- a/packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignup_test.dart
+++ b/packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignup_test.dart
@@ -32,7 +32,10 @@ void main() {
           "isSignUpComplete": false,
           "nextStep": {
             "signUpStep": "DONE",
-            "codeDeliveryDetails": {"atttibuteName": "email"}
+            "codeDeliveryDetails": {
+              "attributeName": "email",
+              "destination": "test@test.test"
+            }
           }
         };
       } else {
--- a/packages/amplify_auth_cognito/test/amplify_auth_cognito_resendSignUpCode_test.dart
+++ b/packages/amplify_auth_cognito/test/amplify_auth_cognito_resendSignUpCode_test.dart
@@ -29,7 +29,10 @@ void main() {
     authChannel.setMockMethodCallHandler((MethodCall methodCall) async {
       if (methodCall.method == "resendSignUpCode") {
         return {
-          "codeDeliveryDetails": {"atttibuteName": "email"}
+          "codeDeliveryDetails": {
+            "attributeName": "email",
+            "destination": "test@test.test"
+          }
         };
       } else {
         return true;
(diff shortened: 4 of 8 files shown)


packages/amplify_auth_cognito/ios/Classes/FlutterSignInResult.swift
```diff
@@ -59,7 +65,7 @@ func setCodeDeliveryDetails(res: AmplifyOperation<AuthSignInRequest, AuthSignInR
           if case let .confirmSignInWithSMSMFACode(deliveryDetails, _) = signInResult.nextStep {
             if case let .sms(e) = deliveryDetails.destination {
               deliveryMap["destination"] = e! as String
-              deliveryMap["attributeName"] = ""
+              deliveryMap["attributeName"] = deliveryDetails.attributeKey != nil ? "" : nil
```
cshfang, line 68: QQ: Kind of confused why we want to explicitly set an empty string here and why we need to discern between that and nil

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
```diff
@@ -277,27 +277,39 @@ class AmplifyAuthCognitoMethodChannel extends AmplifyAuthCognito {
         isSignUpComplete: res["isSignUpComplete"],
         nextStep: AuthNextSignUpStep(
             signUpStep: res["nextStep"]["signUpStep"],
-            codeDeliveryDetails: res["nextStep"]["codeDeliveryDetails"],
+            codeDeliveryDetails: res["nextStep"]["codeDeliveryDetails"] != null
+            ? AuthCodeDeliveryDetails(
+              attributeName: res["nextStep"]["codeDeliveryDetails"]["attributeName"] ?? null,
```
cshfang, line 282: Isn't
```dart
attributeName: res["nextStep"]["codeDeliveryDetails"]["attributeName"] ?? null
```
the same as saying if it's `null` make it `null` thus functionally equivalent to just
```dart
attributeName: res["nextStep"]["codeDeliveryDetails"]["attributeName"]
```

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
```diff
@@ -277,27 +277,39 @@ class AmplifyAuthCognitoMethodChannel extends AmplifyAuthCognito {
         isSignUpComplete: res["isSignUpComplete"],
         nextStep: AuthNextSignUpStep(
             signUpStep: res["nextStep"]["signUpStep"],
-            codeDeliveryDetails: res["nextStep"]["codeDeliveryDetails"],
+            codeDeliveryDetails: res["nextStep"]["codeDeliveryDetails"] != null
```
cshfang, line 280: Nit: Can we extract `res["nextStep"]["codeDeliveryDetails"]` into a variable so that we don't have to keep repeating it? I think it'd help with the readability of the code.

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
```diff
@@ -277,27 +277,39 @@ class AmplifyAuthCognitoMethodChannel extends AmplifyAuthCognito {
...
   SignInResult _formatSignInResponse(Map<String, dynamic> res, String method) {
     return CognitoSignInResult(
         isSignedIn: res["isSignedIn"],
         nextStep: AuthNextSignInStep(
             signInStep: res["nextStep"]["signInStep"],
-            codeDeliveryDetails: res["nextStep"]["codeDeliveryDetails"],
+            codeDeliveryDetails: res["nextStep"]["codeDeliveryDetails"] != null
+                ? AuthCodeDeliveryDetails(
+                attributeName: res["nextStep"]["codeDeliveryDetails"]["attributeName"] ?? null,
+                deliveryMedium: res["nextStep"]["codeDeliveryDetails"]["deliveryMedium"] ?? null,
```
cshfang, line 306: Ditto comments from above

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
```diff
@@ -329,10 +341,16 @@ class AmplifyAuthCognitoMethodChannel extends AmplifyAuthCognito {
         isPasswordReset: res["isPasswordReset"],
         nextStep: ResetPasswordStep(
             updateStep: res["nextStep"]["resetPasswordStep"],
-            codeDeliveryDetails: res["nextStep"]["codeDeliveryDetails"],
+            codeDeliveryDetails: res["nextStep"]["codeDeliveryDetails"] != null
+              ? AuthCodeDeliveryDetails(
+              attributeName: res["nextStep"]["codeDeliveryDetails"]["attributeName"] ?? null,
+              deliveryMedium: res["nextStep"]["codeDeliveryDetails"]["deliveryMedium"] ?? null,
```
cshfang, line 347: Ditto

cshfang:
Thanks for doing this! Left some comments. I think they're mostly technically nitpicky comments but I had some questions and am just trying to make sure we continue to raise the bar on our code and keeping it readable.

packages/amplify_auth_cognito/lib/src/CognitoSignUp/CognitoResendSignUpCodeResult.dart
```diff
@@ -16,5 +16,5 @@
 import 'package:amplify_auth_plugin_interface/amplify_auth_plugin_interface.dart';
 
 class CognitoResendSignUpCodeResult extends ResendSignUpCodeResult {
-  CognitoResendSignUpCodeResult({codeDeliveryDetails}) : super(codeDeliveryDetails: codeDeliveryDetails);
+  CognitoResendSignUpCodeResult(codeDeliveryDetails) : super(codeDeliveryDetails);
```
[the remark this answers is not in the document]
haverchuck: In this case, it is largely the only data being returned.  The 'resend code' APIs' sole responsibility is to resend a confirmation code, and so there will always be codeDeliveryDetails.

packages/amplify_auth_cognito/ios/Classes/FlutterSignInResult.swift
haverchuck: Hmm... I think something did not get committed properly.  This is old.

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterSignUpResult.kt
```diff
@@ -15,22 +15,37 @@
...
 
+import com.amplifyframework.auth.AuthCodeDeliveryDetails
 import com.amplifyframework.auth.result.AuthSignUpResult
 import com.google.gson.Gson
 
 data class FlutterSignUpResult(private var raw: AuthSignUpResult) {
   val isSignUpComplete: Boolean = raw.isSignUpComplete
-  val nextStep: Map<String, Any> = setNextStep();
+  val nextStep: Map<String, Any> = setNextStep()
   private fun setNextStep(): Map<String, Any> {
```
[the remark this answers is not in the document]
haverchuck: Good call... done!

packages/amplify_auth_cognito/ios/Classes/FlutterSignInResult.swift
haverchuck: Fixed this - we do not need the value at all.

fjnoyp (approved):

packages/amplify_auth_cognito/ios/Classes/FlutterSignInResult.swift
```diff
@@ -30,14 +30,20 @@ struct FlutterSignInResult  {
...
-      ]
+      var result: Dictionary<String, Any> = ["isSignedIn": self.isSignedIn]
+      var nextStep: Dictionary<String, Any> = ["signInStep": self.signInStep]
+      
+      if (!self.codeDeliveryDetails.isEmpty) {
+        nextStep["codeDeliveryDetails"] = self.codeDeliveryDetails
+      }
+        
+      if (!self.additionalInfo.isEmpty) {
+        nextStep["additionalInfo"] = self.additionalInfo
```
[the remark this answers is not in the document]
cshfang: Had a similar thought here as well since it's something we ended up doing more consistently in the dart modules. But, in this case, it might be just _a little_ trickier since we're dealing with empty dictionaries rather than just nulls. I would be in favor of trying to remove empties/nulls though just because this pattern can grow unwieldy if more properties were to be added.

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
```diff
@@ -273,31 +273,48 @@ class AmplifyAuthCognitoMethodChannel extends AmplifyAuthCognito {
   }
 
   SignUpResult _formatSignUpResponse(Map<String, dynamic> res, method) {
+    var codeDeliveryDetails = res["nextStep"]["codeDeliveryDetails"];
```
cshfang, line 276: Nit: Would it best practice to use `val` here since it's never mutated?

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
```diff
@@ -273,31 +273,48 @@ class AmplifyAuthCognitoMethodChannel extends AmplifyAuthCognito {
...
+        ));
   }
 
+
   ResendSignUpCodeResult _formatResendSignUpResponse(
       Map<String, dynamic> res, method) {
-    return CognitoResendSignUpCodeResult(
-        codeDeliveryDetails: res["codeDeliveryDetails"]);
+    return CognitoResendSignUpCodeResult(AuthCodeDeliveryDetails(
+        attributeName: res["codeDeliveryDetails"]["attributeName"] ?? null,
```
cshfang, line 297: Might've missed one here with the `?? null`

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
```diff
@@ -325,14 +342,21 @@ class AmplifyAuthCognitoMethodChannel extends AmplifyAuthCognito {
...
+    var codeDeliveryDetails = res["nextStep"]["codeDeliveryDetails"];
     return CognitoResetPasswordResult(
         isPasswordReset: res["isPasswordReset"],
         nextStep: ResetPasswordStep(
             updateStep: res["nextStep"]["resetPasswordStep"],
-            codeDeliveryDetails: res["nextStep"]["codeDeliveryDetails"],
+            codeDeliveryDetails: codeDeliveryDetails != null
+              ? AuthCodeDeliveryDetails(
+              attributeName: codeDeliveryDetails["attributeName"] ?? null,
+              deliveryMedium: codeDeliveryDetails["deliveryMedium"] ?? null,
```
cshfang, line 353: `?? null`

Issue #531, opened 2021-04-22
Is it possible to get the HTTP status code of Amplify.API.get() (i.e.)


blattr:
**Describe the bug**
All status code above 299 and 304 (as mentioned in [here](https://docs.amplify.aws/lib/restapi/fetch/q/platform/js)) are internally "rated" as an error and therefore thrown as an exception (see below). In my opinion it should be possible to handle all status codes explicitly because not all of them are a result of failed execution (i.e.404 / no data).

Could you please at least give me a tip where I can get the status code?

This is the code (just the snippet):

```dart
try {
  RestOptions options = RestOptions(apiName: "api", path: "/users/123");
  RestOperation restOperation = Amplify.API.get(restOptions: options);
  RestResponse response = await restOperation.response;
  print(String.fromCharCodes(response.data));
} on ApiException catch (e) {
  print(e);
}
```

... and this the exception (because user 123 doesn't exist).

```text
ApiException(message: The HTTP response status code is [404]., recoverySuggestion: The metadata associated with the response is contained in the HTTPURLResponse. For more information on HTTP status codes, take a look at https://en.wikipedia.org/wiki/List_of_HTTP_status_codes, underlyingException: null)
```

**Platform**

[] Android
[x] iOS

sivakreddyc:
any work around. even i'm facing the same issue
When i try directly from AWS console API test it's successful, trying from flutter app it's showing this message

blattr:
This is a really nasty workaround but it works for me (atm). Please don't blame me - I know, it highly depends on the underlying error message :)

Basically, I try to extract the status code from the error message of the thrown APIException:
```dart
final RegExp statusPattern = RegExp("\\[(\\d{3})\\]");

int _tryToExtractStatusCodeFromMessage(String message) {
  RegExpMatch match = statusPattern.firstMatch(message);
  if (match != null) {
    return int.parse(match.group(1));
  }
  return -1;
}
```

.... also, I am wrapping the HTTP result for the calling code in a vary basic way:




```dart
class HttpResponse {
  final HttpResponseState state;
  final Uint8List bodyBytes;

  HttpResponse(this.state, [this.bodyBytes]);
}

enum HttpResponseState { OK, NO_DATA, ERROR }
```

... putting all parts together looks like this:

```dart
Future<HttpResponse> doGet(String path,
      [Map<String, String> queryParameters]) async {
    try {
      Stopwatch stopwatch = Stopwatch()..start();
[...]

PR #590, opened 2021-05-21
chore(amplify_api): add httpStatusCode property to ApiException when available from REST response

ragingsquirrel3:
Issue: #531 

Add integer property `httpStatusCode` to `ApiException`. In android and iOS, look for the status code in the response/error object and add to the serialized map. In flutter, add the new property to the serialization logic specifically for `ApiException`.

### DIFF
--- a/packages/amplify_api/lib/method_channel_api.dart
+++ b/packages/amplify_api/lib/method_channel_api.dart
@@ -37,10 +37,10 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
       if (e.code == "AmplifyAlreadyConfiguredException") {
         throw AmplifyAlreadyConfiguredException(
             AmplifyExceptionMessages.alreadyConfiguredDefaultMessage,
-            recoverySuggestion: AmplifyExceptionMessages.alreadyConfiguredDefaultSuggestion);
+            recoverySuggestion:
+                AmplifyExceptionMessages.alreadyConfiguredDefaultSuggestion);
       } else {
-        throw AmplifyException.fromMap(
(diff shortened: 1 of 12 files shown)


packages/amplify_api/lib/method_channel_api.dart
```diff
@@ -271,6 +271,11 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
   // ====== GENERAL METHODS ======
 
   ApiException _deserializeException(PlatformException e) {
+    // TEMP question: Would it also be possible/desirable to define more specific error
+    // exception classes here? For example, could we define HTTPNotFoundError (extends ApiError)
+    // and look at the serialized httpStatusCode to define which one to create.
+    // That way, customers would not need `code == 404` etc...
+
```
ragingsquirrel3, line 278: callout here: this might be more work, but arguable a better DX. lmk thoughts

Jordan-Nelson: IMO the DX improvement is minimal. I think it makes sense to get this in as is and then consider it for a future enhancement.

Jordan-Nelson (commented):
LGTM. Did not approve since it still has a a comment with an open question, but I am happy to approve otherwise.

packages/amplify_api/lib/method_channel_api.dart
ragingsquirrel3: Agreed, will revert change in this file.

Jordan-Nelson (approved):

haverchuck (approved):

PR #592, opened 2021-05-24
refactor: Minor refactor of DataStore methods

cshfang:
*Description of changes:*

* Free up `configure` method for native passthrough
* Move observe configuration into its own explicit method
* Corrected observe setup method in swift to call the invokeMethod callback so the call can be properly awaited upon in dart layer
* Broaden `configureDataStore` naming to convey more than model provider being able to be configured through this method
* Allow plugins to be added in parallel instead of in a series
* Update unit tests to correctly assert observe call result


HuiSF (approved):
Thanks for the refactoring and fixing `addPlugins` not returning a `Future` object as expected.

PR #595, opened 2021-05-26
Null safety master

fjnoyp:
*Description of changes:*

This is a temporary PR to run unit tests for merging all of our current null safety changes into the master branch.  

We have yet to determine how the final tagged release of Null Safety will be handled. 


packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/FlutterNextStepUtility.kt
```diff
@@ -0,0 +1,46 @@
...
+
+import com.amplifyframework.auth.AuthCodeDeliveryDetails
+import com.google.gson.Gson
+
+
+fun setNextStep(resultKey: String, resultValue: String, codeDeliveryDetails: AuthCodeDeliveryDetails?, additionalInfo: Map<String, String>?): Map<String, Any> {
+   val result =  mutableMapOf<String, Any>(resultKey to resultValue)
+   val codeDeliveryDetailsResult = mutableMapOf<String, String?>()
+
+   if (codeDeliveryDetails is AuthCodeDeliveryDetails) {
```
Jordan-Nelson, line 26: Are these changes needed for null safety? Anything passed over the method channel could be null. Do any of the changes in kotlin/swift impact nullability on the dart side?

FYI - Some of these refactors are going to conflict with changes I have in [an open PR](https://github.com/aws-amplify/amplify-flutter/pull/601/files#diff-96450b5b67f95088a777b70333ae932561fcf1cb8940438a6c88fc7771b53469R21). Not a huge deal. We can discuss how we want to reconcile it before either PR is merged potentially. I would suggest we limit how much refactoring is done in the null safety branches though (or any change of this scope). There could be a lot of drift between the null safety branches and master.

packages/amplify_auth_cognito/lib/amplify_auth_cognito.dart
```diff
@@ -121,20 +121,20 @@ class AmplifyAuthCognito extends AuthPluginInterface {
   }
 
   Future<UpdateUserAttributeResult> updateUserAttribute(
-      {@required UpdateUserAttributeRequest request}) async {
+      {UpdateUserAttributeRequest? request}) async {
```
Jordan-Nelson, line 124: I think this should be required? I see all the other requests are also optional so maybe I am missing something. I can't think of a scenario in which this should be optional though.

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
```diff
@@ -275,26 +275,26 @@ class AmplifyAuthCognitoMethodChannel extends AmplifyAuthCognito {
 
   @override
   Future<UpdateUserAttributeResult> updateUserAttribute(
-      {@required UpdateUserAttributeRequest request}) async {
+      {UpdateUserAttributeRequest? request}) async {
```
Jordan-Nelson, line 278: Similar to above, I think this should be required. Not sure if I am missing something.

packages/amplify_auth_plugin_interface/lib/amplify_auth_plugin_interface.dart
```diff
@@ -94,19 +94,19 @@ abstract class AuthPluginInterface extends AmplifyPluginInterface {
   }
 
   Future<UpdateUserAttributeResult> updateUserAttribute(
-      {@required UpdateUserAttributeRequest request}) {
+      {UpdateUserAttributeRequest? request}) {
```
Jordan-Nelson, line 97: Same as above (I would expect this to be required)

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/FlutterNextStepUtility.kt
Jordan-Nelson: After looking again, I don't think these will actually conflict since you didn't consume this utility in `FlutterUpdateUserAttributeResult.kt` or `FlutterUpdateUserAttributesResult.kt`. We will have two similar utils but I guess they could be combined later.

Should `FlutterUpdateUserAttributeResult.kt` and `FlutterUpdateUserAttributesResult.kt` be updated? Was there an issue with how the next step was being set or were you just trying to consolidate duplicated logic?

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
```diff
@@ -419,11 +420,18 @@ class AmplifyAuthCognitoMethodChannel extends AmplifyAuthCognito {
...
         isUpdated: res["isUpdated"],
         nextStep: AuthNextUpdateAttributeStep(
             updateAttributeStep: res["nextStep"]["updateAttributeStep"],
-            codeDeliveryDetails: res["nextStep"]["codeDeliveryDetails"],
+            codeDeliveryDetails: codeDeliveryDetails != null
+                ? AuthCodeDeliveryDetails(
+                    attributeName: codeDeliveryDetails["attributeName"] ?? null,
+                    deliveryMedium:
+                        codeDeliveryDetails["deliveryMedium"] ?? null,
+                    destination: codeDeliveryDetails["destination"])
```
Jordan-Nelson, line 433: This is throwing an error, `type 'Null' is not a subtype of type 'String'`. You are checking that codeDeliveryDetails != null, but `codeDeliveryDetails["destination"]` can still be null. I am guessing this is the main reason you changed the kotlin/swift side and added that util (to consistently return nulls, instead of empty maps)?

This can be fixed by updating the swift/kotlin side for FlutterUpdateUserAttributeResult in the same way that the other results (signUp, signIn, resetPassword, etc.) were updated so that it returns null instead of an empty map. This should be straightforward now, but a heads up that it will cause this PR to conflict with https://github.com/aws-amplify/amplify-flutter/pull/601. My first thought was that I could pull some of these changes into https://github.com/aws-amplify/amplify-flutter/pull/601 so that we don't have to worry about conflicts later but I don't think I can really do that since even just the changes on the swift/kt side could be considered breaking (returning null instead of an empty map). If we go this route, I think we will have to just reconcile some conflicts/inconsistencies when you next pull master into this branch.

Another option would be to add checks for empty maps (or null `codeDeliveryDetails["destination"]`) on the dart side (which maybe should be added anyway?) and then I think the changes on the swift/kt side would not be needed. Returning nulls instead of empty maps and consolidating the logic makes sense but it seems a little out of scope for dart null safety upgrade and if I am understanding the purpose of the changes, it seems easy enough to resolve on the dart side.

It is worth noting that with either approach there is more work that should eventually be done to consolidate logic and return more logical values across auth.

packages/amplify_auth_cognito/lib/src/CognitoSession/AWSCognitoUserPoolTokens.dart
```diff
@@ -22,11 +22,9 @@ class AWSCognitoUserPoolTokens {
...
-  AWSCognitoUserPoolTokens.init({@required LinkedHashMap<dynamic, dynamic> tokens}) {
-    if (tokens != null) {
-      this.accessToken = tokens["accessToken"];
-      this.idToken = tokens["idToken"];
-      this.refreshToken = tokens["refreshToken"];
-    }
-  }
+  AWSCognitoUserPoolTokens.init(
+      {required LinkedHashMap<dynamic, dynamic> tokens})
+      : this.accessToken = tokens["accessToken"],
```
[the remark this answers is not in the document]
Jordan-Nelson: This is a scenario in which null safety ideally should force you to do a null check, but it looks like it doesn't force you to do that with dynamics. It would if the type of tokens was `LinkedHashMap<String, String>`, not `LinkedHashMap<dynamic, dynamic>`. I'm not saying we should necessarily change it. `LinkedHashMap<dynamic, dynamic>` might be the correct type. Its definitely something to pay attention to though.

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/FlutterNextStepUtility.kt
fjnoyp: @haverchuck could you provide your insights here.  I think Dustin might have added this as part of his code updates so I think he should have the final say on the best next step.  From what I understand/see I'm guessing these should be consolidated.

packages/amplify_auth_cognito/lib/amplify_auth_cognito.dart
fjnoyp: @haverchuck can you add your insights here?

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
fjnoyp: Thanks.  @haverchuck I noticed we made destination field of AuthCodeDeliveryDetails class required, can we make it nullable like all of the other fields in that class?

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/FlutterNextStepUtility.kt
haverchuck: We were just trying to consolidate code with this file - it was requested as part of code review. 
I think it would make sense to update the FlutterUpdateUserAttributeResult class to use this same utility.

packages/amplify_auth_cognito/lib/amplify_auth_cognito.dart
haverchuck: Yeah I would think this is required, since the customer has to pass the attribute-to-be-updated.

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
haverchuck: Same as above.

packages/amplify_auth_plugin_interface/lib/amplify_auth_plugin_interface.dart
haverchuck: Same as above.

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
haverchuck: @Jordan-Nelson Which method is throwing this error?

Jordan-Nelson: `updateUserAttribute()`. Only on iOS when you update an attribute that doesn't have a confirmation code (anything other than email/phone)

packages/amplify_auth_cognito/lib/src/CognitoSession/AWSCognitoUserPoolTokens.dart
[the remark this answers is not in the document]
Jordan-Nelson: The method throwing the error is Amplify.Auth.fetchAuthSession(). I could only repro this on iOS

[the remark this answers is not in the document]
haverchuck: Working on a fix for this - for now on the platform side.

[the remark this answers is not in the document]
haverchuck: I also think <String, String> is right here - going to play around with it.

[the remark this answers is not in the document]
haverchuck: Hmm - I think we had this as dynamic because of how the data was typed coming over the method channel.

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
haverchuck: Heh - sorry I realized just now I could have seen this for myself.  This makes sense. I'm taking a look.

PR #599, opened 2021-05-28
feat(datastore): Add ModelField ReadOnly support

fjnoyp:
Add support for "readOnly" modelfield property.  Used for new codegen "createdAt" and "lastUpdatedAt" fields that are auto updated in the background. 

### DIFF
--- a/packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterModelField.kt
+++ b/packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterModelField.kt
@@ -39,6 +39,8 @@ data class FlutterModelField(val map: Map<String, Any>) {
     // True if the field is an instance of model.
     private val isModel: Boolean = type.isModel();
 
+    private val isReadOnly: Boolean = map["isReadOnly"] as Boolean
+
     // An array of rules for owner based authorization
     private val authRules: List<FlutterAuthRule>? =
             (map["authRules"] as List<Map<String, Any>>?)?.map { serializedAuthRule ->
@@ -64,6 +66,7 @@ data class FlutterModelField(val map: Map<String, Any>) {
                 .isArray(isArray)
                 .isEnum(isEnum)
                 .isModel(isModel)
+                .isReadOnly(isReadOnly)
 
         if (!authRules.isNullOrEmpty()) {
             builder = builder.authRules(authRules.map { authRule ->
--- a/packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
+++ b/packages/amplify_datastore/ios/Classes/types/model/FlutterModelField.swift
@@ -23,6 +23,7 @@ public struct FlutterModelField {
     public let type: FlutterModelFieldType
     public let isRequired: Bool
     public let isArray: Bool
+    public let isReadOnly: Bool
     public let association: FlutterModelAssociation?
     public let authRules: [FlutterAuthRule]?
     
@@ -62,6 +63,15 @@ public struct FlutterModelField {
                 desiredType: "Bool")
         }
         self.isArray = isArray
+        
+        guard let isReadOnly = serializedData["isReadOnly"] as? Bool
+        else {
+            throw ModelSchemaError.parse(
+                className: "FlutterModelField",
+                fieldName: "isReadOnly",
+                desiredType: "Bool")
+        }
+        self.isReadOnly = isReadOnly
 
         if let inputAssociationMap = serializedData["association"] as? [String : Any]{
             self.association = try FlutterModelAssociation(serializedData: inputAssociationMap)
@@ -136,6 +146,7 @@ public struct FlutterModelField {
             name: name,
             type: try type.convertToNativeModelField(),
             isRequired: isRequired,
+            isReadOnly: isReadOnly,
             isArray: isArray,
             association: association?.convertToNativeModelAssociation(),
             authRules: authRules?.map{
--- a/packages/amplify_datastore_plugin_interface/lib/src/types/models/model_field.dart
+++ b/packages/amplify_datastore_plugin_interface/lib/src/types/models/model_field.dart
@@ -35,6 +35,8 @@ class ModelField {
 
   final bool isArray;
 
+  final bool isReadOnly;
+
   // An array of rules for owner based authorization
   final List<AuthRule> authRules;
 
@@ -45,6 +47,7 @@ class ModelField {
       this.type,
       this.isRequired,
       this.isArray = false,
+      this.isReadOnly = false,
       this.association,
       this.authRules});
 
@@ -53,6 +56,7 @@ class ModelField {
     String type,
     bool isRequired,
     bool isArray,
+    bool isReadOnly,
     ModelAssociation association,
     List<AuthRule> authRules,
   }) {
@@ -61,6 +65,7 @@ class ModelField {
       type: type ?? this.type,
       isRequired: isRequired ?? this.isRequired,
       isArray: isArray ?? this.isArray,
+      isReadOnly: isReadOnly ?? this.isReadOnly,
       association: association ?? this.association,
       authRules: authRules ?? this.authRules,
     );
@@ -72,6 +77,7 @@ class ModelField {
       'type': type.toMap(),
       'isRequired': isRequired,
       'isArray': isArray,
+      'isReadOnly': isReadOnly,
       'association': association?.toMap(),
       'authRules': authRules?.map((x) => x?.toMap())?.toList(),
     };
@@ -86,6 +92,7 @@ class ModelField {
       type: map['type'],
       isRequired: map['isRequired'],
       isArray: map['isArray'],
+      isReadOnly: map['isReadOnly'],
       association:
           map['association'] ?? ModelAssociation.fromMap(map['association']),
       authRules: map['authRules'] ??
@@ -101,7 +108,7 @@ class ModelField {
 
   @override
   String toString() {
-    return 'ModelField(name: $name, type: $type, isRequired: $isRequired, isArray: $isArray, association: $association, authRules: $authRules)';
+    return 'ModelField(name: $name, type: $type, isRequired: $isRequired, isArray: $isArray, isReadOnly: $isReadOnly, association: $association, authRules: $authRules)';
   }
 
   @override
@@ -114,6 +121,7 @@ class ModelField {
         o.type == type &&
         o.isRequired == isRequired &&
         o.isArray == isArray &&
+        o.isReadOnly == isReadOnly &&
         listEquals(o.authRules, authRules);
   }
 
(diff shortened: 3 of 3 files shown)


packages/amplify_datastore/example/ios/unit_tests/resources/modelSchema/model_schema_maps.json
```diff
@@ -1,279 +1,304 @@
...
-                "type": {
-                    "fieldType": "string"
-                },
-                "isRequired": true,
-                "isArray": false
-            }
-        }
-    }
-}
+{
```
HuiSF, line 1: I think the IDE accidentally converted this file into CRLF mode. (most files are in LF mode in the codebase). Could you double check please?

packages/amplify_datastore/example/lib/models/Blog.dart
```diff
@@ -53,15 +55,33 @@ class Blog extends Model {
...
+        _posts = posts,
+        _createdAt = createdAt,
+        _updatedAt = updatedAt;
+
+  factory Blog(
+      {String? id,
+      required String name,
+      List<Post>? posts,
+      TemporalDateTime? createdAt,
+      TemporalDateTime? updatedAt}) {
```
HuiSF, line 78: A general question: are `createdAt` and `updatedAt` managed by DynamoDB? If this is true, should we expose them as Model fields and allow manual modification?

packages/amplify_datastore/example/lib/models/Blog.dart
```diff
@@ -120,6 +164,8 @@ class Blog extends Model {
       fieldName: "posts",
       fieldType: ModelFieldType(ModelFieldTypeEnum.model,
           ofModelName: (Post).toString()));
+  static final QueryField CREATEDAT = QueryField(fieldName: "createdAt");
+  static final QueryField UPDATEDAT = QueryField(fieldName: "updatedAt");
```
HuiSF, line 168: I may be wrong I think `createdAt` and `updatedAt` should not be query fields.

packages/amplify_datastore/example/lib/models/Blog.dart
```diff
@@ -137,6 +183,18 @@ class Blog extends Model {
         isRequired: false,
         ofModelName: (Post).toString(),
         associatedKey: Post.BLOG));
+
+    modelSchemaDefinition.addField(ModelFieldDefinition.field(
+        key: Blog.CREATEDAT,
```
HuiSF, line 188: Related to my query field comment, `createdAt` and `updatedAt` should not have primary key.

packages/amplify_datastore/test/test_models/Post.dart
```diff
@@ -40,15 +42,42 @@ class Post extends Model {
...
+              .codeGenRequiredFieldForceCastExceptionMessage,
+          recoverySuggestion: DataStoreExceptionMessages
+              .codeGenRequiredFieldForceCastRecoverySuggestion,
+          underlyingException: e.toString());
+    }
   }
 
   int get rating {
-    return _rating!;
+    try {
```
HuiSF, line 58: Thanks for updating the testing models

HuiSF (commented):
Thanks for adding the support of readonly fields 👍
I added a few comments and also question about query fields.

packages/amplify_datastore/example/lib/models/Blog.dart
HuiSF: Update: compared native Android and iOS models, `createdAt` and `updatedAt` are not query fields.

packages/amplify_datastore/example/ios/unit_tests/resources/modelSchema/model_schema_maps.json
fjnoyp: Hey this is quite strange.  I did not find a reason for why this happened. 

Do you know how I might verify if it's in CRLF mode?  I think that if my computer / IDE was making those changes that there might be other issues in all of my other commits / files changed though. 

Looking through the file history you'll see other commits have had the same issue as well: 

image

packages/amplify_datastore/example/lib/models/Blog.dart
fjnoyp: Hi @HuiSF wanted to make sure I understand what you're saying here.  So you're saying we shouldn't have createdAt and updatedAt as parameters in the factory Blog constructor method here because the user technically cannot modify them as they are automatically populated by AppSync?

I think we could make the following change - what are your thoughts?

```
  factory Blog(
      {String? id,
      required String name,
      List<Post>? posts}) {
    return Blog._internal(
        id: id == null ? UUID.getUUID() : id,
        name: name,
        posts: posts != null ? List<Post>.unmodifiable(posts) : posts,
        createdAt: null,
        updatedAt: null);
  }
```

packages/amplify_datastore/example/lib/models/Blog.dart
fjnoyp: Our ModelFieldDefinition.field method is a bit strange - it uses the QueryField.fieldName property to get the name of the field being passed.  I need to update the method to be able to take the field name directly instead of needing a QueryField to be initialized. 

```
 static ModelFieldDefinition field(
      {required QueryField key,
      bool isRequired = true,
      bool isArray = false,
      ModelFieldType ofType = const ModelFieldType(ModelFieldTypeEnum.string),
      ModelAssociation? association,
      List<AuthRule>? authRules}) {
    return ModelFieldDefinition(
        name: key.fieldName,
```

packages/amplify_datastore/example/ios/unit_tests/resources/modelSchema/model_schema_maps.json
HuiSF: The json file was in CRLF mode, and was converted into LF mode to align with codebase in commit `ef250295`.
Could you check you git config if it uses CRLF mode? And the IDE you are using (e.g. prettier in VSCode has option to reformat with LF vs CRLF)

packages/amplify_datastore/example/lib/models/Blog.dart
HuiSF: > because the user technically cannot modify them as they are automatically populated by AppSync?

Yes this is what I meant. These two fields should not be assignable/modifiable via constructor (including internal constructor). For Flutter use case, I think they are only assignable in `Model.fromJson` method (which is technically used internally by Flutter plugin to deserialize model data)

So I think these two fields should not be listed in the `_internal` constructor as well...

packages/amplify_datastore/example/ios/unit_tests/resources/modelSchema/model_schema_maps.json
fjnoyp: Alright where in git config do I check for this CRLF mode.  I don't see any features in the config file that would be enabling that.  I feel like if that was the case I would have had issues in all my other commits though right? 

I installed prettier in VSCode, the default is set to LF mode.  I did auto format on save and switched between CRLF and LF modes but git did not register any file changes so I'm a bit unsure what's going on there.  

How would you switch the file to LF mode?

packages/amplify_datastore/example/lib/models/Blog.dart
fjnoyp: Alright I'll update it like that.  We'll need to talk to codegen team to include those changes as well.

fjnoyp: We should wait on this PR until Aaron confirms he adds those codegen changes in then.  I'll contact him now about that.  Should be quick to make if he has the bandwidth now.

packages/amplify_datastore/example/lib/models/Blog.dart
fjnoyp: Same change needed, I'll need to talk to Aaron to update codegen here.

packages/amplify_datastore/example/lib/models/Blog.dart
fjnoyp: Removing the createdAt and updatedAt fields from the constructor will cause issue with our existing copyWith implementation. 

```
  Blog copyWith(
      {String? id,
      String? name,
      List<Post>? posts,
      TemporalDateTime? createdAt,
      TemporalDateTime? updatedAt}) {
    return Blog(
        id: id ?? this.id,
        name: name ?? this.name,
        posts: posts ?? this.posts,
        createdAt: createdAt ?? this.createdAt,
        updatedAt: updatedAt ?? this.updatedAt);
  }
```

We could switch to using the internal constructor instead so that the public constructor doesn't have createdAt and updatedAt.  However we would need to keep internal constructor to have the createdAt and updatedAt timestamps as parameters then.

packages/amplify_datastore/example/lib/models/Blog.dart
fjnoyp: QueryFields removed

packages/amplify_datastore/example/lib/models/Blog.dart
fjnoyp: QueryFields removed

packages/amplify_datastore/example/lib/models/BelongsToModel.dart
```diff
@@ -95,15 +109,30 @@ class BelongsToModel extends Model {
...
 
-  BelongsToModel copyWith({String? id, String? name, ChildModel? child}) {
-    return BelongsToModel(
-        id: id ?? this.id, name: name ?? this.name, child: child ?? this.child);
+  BelongsToModel copyWith(
+      {String? id,
+      String? name,
+      ChildModel? child,
+      TemporalDateTime? createdAt,
+      TemporalDateTime? updatedAt}) {
```
HuiSF, line 129: As `createdAt` and `updatedAt` are readonly, I think they are not supposed to be parameters of the `copyWith` method, as a user can pass in any value to override the original value.

packages/amplify_datastore/test/test_models/Blog.dart
```diff
@@ -20,121 +20,156 @@ import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_inte
 import 'package:collection/collection.dart';
```
HuiSF, line 20: Probably will need to reformat this codegen generate file to pass format check in CI. Same to other "test models".

packages/amplify_datastore/example/lib/models/BelongsToModel.dart
fjnoyp: Thanks good point.  I asked Aaron to update the amplify codegen to not have these parameters in the copyWith method anymore

HuiSF (approved):
Thanks for adding  the readonly field support!

packages/amplify_datastore/test/test_models/Blog.dart
fjnoyp: Thanks I've reformatted all the files to pass the format check

Jordan-Nelson (approved):

PR #601, opened 2021-05-28
feat(auth): add updateUserAttributes (batch)

Jordan-Nelson:
*Issue #, if available:* #193

*Description of changes:*
- add new auth method for updateUserAttributes (batch update of user attributes)

**NOTE: This requires a fix on Android and should not be merged**
- ~~A PR is open for that fix~~ The PR for the fix has been merged: https://github.com/aws-amplify/amplify-android/pull/1339.
- Once Android releases a new version then the version needs to be bumped in this PR.


packages/amplify_auth_cognito/ios/Classes/FlutterUpdateUserAttributesRequest.swift
```diff
@@ -0,0 +1,48 @@
...
+struct FlutterUpdateUserAttributesRequest {
+    var attributes: Array<AuthUserAttribute>
+    
+    init(dict: NSMutableDictionary) {
+        self.attributes = (dict["attributes"] as! Array<Dictionary<String, String>>).map {
+            return createAuthUserAttribute(key: $0["userAttributeKey"]!, value: $0["value"]!);
+        }
+    }
+    
+    static func validate(dict: NSMutableDictionary) throws {
```
cshfang, line 30: Same general question about sharing code here if possible?

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
```diff
@@ -313,6 +313,25 @@ class AmplifyAuthCognitoMethodChannel extends AmplifyAuthCognito {
...
+  @override
+  Future<Map<String, UpdateUserAttributeResult>> updateUserAttributes(
+      {@required UpdateUserAttributesRequest request}) async {
+    Map<String, UpdateUserAttributeResult> res;
+    try {
+      final Map<String, dynamic> data =
+          await _channel.invokeMapMethod<String, dynamic>(
+        'updateUserAttributes',
+        <String, dynamic>{
+          'data': request != null ? request.serializeAsMap() : null,
```
cshfang, line 325: Nit: I think this is effectively the same as
```suggestion
          'data': request?.serializeAsMap(),
```

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
```diff
@@ -429,6 +448,16 @@ class AmplifyAuthCognitoMethodChannel extends AmplifyAuthCognito {
                 : {}));
   }
 
+  Map<String, UpdateUserAttributeResult> _formatUpdateUserAttributesResponse(
+      Map<String, dynamic> res) {
+    return res.map((key, value) => MapEntry(
+        key,
+        _formatUpdateUserAttributeResponse(
+            new Map<String, dynamic>.from(value))));
```
cshfang, line 456: Nit: Think we should omit the `new` keyword (https://dart.dev/guides/language/effective-dart/usage#dont-use-new)
```suggestion
           Map<String, dynamic>.from(value))));
```

cshfang (commented):
I don't really know enough about Auth so I'm afraid I left more questions than comments. It *seems* pretty straightforward to me and most of my questions were around whether it's possible for us to be more DRY and a few nitpicky things here and there. It lgtm otherwise!

packages/amplify_auth_cognito/example/lib/Widgets/ViewUserAttributes.dart
```diff
@@ -2,6 +2,7 @@ import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
 import 'package:amplify_flutter/amplify.dart';
 import 'package:flutter/material.dart';
 
+import 'UpdateMultipleUserAttributes.dart';
```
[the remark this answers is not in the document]
Jordan-Nelson: Yeah thats fair. I didn't notice how closely the widget names were matching the method names for all the other widgets. Updated.

packages/amplify_auth_plugin_interface/lib/src/Session/UpdateUserAttributesRequest.dart
```diff
@@ -0,0 +1,36 @@
...
+
+import 'package:flutter/foundation.dart';
+import 'AuthUserAttribute.dart';
+
+/// Encapsulates parameters for a update user attributes operation
+class UpdateUserAttributesRequest {
+  /// The list of user attribute to update
+  final List<AuthUserAttribute> attributes;
+
+  // ignore: public_member_api_docs
```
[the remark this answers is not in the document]
Jordan-Nelson: I updated this and `serializeAsMap`.

I could go either way on something like this. On one hand, this isn't meant for public consumption (no harm in it being public, customers just shouldn't need to use it directly) and I don't really want to set a standard of adding documentation that isn't valuable. On the other hand, it does look make the API docs look incomplete to have so many public things undocumented.

Also worth noting that we do lose 10 points on pub.dev for having less than 20% of public APIs documented. I would assume simply adding ignore does not exclude this from that calculation.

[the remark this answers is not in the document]
cshfang: Fwiw, I actually think if it's meant to be internal, we should maybe emphasize that it's not meant to be publicly consumed - i.e. it's subject to change so use at your own risk. I actually wonder if we would benefit from doing something like that across the board.

Generally speaking, I don't like to add documentation that isn't all that meaningful either but if I consider the developer experience it seems like, at worst, extra documentation is annoying but, at best, really helps me navigate foreign code. Since our code is largely intended to be consumed by others, even seemingly low value documentation could be surprisingly helpful... for example, if someone didn't understand Dart at all and was looking at the constructor syntax for the first time, they may find it useful to know the constructor is the constructor.

Anyway, sorry for the long winded way of saying thanks for adding the doc 😂

packages/amplify_auth_cognito/example/lib/Widgets/UpdateMultipleUserAttributes.dart
```diff
@@ -0,0 +1,171 @@
...
+class _UserAttributeController {
+  TextEditingController keyController;
+  TextEditingController valueController;
+  _UserAttributeController({String keyValue}) {
+    keyController = TextEditingController(text: keyValue);
+    valueController = TextEditingController();
+  }
+}
+
+// ignore_for_file: public_member_api_docs
```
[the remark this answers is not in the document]
cshfang: Yeah, it is a nit comment after all. Though I do think there's value in documenting example apps since they are _supposed_ to nudge users towards best practices in theory? It just seems like it would potentially be a good experience to me if I'm perusing this and the code was educational.

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterUpdateUserAttributesRequest.kt
```diff
@@ -0,0 +1,53 @@
...
+import com.amplifyframework.auth.AuthUserAttribute
+
+data class FlutterUpdateUserAttributesRequest(val map: HashMap<String, *>) {
+
+    val attributes: List<AuthUserAttribute> = (map["attributes"] as List<HashMap<*, *>>)
+            .map { createAuthUserAttribute(it["userAttributeKey"] as String, it["value"] as String) }
+
+    companion object {
+        private const val validationErrorMessage: String = "UpdateUserAttributesRequest Request malformed."
+        fun validate(req: HashMap<String, *>?) {
```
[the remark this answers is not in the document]
Jordan-Nelson: I extracted the attribute validation into `validateUserAttribute` used on line 41. Thanks for the suggestion 👍

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/utils/AuthCodeDeliveryDetailsSerialization.kt
```diff
@@ -0,0 +1,28 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+
+package com.amazonaws.amplify.amplify_auth_cognito.utils
+
+import com.amplifyframework.auth.AuthCodeDeliveryDetails
+
+fun serializeAuthCodeDeliveryDetails(deliveryDetails: AuthCodeDeliveryDetails?): Map<String, Any> {
```
Jordan-Nelson, line 21: Note: Right now this is only used for update user attributes (single update and batch update) but this should be reusable for sign up, sign in, reset password, and resend sign up code. I was hesitant to refactor all of those in this PR though.

packages/amplify_auth_cognito/ios/Classes/FlutterUpdateUserAttributesRequest.swift
Jordan-Nelson: Updated.

packages/amplify_auth_cognito/ios/Classes/Utils/AuthCodeDeliveryDetailsSerialization.swift
```diff
@@ -0,0 +1,43 @@
...
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import Foundation
+import Amplify
+
+func serializeAuthCodeDeliveryDetails(deliveryDetails: AuthCodeDeliveryDetails) -> [String: String] {
```
Jordan-Nelson, line 19: Note: Same note I made on the Android side - Right now this is only used for update user attributes (single update and batch update) but this should be reusable for sign up, sign in, reset password, and resend sign up code. I was hesitant to refactor all of those in this PR though.

packages/amplify_auth_cognito/ios/Classes/Utils/UserAttributeDeserialization.swift
```diff
@@ -50,6 +50,12 @@ func createAuthUserAttributeKey(keyName: String) -> AuthUserAttributeKey {
     case "preferred_username":
         return AuthUserAttributeKey.preferredUsername
     default:
-        return AuthUserAttributeKey.custom(keyName)
+        // amplify-ios does not currently include enums for these keyNames
```
Jordan-Nelson, line 53: Note: This isn't needed for the new functionality, but it addresses a related bug.

packages/amplify_auth_cognito/example/lib/Widgets/UpdateUserAttributes.dart
```diff
@@ -0,0 +1,165 @@
...
+class _UserAttributeController {
+  TextEditingController keyController;
+  TextEditingController valueController;
+  _UserAttributeController({String keyValue}) {
+    keyController = TextEditingController(text: keyValue);
+    valueController = TextEditingController();
+  }
+}
+
+// ignore_for_file: public_member_api_docs
```
cshfang, line 15: Nit: Can we move file level ignores to the top of the file?

packages/amplify_auth_cognito/example/lib/Widgets/UpdateUserAttributes.dart
```diff
@@ -0,0 +1,165 @@
...
+      appBar: AppBar(
+        title: Text('Update Attribute'),
+      ),
+      body: Padding(
+        padding: const EdgeInsets.all(8),
+        child: Form(
+          key: _formKey,
+          child: ListView(
+            children: [
+              ..._userAttributeControllers.map((element) {
```
cshfang, line 104: Nit: It might be clearer (debatable) and maybe microscopically more performant to just have
```dart
children: _userAttributeControllers.map().toList()
```
as opposed to spreading the Iterable into an list

packages/amplify_auth_cognito/ios/Classes/FlutterUpdateUserAttributeResult.swift
```diff
@@ -18,99 +18,18 @@ import Foundation
...
-          if case .done = updateResult.nextStep {
-             return  "DONE"
-          }
-          if case .confirmAttributeWithCode = updateResult.nextStep {
-            return "CONFIRM_ATTRIBUTE_WITH_CODE"
-          }
-        case .failure:
-           return "ERROR"
+    func toJSON() -> Dictionary<String, Any> {
+        return self.result == nil ? [:] : serializeAuthUpdateAttributeResult(result: self.result!)
```
cshfang, line 33: Nit suggestion/question: What do you think about moving the nil check into the `serializeAuthUpdateAttributeResult` function itself and have it return the empty dict on a nil result instead? It seems like having the nil guard there would arguably make it a bit more reusable.

packages/amplify_auth_cognito/ios/Classes/Utils/UserAttributeSerialization.swift
```diff
@@ -0,0 +1,49 @@
+/*
```
cshfang, line 1: 💯

cshfang (approved):
Lgtm. Left some minor comments but all pretty nit

packages/amplify_auth_cognito/ios/Classes/FlutterUpdateUserAttributeResult.swift
Jordan-Nelson: I think that is reasonable. I put it here because I viewed it as specific to this use case. In this case, result is only nil if the AmplifyOperation is a failure, and it doesn't really matter what this returns in the case of a failure.

It is admittedly a bit hacky though to just return an empty dictionary in the case of a failure, knowing it won't be used. I don't think there is a clear "correct" return value in the case of a failure, but an empty dict probably isn't appropriate. Looking at what we do for the other results in auth, it looks like we generally return a dictionary that has the same attributes as the success case with all the values as empty string / nil / false. Again, not sure this is clearly "correct" but consistency would be preferred.

tl;dr - I will update serializeAuthUpdateAttributeResult to allow for nil and follow the pattern used elsewhere for serializing results in the case of failures 👍

packages/amplify_auth_cognito/example/lib/Widgets/UpdateUserAttributes.dart
Jordan-Nelson: There are two buttons in the ListView as well on lines 151/155

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterConfirmUserAttributeRequest.kt
```diff
@@ -15,6 +15,7 @@
 
 package com.amazonaws.amplify.amplify_auth_cognito.types
 
+import com.amazonaws.amplify.amplify_auth_cognito.utils.createAuthUserAttributeKey
```
haverchuck, line 18: How was it previously working without this import?

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterUpdateUserAttributeRequest.kt
```diff
@@ -33,21 +35,14 @@ data class FlutterUpdateUserAttributeRequest(val map: HashMap<String, *>) {
...
                 val attribute = req["attribute"] as HashMap<*, *>;
-                if (!attribute.containsKey("value")) {
-                    throw InvalidRequestException(validationErrorMessage, ExceptionMessages.missingAttribute.format( "value" ))
-                } else if (!attribute.containsKey("userAttributeKey")) {
-                    throw InvalidRequestException(validationErrorMessage, ExceptionMessages.missingAttribute.format( "userAttributeKey" ))
-                } else if (attribute["value"] !is String) {
-                    // Android SDK expects a string for user attr values, regardless of the configuration in cognito
-                    throw InvalidRequestException(validationErrorMessage, "Attribute value is not a String.")
-                }
+                validateUserAttribute(attribute, validationErrorMessage)
```
haverchuck, line 45: Good change

packages/amplify_auth_cognito/ios/Classes/Utils/AuthCodeDeliveryDetailsSerialization.swift
haverchuck: Agreed - this should be a seperate PR.  There will be some conflicts with NS, so maybe we can make these changes at that time.

haverchuck (approved):
Left some questions/comments, but overall looks great.

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterConfirmUserAttributeRequest.kt
Jordan-Nelson: It was in the same "types" package before. I created a new package for utils.

PR #603, opened 2021-05-30
Use "flutter pub" in melos.yaml (and everywhere else)

mrhota:
*Issue #, if available:*

None.

*Description of changes:*

`melos.yaml` doesn't work if you don't have `pub` installed globally. I'm a newcomer to Dart and Flutter, and it seems that the global `pub` command is deprecated now: https://dart.dev/tools/pub/cmd


Salakar (approved):

melos.yaml
```diff
@@ -42,14 +42,14 @@ scripts:
...
+      dart pub global run tuneup check
 
   lint:pub: >
     melos exec -c 5 --fail-fast --no-private --ignore="*example*" -- \
-      pub publish --dry-run
+      dart pub publish --dry-run
 
   postbootstrap: >
-    (tuneup --version || pub global activate tuneup) && melos run copy_dummy_config
+    (tuneup --version || dart pub global activate tuneup) && melos run copy_dummy_config
```
cshfang, line 52: Thanks for submitting this change. I think it's a good change considering standalone `pub` is indeed deprecated 😅  but I wonder if the change should be using `flutter pub` instead:
1. This is, after all, intended to be used with Flutter specifically rather than a generic Dart project. Though the root directory does not technically contain Flutter-specific code, I would interpret the recommendation per the docs (https://dart.dev/tools/pub/cmd) applicable to this project.
2. The `dart` command does not necessarily need to be exposed to the user's shell.
3. We apparently already have scripts relying on the `flutter` command.

mrhota: I think the trouble is which tool you have used to activate Melos and/or tuneup. I considered using flutter for these, but I can’t remember why I settled on dart instead

cshfang: > I think the trouble is which tool you have used to activate Melos and/or tuneup

Hmm I see. Could you help me understand what you mean by this better so we can attempt to address this more holistically? If the `flutter pub` approach doesn't work, it would be great if we knew why so we can try to shore that up.

mrhota: Oh I mean simply that if you use `dart pub global activate melos` to install that tool, you cannot use `flutter pub global run melos` to run it, and vice versa.

cshfang: Ah. Given that it appears to be an issue with either path chosen, would the arguments for `flutter pub` I pointed out be valid after all? What are your thoughts on perhaps updating the melos section of the [CONTRIBUTING.md](https://github.com/aws-amplify/amplify-flutter/blob/master/CONTRIBUTING.md) to nudge users towards installing via the `flutter` command?

mrhota:
@cshfang I went ahead and changed the circleci config file to use `flutter pub` instead of `pub _`. Not sure if that'll break builds, but it doesn't seem like it would.

ragingsquirrel3:
I'm not sure if you are able to see the unit test circle CI output but the ios unit tests failed during the setup part. Here is relevant output:

```
Linking workspace packages...
  > SUCCESS

Generating IntelliJ IDE files...
  > SUCCESS
Running postbootstrap script...

melos run postbootstrap
   └> (tuneup --version || flutter pub global activate tuneup) && melos run copy_dummy_config
       └> RUNNING

/bin/sh: line 1: tuneup: command not found
Resolving dependencies...
+ analysis_server_lib 0.1.10 (0.2.0 available)
+ args 1.6.0 (2.1.1 available)
+ charcode 1.2.0
+ cli_util 0.2.0 (0.3.0 available)
+ collection 1.15.0
+ intl 0.16.1 (0.17.0 available)
+ logging 0.11.4 (1.0.1 available)
+ path 1.8.0
+ source_span 1.8.1
+ string_scanner 1.1.0
+ term_glyph 1.2.0
+ tuneup 0.3.7
+ yaml 2.2.1 (3.1.0 available)
Downloading tuneup 0.3.7...
Downloading logging 0.11.4...
Downloading analysis_server_lib 0.1.10...
Downloading intl 0.16.1...
Precompiling executables...
Precompiled tuneup:tuneup.
Installed executable tuneup.
Activated tuneup 0.3.7.
melos run copy_dummy_config
   └> melos exec --scope="*example*" --fail-fast -- \  cp -n ../../../.circleci/dummy_amplifyconfiguration.dart lib/amplifyconfiguration.dart | true
       └> RUNNING


melos run copy_dummy_config
   └> melos exec --scope="*example*" --fail-fast -- \  cp -n ../../../.circleci/dummy_amplifyconfiguration.dart lib/amplifyconfiguration.dart | true
       └> SUCCESS

Too long with no output (exceeded 10m0s): context deadline exceeded
```

I don't understand exactly why, but it seems like it's hanging when running the postbootstrap. I was able to repro this in my local env, not sure exactly why. I am able to run the copy_dummy_config. Are you able to repro this locally? Any idea why it might be hanging?

mrhota:
@ragingsquirrel3 I have been scratching my head about this failure.

Is it significant that the iOS executor uses `bash --login -eo pipefail` where the android executor uses `bash -eo pipefail`?

melos.yaml
```diff
@@ -71,14 +71,14 @@ scripts:
 
   analyze: >
     melos exec -c 1 --no-private --ignore="*example*" -- \
-      pub global run tuneup check
+      flutter pub global run tuneup check
```
ragingsquirrel3, line 74: optional: If removing the global tuneup installation as suggested in below comment, it might make sense to change this task to use deeper YAML syntax with a description block like https://github.com/aws-amplify/amplify-flutter/blob/master/melos.yaml#L51 and include a note in the description that this requires tuneup installed, and to install it, run `flutter pub global activate tuneup`.

.circleci/config.yml
```diff
@@ -37,7 +37,7 @@ commands:
       - run:
           name: Install and set up melos
           command: |
-            pub global activate melos 0.4.5
+            flutter pub global activate melos 0.4.5
```
ragingsquirrel3, line 40: If you remove the tuneup installation from YAML as suggested in below comments, it might make sense to add a line here `flutter pub global activate tuneup` so that it's included in CI env if/when we eventually re-integrate the analyze tasks in the build.

ragingsquirrel3 (changes requested):
Overall looks good, just suggested a way to get around build errors. IMO, better not to install things globally in bootstrap anyway.

ragingsquirrel3:
@mrhota that's a good hypothesis. Frankly, I don't understand that behavior seen in builds 🤷 , so I just suggested a workaround in changes.

mrhota:
@ragingsquirrel3 I have isolated the problem to `flutter pub global activate __anything__`. If we use `dart pub global activate __anything__`, the melos script completes successfully (I repro'd even when trying to globally activate `dartdoc` for example)

I rather like your suggestion of just doing an end-run around the `flutter pub global activate` issue for now.

.circleci/config.yml
mrhota: good idea!

ragingsquirrel3 (approved):
Thanks for the extra info. LGTM and thanks again for improving DX for contributors to this repo. Please wait for a 2nd approval before merging this.

cshfang (approved):
Lgtm :shipit:

mrhota:
@cshfang and @ragingsquirrel3 thank you for the review! It looks like a repo owner will have to merge.

Issue #598, opened 2021-05-27
iOS does not support all of the standard user attributes that Android does


Jordan-Nelson:
**Describe the bug**
Android support more standard user attributes than iOS. Android supports the standard attributes "profile", "updated_at", "website", and "zoneinfo". These attributes are not supported on iOS. If a customer attempts to add these on iOS, amplify-flutter will attempt to add them as custom attributes.

**To Reproduce**
1. Attempt to add create a new user with one of these attributes OR update the attribute on an existing user

**Example w/ Sign Up**
```dart
var userAttributes = {
  // works on Android, but not iOS
  'website': 'www.example.com',
};
var res = await Amplify.Auth.signUp(
    username: 'testuser001',
    password: 'password123',
    options: CognitoSignUpOptions(userAttributes: userAttributes));
```
**Example w/ updateAttribute**
```dart
await Amplify.Auth.updateUserAttribute(
  // works on Android, but not iOS
  userAttributeKey: 'website',
  value: 'www.example.com',
);
```

**Expected behavior**
These four attributes should be supported when targeting iOS as well.

**Platform**
Amplify Flutter current supports iOS and Android.  This issue is reproducible in (check all that apply):
[] Android
[x] iOS

**Additional context**
amplify-ios currently doesn't support these attributes. There may be a workaround that could be implemented in amplify-flutter, which is to add these four attributes using `.unknown()` instead of `.custom()`.

Jordan-Nelson:
Issue opened currently in amplify-android for this: https://github.com/aws-amplify/amplify-android/issues/787

lawmicha:
if amplify flutter adds them as custom, that will actually end up prepending with "custom:" prefix which would mean an app on Android will send "profile", "zoneinfo", etc. and an app on iOS will send "custom:profile", "custom:zoneinfo"

lawmicha:
another scenario is, what happens if i use amplify flutter and create a .custom attribute? is this possible? If so, that would mean iOS sends it as "custom:[custom_value]" and Android sends it as "[custom_value]"

lawmicha:
> amplify-ios currently doesn't support these attributes. There may be a workaround that could be implemented in amplify-flutter, which is to add these four attributes using .unknown() instead of .custom().

oh i forgot about this, so iOS has `unknown` and `custom` to support both scenarios. I think iOS should update it's enum case to include the extra fields that Android has to be in parity

Jordan-Nelson:
[...]

Issue #576, opened 2021-05-13
Need exactly the same completeNewPassword  function as the JavaScript Library


rhinozD:
Hi,
Our case is similar to the one below: 

https://github.com/aws-amplify/amplify-flutter/issues/560#issuecomment-834057529

but our attribute is `phone_number`.  Furthermore, we use `SMS_MFA_CODE` so the `phone_number` attribute must be set before login.
As your comment in the issue above, you are adding the `updateUserAtributes` function. Does it have to be done after login or not? If that's the case then it doesn't match our case.
Also is there any alternative solution?
Our customer wants an answer as soon as possible so please reply as soon as possible to us.
Thanks in advance for your help.

@Amplifiyer @haverchuck

rhinozD:
Is there any update? 
Will it be resolved in the near future?

haverchuck:
@rhinozD What exactly is the use case that is not satisfied?  Are you unable to login as a user who is in the 'FORCE_CHANGE_PASSWORD' state?  If so, you can use the confirmSignIn API to pass in the new password.  If this is not the problem you are having, can you lay out the specific API that is not working?

rhinozD:
@haverchuck 
Thanks for the reply.

> Are you unable to login as a user who is in the 'FORCE_CHANGE_PASSWORD' state? If so, you can use the confirmSignIn API to pass in the new password. 

I tried confirmSignIn method. It works fine except that I want to add the `phone_number` attribute too.
Like this:
```
Map<String, String> userAttributes = { 'phone_number': phoneNumber, };
var options = CognitoConfirmSignInOptions(clientMetadata: userAttributes);
Amplify.Auth.confirmSignIn( confirmationValue: newPassword, options: options);
```
As I said above,
since I set SMS MFA as required the flow will be like this: 
`confirmSignIn → receive SMS text with OTP code → go to SMS OTP input screen → home Page`
so I need to set the `phone_number` attribute when doing confirmSignIn - What the javascript Library can do using the `completeNewPassword` function:
```
       Auth.completeNewPassword(
            user,               // the Cognito User Object
            newPassword,       // the new password
            // OPTIONAL, the required attributes
            {
              phone_number: '1234567890'
            }
```
Note that my application does not have the SignUp flow. Admin will run the batch and import users into the Cognito User Pool.
[...]

PR #607, opened 2021-06-01
fix(amplify_auth_cognito): adds userAttributes to confirmSignIn

haverchuck:
*Issue #, if available:*
#576 
#598 

*Description of changes:*
Adds userAttributes to CognitoConfirmSignInOptions. 

### DIFF
--- a/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterConfirmSignInRequest.kt
+++ b/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterConfirmSignInRequest.kt
@@ -17,6 +17,8 @@ package com.amazonaws.amplify.amplify_auth_cognito.types
 
 import com.amazonaws.amplify.amplify_core.exception.ExceptionMessages
 import com.amazonaws.amplify.amplify_core.exception.InvalidRequestException
+import com.amplifyframework.auth.AuthUserAttribute
+import com.amplifyframework.auth.AuthUserAttributeKey
 import com.amplifyframework.auth.cognito.options.AWSCognitoAuthConfirmSignInOptions
 
 data class FlutterConfirmSignInRequest(val map: HashMap<String, *>) {
@@ -28,6 +30,15 @@ data class FlutterConfirmSignInRequest(val map: HashMap<String, *>) {
 
     if(rawOptions?.get("clientMetadata") != null)
       options.metadata(rawOptions["clientMetadata"] as HashMap<String, String>);
+
+    if(rawOptions?.get("userAttributes") != null) {
+      var attributes = mutableListOf<AuthUserAttribute>();
+      var rawAttributes = rawOptions["userAttributes"] as HashMap<String, String>;
+      rawAttributes.forEach { (key, value) ->
+        attributes.add(createAuthUserAttribute(key, value))
+      }
+      options.userAttributes(attributes);
+    }
 
     return options.build();
   }
--- a/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterConfirmSignInRequest.kt
+++ b/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterConfirmSignInRequest.kt
@@ -31,14 +31,12 @@ data class FlutterConfirmSignInRequest(val map: HashMap<String, *>) {
     if(rawOptions?.get("clientMetadata") != null)
       options.metadata(rawOptions["clientMetadata"] as HashMap<String, String>);
 
-    if(rawOptions?.get("userAttributes") != null) {
-      var attributes = mutableListOf<AuthUserAttribute>();
-      var rawAttributes = rawOptions["userAttributes"] as HashMap<String, String>;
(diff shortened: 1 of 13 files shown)


haverchuck:
Tests will continue to fail until build.gradle is updated in the cognito plugin.

Jordan-Nelson (approved):
LGTM

offlineprogrammer (approved):

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterConfirmSignInRequest.kt
```diff
@@ -29,6 +31,15 @@ data class FlutterConfirmSignInRequest(val map: HashMap<String, *>) {
     if(rawOptions?.get("clientMetadata") != null)
       options.metadata(rawOptions["clientMetadata"] as HashMap<String, String>);
 
+    if(rawOptions?.get("userAttributes") != null) {
+      var attributes = mutableListOf<AuthUserAttribute>();
+      var rawAttributes = rawOptions["userAttributes"] as HashMap<String, String>;
+      rawAttributes.forEach { (key, value) ->
+        attributes.add(createAuthUserAttribute(key, value))
+      }
```
cshfang, line 39: Nit: Rather than declaring a mutable list and then adding elements in, would it be possible to just map the userAttributes? E.g.
```suggestion
    if(rawOptions?.get("userAttributes") != null) {
      val rawAttributes = rawOptions["userAttributes"] as HashMap<String, String>
      val attributes = rawAttributes.map { (key, value) ->
        createAuthUserAttribute(key, value)
      }
```

Also noticed this is slightly different from the Swift implementation where it seemed ok to have the attributes be an empty list. If it's ok here too, we can make this even more concise

```suggestion
    val rawAttributes = rawOptions?.get("userAttributes") as HashMap<String, String>? ?: emptyMap<String, String>()
    val attributes = rawAttributes.map { (key, value) ->
      createAuthUserAttribute(key, value)
    }
```

packages/amplify_auth_cognito/ios/Classes/FlutterConfirmSignInRequest.swift
```diff
@@ -29,8 +29,16 @@ struct FlutterConfirmSignInRequest {
...
   func formatOptions(options: Dictionary<String, Any>?) -> AuthConfirmSignInOperation.Request.Options {
-     
+    let rawAttributes = options?["userAttributes"] as? [String : String]
+    var userAttributes: [AuthUserAttribute] = [];
+    if (rawAttributes != nil) {
+      for (key, value) in rawAttributes ?? [:] {
+        userAttributes.append(AuthUserAttribute(createAuthUserAttributeKey(keyName: key), value: value))
+      }
+    }
+
```
cshfang, line 39: Similar to kotlin counterpart
```suggestion
    var rawAttributes = options?["userAttributes"] as? [String : String] ?? [:]
    var userAttributes: [AuthUserAttribute] = rawAttributes.map { key, value in
      AuthUserAttribute(createAuthUserAttributeKey(keyName: key), value: value)
    }
```

Even if we don't take this suggestion, I think we should move the `?? [:]` from line 35 to line 32. I think that might make it possible to have `rawAttributes` be a `var` instead of `let` as well?

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
```diff
@@ -39,6 +39,10 @@ void main() {
 
   AmplifyAuthCognito auth = AmplifyAuthCognito();
   TestWidgetsFlutterBinding.ensureInitialized();
+  const testAttributeKey = 'email';
+  const testEmailValue = 'test@test.test';
+  const testMetadataKey = 'key';
+  const testMetaDataAttribute = 'val';
```
cshfang, line 45: 💯

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
```diff
@@ -84,11 +88,13 @@ void main() {
 
   test('confirmSignIn request accepts and serializes options',
           () async {
-        var options = CognitoConfirmSignInOptions(clientMetadata: {'key': 'val'});
+        var options = CognitoConfirmSignInOptions(clientMetadata: {testMetadataKey: testMetaDataAttribute}, userAttributes: {testAttributeKey: 'test@test.test'});
```
cshfang, line 91:
```suggestion
        var options = CognitoConfirmSignInOptions(clientMetadata: {testMetadataKey: testMetaDataAttribute}, userAttributes: {testAttributeKey: testEmailValue});
```

cshfang (commented):
LGTM for the most part! Left a few suggestions

cshfang (approved):

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
haverchuck: Done

packages/amplify_auth_cognito/ios/Classes/FlutterConfirmSignInRequest.swift
haverchuck: Done

cshfang (approved):
💯

PR #614, opened 2021-06-04
Null safety master fixes pr

fjnoyp:
Fix unit tests issues in Dart and remove redundant exception. 

### DIFF
--- a/packages/amplify_datastore_plugin_interface/lib/src/types/models/model.dart
+++ b/packages/amplify_datastore_plugin_interface/lib/src/types/models/model.dart
@@ -17,17 +17,13 @@ import 'model_schema.dart';
 import 'model_schema_definition.dart';
 
 abstract class Model {
-  final String id;
-
   ModelType getInstanceType();
 
-  const Model({required this.id});
-
-  String getId() {
-    return id;
-  }
+  String getId();
 
   Map<String, dynamic> toJson();
+
+  const Model();
 
   static ModelSchema defineSchema(
       {required Function(ModelSchemaDefinition) define}) {
--- a/packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_date.dart
+++ b/packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_date.dart
@@ -74,7 +74,7 @@ class TemporalDate {
 
     // Validate
     String? regexString = regExp.stringMatch(iso8601String);
-    if (regexString != iso8601String) {
+    if (regexString == null || regexString != iso8601String) {
       throw AmplifyException("Invalid ISO8601 String Input",
           recoverySuggestion:
               "Please provide an extended ISO 8601 date string in the format YYYY-MM-DD with an optional time zone offset ±hh:mm:ss.  " +
--- a/packages/amplify_datastore_plugin_interface/test/testData/Blog.dart
+++ b/packages/amplify_datastore_plugin_interface/test/testData/Blog.dart
@@ -13,19 +13,28 @@
 * permissions and limitations under the License.
 */
 
+// ignore_for_file: public_member_api_docs
+
+import 'ModelProvider.dart';
 import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart';
 import 'package:collection/collection.dart';
 import 'package:flutter/foundation.dart';
-
-import 'ModelProvider.dart';
 
 /** This is an auto generated class representing the Blog type in your schema. */
 @immutable
 class Blog extends Model {
   static const classType = const BlogType();
   final String id;
-  final String name;
-  final List<Post>? posts;
+  final String? _name;
+  final List<Post>? _posts;
+
+  String get name {
+    return _name!;
+  }
+
+  List<Post>? get posts {
+    return _posts;
+  }
 
   @override
   getInstanceType() => classType;
@@ -35,9 +44,11 @@ class Blog extends Model {
     return id;
   }
 
-  const Blog._internal({required this.id, required this.name, this.posts});
+  const Blog._internal({required this.id, required name, posts})
+      : _name = name,
+        _posts = posts;
 
-  factory Blog({required String id, required String name, List<Post>? posts}) {
+  factory Blog({String? id, required String name, List<Post>? posts}) {
     return Blog._internal(
         id: id == null ? UUID.getUUID() : id,
         name: name,
@@ -53,8 +64,8 @@ class Blog extends Model {
     if (identical(other, this)) return true;
     return other is Blog &&
         id == other.id &&
-        name == other.name &&
-        DeepCollectionEquality().equals(posts, other.posts);
+        _name == other.name &&
+        DeepCollectionEquality().equals(_posts, other.posts);
   }
 
   @override
@@ -65,29 +76,32 @@ class Blog extends Model {
     var buffer = new StringBuffer();
 
     buffer.write("Blog {");
-    buffer.write("id=" + id + ", ");
-    buffer.write("name=" + name);
+    buffer.write("id=" + "$id" + ", ");
+    buffer.write("name=" + "$_name");
     buffer.write("}");
 
     return buffer.toString();
   }
 
-  Blog copyWith({required String id, required String name, List<Post>? posts}) {
+  Blog copyWith({String? id, String? name, List<Post>? posts}) {
     return Blog(
         id: id ?? this.id, name: name ?? this.name, posts: posts ?? this.posts);
   }
 
   Blog.fromJson(Map<String, dynamic> json)
       : id = json['id'],
-        name = json['name'],
-        posts = json['posts'] is List
+        _name = json['name'],
+        _posts = json['posts'] is List
             ? (json['posts'] as List)
                 .map((e) => Post.fromJson(new Map<String, dynamic>.from(e)))
                 .toList()
             : null;
 
-  Map<String, dynamic> toJson() =>
-      {'id': id, 'name': name, 'posts': posts?.map((e) => e?.toJson())};
+  Map<String, dynamic> toJson() => {
+        'id': id,
+        'name': _name,
+        'posts': _posts?.map((e) => e.toJson()).toList()
+      };
 
   static final QueryField ID = QueryField(fieldName: "blog.id");
   static final QueryField NAME = QueryField(fieldName: "name");
--- a/packages/amplify_datastore_plugin_interface/test/testData/Blog.dart
+++ b/packages/amplify_datastore_plugin_interface/test/testData/Blog.dart
@@ -76,8 +76,8 @@ class Blog extends Model {
     var buffer = new StringBuffer();
 
     buffer.write("Blog {");
-    buffer.write("id=" + "$id" + ", ");
-    buffer.write("name=" + "$_name");
+    buffer.write("id=$id" + ", ");
+    buffer.write("name=$_name");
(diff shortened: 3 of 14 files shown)


packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_date.dart
```diff
@@ -74,7 +74,7 @@ class TemporalDate {
 
     // Validate
     String? regexString = regExp.stringMatch(iso8601String);
-    if (regexString != iso8601String) {
+    if (regexString == null || regexString != iso8601String) {
```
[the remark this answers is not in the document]
fjnoyp: @HuiSF we will be merging in null safety changes to the datastore plugin **interface** now but no changes to the datastore plugin. 

This is because the amplify flutter plugin has dependencies on all plugin interfaces but not the actual plugins themselves.

packages/amplify_datastore_plugin_interface/lib/src/types/models/model.dart
```diff
@@ -17,18 +17,14 @@ import 'model_schema.dart';
...
-  final String id;
-
   ModelType getInstanceType();
 
-  const Model({required this.id});
-
-  String getId() {
-    return id;
-  }
+  String getId();
```
Jordan-Nelson, line 22: Should this throw an UnimplementedError if it no longer has a default implementation?

packages/amplify_datastore_plugin_interface/test/testData/Blog.dart
```diff
@@ -13,19 +13,28 @@
...
 class Blog extends Model {
   static const classType = const BlogType();
   final String id;
-  final String name;
-  final List<Post>? posts;
+  final String? _name;
+  final List<Post>? _posts;
+
+  String get name {
+    return _name!;
```
Jordan-Nelson, line 32: We are already discussing this one offline, but I don't see how this could be a safe cast.

packages/amplify_datastore_plugin_interface/test/testData/Blog.dart
```diff
@@ -65,29 +76,32 @@ class Blog extends Model {
     var buffer = new StringBuffer();
 
     buffer.write("Blog {");
-    buffer.write("id=" + id + ", ");
-    buffer.write("name=" + name);
+    buffer.write("id=" + "$id" + ", ");
```
Jordan-Nelson, line 79: *nit:* If you are going to use string interpolation
```dart
buffer.write("id=$id, ");
```

packages/amplify_datastore_plugin_interface/test/testData/ModelProvider.dart
```diff
@@ -27,14 +28,9 @@ export 'PostAuthComplex.dart';
...
 class ModelProvider implements ModelProviderInterface {
   @override
-  String version = "6319a5b46c3ea5e5817b63396bc464bb";
+  String version = "33b1717ef9afb390e3f7431e8f3859fc";
   @override
-  List<ModelSchema> modelSchemas = [
-    Blog.schema,
-    Comment.schema,
-    Post.schema,
-    PostAuthComplex.schema
```
ragingsquirrel3, line 36: question: Why is `PostAuthComplex` omitted now?

packages/amplify_datastore_plugin_interface/test/testData/Post.dart
```diff
@@ -13,22 +13,43 @@
...
 @immutable
 class Post extends Model {
   static const classType = const PostType();
   final String id;
-  final String title;
-  final int? rating;
-  final DateTime? created;
-  final Blog? blog;
-  final List<Comment>? comments;
+  final String? _title;
```
ragingsquirrel3, line 28: question: Should `title` be required here? Is it required in `_internal`? Same question for `rating`? I'm just a little confused by different syntax for accessors below but I'm likely missing something.

packages/amplify_datastore_plugin_interface/test/testData/PostAuthComplex.dart
```diff
@@ -21,8 +21,16 @@ import 'package:flutter/foundation.dart';
 class PostAuthComplex extends Model {
   static const classType = const PostAuthComplexType();
   final String id;
-  final String title;
-  final String? owner;
+  final String? _title;
+  final String? _owner;
+
+  String get title {
```
ragingsquirrel3, line 27: Again, I guess I'm a little confused as `_title` is optional but not sure why this does not have `String?` syntax like owner below.

packages/amplify_datastore_plugin_interface/test/testData/Blog.dart
```diff
@@ -65,29 +76,32 @@ class Blog extends Model {
...
+    buffer.write("name=" + "$_name");
```
HuiSF, line 80: nit: `buffer.write("name=$_name");`

packages/amplify_datastore_plugin_interface/test/testData/Post.dart
```diff
@@ -1,5 +1,5 @@
 /*
-* Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+* Copyright 201 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
ragingsquirrel3, line 2: nit: Is this change intentional?

packages/amplify_datastore_plugin_interface/lib/src/types/models/model.dart
ragingsquirrel3, line 22: question: Why continue to include this method written this way? Extended classes will continue to use @override, correct? Not requesting a change, just trying to understand.

ragingsquirrel3 (commented):
Overall, looks good, just nits/questions to help me understand. Likely I can approve if I understand correctly, just trying to make sure.

packages/amplify_datastore_plugin_interface/test/testData/Comment.dart
```diff
@@ -79,13 +90,13 @@ class Comment extends Model {
 
   Comment.fromJson(Map<String, dynamic> json)
       : id = json['id'],
-        post = json['post'] != null
+        _post = json['post'] != null
             ? Post.fromJson(new Map<String, dynamic>.from(json['post']))
             : null,
-        content = json['content'];
+        _content = json['content'];
```
HuiSF, line 96: Should we validate `json['content']` here to ensure it's non-null? Asking because that it looks like class `Comment` ensures `_content` cannot be assigned `null`, except this constructor. Thinking it may be better to keep the behavior consistent. (same question to other `fromJson` methods in other classes)

packages/amplify_datastore_plugin_interface/test/testData/Post.dart
HuiSF, line 2: nit: `2021`

packages/amplify_datastore_plugin_interface/lib/src/types/models/model.dart
fjnoyp: Not a major concern as all models will be generated by codegen team which is guaranteed to have that implementation.

packages/amplify_datastore_plugin_interface/test/testData/Blog.dart
fjnoyp: Yup it's not.  The Blog graphqL schema has a required name field.  Customers wouldn't expect a required field to be nullable.  However, we make the underlying private field nullable to handle the fact that the model might not have that field if it was a nested model with only the id field set.

packages/amplify_datastore_plugin_interface/test/testData/Blog.dart
fjnoyp: good call. thanks!

packages/amplify_datastore_plugin_interface/test/testData/ModelProvider.dart
fjnoyp: good call.  it shouldn't ... added it back in

packages/amplify_datastore_plugin_interface/test/testData/Post.dart
fjnoyp: the codegen model fields are generated based on a graphql schema that unfortunately isn't shown in the file diff here.  The underlying fields (private _) are all nullable because we might create a model with just an id.  However, the field can be marked as required in the graphql schema.  Thus the public accessor makes a force cast "!"

packages/amplify_datastore_plugin_interface/test/testData/PostAuthComplex.dart
fjnoyp: see answer above.

packages/amplify_datastore_plugin_interface/test/testData/Post.dart
fjnoyp: Nope thanks for catching!

packages/amplify_datastore_plugin_interface/lib/src/types/models/model.dart
fjnoyp: We need to include this method since Model is an abstract class.  And abstract class needs to have method so we can call it without need to cast to the underlying Model type like Blog, Post, etc.

packages/amplify_datastore_plugin_interface/test/testData/Blog.dart
fjnoyp: Thanks!

packages/amplify_datastore_plugin_interface/test/testData/Comment.dart
fjnoyp: _content can be null as it is a nullable type.   Typically the field should not be null, but in fromJson we don't know if all fields are going to be set as it might be a nested model in which only the id field will be set.

packages/amplify_datastore_plugin_interface/lib/src/types/models/model.dart
Jordan-Nelson: Yeah, I consider this to be more of a nit/non-blocking comment. It is hard to imagine a scenario where this would actually be an issue, but if there were some bug in the code generation failing fast by throwing an error would be preferred to returning null.

packages/amplify_datastore_plugin_interface/test/testData/Blog.dart
Jordan-Nelson: I think it is fine for now since datastore won't be in the initial preview and since this looks like it is just for tests. I assume there will be more discussion about datastore offline.

packages/amplify_datastore_plugin_interface/lib/src/types/models/model.dart
ragingsquirrel3: thanks

packages/amplify_datastore_plugin_interface/test/testData/Post.dart
ragingsquirrel3: thanks

HuiSF (approved):
Thanks for resolving/answering comments

ragingsquirrel3 (approved):
LGTM, thanks for info on questions I had.

Jordan-Nelson (approved):
Approving, but noting that there needs to be more discussion about the nullable/non-nullable fields in the generate models for datastore.

Issue #575, opened 2021-05-12
Can not pass data in ConfirmSignUpOptions()


shajedulislam:
```
SignUpResult signUpResult = await Amplify.Auth.confirmSignUp(
          username: widget.email,
          confirmationCode: _pinController.text,
          options: ConfirmSignUpOptions(),
        );
```

I need to pass clientMetada here but can not pass data in confirm signup option. Like,

```
{
  'clientMetada': {
    'IsDoctor': false,
    
  }}
```

 It says,
### Abstract classes can't be instantiated. Try creating an instance of a concrete subtype

Any solution?

HuiSF:
Hi @shajedulislam `ConfirmSignUpOptions` is an abstract class, initiating an abstract class causes the error you encountered.

The auth plugin provides a concrete subtype [`CognitoConfirmSignInOptions`](https://github.com/aws-amplify/amplify-flutter/blob/master/packages/amplify_auth_cognito/lib/src/CognitoSignIn/CognitoConfirmSignInOptions.dart#L18) of `ConfirmSignUpOptions`, we can use this class to initiate the options. (Please note that `CognitoConfirmSignUpOptions` currently accept only named parameter `validationData`.)

e.g.

```dart
SignUpResult res = await Amplify.Auth.confirmSignUp(
          username: usernameController.text.trim(),
          confirmationCode: confirmationCodeController.text.trim(),
          options: CognitoConfirmSignUpOptions(
              validationData: {'testField': 'doctor'}));
```

Is this sufficient for your use case please?

HuiSF:
In addition, there might be a feature gap here that `confirmSignUp` doesn't properly work with `clientMetadata`, will continue investigating.

shajedulislam:
@HuiSF Thanks fro your response. Our web version already developed using client meta data and yes currently it only accepts validation data for flutter. We will try to manage using validation data but it would be better to bring it with client meta data also.

HuiSF:
Hello @shajedulislam amplify-flutter 0.16.0 just released, which includes the support of passing `clientMetadata` as `confirmSignUp` options, please try it out, and feel free to follow if you are experiencing any issues. Thanks!

PR #619, opened 2021-06-07
fix(auth): Add clientMetadata to confirmSignUp API options

HuiSF:
**Issue #, if available:**

Fix part 2 for #575 

`Amplify.Auth.confirmSignUp` API now supports options parameter to include `clientMetadata`.

Example of usage

```dart
final options =
    CognitoConfirmSignUpOptions(clientMetadata: {'key': 'value'});

SignUpResult res = await Amplify.Auth.confirmSignUp(
    username: 'testUser',
    confirmationCode: '123456',
    options: options
);
```

**Description of changes:**
* Updated `CognitoConfirmSignUpOptions` to hold `clientMetadata`
* Updated method channel implementation to actually pass `clientMetadata` to `amplify-android` confirmSignUp API
* Removed `validationData` from `CognitoConfirmSignUpOptions` which is not supported by either amplify-android and ampliy-ios


packages/amplify_analytics_pinpoint/android/build.gradle
```diff
@@ -60,8 +60,8 @@ dependencies {
     api amplifyCore
 
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
-    implementation 'com.amplifyframework:aws-analytics-pinpoint:1.17.4'
-    implementation 'com.amplifyframework:aws-auth-cognito:1.17.4'
+    implementation 'com.amplifyframework:aws-analytics-pinpoint:1.17.8'
```
haverchuck, line 63: We have several PRs in flight that are all going to be changing these versions...we should figure out a good way to coordinate this, and also a way to a global variable.

packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthCognitoPluginTest.kt
```diff
@@ -121,12 +123,18 @@ class AmplifyAuthCognitoPluginTest {
...
     @Test
-    fun confirmSignUp_returnsSuccess() {
+    fun confirmSignUpWithoutOptions_returnsSuccess() {
         // Arrange
         doAnswer { invocation: InvocationOnMock ->
             plugin.prepareSignUpResult(mockResult, mockSignUpResult)
             null as Void?
-        }.`when`(mockAuth).confirmSignUp(anyString(), anyString(), ArgumentMatchers.any<Consumer<AuthSignUpResult>>(), ArgumentMatchers.any<Consumer<AuthException>>())
+        }.`when`(mockAuth).confirmSignUp(
+            anyString(),
```
haverchuck, line 132: Just to confirm - we are only making this change because under the hood amplify-android will always have a default value and/or empty map here, so we are changing the pattern matcher?

haverchuck (approved):
Left a question, but otherwise looks good!

packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthCognitoPluginTest.kt
HuiSF: amplify-android will invoke `AuthConfirmSignUpOptions.default()` to get an empty map for the `options` parameter if it's not supplied (if this is what you were confirming about :))

packages/amplify_analytics_pinpoint/android/build.gradle
HuiSF: Good call out, thanks! I was thinking to define the version in `gradle.properties` to share across android packages, though I couldn't find a good location to hold this file.

packages/amplify_auth_cognito/example/ios/unit_tests/amplify_auth_cognito_tests.swift
```diff
@@ -348,6 +348,30 @@ class amplify_auth_cognito_tests: XCTestCase {
             }
         })
     }
+
+    func test_confirmSignUpForwardOptions() {
+        let mockOptions: Dictionary<String, Any> = ["clientMetadata": ["key": "value"]]
```
cshfang, line 353: Nit: Maybe we could extract to a `mockClientMetadata` to reuse on the assert

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignup_test.dart
```diff
@@ -45,6 +45,20 @@ void main() {
...
+    final mockValue = 'value';
+    final mockOptions =
+        CognitoConfirmSignUpOptions(clientMetadata: {mockKey: mockValue});
+    final requst = ConfirmSignUpRequest(
+            username: 'user', confirmationCode: '123456', options: mockOptions)
+        .serializeAsMap();
+
+    expect(requst['options'], isInstanceOf<Map>());
+    expect(requst['options']['clientMetadata'], isInstanceOf<Map>());
+    expect(requst['options']['clientMetadata'][mockKey], mockValue);
```
cshfang, line 59: Nit: Would it be possible to combine `mockKey` and `mockValue` into a `mockClientMetadata` and check for that equality directly? Just think we could align it a little more with the kt/swift tests that way

packages/amplify_auth_cognito/ios/Classes/FlutterConfirmSignUpRequest.swift
```diff
@@ -14,24 +14,37 @@
...
+        self.username = dict["username"] as! String
+        self.confirmationCode = dict["confirmationCode"] as! String
+        self.options = formatOptions(options: dict["options"] as! Dictionary<String, Any>?)
     }
-    if (dict["confirmationCode"] == nil && dict["options"] == nil) {
-        throw InvalidRequestError.auth(comment: validationErrorMessage,
-                                          suggestion: String(format: ErrorMessages.missingAttribute, "confirmationCode"))
+    
+    func formatOptions(options: Dictionary<String, Any>?) -> AuthConfirmSignUpRequest.Options {
+        let pluginOptions =  AWSAuthConfirmSignUpOptions(
```
cshfang, line 33: Super nit:
```suggestion
        let pluginOptions = AWSAuthConfirmSignUpOptions(
```

cshfang (approved):
Lgtm for the most part. Just a few minor comments

packages/amplify_auth_cognito/example/ios/unit_tests/amplify_auth_cognito_tests.swift
HuiSF: This is left as this on purpose... the assertion is happening inside the mock class `ConfirmSignUpMock`, the assertion function doesn't have access to variables declared outside of this mock class. So I just used the literal inside the mock class and without reusing the mocked variable.

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignup_test.dart
```diff
@@ -45,6 +45,20 @@ void main() {
...
+  test('confirmSignUp request accepts and serializes options', () async {
+    final mockKey = 'key';
+    final mockValue = 'value';
+    final mockOptions =
+        CognitoConfirmSignUpOptions(clientMetadata: {mockKey: mockValue});
+    final request = ConfirmSignUpRequest(
+            username: 'user', confirmationCode: '123456', options: mockOptions)
+        .serializeAsMap();
+
+    expect(requst['options'], isInstanceOf<Map>());
```
Jordan-Nelson, line 57:
```suggestion
    expect(request['options'], isInstanceOf<Map>());
```

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignup_test.dart
```diff
@@ -45,6 +45,20 @@ void main() {
...
+    expect(requst['options']['clientMetadata'], isInstanceOf<Map>());
```
Jordan-Nelson, line 58:
```suggestion
    expect(request['options']['clientMetadata'], isInstanceOf<Map>());
```

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignup_test.dart
```diff
@@ -45,6 +45,20 @@ void main() {
...
+    expect(requst['options']['clientMetadata'][mockKey], mockValue);
```
Jordan-Nelson, line 59:
```suggestion
    expect(request['options']['clientMetadata'][mockKey], mockValue);
```

Jordan-Nelson (approved):

Issue #596, opened 2021-05-26
Custom User Attributes: Inconsistent behavior on iOS and Android


Jordan-Nelson:
**Describe the bug**

Custom user attributes on Android can be created/updated by constructing a key that has a prefix of `custom:`, but this will fail on iOS. For example, if a custom user attribute of "favorite_number" exists in cognito the attribute this attribute needs to be added/updated using the key "favorite_number" on iOS. However, on Android this attribute can be added/updated using the key "favorite_number" OR "custom:favorite_number".

**To Reproduce**
1. Create a custom user attribute in cognito, for example, "favorite_number"
2. Attempt to add create a new user with this attribute OR update the attribute on an existing user

**Example w/ Sign Up**
```dart
var userAttributes = {
  // "custom:favorite_color" works on Android, but not iOS. "favorite_color" works on both platforms
  'custom:favorite_color': 'blue',
};
var res = await Amplify.Auth.signUp(
    username: 'testuser001',
    password: 'password123',
    options: CognitoSignUpOptions(userAttributes: userAttributes));
```
**Example w/ updateAttribute**
```dart
await Amplify.Auth.updateUserAttribute(
  // "custom:favorite_color" works on Android, but not iOS. "favorite_color" works on both platforms
  userAttributeKey: 'custom:favorite_color',
  value: 'blue',
);
```

**Expected behavior**
1. Consistency across platforms.
2. Consistency between updating and fetching user attributes. `fetchUserAttributes()` returns custom user attributes with a key that has a prefix of "custom:". It would be expected that the same key returned from fetchUserAttributes() could be used to update the user attribute. Currently `"custom:"` has to be removed.

**Platform**
Amplify Flutter current supports iOS and Android.  This issue is reproducible in (check all that apply):
[x] Android
[x] iOS

**Additional context**
This issue is caused by an underlying inconsistency between amplify-android and amplify-ios. amplify-android requires custom attribute keys to be prepended with "custom:" but amplify-ios requires them to **not** be prepended with "custom:". To handle this inconsistency, amplify-flutter (for Android) checks to see if the key is already prepended with "custom:" and if it is not, it is added. This was done in attempt to make the flutter API consistent across platforms, but does not fully address the inconsistency.

Jordan-Nelson:
This has been fixed in version 0.1.6. Both of the following code snippets will now work on iOS and Android for a custom attribute of `favorite_color`.

w/ `custom:` prefix`
```dart
[...]

PR #620, opened 2021-06-08
fix(auth): iOS/Android user attribute inconsistencies 

Jordan-Nelson:
*Issue #, if available:* https://github.com/aws-amplify/amplify-flutter/issues/596

*Description of changes:*
- remove "custom:" from custom attribute keys on iOS
- pulled in fix for missing attribute keys on iOS from https://github.com/aws-amplify/amplify-flutter/pull/607


packages/amplify_auth_cognito/ios/Classes/Utils/UserAttributeDeserialization.swift
```diff
@@ -50,6 +50,14 @@ func createAuthUserAttributeKey(keyName: String) -> AuthUserAttributeKey {
...
         return AuthUserAttributeKey.preferredUsername
     default:
-        return AuthUserAttributeKey.custom(keyName)
+        // amplify-ios does not currently include enums for these keyNames
+        let unknownKeys = ["profile", "updated_at", "website", "zoneinfo"]
+        if (unknownKeys.contains(keyName)) {
+            return AuthUserAttributeKey.unknown(keyName)
+        } else {
+            // amplify-ios prepends 'custom:' to custom keys
+            let customKeyName = keyName.starts(with: "custom:") ? String(keyName.dropFirst(7)) : keyName
```
Jordan-Nelson, line 59: Adding comment again for visibility.
___

This change will make amplify-flutter consistent on iOS and Android. On both platforms, prepending custom: will be optional with this change. Amplify-android requires that the value be prepended with custom: so we add it if it is missing. This is essentially the opposite. amplify-ios requires that it is not present, so we remove it.

Note: When the key is created in Cognito, Cognito does not allow for users to prepend custom: to the key name since it will be added automatically. There should be no risk of an edge case scenario where a users key prepended custom: during creating, resulting in a key like custom:custom:my_key.

haverchuck (approved):

packages/amplify_auth_cognito/ios/Classes/Utils/UserAttributeDeserialization.swift
Jordan-Nelson: I think thats fair. Something like this?
```suggestion
            let customKeyPrefix = "custom:"
            let customKeyName = keyName.starts(with: customKeyPrefix) ? String(keyName.dropFirst(customKeyPrefix.count)) : keyName
```

Issue #623, opened 2021-06-09
Generic AuthException being thrown 


brunovsiqueira:
Hi there.

When user is already signed in, an AuthException is thrown instead of a FlutterInvalidStateException. I can only see that it is a FlutterInvalidStateException if I open the "underlyingException" attribute inside the AuthException.

Shouldn't it throw a FlutterInvalidStateException so I could handle it easier?

Furthermore, it would be nice if every exception thrown by Amplify  were specific and not a generic AuthException.

Jordan-Nelson:
Hi @brunovsiqueira - Thanks for opening this issue. Would you mind filling out the issue from the issue template? I am particularly looking for steps to reproduce (are you simply calling signIn when a user is already signed in?), platform (iOS, Android, or both) and the version of amplify flutter you are using. Thanks.

brunovsiqueira:
Sure @Jordan-Nelson.

Yes, I am simply calling signIn when a user is already signed in. I tested on Andorid, but I guess it also happens on iOS. We are using Amplify 0.1.5 now.

Jordan-Nelson:
@brunovsiqueira - this has been resolved as of version 0.2.1. You should now see an InvalidStateException on both platforms when calling signIn with a user already authenticated.

PR #625, opened 2021-06-09
fix(auth): handle InvalidStateException

Jordan-Nelson:
*Issue #, if available:* https://github.com/aws-amplify/amplify-flutter/issues/623

*Description of changes:*
- Create a new exception on the Flutter side, InvalidStateException
- On iOS, map AuthException.invalidState to InvalidStateException
- On Android, map FlutterInvalidStateException to InvalidStateException


packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterInvalidStateException.kt
```diff
@@ -18,5 +18,5 @@ package com.amazonaws.amplify.amplify_auth_cognito.types
 import com.amplifyframework.auth.AuthException
 
 class FlutterInvalidStateException: AuthException {
-    constructor(message: String, recoverySuggestion: String) : super(message, Exception("invalidSate"), recoverySuggestion)
+    constructor(message: String, recoverySuggestion: String) : super(message, Exception("invalidState"), recoverySuggestion)
```
cshfang, line 21: 🕵️   Is it possible that someone could be looking for `invalidSate` for some reason in their code?

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterSignInRequest.kt
```diff
@@ -55,7 +55,7 @@ data class FlutterSignInRequest(val map: HashMap<String, *>) {
       try {
         var user: AuthUser? = Amplify.Auth.currentUser;
         if (user != null) {
-          throw FlutterInvalidStateException("There is already a user  signed in.", "Sign out before calling sign in.")
+          throw FlutterInvalidStateException("There is already a user signed in.", "Sign out before calling sign in.")
```
cshfang, line 58: 🕵️  Is it possible that someone could be looking for the previous string with the extra space for some reason in their code?

packages/amplify_auth_plugin_interface/lib/src/Exceptions/InvalidStateException.dart
```diff
@@ -0,0 +1,40 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import 'package:amplify_auth_plugin_interface/amplify_auth_plugin_interface.dart';
+
+/// Thrown when an operation is not valid with the current state of Auth category 
+class InvalidStateException extends AuthException {
+  /// Named constructor
```
cshfang, line 20: Nit: Don't think this is a named constructor
```suggestion
  /// Default constructor
```

packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthErrorHandlerTest.kt
```diff
@@ -267,6 +268,30 @@ class AuthErrorHandlerTest {
         verify(mockResult, times(1)).error(expectedCode, defaultMessage, details);
     }
 
+    @Test
+    fun invalidStateException() {
+        // Arrange
+        val expectedCode = "InvalidStateException"
+        val exception: AuthException = FlutterInvalidStateException("Invalid state message", "Invalid state recovery message")
```
cshfang, line 275: Nit: Extract the strings for reuse

cshfang (approved):
Minor questions and a couple nits but lgtm otherwise 🚀

packages/amplify_auth_plugin_interface/lib/src/Exceptions/InvalidStateException.dart
Jordan-Nelson: You are right. This is what I get for copy and pasting code 🙂

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterInvalidStateException.kt
Jordan-Nelson: 🤯  ... Yes, it is definitely possible.

I am not sure if we would consider the message attached to an error part of the public API or not. It depends how you define public API (anything that is technically exposed as public or only the intended use).

Either way, this IMO is a bug fix and bug fixes should not be considered breaking in the same way that changing working functionality is considered breaking. Bug fixes are fixes to something that is currently broken. This entire PR is actually a good example of that. Currently the wrong exception is being thrown. This PR changes which exception type is being thrown. That could conceivably break code that a customer has written, but it is a fix to something that is broken and was never intended to be used that way.

dnys1 (approved):

haverchuck (approved):

Issue #633, opened 2021-06-11
Organize packages according to Dart package layout conventions


mrhota:
**Describe the chore**

Many of the amplify-flutter packages do not have `lib/src` directories, which seems to lead to API leakage in docs.

It isn't always clear what code is public API and what code is implementation for a package.

In an effort to clarify and tidy things up, I'd like to ensure all amplify-flutter packages follow the conventions laid out in the [Dart package layout page](https://dart.dev/tools/pub/package-layout).

**Expected behavior**

Once done, dartdocs should be tidier and clearer, and it should be easier to navigate the code, both as a client of this package, and a contributor.

**Additional context**

Once this chore is done, my near-term goal is also to begin refactoring packages according to the [Flutter Federated Plugins guidelines](https://flutter.dev/docs/development/packages-and-plugins/developing-packages#plugin).

Beyond that, these tasks are in service to my medium-term goal of implementing web support.

dnys1:
Resolved as part of 0.3.0 release. Thanks again for raising this issue!

PR #638, opened 2021-06-12
Chore/restructure amplify core

mrhota:
# Issue

#633

# Description of changes

We try to tidy up `amplify_core` according to conventions outlined within Dart documentation: https://dart.dev/guides/libraries/create-library-packages

We now have 3 libraries exposed in the `amplify_core` package:

1. `amplify_core`, exporting `hub` and `exception` types
2. `amplify_plugin_interface` exporting the class of the same name
3. `test_utils` exporting the `getJsonFromFile` function

Types are kept under the new `lib/src` dir, though they are exported, and the libraries live under `lib`.

## Notes

I should note that it isn't clear to me what purpose the `AmplifyCorePlatform` class serves. I don't see any uses outside of the single use in this package's tests.

---


mrhota:
The errors about the way I restructured amplify_core baffle me.

I thought every package here depended on a fixed version of amplify_core from pub.dev. Therefore, why does every package fail to compile when I neither published to pub.dev, nor changed any other package’s dependency version specifier?

dnys1:
Thank you so much for your contribution. We have resolved this as part of the forthcoming 0.3.0 release. Really appreciate it and look forward to seeing more ideas for how we can improve our codebase!

PR #649, opened 2021-06-17
Null safety datastore

fjnoyp:
*Issue #, if available:*

*Description of changes:*

Upgrade Datastore to support null safety


packages/amplify_datastore/example/lib/main.dart
```diff
@@ -27,7 +27,7 @@ import 'package:amplify_flutter/amplify.dart';
 import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart';
 import 'package:flutter/rendering.dart';
 import 'package:flutter/services.dart';
-import 'amplifyconfiguration.dart';
+//import 'amplifyconfiguration.dart';
```
haverchuck, line 30: Why are we doing this?

packages/amplify_datastore/example/lib/main.dart
```diff
@@ -100,7 +100,7 @@ class _MyAppState extends State<MyApp> {
       // await Amplify.addPlugin(AmplifyAPI());
 
       // Configure
-      await Amplify.configure(amplifyconfig);
+      await Amplify.configure("{}");
```
haverchuck, line 103: Same as above.

packages/amplify_datastore/example/lib/main.dart
fjnoyp: ah thanks - removed that for testing and forgot to change it back.

packages/amplify_datastore/lib/amplify_datastore.dart
```diff
@@ -28,16 +28,19 @@ class AmplifyDataStore extends DataStorePluginInterface {
   static final Object _token = Object();
 
   /// Constructs an AmplifyDataStore plugin
-  AmplifyDataStore({@required ModelProviderInterface modelProvider})
+  AmplifyDataStore({required ModelProviderInterface modelProvider})
       : super(token: _token, modelProvider: modelProvider);
 
+  /// Internal use constructor
+  @protected
+  AmplifyDataStore.protected() : super.protected(token: _token);
```
haverchuck, line 36: For my own edification, could you explain this change?

packages/amplify_datastore/pubspec.yaml
```diff
@@ -4,13 +4,14 @@ version: 0.2.0-nullsafety.0
...
-  sdk: ">=2.7.0 <3.0.0"
+  sdk: ">=2.12.0 <3.0.0"
   flutter: ">=1.17.0"
 
 dependencies:
   flutter:
     sdk: flutter
-  amplify_datastore_plugin_interface: 0.2.0-nullsafety.0
+  amplify_datastore_plugin_interface:
+    path: ../amplify_datastore_plugin_interface
```
haverchuck, line 14: Seems like we should leave this as is?

packages/amplify_datastore/test/amplify_datastore_query_test.dart
```diff
@@ -49,7 +49,9 @@ void main() {
         Comment(
             id: '39c3c0e6-8726-436e-8cdf-bff38e9a62da',
             content: 'Loving Amplify Datastore!',
-            post: Post(id: 'e50ffa8f-783b-4780-89b4-27043ffc35be')));
+            post: Post(
+                id: 'e50ffa8f-783b-4780-89b4-27043ffc35be',
+                title: ""))); // TODO POSSIBLE ISSUE HERE CONSTRUCTOR ID ONLY
```
haverchuck, line 54: Do we still need this comment?

packages/amplify_datastore/pubspec.yaml
fjnoyp: ah thanks again another testing artifact I forgot to remove

packages/amplify_datastore/test/amplify_datastore_query_test.dart
fjnoyp: no I'll remove that

packages/amplify_datastore/example/lib/main.dart
```diff
@@ -260,7 +260,11 @@ class _MyAppState extends State<MyApp> {
 
   savePost(String title, int rating, Blog associatedBlog) async {
     try {
-      Post post = Post(title: title, rating: rating, blog: associatedBlog);
+      Post post = Post(
+          title: title,
+          rating: rating,
+          created: TemporalDateTime.now(),
```
HuiSF, line 266: I'm wondering if it would be better that the `created` timestamp can be generated internally by the `Post` class, as it's more an "internal" field, devs could be less concerned about how to assign it a correct value.

haverchuck (commented):
Mostly LGTM; see inline comments

packages/amplify_datastore/example/lib/models/Blog.dart
```diff
@@ -64,30 +76,32 @@ class Blog extends Model {
     var buffer = new StringBuffer();
 
     buffer.write("Blog {");
-    buffer.write("id=" + id + ", ");
-    buffer.write("name=" + name);
+    buffer.write("id=" + "$id" + ", ");
```
HuiSF, line 79: nit: `buffer.write("id=$id, ");` ditto to other string interpolation

packages/amplify_datastore/example/lib/models/Blog.dart
```diff
@@ -23,8 +25,16 @@ import 'package:flutter/foundation.dart';
...
+  final String? _name;
+  final List<Post>? _posts;
+
+  String get name {
+    return _name!;
+  }
+
+  List<Post>? get posts {
+    return _posts;
+  }
```
HuiSF, line 37: Not directly related to this PR -
In previous discussion, we mentioned that for one-to-many relationship, we would return an empty array for the nested field. Does that imply that the `posts` field is always an array type value and is not nullable?

packages/amplify_datastore/lib/method_channel_datastore.dart
```diff
@@ -51,25 +55,24 @@ class AmplifyDataStoreMethodChannel extends AmplifyDataStore {
...
-          await _channel.invokeListMethod('query', <String, dynamic>{
+          (await (_channel.invokeListMethod('query', <String, dynamic>{
         'modelName': modelType.modelName(),
         'queryPredicate': where?.serializeAsMap(),
         'queryPagination': pagination?.serializeAsMap(),
-        'querySort':
-            sortBy?.map((element) => element?.serializeAsMap())?.toList()
-      });
+        'querySort': sortBy?.map((element) => element.serializeAsMap()).toList()
+      })))!;
```
HuiSF, line 75: nit: indentation for readability

packages/amplify_datastore/lib/types/DataStoreHubEvents/HubEventElement.dart
```diff
@@ -17,11 +17,17 @@ import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_inte
...
+      Map<dynamic, dynamic> serializedData, ModelProviderInterface provider) {
+    Map<String, dynamic> serializedElement =
+        new Map<String, dynamic>.from(serializedData["element"]);
+    Map<String, dynamic> modelData =
+        new Map<String, dynamic>.from(serializedElement["model"]);
+    Map<String, dynamic> serializedModelData =
+        new Map<String, dynamic>.from(modelData["serializedData"]);
+    model = provider
+        .getModelTypeByModelName(serializedData['modelName'])
+        .fromJson(serializedModelData);
```
HuiSF, line 31: nit: indentation for readability

packages/amplify_datastore_plugin_interface/test/testData/Post.dart
```diff
@@ -1,5 +1,5 @@
 /*
-* Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+* Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
HuiSF, line 2: nit: 2021 -> 2020

packages/amplify_datastore/lib/amplify_datastore.dart
fjnoyp: There is a compile issue without this change.  The AmplifyDataStore.protected() will implicitly try to call a constructor with no constructors so we have to explicitly tell it to call the super's protected method with the token parameter.

packages/amplify_datastore/example/lib/main.dart
fjnoyp: Not quite, this is a requirement from nullsafety.  Post has a required created field defined in its schema graphql that must be set, before that wasn't being enforced by the compiler so we missed creating their created field. 

You have a great point though.  There is actually some new Codegen work from Aaron Lee to automatically create "createdAt" and "lastUpdatedAt" fields.  We need to update our code to do this which will be done in the future. 

The created field you see here is more just to check that our internal TemporalDateTime class is working.

packages/amplify_datastore/example/lib/models/Blog.dart
fjnoyp: thanks this is a codegen change.  If the codegen team has time I can ask them to make this change.

packages/amplify_datastore/example/lib/models/Blog.dart
fjnoyp: Yes we mentioned that but only need to do that for "required" list fields.  If the list is not required and is nullable then we can return a null value no problem.  So no, if the list field is not required we can return null and make it nullable.

packages/amplify_datastore/lib/method_channel_datastore.dart
fjnoyp: Thanks!  Updated indentation for that file

packages/amplify_datastore/example/lib/main.dart
HuiSF: Thanks for the explanation!
Hence... before this work gets completed:
>We need to update our code to do this which will be done in the future.

Developers would need to insert `created` timestamp when create every new object, correct? And after the work the developers can omit `created`. If that's the case, I think to avoid this inconsistency would be nice. Thoughts?

fjnoyp: Yes as it would be created automatically.  How do you propose we avoid this inconsistency for now though?  Please remember these are our example app codegen models what fields the users create is up to them.

fjnoyp:
Hi guys please note that the unit_test_flutter is failing because the amplify_datastore plugin is using amplify_datastore_plugin_interface 0.2.0 which doesn't have the additional changes included in this PR.  I manually ran the unit tests and verified that they were working.

packages/amplify_datastore/lib/method_channel_datastore.dart
```diff
@@ -66,14 +69,16 @@ class AmplifyDataStoreMethodChannel extends AmplifyDataStore {
...
-      final List<Map<dynamic, dynamic>> serializedResults =
+      final List<Map<dynamic, dynamic>>? serializedResults =
           (await (_channel.invokeListMethod('query', <String, dynamic>{
         'modelName': modelType.modelName(),
         'queryPredicate': where?.serializeAsMap(),
         'queryPagination': pagination?.serializeAsMap(),
         'querySort': sortBy?.map((element) => element.serializeAsMap()).toList()
-      })))!;
-
+      })));
```
HuiSF, line 78: nit: As the `!` is removed, the outer `()` should be no longer needed

fjnoyp: ah thanks good catch

packages/amplify_datastore/test/test_models/Post.dart
```diff
@@ -85,23 +87,25 @@ class Post extends Model {
...
-    buffer.write("rating=" + rating?.toString() + ", ");
-    buffer.write("created=" + created?.toDateTimeIso8601String() + ", ");
-    buffer.write("blog=" + blog?.toString());
+    buffer.write("id=" + "$id" + ", ");
+    buffer.write("title=" + "$title" + ", ");
+    buffer.write(
+        "rating=" + (rating != null ? rating.toString() : "null") + ", ");
+    buffer.write(
+        "created=" + (created != null ? created!.format() : "null") + ", ");
+    buffer.write("blog=" + (blog != null ? blog.toString() : "null"));
```
cshfang, line 96: Since this test model (along with others above) isn't codegen'd, maybe we should clean up the string interpolation. Also should not need to explicitly convert `null` to `"null"` just to stringify it.
```suggestion
    buffer.write("id=$id, ");
    buffer.write("title=$title, ");
    buffer.write("rating=${rating?.toString()}, ");
    buffer.write("created=${created?.format()}, ");
    buffer.write("blog=${blog?.toString()}"));
```

packages/amplify_datastore/test/test_models/Comment.dart
```diff
@@ -62,15 +64,15 @@ class Comment extends Model {
     var buffer = new StringBuffer();
 
     buffer.write("Comment {");
-    buffer.write("id=" + id + ", ");
-    buffer.write("post=" + post?.toString() + ", ");
-    buffer.write("content=" + content);
+    buffer.write("id=" + "$id" + ", ");
+    buffer.write("post=" + (post != null ? post.toString() : "null") + ", ");
+    buffer.write("content=" + "$content");
```
cshfang, line 69: Ditto here about string interpolation and `null`s

packages/amplify_datastore/test/amplify_datastore_save_test.dart
```diff
@@ -48,7 +48,7 @@ void main() {
         id: '9fc5fab4-37ff-4566-97e5-19c5d58a4c22',
         title: 'New Post being saved',
         rating: 10,
-        created: DateTime.parse('2020-11-12T03:15:48-08:00'));
+        created: TemporalDateTime.fromString('2020-11-12T03:15:48+03:18'));
```
cshfang, line 51: Maybe it doesn't matter but same question - what necessitated the actual timestamp change?

packages/amplify_datastore/test/amplify_datastore_query_test.dart
cshfang: Nit: Since this is test data and we expect Title to be non-null.. maybe just give this a meaningful string?

packages/amplify_datastore/lib/types/DataStoreHubEvents/HubEventElement.dart
```diff
@@ -17,11 +17,17 @@ import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_inte
...
-    model = provider.getModelTypeByModelName(serializedData['modelName']).fromJson(serializedModelData);
+  late Model model;
+  HubEventElement(
+      Map<dynamic, dynamic> serializedData, ModelProviderInterface provider) {
+    Map<String, dynamic> serializedElement =
+        new Map<String, dynamic>.from(serializedData["element"]);
+    Map<String, dynamic> modelData =
+        new Map<String, dynamic>.from(serializedElement["model"]);
+    Map<String, dynamic> serializedModelData =
+        new Map<String, dynamic>.from(modelData["serializedData"]);
```
cshfang, line 28: Nit: Omit `new` keyword

packages/amplify_datastore/lib/method_channel_datastore.dart
```diff
@@ -23,18 +23,22 @@ const MethodChannel _channel = MethodChannel('com.amazonaws.amplify/datastore');
...
   /// plugins are needed to be added before that.
   Future<void> configureModelProvider(
-      {ModelProviderInterface modelProvider}) async {
+      {ModelProviderInterface? modelProvider}) async {
     try {
       return await _channel
           .invokeMethod('configureModelProvider', <String, dynamic>{
-        'modelSchemas':
-            modelProvider.modelSchemas.map((schema) => schema.toMap()).toList(),
+        'modelSchemas': modelProvider!.modelSchemas
```
cshfang, line 41: Should we be casting to non-null here when the param is nullable?

packages/amplify_datastore/lib/method_channel_datastore.dart
```diff
@@ -23,18 +23,22 @@ const MethodChannel _channel = MethodChannel('com.amazonaws.amplify/datastore');
...
+            .map((schema) => schema.toMap())
+            .toList(),
         'modelProviderVersion': modelProvider.version
```
cshfang, line 44: Weird that this didn't need a cast or null-safe access ...

packages/amplify_datastore/test/amplify_datastore_query_test.dart
fjnoyp: sounds good, update title to "some title"

packages/amplify_datastore/example/lib/main.dart
```diff
@@ -293,7 +297,8 @@ class _MyAppState extends State<MyApp> {
   deletePost(String id) async {
     try {
       _selectedPostForNewComment = null;
-      await Amplify.DataStore.delete(Post(id: id, title: null));
+      await Amplify.DataStore.delete(
+          Post(id: id, title: "", rating: 0, created: TemporalDateTime.now()));
```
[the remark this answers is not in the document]
fjnoyp: Sure that's a great point.  And leads us to other considerations as well.  

The delete really just needs the id field set to know which model to delete.  But because of null safety now requiring fields to be set, we have to set a bunch of junk fields to just delete correctly ... This isn't good user experience.  

We probably need to have a meeting for this actually where we consider if we should have a "justId" constructor for this deletion case or if we can update the delete method to just take an id. 

@haverchuck @Ashish-Nanda what are your thoughts on this?

In the meantime, as this is a relic of the past as you said, I think in the interests of time we should look to resolve this in another PR.  Perhaps a hashmap of id->model instances.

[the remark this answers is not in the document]
haverchuck: @cshfang @fjnoyp The docs from [amplify-android ](https://docs.amplify.aws/lib/datastore/data-access/q/platform/android#delete) assume the use of a full instance to delete, while[ amplify-ios](https://docs.amplify.aws/lib/datastore/data-access/q/platform/ios#delete) seems to describe the ability to delete with an instance, plus a 'justId'-type method. We could potentially do the ios version as well, but the lack of method overloading would be a challenge here.  Either we have a second api method (we tend to have a high bar for introducing these), or we would use optional parameters in a very hacky way - correct me if you see any alternatives.  

In any case, to @fjnoyp's point about "a bunch of junk fields", we should probably change our docs to show the query-then-delete pattern that native recommends if we go forward with this as-is.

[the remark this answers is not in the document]
fjnoyp: Sounds good.  I spoke with Ashish yesterday about it and we agreed we'll go with this as is for now.  

We'll make a note of it for now and address it later down the road.

packages/amplify_datastore/example/lib/main.dart
```diff
@@ -423,7 +428,7 @@ class _MyAppState extends State<MyApp> {
   @override
   void didChangeDependencies() {
     super.didChangeDependencies();
-    WidgetsBinding.instance.addPostFrameCallback((_) => executeAfterBuild());
+    WidgetsBinding.instance!.addPostFrameCallback((_) => executeAfterBuild());
```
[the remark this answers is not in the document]
fjnoyp: Yeah safer to add a null check beforehand

packages/amplify_datastore/example/lib/models/Blog.dart
```diff
@@ -23,8 +25,16 @@ import 'package:flutter/foundation.dart';
 class Blog extends Model {
   static const classType = const BlogType();
   final String id;
-  final String name;
-  final List<Post> posts;
+  final String? _name;
+  final List<Post>? _posts;
+
+  String get name {
+    return _name!;
```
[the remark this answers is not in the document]
fjnoyp: Yes that's correct we can't guarantee it's actually non null.

packages/amplify_datastore/example/lib/models/Blog.dart
```diff
@@ -64,30 +76,32 @@ class Blog extends Model {
     var buffer = new StringBuffer();
 
     buffer.write("Blog {");
-    buffer.write("id=" + id + ", ");
-    buffer.write("name=" + name);
+    buffer.write("id=" + "$id" + ", ");
+    buffer.write("name=" + "$_name");
```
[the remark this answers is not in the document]
fjnoyp: Yeah spoke to codegen team about this and mentioned it wasn't super urgent.  They responded saying that the codegen right now is easier on their side to generate in the current format so far.

packages/amplify_datastore/example/lib/models/Blog.dart
```diff
@@ -64,30 +76,32 @@ class Blog extends Model {
...
     buffer.write("}");
 
     return buffer.toString();
   }
 
-  Blog copyWith(
-      {@required String id, @required String name, List<Post> posts}) {
+  Blog copyWith({String? id, String? name, List<Post>? posts}) {
```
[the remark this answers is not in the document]
fjnoyp: The fields don't need to be required because we default to the local model instance's values otherwise.

packages/amplify_datastore/example/lib/queries_display_widgets.dart
```diff
@@ -119,7 +119,7 @@ Widget getWidgetToDisplayPost(
                   _postsToView[i].rating.toString() +
                   ", blog: " +
                   allBlogs
-                      .firstWhere((blog) => blog.id == _postsToView[i].blog.id)
+                      .firstWhere((blog) => blog.id == _postsToView[i].blog?.id)
```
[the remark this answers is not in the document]
fjnoyp: We're not casting blog here.  We're doing a conditional access, in this case I believe null would be returned based on this quick test I've done on Dartpad: 

`void main() {
  Container container = Container();
  
  print("hello" == container.test?.someString);
  
  print(container.test?.someString);
}

class Container {
  Child? test; 
}

class Child {
  late String someString; 
  
  Child(){
    someString = "stuff";
  }
}`

packages/amplify_datastore/example/lib/save_model_widgets.dart
```diff
@@ -32,7 +32,7 @@ Widget addPostWidget(
     TextEditingController titleController,
     TextEditingController ratingController,
     bool isAmplifyConfigured,
-    Blog defaultBlog,
+    Blog? defaultBlog,
```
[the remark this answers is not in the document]
fjnoyp: sounds good switched to named params

packages/amplify_datastore/test/resources/save_api/request/instance_without_predicate.json
```diff
@@ -4,7 +4,7 @@
         "id": "9fc5fab4-37ff-4566-97e5-19c5d58a4c22", 
         "title": "New Post being saved",
         "rating": 10,
-        "created": "2020-11-12T11:15:48+00:00",
+        "created": "2020-11-12T03:15:48+03:18",
```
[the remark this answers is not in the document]
fjnoyp: discrepancy between expected and actual timestamp.  Somehow they differed.

packages/amplify_datastore/example/lib/models/Blog.dart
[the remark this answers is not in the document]
cshfang: Yep just wanted to make sure that the codegen'd models will be doing this.

packages/amplify_datastore/example/lib/models/Blog.dart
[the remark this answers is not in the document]
cshfang: Will that be a problem if this is how codegen'd models look? Unfortunately, I haven't been able to keep up with the codegen changes and I just wanted to double check we're not generating models that have potential to cause app crashing.

packages/amplify_datastore/test/resources/save_api/request/instance_without_predicate.json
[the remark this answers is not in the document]
cshfang: Hmm, is that something we should be concerned about? I think we should try to understand what caused this discrepancy to make sure we haven't somehow affected something we did not intend to with our changes. For example (and this is just an example, I'm not sure what's actually happening), if a change to TemporalDateTime is somehow translating the underlying data model timestamps differently, we could be introducing weird bugs to customers that have nothing to do with null safety.

packages/amplify_datastore/example/lib/models/Blog.dart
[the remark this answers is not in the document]
fjnoyp: Yes we agreed that it's okay for now.  

When you query all of its fields will be returned.  Required fields must be set because you couldn't have saved a model without those required fields.  Thus when we force cast the underlying value should not be null. 

However, nested models currently have only their id field returned.  This means it others required fields can be null. 

Our resolution so far is to release this as is and to work on updating the iOS and Android bridge code to actually send back the fields of the nested models.  

Currently iOS does this but Android does not so we are coordinating with Android team to get that done.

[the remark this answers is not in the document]
cshfang: Ok gotcha. Sometimes I just need to put the conversation in context of code.

packages/amplify_datastore/lib/amplify_datastore_stream_controller.dart
```diff
@@ -105,7 +105,7 @@ _rebroadcast(String eventName, { HubEventPayload payload}) {
 
 _onCancel() {
   if (!_controller.hasListener) {
-    eventStream.cancel();
+    eventStream!.cancel();
```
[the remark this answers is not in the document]
fjnoyp: Probably always better to check if it's null before.  I'll update code to add null check thanks.

packages/amplify_datastore_plugin_interface/test/testData/Blog.dart
```diff
@@ -1,5 +1,5 @@
 /*
-* Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+* Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
fjnoyp: updated

packages/amplify_datastore_plugin_interface/lib/amplify_datastore_plugin_interface.dart
```diff
@@ -42,11 +43,15 @@ export 'src/types/models/subscription_event.dart';
...
-  /// Constructs an AmplifyPlatform.
-  DataStorePluginInterface(
-      {required Object token, required this.modelProvider})
+  /// Constructs an AmplifyPlatform
+  DataStorePluginInterface({required Object token, required this.modelProvider})
+      : super(token: token);
+
+  /// Internal use constructor
+  @protected
+  DataStorePluginInterface.protected({required Object token})
```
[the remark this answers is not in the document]
fjnoyp: That's a great point.  I'll update to use tokenOnly instead.

packages/amplify_datastore/test/test_models/Post.dart
fjnoyp: Hmm why would we change the models to be different from codegen models?  Seems strange to prettify and clean up a model that should be representative of what codegen is.

packages/amplify_datastore/test/amplify_datastore_save_test.dart
fjnoyp: Like before, somehow the timestamps were different and not aligned.  I'm not too sure how it was working in the first place (unit test wise) as it was returning a DateTime instead of our TemporalDateTime class ..

packages/amplify_datastore/lib/method_channel_datastore.dart
fjnoyp: This is a bit weird.  See comments: 

/// NOTE: modelProvider is nullable because the parent AmplifyDataStore class
  /// provides a default value

Since they extend from each other I can't make one required.. even though the implementation guarantees that it woudn't be null in AmplifyDataStore: 

```
@override
  Future<void> configureModelProvider(
      {ModelProviderInterface? modelProvider}) async {
    ModelProviderInterface provider =
        modelProvider == null ? this.modelProvider : modelProvider;
    if (provider.modelSchemas.isEmpty) {
```

packages/amplify_datastore/lib/method_channel_datastore.dart
fjnoyp: Yeah it seems that if you force cast previously the compiler doesn't both enforcing it for following accesses...

cshfang (approved):

HuiSF (approved):

packages/amplify_datastore/example/lib/models/TestEnum.dart
```diff
@@ -0,0 +1,18 @@
...
+*
+* or in the "license" file accompanying this file. This file is distributed
+* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+*/
+
+// ignore_for_file: public_member_api_docs
+
+enum TestEnum { yes, no, maybe }
```
haverchuck, line 18: Lol - reminds me of [this](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/canPlayType#return_value)

packages/amplify_datastore/example/lib/models/schema.graphql
```diff
@@ -0,0 +1,93 @@
+type Blog @model {
```
haverchuck, line 1: Add a license?

haverchuck (approved):
Made a note that you should add a license on one file; but I'm approving since I think you can add that without re-review.

packages/amplify_datastore/example/lib/main.dart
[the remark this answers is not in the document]
dnys1: It's initialized as part of [`runApp`](https://api.flutter.dev/flutter/widgets/runApp.html), so it's safe to unwrap outside of `main`.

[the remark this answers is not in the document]
fjnoyp: alright though does it hurt to switch to ? instead of !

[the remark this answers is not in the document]
dnys1: For sure, just wanted to share a tidbit.

Issue #642, opened 2021-06-14
Easily Access Nested Data with DataStore


Kilo-Loco:
TL;DR - I want to access nested data easily (`post.author.name`)

I'm running into the issue of loading the nested/associated data (i.e. Post Author) with the current Flutter DataStore implementation. I believe that most devs would expect to get associated data back with a single query as opposed to multiple queries to fill in the missing data of in individual object in the list of objects. 

Building something like a social media post feed, you would need to query the list of Posts, and for each Post, the Author would have to be queried for each Post. So then Amplify is forcing the dev into the position of figuring out when to make that second query for each post: 

1. Immediately after the query for the Posts, which could then lead to unnecessary queries for Authors if those Posts are never shown on screen.
2. Just before the next Post is displayed on the screen, which makes for a complex setup for managing the data.

Regardless of the approach, the data should be cached into memory for faster lookups, and this further complicates a developer's implementation of a social media post feed. This process would only get more complicated if there are multiple `hasOne` relationships on Post.

However, if the implementation was to follow the Amplify iOS approach of lazy loading the data, a lot of these issues are resolved for the developer and make Amplify much easier to use. The ideal solution would be to access two levels of data using a single query (i.e `post.author.name`) and three levels should require the additional query, since that data isn't directly relevant to the initial query (i.e `post.author.company.name`). Having the ability to query three levels deep would be nice, but two levels is a must.

Example Schema:
```graphql
type Post @model @auth(rules: [{allow: public}]) {
  id: ID!
  body: String!
  author: User @connection
}

type User @model @auth(rules: [{allow: public}]) {
  id: ID!
  name: String!
}
```

dnys1:
Hi @Kilo-Loco - apologies for your frustration and thank you for taking the time to create this issue and provide some feedback. You're absolutely right on all accounts and the Amplify team is working hard to fix this in coming versions. It's looking like we will be moving towards an implementation similar to what you've described which piggybacks off of iOS's lazy loading of nested models.

Please follow our tracking ticket for this issue and stay tuned in the coming releases! https://github.com/aws-amplify/amplify-flutter/issues/260

HuiSF:
With amplify-flutter 0.2.2 release, nested models (root model `belongsTo` nested models) will be populated with data to provide easy access.

OLTApp:
[...]

PR #658, opened 2021-06-21
feat(datastore): Populate belongs-to nested models

HuiSF:
**Issue #, if available:**

Related to #642 

**Description of changes:**

Add implementation to populate nested models in iOS. With this PR, nested models that are only with below model relationships are populated with data.

* ~has one~ (amplify-ios doesn't populate hasOne nested models [#1107](https://github.com/aws-amplify/amplify-ios/issues/1107))
* belongs to

Nested models with has-many relationship will be populated as an empty array.

In order to persist `modelName` in each resolved nested model (see below example), `flutterModelRegistration` object is needed to pass into functions that need access to it. Choosing passing `flutterModelRegistration` following Swift preferred dependency injection pattern...

Other changes:
- Reformatted code
- Convert CRLF files to LF to align with codebase

<details>
<summary>Before/After examples</summary>

<b>Query result before this change</b>
```json
{
  "author": {
    "id": "f2012597-a5f3-44ab-96ca-323770c517fb",
  },
  "rating": 10,
  "id": "ef675e9e-17a6-4b09-8b6d-84ef5283d8f3",
  "blog": {
    "id": "f188039d-c745-4e28-b9da-cfd002da284d"
  },
  "title": "Post 3"
}
```

<b>Query result after this change</b>
```json
{
  "author": {
    "posts": [],
    "department": {
      "org": {
        "departments": [],
        "name": "Writing Guild",
        "id": "27690304-347c-4b4d-a626-1bd5fdd0a286"
      },
      "id": "39a9da32-d7d1-4907-a84a-db92401849e2",
      "people": [],
      "name": "First Writing Dept"
    },
    "id": "f2012597-a5f3-44ab-96ca-323770c517fb",
    "name": "Someone Testing"
  },
  "rating": 10,
  "id": "ef675e9e-17a6-4b09-8b6d-84ef5283d8f3",
  "blog": {
    "posts": [],
    "name": "Nature Blog",
    "id": "f188039d-c745-4e28-b9da-cfd002da284d"
  },
  "title": "Post 3"
}
```

<b>iOS returned query result via FlutterMethodChannel that is going to be parsed by Dart models</b>
```json
{
  "author": {
    "id": "f2012597-a5f3-44ab-96ca-323770c517fb",
    "modelName": "Person",
    "serializedData": {
      "department": {
        "id": "39a9da32-d7d1-4907-a84a-db92401849e2",
        "modelName": "Department",
        "serializedData": {
          "id": "39a9da32-d7d1-4907-a84a-db92401849e2",
          "name": "First Writing Dept",
          "org": {
            "id": "27690304-347c-4b4d-a626-1bd5fdd0a286",
            "modelName": "Organization",
            "serializedData": {
              "departments": [],
              "id": "27690304-347c-4b4d-a626-1bd5fdd0a286",
              "name": "Writing Guild"
            }
          },
          "people": []
        }
      },
      "id": "f2012597-a5f3-44ab-96ca-323770c517fb",
      "name": "Simon Guck",
      "posts": []
    }
  },
  "blog": {
    "id": "f188039d-c745-4e28-b9da-cfd002da284d",
    "modelName": "Blog",
    "serializedData": {
      "id": "f188039d-c745-4e28-b9da-cfd002da284d",
      "name": "Nature Blog",
      "posts": []
    }
  },
  "id": "59f2c853-621c-44f0-b825-730443787885",
  "rating": 10,
  "title": "Post 1"
}
```

<b>Schema</b>

```graphql
type Blog @model {
  id: ID!
  name: String!
  posts: [Post] @connection(keyName: "byBlog", fields: ["id"])
}
type Post
  @model
  @key(name: "byBlog", fields: ["blogID"])
  @key(name: "byAuthor", fields: ["authorId"]) {
  id: ID!
  title: String!
  rating: Int!
  created: AWSDateTime
  blogID: ID!
  blog: Blog @connection(fields: ["blogID"])
  comments: [Comment] @connection(keyName: "byPost", fields: ["id"])
  authorId: ID!
  author: Person! @connection(fields: ["authorId"])
}
type Comment @model @key(name: "byPost", fields: ["postID", "content"]) {
  id: ID!
  postID: ID!
  post: Post @connection(fields: ["postID"])
  content: String!
}
type Person @model @key(name: "byDepartmentId", fields: ["departmentId"]) {
  id: ID!
  name: String!
  posts: [Post] @connection(keyName: "byAuthor", fields: ["id"])
  departmentId: ID!
  department: Department! @connection(fields: ["departmentId"])
}
type Department @model @key(name: "byOrgId", fields: ["orgId"]) {
  id: ID!
  name: String!
  people: [Person] @connection(keyName: "byDepartmentId", fields: ["id"])
  orgId: ID!
  org: Organization! @connection(fields: ["orgId"])
}
type Organization @model {
  id: ID!
  name: String!
  departments: [Department] @connection(keyName: "byOrgId", fields: ["id"])
}
```
</details>

----


packages/amplify_datastore/example/ios/unit_tests/resources/modelSchema/model_schema_maps.json
```diff
@@ -1,264 +1,279 @@
...
+                },
+                "isRequired": true,
+                "isArray": false,
+                "association": {
+                    "associationType": "BelongsTo",
+                    "targetName": "authorId"
+                }
+            }
+        }
+    },
```
HuiSF, line 152: Converted this file to LF mode to align with other files in the codebase. The actual change is to add the `author` data

packages/amplify_datastore/example/ios/unit_tests/resources/modelSchema/serialized_model_maps.json
```diff
@@ -1,49 +1,71 @@
...
+                        "id": "777",
+                        "modelName": "Department",
+                        "serializedData": {
+                            "id": "777",
+                            "name": "Department 1",
+                            "description": "Testing department 1"
+                        }
+                    }
+                }
+            }
```
HuiSF, line 52: Converted this file to LF mode to align with other files in the codebase. The actual change is to add the `author` data

packages/amplify_datastore/ios/Classes/types/model/FlutterSerializedModel.swift
```diff
@@ -156,37 +155,74 @@ struct FlutterSerializedModel: Model, JSONValueHolder {
...
+    private static func getModelSchema(flutterModelRegistration: FlutterModels, modelName: String) throws -> ModelSchema {
+        do {
+            return try FlutterDataStoreRequestUtils.getModelSchema(modelSchemas: flutterModelRegistration.modelSchemas, modelName: modelName)
+        } catch let error {
+            print("model \(modelName) is not registered.")
+            throw error
+        }
+    }
+
+    private static func generateSerializedJsonData(values: [String: JSONValue], flutterModelRegistration: FlutterModels, modelName: String) throws -> [String: Any] {
```
Jordan-Nelson, line 167: > In order to persist modelName in each resolved nested model (see below example), flutterModelRegistration object is needed to pass into functions that need access to it

Is this line from the PR description referring to these changes? is iOS not passing the model type back for the nested models? I may be wrong, but I believe Android is (with the updates that were recently merged in amplify-android). Associating the nested model with a model type seems like logic we would want done on the native side if possible.

packages/amplify_datastore/ios/Classes/types/model/FlutterSerializedModel.swift
```diff
@@ -156,37 +155,74 @@ struct FlutterSerializedModel: Model, JSONValueHolder {
         }
     }
 
-    private func generateSerializedData(modelSchema: ModelSchema) -> [String: Any]{
-        
+    private static func getModelSchema(flutterModelRegistration: FlutterModels, modelName: String) throws -> ModelSchema {
+        do {
+            return try FlutterDataStoreRequestUtils.getModelSchema(modelSchemas: flutterModelRegistration.modelSchemas, modelName: modelName)
+        } catch let error {
+            print("model \(modelName) is not registered.")
```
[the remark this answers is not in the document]
HuiSF: According to [the implementation](https://github.com/aws-amplify/amplify-flutter/blob/f67efea8d19b0ce4635bd8161363d730e1023358/packages/amplify_datastore/ios/Classes/FlutterDataStoreRequestUtils.swift#L43-L50) of the `getModelSchema` function, it only throws one type of exception which is no model schema is found by given schema name.

packages/amplify_datastore/ios/Classes/types/model/FlutterSerializedModel.swift
Jordan-Nelson: Given the update [to the test here](https://github.com/aws-amplify/amplify-android/pull/1390/files#diff-b7e5dd19e63c6d95721c95ab2d982fb2da430b6c29b190ee7a6fc2125c502edbR169) to add the modelSchema to the SerializedModel builder, I am fairly certain this data is being returned from amplify-android now.

A few reasons for wanting this on the native side:
- I haven't looked at the code in iOS, but I would guess that this is actually less complex and error prone to do on the native side. Getting the correct model type was pretty simple in amplify-android. It is essentially [these three lines](https://github.com/aws-amplify/amplify-android/pull/1390/files#diff-4a19dd8ac30d07afb369627a6e8c6d7018ff8cdd35a84c55f4b830c91e920aefR829-R831) of code. 
- Cleaner boundaries and less coupling between amplify-ios and amplify-flutter. With the logic to get the model name in amplify-flutter, the logic for serialized models is sort of split between the two libraries without a clean division.
- Consistency between amplify-iOS and amplify-Android

packages/amplify_datastore/ios/Classes/types/model/FlutterSerializedModel.swift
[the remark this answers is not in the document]
Jordan-Nelson: Got it. In that case, this is definitely a *nit* comment, but you might want to catch that specific error if you are going to include the print statement. If that method was updated to throw a different error, this print statement might not make sense.

packages/amplify_datastore/ios/Classes/types/model/FlutterSerializedModel.swift
HuiSF: > Consistency between amplify-iOS and amplify-Android

Good point, let me take a look.

HuiSF: Looked amplify-ios [`StorageEngineAdapter`](https://github.com/aws-amplify/amplify-ios/blob/main/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter%2BSQLite.swift#L236) and [SQL statement + result conversion](https://github.com/aws-amplify/amplify-ios/blob/c622147b17b198cdc9d15a4ff049d7aa520139e1/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement%2BModel.swift#L71-L109) implemenation, also compared it to amplify-android. There seem to be a major difference between the two platform - amplify-ios doesn't have an equivalent of `SerializedModel` as in amplify-android.

In amplify-ios, it extracts fields from sql result, and then reads Flutter passed `modelSchemas` to convert fields into desired types, and ultimately it constructs a dictionary with nested values. This dictionary then gets decoded into `JSONValue`s by [`FlutterSerializedModel`](https://github.com/aws-amplify/amplify-flutter/blob/f67efea8d19b0ce4635bd8161363d730e1023358/packages/amplify_datastore/ios/Classes/types/model/FlutterSerializedModel.swift#L30-L45) We can see here `FlutterSerializedModel` in amplify-ios expects a json like payload which is completely different from amplify-android expecting a `SerializedModel`.

Given the level of difference, I think align amplify-ios with Android is not an easy task that can be included with this PR, and it may be the call of amplify-ios maintainer considering there are many differences between the two libraries.

In addition, I found `ModelRegistry` provides similar functionality as `ModelSchemaRegistry`, I will see if I can use this class instead of injecting `flutterModelRegistration` in multiple method functions.

Jordan-Nelson: > and ultimately it constructs a dictionary with nested values

From that description, the iOS implementation doesn't sound that different from Android. A SerializedModel in amplify-android is just an object that wraps a dictionary of data, a model ID, and schema info. It sounds like iOS is returning the dictionary of data and the model ID, but not any data about the schema.

I will take a look at the iOS code tomorrow. Maybe the changes are more involved than I am thinking.

packages/amplify_datastore/ios/Classes/types/model/FlutterSerializedModel.swift
```diff
@@ -156,37 +155,74 @@ struct FlutterSerializedModel: Model, JSONValueHolder {
...
+                // If a field that has one or belongs to a model
+                else if case .string(let modelId) = deserializedValue["id"],
+                        case .model(let nextModelName) = field!.type {
+                    result[key] = [
+                        "id": modelId,
+                        "modelName": nextModelName,
+                        "serializedData": try generateSerializedJsonData(values: deserializedValue, flutterModelRegistration: flutterModelRegistration, modelName: nextModelName)
+                    ]
+                }
+            } else if case .string(let deserializedValue) = value {
```
Jordan-Nelson, line 187: *Question:* Can these `else if` statements be combined into one `else` ? The other cases are mostly handled the same way.

HuiSF: Yes absolutely :) I was pretty much following the examples in the codebase, it looks like a convention that to explicitly list all possible types and the way to handle the value decoding/casting respectively. I can merge them into one `else`

packages/amplify_datastore/ios/Classes/types/model/FlutterSerializedModel.swift
HuiSF: Yes the concepts are the same :D the major difference I meant is there is no middle layer (`SerializedModel`) to hold model schema info for the nested models. That's the reason I added `modelSchema` look up within `FlutterSerializedModel`. Originally, `FlutterSerializedModel` knows only the schema (passed in as a parameter of the `toMap` function) of the root model. 

Thanks for looking!

Jordan-Nelson: Got it. This definitely doesn't have to block this PR, but I think is at least worth more discussion. I think it will become difficult to maintain long term with the logic split up across amplify-ios and amplify-flutter, and the two native platforms returning different things. Maybe the decision to not create a SerializedModel in iOS was intentional though.

Jordan-Nelson (approved):
Overall the changes look good. I would prefer if amplify-ios were updated to return the schema info for the nested models rather than putting logic in amplify-flutter for that, but I don't think it is worth holding the PR up for that.

packages/amplify_datastore/ios/Classes/types/model/FlutterSerializedModel.swift
HuiSF: I'm going to keep it as is, as the if else flow also includes value parsing using `JSONValue` method, which cannot be combined. Thanks!

cshfang (approved):

Issue #612, opened 2021-06-03
some REST API method have a fatal error on android when body not included in request


ragingsquirrel3:
**Describe the bug**
 put, post, patch crash fatally on android when body not set. Error is `java.lang.IllegalArgumentException: method PATCH must have a request body.`

**To Reproduce**
1. Run the example api app in android in debug mode `flutter run -d sdk`
2. Click on "Patch" button (which calls API with no body in the request)
3. Get error and flutter crashes.

```
E/AndroidRuntime(31856): java.lang.IllegalArgumentException: method PATCH must have a request body.
E/AndroidRuntime(31856): 	at okhttp3.Request$Builder.method(Request.kt:254)
E/AndroidRuntime(31856): 	at com.amplifyframework.api.aws.sigv4.AppSyncSigV4SignerInterceptor.intercept(AppSyncSigV4SignerInterceptor.java:237)
E/AndroidRuntime(31856): 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
E/AndroidRuntime(31856): 	at okhttp3.internal.connection.RealCall.getResponseWithInterceptorChain$okhttp(RealCall.kt:201)
E/AndroidRuntime(31856): 	at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:517)
E/AndroidRuntime(31856): 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
E/AndroidRuntime(31856): 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
E/AndroidRuntime(31856): 	at java.lang.Thread.run(Thread.java:923)
```

**Expected behavior**
Certainly not a fatal error, though not sure exactly best way out, or why exactly it works on iPhone. Possible ways out:
* catch error and throw a flutter exception which is not fatal
* make the body a required parameter for `RestOptions` so this becomes a compile time error and forces developers to provide a body
* provide a default body, such as `"{}"`
* understand how this is not error in iOS and do that

**Platform**
Amplify Flutter current supports iOS and Android.  This issue is reproducible in (check all that apply):
[x] Android
[] iOS

<details>

  <summary>Output of <code>flutter doctor -v</code></summary>
  
  ```
Doctor summary (to see all details, run flutter doctor -v):
[...]

PR #661, opened 2021-06-22
fix(amplify_api): throw ApiException in android when PUT, POST, and PATCH REST requests have no body to prevent fatal error

ragingsquirrel3:
*Issue #, if available:* #612

*Description of changes:* In android, PUT, POST, and PATCH methods throw a fatal error if no body is included (iOS works fine) https://github.com/aws-amplify/amplify-android/issues/1355. This is because of okhttp3 library, underlying HTTP library for amplify-android. This PR smoothes that over by making it an API exception in android that will not crash the app.

Alternatively, this could be changed to just add a blank body in this scenario but that might have some unexpected behavior for customers if they are not aware. It could also have some handling in dart layer to make it consistent across platforms. However, I would rather not change the iOS behavior here to introduce an exception for something that does not currently throw one.

### DIFF
--- a/packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/FlutterApiRequest.kt
+++ b/packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/FlutterApiRequest.kt
@@ -58,7 +58,7 @@ class FlutterApiRequest {
             return request[CANCEL_TOKEN_KEY] as String
         }
 
-        fun getRestOptions(request: Map<String, Any>) : RestOptions {
+        fun getRestOptions(request: Map<String, Any>, methodName: String) : RestOptions {
 
             try {
                 val builder: RestOptions.Builder = RestOptions.builder()
@@ -80,6 +80,10 @@ class FlutterApiRequest {
                             builder.addHeaders(value as Map<String, String>)
                         }
                     }
+                }
+
+                if (methodName == "PATCH" && request[BODY_KEY] == null) {
+                    builder.addBody("".toByteArray())
                 }
                 return builder.build()
             } catch (cause: Exception) {
--- a/packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/FlutterApiRequest.kt
+++ b/packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/FlutterApiRequest.kt
@@ -15,7 +15,9 @@
 
 package com.amazonaws.amplify.amplify_api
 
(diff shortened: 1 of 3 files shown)


packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/rest_api/FlutterRestApi.kt
```diff
@@ -59,6 +65,18 @@ class FlutterRestApi {
...
+                if (!options.hasData() && (methodName == PUT || methodName == POST || methodName == PATCH)) {
+                    handler.post {
+                        ExceptionUtil.postExceptionToFlutterChannel(flutterResult, "ApiException",
+                                ExceptionUtil.createSerializedError(
+                                        ApiException("$methodName request must have a body", "Add a body to the request.")
+                                )
+                        )
+                    }
+                    return
+                }
```
HuiSF, line 78: Nit: could extract the logic into a throwable method similar to line 55-64, and indentations

HuiSF (commented):
Thanks for adding the request body check. 1 nit-picking comment

packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/rest_api/FlutterRestApi.kt
ragingsquirrel3: good suggestion, corrected in latest revisions

packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/FlutterApiRequest.kt
```diff
@@ -90,16 +92,24 @@ class FlutterApiRequest {
             }
         }
 
+        @JvmStatic
+        fun checkForEmptyBodyIfRequired(options: RestOptions, methodName: String): Void? {
```
HuiSF, line 96: nit: May omit the return type `Void?` so it returns `Unit` (not anything meaningful) by default in Kotlin, then can omit the `return` statement.

HuiSF (approved):
Thanks for adding the check and better formatting. Looks good!

packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/FlutterApiRequest.kt
ragingsquirrel3: good suggestion, done for latest revision

dnys1 (approved):
Thanks!

PR #668, opened 2021-06-23
Null safety datastore force cast exception for codegen

fjnoyp:
*Issue #, if available:*

Add new exception to be thrown when force casting in codegen models. 

*Description of changes:*

### DIFF
--- a/packages/amplify_datastore_plugin_interface/lib/src/types/exception/DataStoreExceptionMessages.dart
+++ b/packages/amplify_datastore_plugin_interface/lib/src/types/exception/DataStoreExceptionMessages.dart
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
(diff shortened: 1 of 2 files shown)


packages/amplify_datastore_plugin_interface/lib/src/types/exception/DataStoreExceptionMessages.dart
```diff
@@ -0,0 +1,27 @@
...
+ * permissions and limitations under the License.
+ */
+
+// ignore_for_file: public_member_api_docs
+
+/// Commonly used cross-category error messages.
+class DataStoreExceptionMessages {
+  static const codeGenRequiredFieldForceCastExceptionMessage =
+      "The field you are accessing is not nullable but has a null value."
+      "It was marked as required (!) in your schema.graphql but the containing model class was initialized without settings its value.";
```
Jordan-Nelson, line 22:
```suggestion
      "It was marked as required (!) in your schema.graphql but the containing model class was initialized without setting its value.";
```

packages/amplify_datastore_plugin_interface/lib/src/types/exception/DataStoreExceptionMessages.dart
```diff
@@ -0,0 +1,27 @@
...
+
+  static const codeGenRequiredFieldForceCastRecoverySuggestion =
+      "Please validate that the containing model class was initialized properly with all requried fields being initialized."
+      "This can happen when a nested model is returned but only its id field has been set";
```
Jordan-Nelson, line 26: I might add something a little more specific about performing another query.
```suggestion
      "This can happen when a nested model is returned but only its id field has been set. In this case, you may need to perform an additional query for the nested model.";
```

fjnoyp:
Already merged into null safety master during the rebase PR

PR #670, opened 2021-06-24
bug(datastore): HubEvent null check

dnys1:
*Issue #, if available:*
`_deleted` can be null on `HubEventWithMetadata`. This is by design, and is shared between both iOS and Android. Both platforms interpret this as being `false`.

*Description of changes:*
- Default to `_deleted = false`.
- Refactor HubEvent/HubEventWithMetadata for better testing.
- Add unit tests.
- Prefer `.cast()` or plain `Map` instead of `Map.from` which creates a new Map.

### DIFF
--- a/packages/amplify_datastore/lib/types/DataStoreHubEvents/HubEventElement.dart
+++ b/packages/amplify_datastore/lib/types/DataStoreHubEvents/HubEventElement.dart
@@ -14,6 +14,7 @@
  */
 
 import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart';
+import 'package:flutter/material.dart';
 
 part 'HubEventElementWithMetadata.dart';
 
@@ -32,6 +33,14 @@ class HubEventElement {
     var model = _parseModelFromMap(serializedHubEventElement, provider);
     return HubEventElement(model);
   }
+
+  @override
+  bool operator ==(Object other) =>
+      identical(this, other) ||
+      other is HubEventElement && model == other.model;
+
+  @override
+  int get hashCode => model.hashCode;
 }
 
 /// Retrieves the model instance from [serializedHubEventElement].
--- a/packages/amplify_datastore/lib/types/DataStoreHubEvents/HubEventElement.dart
+++ b/packages/amplify_datastore/lib/types/DataStoreHubEvents/HubEventElement.dart
@@ -13,8 +13,6 @@
  * permissions and limitations under the License.
  */
 
-import 'dart:ui' show hashValues;
-
 import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart';
 
 part 'HubEventElementWithMetadata.dart';
--- a/packages/amplify_datastore/lib/types/DataStoreHubEvents/HubEventElement.dart
+++ b/packages/amplify_datastore/lib/types/DataStoreHubEvents/HubEventElement.dart
@@ -13,7 +13,7 @@
  * permissions and limitations under the License.
  */
 
-import 'dart:ui';
+import 'dart:ui' show hashValues;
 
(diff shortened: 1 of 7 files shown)


packages/amplify_datastore/lib/types/DataStoreHubEvents/HubEventElement.dart
```diff
@@ -13,22 +13,48 @@
...
+
+  factory HubEventElement.fromMap(
+    Map serializedHubEventElement,
+    ModelProviderInterface provider,
+  ) {
+    var model = _parseModelFromMap(serializedHubEventElement, provider);
+    return HubEventElement(model);
   }
+
+  @override
```
haverchuck, line 38: We definitely need to start doing this more consistently.

packages/amplify_datastore/lib/types/DataStoreHubEvents/HubEventElementWithMetadata.dart
```diff
@@ -13,23 +13,56 @@
...
-    Map<String, dynamic> metadata =
-        new Map<String, dynamic>.from(serializedElement["syncMetadata"]);
-    version = metadata["_version"];
-    lastChangedAt = metadata["_lastChangedAt"];
-    deleted = metadata["_deleted"];
+  /// The version of the model.
+  final int version;
+
+  /// The last time the model was updated locally.
+  final DateTime lastChangedAt;
```
haverchuck, line 25: I think we agreed to try to keep the Hub event as consistent as possible with other platforms; in this case, Android and iOS express lastChangedAt as an Int (I am not sure why). I would suggest leaving this as is (and if we do end up keeping this change we need to remember to list it as a breaking change).

haverchuck (commented):
Left a few inline comments.

packages/amplify_datastore/lib/types/DataStoreHubEvents/HubEventElement.dart
dnys1: Yeah, it makes for easier testing and just trying to leave things better than I found found em. I can revert it.

packages/amplify_datastore/lib/types/DataStoreHubEvents/HubEventElementWithMetadata.dart
dnys1: Yeah that was my bad, I forgot about breaking changes. Although chances are, if anyone is using that value, they are converting it to a datetime object first. In any case, I've reverted the change.

haverchuck (approved):

haverchuck:
@dnys1  Approving but did not have time to test; please make sure you've tested thoroughly.

packages/amplify_datastore/lib/types/DataStoreHubEvents/HubEventElement.dart
haverchuck: Definetely would be good to add, but in a consistent manner.

packages/amplify_datastore/test/amplify_datastore_stream_controller_test.dart
```diff
@@ -275,7 +274,7 @@ void main() {
     expect((element.model as Post).created, equals(parsedDate));
     expect(element.deleted, equals(false));
     expect(element.version, equals(1));
-    expect(element.lastChangedAt, equals(1));
+    expect(element.lastChangedAt, equals(1624492860));
```
HuiSF, line 277: nit: this is the original code, think that we should get the expected value from `json` instead hardcoding it here.

HuiSF (approved):
Thanks for adding the unit tests! LGTM with 1 nit-picking comment

packages/amplify_datastore/test/amplify_datastore_stream_controller_test.dart
dnys1: Sure, just didn't feel like changing too much in this PR.

Issue #669, opened 2021-06-23
[amplify_auth_cognito][confirmPassword] throws the wrong exception


BerndWessels:
`Amplify.Auth.confirmPassword` throws `AuthException` when it should actually throw `InvalidPasswordException`.

The `underlyingException` is
```
com.amazonaws.services.cognitoidentityprovider.model.InvalidPasswordException: Password does not conform to policy: Password must have symbol characters (Service: AmazonCognitoIdentityProvider; Status Code: 400; Error Code: InvalidPasswordException; Request ID: e5e335f2-2553-44b9-a589-f1ca2d57fa24)
```

Because of this, it is unnecessary more complicated for the app to figure out what went wrong and give the correct feedback to the end user.

UPDATE:
After some further digging, this seems to be true for all other exceptions too :(

<details>

  <summary>Output of <code>flutter doctor -v</code></summary>
  
  ```
C:\Users\admin>flutter doctor -v
[√] Flutter (Channel stable, 2.0.4, on Microsoft Windows [Version 10.0.19042.1081], locale en-NZ)
    • Flutter version 2.0.4 at C:\Users\admin\flutter
    • Framework revision b1395592de (3 months ago), 2021-04-01 14:25:01 -0700
    • Engine revision 2dce47073a
    • Dart version 2.12.2

[!] Android toolchain - develop for Android devices (Android SDK version 30.0.3)
    • Android SDK at C:\Users\admin\AppData\Local\Android\sdk
    • Platform android-30, build-tools 30.0.3
    • Java binary at: C:\Program Files\Android\Android Studio\jre\bin\java
    • Java version OpenJDK Runtime Environment (build 11.0.8+10-b944.6842174)
    X Android license status unknown.
      Run `flutter doctor --android-licenses` to accept the SDK licenses.
      See https://flutter.dev/docs/get-started/install/windows#android-setup for more details.

[√] Chrome - develop for the web
    • Chrome at C:\Program Files\Google\Chrome\Application\chrome.exe

[√] Android Studio (version 4.1.0)
    • Android Studio at C:\Program Files\Android\Android Studio
    • Flutter plugin can be installed from:
       https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
       https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 11.0.8+10-b944.6842174)

[√] VS Code (version 1.57.0)
[...]

PR #672, opened 2021-06-24
fix(auth): refactor error handling in auth to match class names

Jordan-Nelson:
*Issue #, if available:* https://github.com/aws-amplify/amplify-flutter/issues/669

*Description of changes:*
- update auth error handling on Android to match on the class name, based on the list of exceptions that are supported in Dart.


packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthErrorHandlerTest.kt
```diff
@@ -366,7 +366,7 @@ class AuthErrorHandlerTest {
     @Test
     fun sessionUnavailableOfflineException() {
         // Arrange
-        val expectedCode = "AuthException"
+        val expectedCode = "SessionUnavailableOfflineException"
```
Jordan-Nelson, line 369: We previously were not mapping this code, but I think we want to given that we handle this code on the Dart side.

PR #676, opened 2021-06-25
Null safety master rebase

fjnoyp:
*Issue #, if available:*

*Description of changes:*


packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
```diff
@@ -69,11 +79,20 @@ void main() {
...
               deliveryMedium: "EMAIL",
               destination: "test@test.test"),
+=======
+          additionalInfo: {},
+          codeDeliveryDetails: {
+            "deliveryMedium": "EMAIL",
+            "attributeName": testAttributeKey,
+            "destination": testEmailValue
+          },
+>>>>>>> b432213fdfbacf7e0ec29981e95b81c1cbfd22c3
```
HuiSF, line 95: Unresolved conflict...

fjnoyp:
Hi guys this is the old PR with too many file diffs.  Deleting it and creating the other one now.

packages/amplify_datastore_plugin_interface/lib/amplify_datastore_plugin_interface.dart
```diff
@@ -43,26 +43,58 @@ export 'src/types/models/subscription_event.dart';
...
-  /// Constructs an AmplifyPlatform
-  DataStorePluginInterface({required Object token, required this.modelProvider})
+  /// Datastore sync interval (in seconds)
+  late final int? syncInterval;
+
+  /// Datastore max number of records to sync
+  late final int? syncMaxRecords;
+
+  /// Datastore page size to sync
+  late final int? syncPageSize;
```
HuiSF, line 56: Should these fields actually be nullable for the use case of `tokenOnly`? as they will never be assigned within this use case

packages/amplify_api_plugin_interface/lib/src/exceptions/ApiException.dart
```diff
@@ -17,22 +17,36 @@ import 'package:amplify_core/types/exception/AmplifyException.dart';
...
 class ApiException extends AmplifyException {
+  /// HTTP status of response, only available if error
+  final int? httpStatusCode;
+
   /// Named constructor
   const ApiException(String message,
-      {String? recoverySuggestion, String? underlyingException})
+      {String? recoverySuggestion,
+      String? underlyingException,
+      this.httpStatusCode})
```
cshfang, line 27: _Might_ be nit
I think the original constructor on `master` was:
```dart
ApiException(String message,
    {String recoverySuggestion, String underlyingException, int statusCode})
    : httpStatusCode = statusCode,
      super(message,
          recoverySuggestion: recoverySuggestion,
          underlyingException: underlyingException);
```
So doing `this.httpStatusCode` would change that named parameter from `statusCode` to `httpStatusCode`. I don't think that's what we want to do ... but I'm also not seeing this actually being used anywhere so the change might be fine?

Side note nit: This is original code but the comment calls this a `Named constructor` but it's literally not named 😅

Jordan-Nelson (approved):
The changes within Auth that I had made to master look good to me. I did not review the entire PR

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
```diff
@@ -320,6 +320,27 @@ class AmplifyAuthCognitoMethodChannel extends AmplifyAuthCognito {
     }
   }
 
+  @override
+  Future<Map<String, UpdateUserAttributeResult>> updateUserAttributes(
+      {required UpdateUserAttributesRequest request}) async {
+    Map<String, UpdateUserAttributeResult> res;
```
cshfang, line 326: Is this used anywhere anymore in this method?

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
```diff
@@ -320,6 +320,27 @@ class AmplifyAuthCognitoMethodChannel extends AmplifyAuthCognito {
...
+      final Map<String, dynamic>? data =
+          await _channel.invokeMapMethod<String, dynamic>(
+        'updateUserAttributes',
+        <String, dynamic>{
+          'data': request.serializeAsMap(),
+        },
+      );
+      if (data == null)
+        throw AmplifyException(
+            AmplifyExceptionMessages.nullReturnedFromMethodChannel);
```
cshfang, line 337: Nit: Wrap if with curly braces for readability
```suggestion
      if (data == null) {
        throw AmplifyException(
            AmplifyExceptionMessages.nullReturnedFromMethodChannel);
      }
```

packages/amplify_flutter/lib/categories/amplify_auth_category.dart
```diff
@@ -180,6 +180,16 @@ class AuthCategory {
         : throw _pluginNotAddedException("Auth");
   }
 
+  /// Updates multiple user attributes and returns a map of  [UpdateUserAttributeResult]
```
cshfang, line 183: Nit:
```suggestion
  /// Updates multiple user attributes and returns a map of [UpdateUserAttributeResult]
```

packages/amplify_auth_plugin_interface/lib/src/Session/UpdateUserAttributesRequest.dart
```diff
@@ -0,0 +1,36 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import 'package:flutter/foundation.dart';
```
cshfang, line 16: Nit: Don't think this is used

packages/amplify_datastore_plugin_interface/lib/amplify_datastore_plugin_interface.dart
cshfang: +1 I think the `late` implies that these could be initialized post-constructor but we do not actually have mechanisms for that so we could just be creating situations where we assume something is non-nullable when it's actually very possible it's never defined (hence null).

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterUpdateUserAttributeResult.kt
```diff
@@ -16,6 +16,7 @@
 package com.amazonaws.amplify.amplify_auth_cognito.types
 
 import com.amazonaws.amplify.amplify_auth_cognito.setNextStep
+import com.amazonaws.amplify.amplify_auth_cognito.utils.serializeAuthUpdateAttributeResult
```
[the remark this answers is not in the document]
haverchuck: @fjnoyp I think Chris is correct- this should probably just use the `FlutterUpdateUserAttributeResult.kt` file from the master branch.

packages/amplify_auth_cognito/ios/Classes/FlutterUpdateUserAttributeResult.swift
```diff
@@ -120,3 +108,4 @@ func setState(res: AmplifyOperation<AuthUpdateUserAttributeRequest, AuthUpdateAt
     }
     return state
 }
+
```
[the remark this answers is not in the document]
haverchuck: @fjnoyp  Yeah - again I think you can take what's in master branch for this one.

packages/amplify_datastore/lib/amplify_datastore.dart
```diff
@@ -61,6 +78,27 @@ class AmplifyDataStore extends DataStorePluginInterface {
...
   }
 
+  @override
+  Future<void> configureDataStore(
+      {ModelProviderInterface? modelProvider,
+      int? syncInterval,
+      int? syncMaxRecords,
+      int? syncPageSize}) async {
+    ModelProviderInterface provider =
+        modelProvider == null ? this.modelProvider : modelProvider;
```
cshfang, line 88: When making the changes in the interface for `modelProvider` to be nullable you will probably come across some issues here:
```suggestion
    ModelProviderInterface provider =
        modelProvider ?? this.modelProvider!;
```
Since the only usage of the `tokenOnly` constructor is internal, we should be confident in saying that the only way to construct `AmplifyDataStore` is via the default constructor where we've made the modelProvider required, so it should be safe to cast `this.modelProvider` to non-null.

haverchuck (changes requested):
I mainly looked at auth.  Two instances of incorrect platform side conflict resolution, but I think you can just take what is in master. Otherwise looks good.  This was a ton of work - thanks @fjnoyp !

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
cshfang: Oops I was only referring to line 326

packages/amplify_datastore_plugin_interface/lib/src/types/exception/DataStoreExceptionMessages.dart
```diff
@@ -0,0 +1,27 @@
...
+ * permissions and limitations under the License.
+ */
+
+// ignore_for_file: public_member_api_docs
+
+/// Commonly used cross-category error messages.
+class DataStoreExceptionMessages {
+  static const codeGenRequiredFieldForceCastExceptionMessage =
+      "The field you are accessing is not nullable but has a null value."
+      "It was marked as required (!) in your schema.graphql but the containing model class was initialized without settings its value.";
```
cshfang, line 22: Nit:
```suggestion
      "It was marked as required (!) in your schema.graphql but the containing model class was initialized without setting its value.";
```

packages/amplify_datastore_plugin_interface/lib/amplify_datastore_plugin_interface.dart
fjnoyp: Shouldn't they be nullable because they won't always be passed in the constructor?

@HuiSF 

Agree with cshfang, I'll remove the late keyword

HuiSF: Yes @fjnoyp I meant to remove `late` sorry for the confusion

fjnoyp: Great removed the late modifier

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterUpdateUserAttributeResult.kt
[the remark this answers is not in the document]
fjnoyp: Took the master branch version.

packages/amplify_auth_cognito/ios/Classes/FlutterUpdateUserAttributeResult.swift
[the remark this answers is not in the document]
fjnoyp: Thanks, took the master branch version.

packages/amplify_api_plugin_interface/lib/src/exceptions/ApiException.dart
fjnoyp: Yes that's correct and we don't use that anywhere except within the ApiException class itself.  I found it very weird that a named parameter has a different name for the same thing so wanted to realign the code there. 

Ah good catch, I'll remove that comment as well.

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
fjnoyp: It's not thanks!

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
fjnoyp: Thanks added {}

packages/amplify_auth_plugin_interface/lib/src/Session/UpdateUserAttributesRequest.dart
fjnoyp: Thanks for catching that, removed it

packages/amplify_datastore/lib/amplify_datastore.dart
fjnoyp: Sounds good added that change in

packages/amplify_datastore_plugin_interface/lib/src/types/exception/DataStoreExceptionMessages.dart
fjnoyp: thanks!

haverchuck (approved):

packages/amplify_datastore_plugin_interface/lib/amplify_datastore_plugin_interface.dart
```diff
@@ -47,13 +47,13 @@ abstract class DataStorePluginInterface extends AmplifyPluginInterface {
   late final ModelProviderInterface modelProvider;
```
cshfang, line 47: Hmm what is the reasoning for keeping this as `late final`? I think it should also just be nullable.

Ashish-Nanda (approved):

cshfang (approved):

offlineprogrammer (approved):

packages/amplify_datastore_plugin_interface/lib/amplify_datastore_plugin_interface.dart
fjnoyp: It should always be initialized and have a value right?

PR #687, opened 2021-06-29
Fix enum string function

dnys1:
*Issue #, if available:*
Exception if decoding an null enum string, like the line below from `AllTypeOptionalModel` where `json['enumType']` is null.

Screen Shot 2021-06-29 at 3 54 18 PM

*Description of changes:*
- Make argument nullable

### DIFF
--- a/packages/amplify_datastore_plugin_interface/lib/src/types/utils/parsers.dart
+++ b/packages/amplify_datastore_plugin_interface/lib/src/types/utils/parsers.dart
@@ -23,4 +23,4 @@ String? enumToString(Object? o) =>
 
 // only to be used internally by amplify-flutter library
 T? enumFromString<T>(String? key, List<T> values) =>
-    values.firstWhereOrNull((v) => key == enumToString(v));
+    key == null ? null : values.firstWhereOrNull((v) => key == enumToString(v));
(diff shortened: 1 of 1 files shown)


packages/amplify_datastore_plugin_interface/lib/src/types/utils/parsers.dart
```diff
@@ -16,11 +16,11 @@
...
+import 'package:collection/collection.dart' show IterableExtension;
+
 String? enumToString(Object? o) =>
     o != null ? o.toString().split('.').last : null;
 
 // only to be used internally by amplify-flutter library
-T? enumFromString<T>(String key, List<T> values) =>
-    values.firstWhereOrNull((v) => key == enumToString(v));
+T? enumFromString<T>(String? key, List<T> values) =>
+    key == null ? null : values.firstWhereOrNull((v) => key == enumToString(v));
```
cshfang, line 26: Can't verify from my phone but it seems like the ternary isn't even needed unless there are compiler complaints about trying to compare a nullable String. But this is nitpicky anyway

cshfang (approved):

HuiSF (approved):

fjnoyp (approved):

packages/amplify_datastore_plugin_interface/lib/src/types/utils/parsers.dart
dnys1: Yeah, you're right. It was to protect against the case where the enum has a case named `null`, but I just realized Dart doesn't let you do that anyway.

PR #697, opened 2021-07-03
chore(auth): User attributes/sign up docs

dnys1:
*Issue #, if available:*

*Description of changes:*
- Add user attributes for easy use.
- Make necessary fields required.
- Add doc comments to sign up options.


packages/amplify_auth_cognito/lib/src/CognitoSignUp/cognito_user_attributes.dart
```diff
@@ -0,0 +1,24 @@
+/// Standard user attributes available for configuring via `Amplify.Auth.signUp`,
+/// or updating via `Amplify.Auth.updateUserAttribute` and
+/// `Amplify.Auth.updateUserAttributes`.
+///
+/// More information about these attributes can be found
+/// [here](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-attributes.html).
+abstract class CognitoUserAttributes {
```
Jordan-Nelson, line 7: *Question:* Probably out of scope for your changes, but curious to get thoughts on at some point aligning this with iOS/Android. By that I mean requiring an enum/class for .signUp, .updateAttribute, etc. instead of using a string for the key name.

iOS uses enums, which still works for custom attributes because swift enums can have associated data. I think we would probably have do something similar to Android.

Pro: By requiring something more specific than a string, I think it is clearer to the customer what to use. This PR helps with that as well, but without looking at the docs I would probably not know that the CognitoUserAttributes class exists.

Con: More verbose.

Example:

```dart
// new class, similar to Android
class UserAttributeKey {
  String keyName;
  UserAttributeKey._(this.keyName);
  factory UserAttributeKey.email() => UserAttributeKey._('email');
  factory UserAttributeKey.phone() => UserAttributeKey._('phone');
  factory UserAttributeKey.custom(String keyName) =>
      UserAttributeKey._(keyName);
}

// existing class (with code removed for simplification) w/ updates to use UserAttributeKey
class AuthUserAttribute {
  UserAttributeKey userAttributeKey;
  var value;
  AuthUserAttribute({required this.userAttributeKey, required this.value});

    final Map<String, dynamic> pendingRequest = <String, dynamic>{};
    pendingRequest['userAttributeKey'] = userAttributeKey.keyName;
    pendingRequest['value'] = value;
    return pendingRequest;
  }
}

// example w/ sign up
void signUp() async {
  var res = await Amplify.Auth.signUp(
      username: 'username',
      password: 'password',
      options: CognitoSignUpOptions(userAttributes: {
        UserAttributeKey.email(): 'email@example.com',
      }));
}

// alternate example using an array of AuthUserAttribute (matches iOS)
void signUpAlternate() async {
  var res = await Amplify.Auth.signUp(
      username: 'username',
      password: 'password',
      options: CognitoSignUpOptions(userAttributes: [
        AuthUserAttribute(
          userAttributeKey: UserAttributeKey.email(),
          value: 'email@example.com',
        )
      ]));
}
```

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
```diff
@@ -13,6 +13,7 @@
  * permissions and limitations under the License.
  */
 
+import 'package:amplify_auth_cognito/src/CognitoSignUp/cognito_user_attributes.dart';
```
Jordan-Nelson, line 16: Can you export the class as part of the main amplify_auth_cognito package?

Jordan-Nelson:
Changes mostly LGTM. I left a long question that you can feel to ignore if you want and one request for a change to export the class

packages/amplify_auth_cognito/lib/src/CognitoSignUp/cognito_user_attributes.dart
dnys1: That's actually an awesome idea - I forgot about that pattern for enum classes in Dart. Gonna make the change!

packages/amplify_auth_cognito/test/amplify_auth_cognito_confirmSignIn_test.dart
dnys1: Yep - good catch!

packages/amplify_auth_cognito/lib/src/CognitoSignUp/cognito_user_attributes.dart
haverchuck: @Jordan-Nelson @dnys1  Wouldn't this be a breaking change?

haverchuck (commented):
Left one question

packages/amplify_auth_cognito/lib/src/CognitoSignUp/cognito_user_attributes.dart
Jordan-Nelson: Yes, it would be breaking. Sorry, I should have called that out!

We could go with what @dnys1 originally had for now and then roll the other change into a later update if we think there are other breaking changes we want to group it with.

Jordan-Nelson: I also would not be opposed to merging it and bumping to 0.3.x. I sort of think that is an advantage of not being 1.x.x yet (semi frequent breaking changes shouldn't be too unexpected). We should definitely try to group breaking changes, but I don't think we need to follow the frequency that Android or iOS is following.

dnys1: Cool, no worries. I'll put the changes in another branch for 0.3. Good thinking!

Jordan-Nelson: We probably need to establish norms around handling breaking changes. So, @dnys1 If you don't want that to hold you up, you could always open it as a separate PR.

packages/amplify_auth_cognito/test/amplify_auth_cognito_signup_test.dart
```diff
@@ -64,7 +65,7 @@ void main() {
             username: 'testUser',
             password: '123',
             options: CognitoSignUpOptions(userAttributes: {
-              "email": "test@test.com",
+              CognitoUserAttributes.email: "test@test.com",
```
cshfang, line 68: Nit: Kinda seems like we could just extract and use the same request three times? Ditto with some of the other tests.

cshfang (approved):

Jordan-Nelson (approved):

packages/amplify_auth_cognito/test/amplify_auth_cognito_signup_test.dart
dnys1: Probably, yeah. It would be good to do a sweep through all the tests at some point and clean them up.

Issue #688, opened 2021-06-30
No HTTP response data when Amplify.API request is erroneous


ioantsaf:
**Describe the bug**
When an API status code > 299 is returned, an `ApiException` is thrown. The exception now contains the `int httpStatusCode`, as implemented [here](https://github.com/aws-amplify/amplify-flutter/pull/590), but it does not contain the response data (if available).

It would be a good improvement to include the HTTP response data, if available, like it is implemented in the [js SDK, where one can access the (erroneous) response data in the `error.response` object](https://docs.amplify.aws/lib/restapi/fetch/q/platform/js).

**To Reproduce**
1. Perform an `Amplify.API.get()` request (or `.put()`, `.post()` etc.).
2. If the HTTP status code returned is > 299, an `ApiException` is thrown.
3. The exception does not contain the error response data, nor is it accessible as part of the exception message.

**Expected behavior**
The error response data is available as part of the `ApiException` object.

**Platform**
This issue is reproducible in:
[X] Android
[X] iOS

ragingsquirrel3:
@ioantsaf thanks for bringing this request to our attention. I can see how the raw response might be useful. Offhand, I'm not sure how much effort might be required to add that to the exception in the Flutter library in a similar manner to JS code you mentioned. We will put this in our backlog for further consideration/investigation.

ragingsquirrel3:
@ioantsaf the request was implemented in PR #721 and should go out with the next release.

dnys1:
This is now available in amplify_api `0.2.1`: https://pub.dev/packages/amplify_api/changelog

PR #709, opened 2021-07-08
chore(api): Add API response headers/status

dnys1:
*Issue #, if available:*
- #688 
- #612 

*Description of changes:*
- Add `statusCode` and `headers` to `RestResponse`
- Do not throw `ApiException` on HTTP 4xx/5xx anymore. Instead, return `RestResponse`

_Limitations_:
- Currently, `headers` are only available on iOS for successful calls (2xx?).
- Getting the `statusCode` on Android is a little hacky and could benefit from a public getter.

### DIFF
--- a/packages/amplify_api/android/build.gradle
+++ b/packages/amplify_api/android/build.gradle
@@ -59,7 +59,7 @@ dependencies {
     implementation "com.amplifyframework:aws-api-appsync:1.19.0"
 
     testImplementation 'junit:junit:4.13'
-    testImplementation 'org.mockito:mockito-core:3.1.0'
+    testImplementation 'org.mockito:mockito-core:3.10.0'
     testImplementation 'org.mockito:mockito-inline:3.1.0'
     testImplementation 'androidx.test:core:1.2.0'
     testImplementation 'org.robolectric:robolectric:4.3.1'
--- a/packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/rest_api/FlutterRestApi.kt
+++ b/packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/rest_api/FlutterRestApi.kt
@@ -3,7 +3,6 @@ package com.amazonaws.amplify.amplify_api.rest_api
 import android.os.Handler
 import android.os.Looper
 import com.amazonaws.amplify.amplify_api.FlutterApiRequest
-import com.amazonaws.amplify.amplify_api.FlutterGraphQLApi
 import com.amazonaws.amplify.amplify_api.OperationsManager
 import com.amazonaws.amplify.amplify_core.exception.ExceptionUtil
 import com.amplifyframework.api.ApiException
@@ -13,6 +12,8 @@ import com.amplifyframework.api.rest.RestResponse
 import com.amplifyframework.core.Amplify
 import com.amplifyframework.core.Consumer
 import io.flutter.plugin.common.MethodChannel.Result
+import okhttp3.internal.http.HttpMethod
+import java.util.*
 import kotlin.reflect.KFunction3
 import kotlin.reflect.KFunction4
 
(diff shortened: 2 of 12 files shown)


packages/amplify_api/example/.gitignore
```diff
@@ -57,7 +57,8 @@ node_modules/
 aws-exports.js
 awsconfiguration.json
 amplifyconfiguration.json
```
Jordan-Nelson, line 59: *nit:* You could probably remove `amplifyconfiguration.json`. I believe that is what the CLI does for flutter projects now. I think `.dart` is missing and `.json` is present because the CLI use to add `.json` even for flutter projects, but that has been updated.

packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/rest_api/FlutterRestApi.kt
```diff
@@ -58,14 +59,25 @@ class FlutterRestApi {
                 return
             }
 
+            // Protect against empty bodies, causing Okhttp3 to crash. ByteArray must be non-empty
+            // so that AppSync SDK does not swap it for null.
+            if (HttpMethod.requiresRequestBody(methodName.toUpperCase(Locale.ROOT)) && !options.hasData()) {
```
ragingsquirrel3, line 64: a few questions about this part and related logic:
* I have an open PR for this bug, but in a different way https://github.com/aws-amplify/amplify-flutter/pull/661 though that one can be closed, my bad for not self-assigning issue. 
* Do we want to provide a blank string for the user as the response body, only in Android? It may be unlikely for most folks, but adding a body to their request without them knowing could have some unintended outcomes, depending on their backend implementation. Sure, we want to prevent a fatal error (current behavior), but another possibly is a non-fatal exception so they know they have to fix the body on their end. In most cases, I would expect these requests to require bodies, although HTTP itself does not require it.
* nice part here is `requiresRequestBody`. I did not know about that and it's useful here.

packages/amplify_api_plugin_interface/lib/src/RestAPI/RestResponse.dart
```diff
@@ -13,9 +13,37 @@
...
 
+import 'dart:convert';
 import 'dart:typed_data';
 
+/// An HTTP response from a REST API.
 class RestResponse {
-  Uint8List data;
-  RestResponse({required this.data});
+  /// The response status code.
+  final int statusCode;
```
ragingsquirrel3, line 22: question: IIRC the status code is only available for error on iOS. PR description mentions adding stuff when available from platform. Am I correct in understanding the status code here is only available on IOS? Are there other platform discrepancies here?

ragingsquirrel3 (commented):
General question: This PR looks at 2 issues. Would it make sense to split into 2 PRs? The thing about android requests needing bodies has some more questions IMO and is kinda separate, already has open PR, so it might help to split into 2 so we can merge them separately, or revert them separately if needed (unlikely), so there would be separate PR for the response data feature.

packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/rest_api/FlutterRestApi.kt
dnys1: I'm open to either. iOS does not throw on an empty body for these methods, though. This seems like an Okhttp3 thing, only.

packages/amplify_api/android/build.gradle
```diff
@@ -59,7 +59,7 @@ dependencies {
     implementation "com.amplifyframework:aws-api-appsync:1.19.0"
 
     testImplementation 'junit:junit:4.13'
-    testImplementation 'org.mockito:mockito-core:3.1.0'
+    testImplementation 'org.mockito:mockito-core:3.10.0'
```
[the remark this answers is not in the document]
dnys1: Yeah, we should probably update all of them - I couldn't get tests to run before.

packages/amplify_api_plugin_interface/lib/src/RestAPI/RestResponse.dart
dnys1: Yeah, you're right - my mistake. The status code is not available on iOS. I had put "200" in for success cases but that's technically incorrect.

dnys1:
Changes need to be addressed. Will re-open at a later point.

packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/rest_api/FlutterRestApi.kt
ragingsquirrel3: You are right only an okhttp3 (thus android) thing, but I do not believe ios adds bodies to requests like this (correct me if I'm wrong) so adding it only in android could just make it more opaque and unexpected. You have a point though, that any android solutions adds to platform discrepancy, even if it makes it less bad. Another option would be to add a default body in dart so that it avoids the error and makes it consistent on platforms.

PR #711, opened 2021-07-09
chore: make .gitignore more specific for amplify dirs and tweak integ…

ragingsquirrel3:
*Description of changes:*

* Make a line in .gitignore more specific as I believe it is ignoring files too aggressively.
* There is also a tweak to the melos tasks for integration tests. I learned they do not work as expected when there are multiple packages with integration tests in that it then prompts the user to choose which package, which then causes issues when trying to automatically do all. The change makes it do them all one at a time with no further prompt.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,7 @@ Podfile.lock
 
 
 # amplify resources from example apps
-amplify/
+**/example/amplify/
 
 dist/
 node_modules/
--- a/melos.yaml
+++ b/melos.yaml
@@ -37,18 +37,10 @@ scripts:
       - Requires running Android and iOS simulators.
 
   test:integration:android:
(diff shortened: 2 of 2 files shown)


haverchuck (approved):

offlineprogrammer (approved):

Jordan-Nelson (approved):

Salakar:
> There is also a tweak to the melos tasks for integration tests. I learned they do not work as expected when there are multiple packages with integration tests in that it then prompts the user to choose which package, which then causes issues when trying to automatically do all. The change makes it do them all one at a time with no further prompt.

Extra context on how this works in melos; `melos run test:integration:android --no-select` - the `no-select` arg for this works e.g. for CI environments so there's no prompts - a few people have found it helpful locally to be able to test a specific package that they may have been working on hence the package selection prompt, the default option in the prompt (`[ENTER]`) however will run all packages locally.

PR #724, opened 2021-07-15
chore(auth): add integration test coverage for auth

Jordan-Nelson:
*Issue #, if available:* n/a

*Description of changes:*
- add tests for auth hub events
- add tests for update password
- add tests for exceptions in Auth.signUp()
- add tests for fetchSession
- add tests for getCurrentUser

This PR adds integration tests for everything in auth that doesn't require a confirmation code or a particular config


packages/amplify_auth_cognito/example/integration_test/sign_in_sign_out_test.dart
```diff
@@ -48,18 +50,84 @@ void main() {
...
+      expect(finalAuthSession.isSignedIn, isFalse);
+    });
+
+    testWidgets('should not throw even if there is no user to sign out',
+        (WidgetTester tester) async {
+      // ensure that no user is currently logged in
+      final authSession = await Amplify.Auth.fetchAuthSession();
+      expect(authSession.isSignedIn, isFalse);
 
+      // call signOut without and expection for an exception
```
Jordan-Nelson, line 129:
```suggestion
      // call signOut without an expectation for an exception
```

packages/amplify_auth_cognito/example/integration_test/sign_in_sign_out_test.dart
```diff
@@ -24,13 +24,15 @@ import 'utils/setup_utils.dart';
...
-
-  group('signIn and signOut', () {
-    setUpAll(() async {
+  group('signIn', () {
+    late String username;
+    late String password;
+    setUp(() async {
       await configureAuth();
 
+      username = generateUsername();
```
Jordan-Nelson, line 33:
```suggestion
      // create new user for each test
      username = generateUsername();
```

packages/amplify_auth_cognito/example/integration_test/utils/validation_utils.dart
```diff
@@ -0,0 +1,42 @@
...
+// validates that the user sub exists and is a valid uuid
+bool isValidUserSub(String? value) {
+  if (value == null) {
+    return false;
+  }
+  final uuidRegExp = RegExp('^$_uuidMatch\$');
+  return uuidRegExp.hasMatch(value);
+}
+
+// validates that the identityId exists and matches an the pattern region:uuid
```
Jordan-Nelson, line 16:
```suggestion
// validates that the identityId exists and matches the pattern region:uuid
```

packages/amplify_auth_cognito/example/integration_test/utils/validation_utils.dart
```diff
@@ -0,0 +1,42 @@
...
+        credentials.awsSecretKey!.isNotEmpty &&
+        credentials.sessionToken!.isNotEmpty;
+  } catch (e) {
+    return false;
+  }
+}
+
+// validates that the tokens exist and are non empty strings
+bool isValidAWSCognitoUserPoolTokens(AWSCognitoUserPoolTokens? tokens) {
+  return tokens != null &&
```
ragingsquirrel3, line 38: nit: This seems like a different pattern to do the same thing as previous method, unless I'm missing something. It might be easier to just pick one of the patterns.

packages/amplify_auth_cognito/example/integration_test/utils/validation_utils.dart
```diff
@@ -0,0 +1,42 @@
+import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
+
+final _hex = '[0-9a-f]';
+final _uuidMatch = '$_hex{8}-$_hex{4}-$_hex{4}-$_hex{4}-$_hex{12}';
+final _regionMatch = '[a-z]{2}-[a-z]{4,9}-[0-9]{1}';
+
```
ragingsquirrel3, line 6: Nice job on these utils!

packages/amplify_auth_cognito/example/integration_test/get_current_user_test.dart
```diff
@@ -0,0 +1,74 @@
...
+    testWidgets('should throw SignedOutException if the user is signed out',
+        (WidgetTester tester) async {
+      await Amplify.Auth.signOut();
+      try {
+        await Amplify.Auth.getCurrentUser();
+      } catch (e) {
+        expect(e, TypeMatcher<SignedOutException>());
+        return;
+      }
+      throw Exception('Expected SignedOutException');
```
ragingsquirrel3, line 71: kinda dumb question: Is it better to throw exception like this or call `fail()`? Can we confirm the test will fail and the task returns non 0 exit status when this exception is thrown? It would seem obvious that it would, just I've just noticed weird stuff with the integration test driver related to exceptions/failures, so just wanted to confirm.

I realize this pattern already exists in other integ tests that I approved. If this makes the test fail, as expected, then this pattern is fine IMO and `fail()` has no advantage.

packages/amplify_auth_cognito/example/integration_test/update_password_test.dart
```diff
@@ -0,0 +1,107 @@
...
+
+      await signOutUser();
+      // sign in with current password
+      await Amplify.Auth.signIn(
+        username: username,
+        password: password,
+      );
+    });
+
+    testWidgets('should update a users password', (WidgetTester tester) async {
```
ragingsquirrel3, line 55: uber nit: `update a user\'s password'`

ragingsquirrel3 (approved):
Passed locally for me. Thanks for adding all this coverage! I think the auth category is getting pretty decent integ cov thanks to all your efforts. Also, great job on the utils and general code hygiene here.

I had a small question about the the throw/fail thing. Assuming that works as expected, just a few nits.

packages/amplify_auth_cognito/example/integration_test/utils/validation_utils.dart
Jordan-Nelson: Yeah I was being lazy when writing `isValidAWSCredentials` and didn't want to do all the extra null checks 🙂  (everything on AWSCredentials is nullable). That wasn't very clear what I was doing with the try/catch though. Updated.

packages/amplify_auth_cognito/example/integration_test/get_current_user_test.dart
Jordan-Nelson: It will fail the tests, but `fail()` is a little clearer to understand the intention of, which is probably the reason it exists. Updated 👍

packages/amplify_auth_cognito/example/integration_test/hub_events_test.dart
```diff
@@ -0,0 +1,84 @@
...
+            'email': generateEmail(),
+            'phone_number': mockPhoneNumber
+          }));
+
+      await signOutUser();
+    });
+
+    testWidgets(
+      'should broadcast events for sign in and sign out',
+      (WidgetTester tester) async {
```
dnys1, line 47: Check out [stream matchers](https://pub.dev/packages/test#stream-matchers). This test is correct, and, for more complex tests, it might be easier to structure it with those.

dnys1 (approved):
Looks good. Tests passed on both iOS and Android for me. Only minor comments.

packages/amplify_auth_cognito/example/integration_test/sign_in_sign_out_test.dart
```diff
@@ -48,18 +51,84 @@ void main() {
...
-      if (!initalAuthRes.isSignedIn) {
+    testWidgets(
+        'should throw a NotAuthorizedException with an incorrect password',
+        (WidgetTester tester) async {
+      final incorrectPassword = generatePassword();
+      try {
+        await Amplify.Auth.signIn(
+            username: username, password: incorrectPassword);
+      } catch (e) {
+        expect(e, TypeMatcher<NotAuthorizedException>());
```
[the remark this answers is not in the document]
Jordan-Nelson: Ah I did not know about `expectLater`. I didn't know how to do this with async functions. Thanks for that. I think I am going to leave it in this PR, but will add a note to open a follow up when I get a chance. There are a handful of tests that I added in previous PRs that use the same try/catch pattern that could be cleaned up too.

Issue #722, opened 2021-07-15
Getting 'Error: Invalid argument' when querying data with predicate containing enum.


EmmanuelOgbewe:
**Describe the bug**
I have a schema that contains an enum. When querying this data and providing an enum as a predicate I receive an error saying Invalid argument. 

**Expected behavior**
The enum provided as a predicate matches the enum type used in the schema. I do not expect to get the error.

**Screenshots**
ActivityNotification model contains enum NotificationType.

![Screen Shot 2021-07-15 at 1 21 35 AM](https://user-images.githubusercontent.com/29666179/125733346-9d27a326-3a1e-4b85-82f7-ca447b257f9f.png)

Querying for recipient notifications matching the type NotificationType.FriendRequest
![Screen Shot 2021-07-15 at 1 20 42 AM](https://user-images.githubusercontent.com/29666179/125733254-e54050c4-4d5e-443f-a5e5-dd35dd9c6644.png)

Error received is `Invalid argument: Instance of 'NotificationType'`

**Platform**
Amplify Flutter current supports iOS and Android.  This issue is reproducible in (check all that apply):
[] Android
[X] iOS

<details>

  <summary>Output of <code>flutter doctor -v</code></summary>
  
  ```[✓] Flutter (Channel stable, 2.0.6, on Mac OS X 10.15.4 19E266
    darwin-x64, locale en-US)
    • Flutter version 2.0.6 at /Users/emmanuelogbewe/flutter
    • Framework revision 1d9032c7e1 (3 months ago), 2021-04-29
      17:37:58 -0700
    • Engine revision 05e680e202
    • Dart version 2.12.3

[✓] Android toolchain - develop for Android devices (Android
    SDK version 29.0.3)
    • Android SDK at /Users/emmanuelogbewe/Library/Android/sdk
    • Platform android-29, build-tools 29.0.3
    • Java binary at: /Applications/Android
      Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build
      1.8.0_202-release-1483-b49-5587405)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 12.0.1, Build version 12A7300
    • CocoaPods version 1.10.0

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google
      Chrome.app/Contents/MacOS/Google Chrome

[...]

PR #726, opened 2021-07-16
Fix/enum in predicate

HuiSF:
*Issue #, if available:* #722 

*Description of changes:*

Originally enum type value was not correctly serialized while serializing a predicate for query operation. Changes for the fix:

* Add `isEnum` util function
* Check if a value is enum type and try to serialize if it's true

This PR converts 3 files from CRLF to LF mode to align with other files in the codebase. Please check the [second commit](https://github.com/aws-amplify/amplify-flutter/commit/56346eba1bd50711df715e22e3521b670edc4a55) to see the actual changes. Thanks!

----


packages/amplify_datastore_plugin_interface/lib/src/types/utils/parsers.dart
```diff
@@ -17,10 +17,22 @@
...
+  }
+
+  return describeEnum(obj);
+}
 
 // only to be used internally by amplify-flutter library
 T? enumFromString<T>(String? key, List<T> values) =>
     values.firstWhereOrNull((v) => key == enumToString(v));
+
+/// check wehather an dynamic type [value] is an enum
```
cshfang, line 34: Nit:
```suggestion
/// check whether a dynamic type [value] is an enum
```

cshfang (approved):

packages/amplify_datastore_plugin_interface/lib/src/types/utils/parsers.dart
```diff
@@ -1,26 +1,38 @@
...
+  }
+
+  return describeEnum(obj);
+}
+
+// only to be used internally by amplify-flutter library
+T? enumFromString<T>(String? key, List<T> values) =>
+    values.firstWhereOrNull((v) => key == enumToString(v));
+
+/// check whether a dynamic type [value] is an enum
```
ragingsquirrel3, line 34: nit/question: should this comment say "type [data]" or should the variable below be renamed "value" in line 35, or am I missing something?

ragingsquirrel3 (approved):
🚢  with small question about a comment/variable name.

packages/amplify_datastore_plugin_interface/lib/src/types/utils/parsers.dart
HuiSF: Oops mean to rename to `value`, will update

PR #728, opened 2021-07-19
chore(amplify_datastore): make SubscriptionEvent directly available via datastore plugin by adding to publicTypes in interface

ragingsquirrel3:
*Problem statement:* On discord, someone was having trouble resolving type `SubscriptionEvent`. Turns out they had to import datastore interface into their file in addition to datastore, which they found confusing. I thought it would make sense to add this class ~~as well as some other similar things~~ into `publicTypes` which are exported in datastore plugin itself (as well as datastore interface). That way, customers will not have to import the interface for stuff like this, but it should continue resolving for anyone who already imported the interface.

*Description of changes:* Moves ~~some exports~~ `SubscriptionEvent` from datastore_plugin_interface to `publicTypes` file which is exported both for datastore_plugin_interface and datastore.


packages/amplify_datastore_plugin_interface/lib/src/publicTypes.dart
```diff
@@ -1,8 +1,20 @@
...
 export 'types/exception/DataStoreExceptionMessages.dart';
+export 'types/models/auth_rule.dart';
+export 'types/models/model.dart';
+export 'types/models/model_field.dart';
+export 'types/models/model_field_definition.dart';
+export 'types/models/model_field_type.dart';
+export 'types/models/model_provider.dart';
+export 'types/models/model_schema.dart';
+export 'types/models/model_schema_definition.dart';
+export 'types/models/subscription_event.dart';
```
ragingsquirrel3, line 11: callout: `SubscriptionEvent` was the type that started this conversation, but when I saw these I thought to publicly export all. LMK if makes sense to limit this list a little more.

cshfang (commented):
A lot of these definitely make sense imo. However, could we just double check that there's value in exporting some of the more utility things like uuid or parsers? I.e. we're using them outside of the DS library itself?

packages/amplify_datastore_plugin_interface/lib/src/publicTypes.dart
```diff
@@ -1,8 +1,20 @@
...
+export 'types/models/uuid.dart';
+export 'types/query/query_field.dart';
 export 'types/sync/DataStoreSyncExpression.dart';
+export 'types/temporal/datetime_parse.dart';
 export 'types/temporal/temporal_date.dart';
 export 'types/temporal/temporal_time.dart';
 export 'types/temporal/temporal_datetime.dart';
 export 'types/temporal/temporal_timestamp.dart';
-export 'types/query/query_field.dart';
+export 'types/utils/parsers.dart';
```
HuiSF, line 20: I was looking into this file working on something else, there is one util function used only by internal functions. Not sure if should export it at all...

ragingsquirrel3: Ok, will remove this one for sure. The more I look at this look, I'm not sure if anything besides `SubscriptionEvent` really needs to be added, but will try to cull this list a little.

packages/amplify_datastore_plugin_interface/lib/src/publicTypes.dart
ragingsquirrel3: ok, now just limiting to this one.

ragingsquirrel3:
> A lot of these definitely make sense imo. However, could we just double check that there's value in exporting some of the more utility things like uuid or parsers? I.e. we're using them outside of the DS library itself?

After a 2nd look, I can't think of anything besides `SubscriptionEvent` that needs to be added like this, so just limiting it to that for now. Customer got confused from this snippet https://docs.amplify.aws/lib/datastore/real-time/q/platform/flutter so seems like this type might be worst offender. Same issue exists for example app and is solved with only adding `SubscriptionEvent`.

HuiSF (approved):
Thanks for the improvement. I think it's worth to review the other changes you suggested later to reduce confusion around the interface.

cshfang (approved):

Issue #308, opened 2021-01-15
Amplify API: Model based APIs and OIDC support


fjnoyp:
The current version of Amplify API has some features that will be supported in the future: 

1) OIDC support : https://docs.amplify.aws/lib/restapi/authz/q/platform/android#cognito-user-pool-authorization

2) GraphQL Request Helpers that automatically generate GraphQL requests for you.  Instead, you need to create the GraphQL request document yourself when calling the API. 
(ie. only Advanced Workflows is supported: https://docs.amplify.aws/lib/graphqlapi/advanced-workflows/q/platform/android) 

3) GraphQL Subscriptions

rlee1990:
Is there a timeline on GraphQL Request Helpers?

JuanDelgadillo:
Hey heyy guys! Does someone know alternative ways of automatically generating GraphQL Request Helpers?

I am thinking of automatically generating them in JavaScript and then translating them to Dart... but I am not sure if this will work, I'll test it out and comment back how it went.

If you know other more efficient ways of doing this, please let us know.

Thanks in advance.

Cheers!

cto-leaps:
Hey guys,

The doc at this url:
shows a way to use different auth modes for the same GraphQL Endpoint, which is exactly what I am looking for.
```
{
    "UserAgent": "aws-amplify-cli/2.0",
    "Version": "1.0",
    "api": {
        "plugins": {
            "awsAPIPlugin": {
                "[FRIENDLY-NAME-API-WITH-IAM": {
                    "endpointType": "GraphQL",
                    "endpoint": "[GRAPHQL-ENDPOINT]",
                    "region": "[REGION]",
                    "authorizationType": "AWS_IAM",
                },
                "[FRIENDLY-NAME-API-WITH-USER-POOLS]": {
                    "endpointType": "GraphQL",
                    "endpoint": "https://xyz.appsync-api.us-west-2.amazonaws.com/graphql",
                    "region": "[REGION]",
                    "authorizationType": "AMAZON_COGNITO_USER_POOLS",
                },
            }
        }
    }
}
```

One model in my scheme is set as follows:

```
type Exercise
  @model(subscriptions: null)
  @searchable
  @auth(rules: [
    {allow: groups, groups: ["Athletes"], operations: [read]}
    {allow: groups, groups: ["Administrators"], operations: [create, read, update, delete]}
    {allow: public, provider: iam, operations: [read] }
  ])
[...]

PR #729, opened 2021-07-19
chore(amplify_api): GraphQL helper move dependencies

Equartey:
*Issue #, if available:*

*Description of changes:*
Temporarily added the DataStore as a dependency to both `amplify_api` and `amplify_api_plugin_interface`. This PR specifically exposes the Model and associated classes from `amplify_datastore_plugin_interface/types` to locations we need it to start work on implementing the GraphQL model helpers.


packages/amplify_api_plugin_interface/lib/src/types.dart
```diff
@@ -13,6 +13,7 @@
  * permissions and limitations under the License.
  */
 
+export 'GraphQL/GraphQLHelpers.dart';
```
ragingsquirrel3, line 16: Does this file exist yet? If not, could this export be delayed (removed from this PR and re-added when the file exists) so all these exports resolve?

packages/amplify_api_plugin_interface/lib/src/types.dart
```diff
@@ -24,6 +25,18 @@ export 'RestAPI/RestOptions.dart';
 export 'RestAPI/RestResponse.dart';
 export 'RestAPI/RestException.dart';
 
+export 'types/models/model.dart';
```
ragingsquirrel3, line 28: I would suggest adding a prominent TODO comment for all the places where we are temporarily duplicating dependencies in API so it's super obvious in the code and no chance to miss if someone were to try to merge back to main. I would suggest something like:

```TODO: Datastore dependencies temporarily duplicated in API. Eventually they should be moved to core or otherwise reconciled to avoid duplication.```

I would add this comment here, as well as in every file this applies to. One benefit of this is that we can later search the code for this comment to fix it and reduce the chances of missing a file.

File: packages/amplify_api_plugin_interface/lib/src/types.dart [hunk 3]
```diff
@@ -24,6 +25,18 @@ export 'RestAPI/RestOptions.dart';
...
+export 'types/models/model_field.dart';
+export 'types/models/model_field_definition.dart';
+export 'types/models/model_field_type.dart';
+export 'types/models/model_provider.dart';
+export 'types/models/model_schema.dart';
+export 'types/models/model_schema_definition.dart';
+export 'types/models/subscription_event.dart';
+export 'types/query/query_field.dart';
+
+
```
ragingsquirrel3, line 39: nit: extra space here

packages/amplify_api_plugin_interface/lib/src/types/models/auth_rule.dart
```diff
@@ -0,0 +1,123 @@
...
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
```
ragingsquirrel3, line 15: I would suggest adding the same TODO comment here, or at top of file for all these files.

ragingsquirrel3 (changes requested):
Why are all the tests failing? If possible, should address that. If this branch has some period of time in the beginning where things can not be reconciled and tests failed, that might be ok. However, if possible we should try to keep it passing. Maybe bc `GraphQLHelpers` not here yet? 

Otherwise, just suggested adding a comment to these files.

packages/amplify_api_plugin_interface/lib/src/types.dart
Equartey: Good call out, I've added it to all files.

packages/amplify_api_plugin_interface/lib/src/types.dart
Equartey: Not yet, this was an artifact of cherry picking commits.

packages/amplify_api_plugin_interface/lib/src/types.dart
```diff
@@ -24,6 +24,18 @@ export 'RestAPI/RestOptions.dart';
 export 'RestAPI/RestResponse.dart';
 export 'RestAPI/RestException.dart';
 
+// TODO: Datastore dependencies temporarily duplicated in API. Eventually they should be moved to core or otherwise reconciled to avoid duplication.
```
ragingsquirrel3, line 27: question: Can commented code be deleted?

Equartey:
Placing this comment here to call out that this PR has been refactored to a less invasive approach. While still not permanent, the changes here will allow us to get developing the feature faster with less repercussions down stream. 

This PR now also adds interfaces and adjusts constructors that will be used to implement this feature. Feed back for this will extra helpful.

@ragingsquirrel3 @dnys1 Ready for review

packages/amplify_api_plugin_interface/pubspec.yaml
```diff
@@ -14,6 +14,9 @@ dependencies:
   meta: ^1.3.0
   uuid: ^3.0.1
 
+  # TODO: Datastore dependencies temporarily added in API. Eventually they should be moved to core or otherwise reconciled to avoid duplication.
```
ragingsquirrel3, line 17: I would think this is the only place this needs to be added as it's a dependency for API plugin, which is a dep for the example app. Could you try removing the other additions to pubspec to see if it works as expected?

packages/amplify_api/example/pubspec.yaml
```diff
@@ -30,6 +30,9 @@ dependencies:
   # Use with the CupertinoIcons class for iOS style icons.
   cupertino_icons: ^0.1.3
 
+  # TODO: Datastore dependencies temporarily added in API. Eventually they should be moved to core or otherwise reconciled to avoid duplication.
```
ragingsquirrel3, line 33: question: Does this need to be in example app if you are adding it to api plugin interface? See below comments.

packages/amplify_api/pubspec.yaml
```diff
@@ -14,6 +14,9 @@ dependencies:
   amplify_api_plugin_interface: 0.2.0
   plugin_platform_interface: ^2.0.0
 
+  # TODO: Datastore dependencies temporarily added in API. Eventually they should be moved to core or otherwise reconciled to avoid duplication.
```
ragingsquirrel3, line 17: see below comment in api plugin interface pubspec

packages/amplify_api_plugin_interface/lib/amplify_api_graphql_helpers_interface.dart
```diff
@@ -0,0 +1,62 @@
...
+
+import 'src/types.dart';
+import 'src/GraphQL/GraphQLRequest.dart';
+import 'src/types/pagination/model_pagination.dart';
+import 'src/types/pagination/paginated_result.dart';
+
+abstract class ModelQueriesInterface {
+  // Get
+  static GraphQLRequest<T> get<T extends Model>(
+      ModelType modelType, String id) {
```
ragingsquirrel3, line 27: Question: Can this be changed to be `ModelType<T> modelType, ...` to indicate the model type is for the same model used as generic for request?

packages/amplify_api_plugin_interface/lib/amplify_api_graphql_helpers_interface.dart
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
ragingsquirrel3, line 2: uber nit: might as well update the year from 2020 to 2021 in these comments for all these files.

packages/amplify_api_plugin_interface/pubspec.yaml
Equartey: I tried removing it as a dependency from the locations you specified, but imports are unable to resolve to amplify_api_plugin_interface. My hunch is that's due to only importing the datastore interface & not exporting the types too. That being said I think this is better because it doesn't allow the types to propagate down and cause potential issues like I saw with previous approaches. Additionally Datastore is an optional dependency.

packages/amplify_api/example/pubspec.yaml
Equartey: Yes, see comment above.

packages/amplify_api_plugin_interface/lib/amplify_api_graphql_helpers_interface.dart
Equartey: Yeah it can, good idea.

packages/amplify_api_plugin_interface/lib/amplify_api_graphql_helpers_interface.dart
Equartey: No worries, updated.

packages/amplify_api_plugin_interface/pubspec.yaml
ragingsquirrel3: Cool, thanks for looking into this. It seems weird, but I guess as a temporary solution it's fine and we can sort this out when we eventually resolve this in a shared place.

ragingsquirrel3 (approved):
🚢  it, good start here! There might be a few details that need adjusting in the future. I still have some questions about the signature for delete, but I think these things can be adjusted as the implementation progresses, especially when we have some unit tests.

packages/amplify_api/lib/graphql_helpers.dart
```diff
@@ -0,0 +1,25 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import 'package:amplify_api_plugin_interface/amplify_api_plugin_interface.dart';
+import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart';
+
+class ModelQueries extends ModelQueriesInterface {
+  @override
+  static GraphQLRequest<T> get<T extends Model>(
```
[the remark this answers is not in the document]
dnys1: Although, now I'm wondering if `ModelQueries` even needs an interface. Do you think we could put all the static methods on `ModelQueries` and remove the interface class?

packages/amplify_api_plugin_interface/lib/amplify_api_graphql_helpers_interface.dart
```diff
@@ -0,0 +1,62 @@
...
+      ModelType<T> modelType,
+      {QueryPredicate? where,
+      ModelPagination? modelPagination}) {
+    throw UnimplementedError("list() has not been implemented.");
+  }
+}
+
+abstract class ModelMutationsInterface {
+  // Create
+  static GraphQLRequest<T> create<T extends Model>(Model model) {
```
dnys1, line 42: These should all take in `T model` as an argument, instead of `Model model`.

packages/amplify_api/lib/graphql_helpers.dart
```diff
@@ -0,0 +1,25 @@
...
+      ModelType modelType, String id) {
```
dnys1, line 22: `ModelType<T>`

packages/amplify_api_plugin_interface/lib/src/types/pagination/model_pagination.dart
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+class ModelPagination {
+  // limit is used for the pagination page limit
+  late int limit;
```
dnys1, line 18: Should be `final` with a constructor.

packages/amplify_api_plugin_interface/lib/src/types/pagination/paginated_result.dart
```diff
@@ -0,0 +1,37 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+// TODO: Datastore dependencies temporarily added in API. Eventually they should be moved to core or otherwise reconciled to avoid duplication.
+import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart';
+
+class PaginatedResult<T extends Model> {
+  @override
+  String getId() {
```
dnys1, line 21: Prefer getters over `getX()` methods, for example:

```dart
String get id;
```

instead of

```dart
String getId();
```

dnys1: These methods aren't overriding anything, so we can remove `@override`

dnys1: Actually, this can just be an abstract class since all it is doing is setting up an interface for subclasses to override.

```dart
abstract class PaginatedResult<T extends Model> {
  String get id;

  ModelType<T> get instanceType;

  Map<String, dynamic> toJson();
}
```

packages/amplify_api_plugin_interface/lib/src/types/pagination/paginated_result.dart
```diff
@@ -0,0 +1,37 @@
...
+    // TODO: implement getId
+    throw UnimplementedError("getId() has not been implemented.");
+  }
+
+  @override
+  ModelType<Model> getInstanceType() {
```
dnys1, line 27: `ModelType<T>`

packages/amplify_api_plugin_interface/lib/amplify_api_graphql_helpers_interface.dart
```diff
@@ -0,0 +1,62 @@
...
+  // List
+  static GraphQLRequest<PaginatedResult<T>> list<T extends Model>(
+      ModelType<T> modelType,
+      {QueryPredicate? where,
+      ModelPagination? modelPagination}) {
+    throw UnimplementedError("list() has not been implemented.");
+  }
+}
+
+abstract class ModelMutationsInterface {
```
dnys1, line 40: Same comments regarding interfaces and static types as with ModelQueries.

dnys1 (changes requested):
Good work! I think these PR reviews are great opportunities to discuss the nitty gritty, so please don't be discouraged by my comments. Dart is a unique language with fairly strict ways of doing things, and I wouldn't expect you to have known these on the first go.

Seriously, great work - I'm excited to see the rest 😃

packages/amplify_api_plugin_interface/lib/src/types/pagination/paginated_result.dart
Equartey: Great feedback, totally missed this not being an abstract. Fixed 👍

packages/amplify_api_plugin_interface/lib/src/types/pagination/model_pagination.dart
Equartey: Good callout, added final and constructor.

packages/amplify_api/lib/graphql_helpers.dart
Equartey: Fixed.

packages/amplify_api_plugin_interface/lib/amplify_api_graphql_helpers_interface.dart
Equartey: Fixed.

packages/amplify_api/lib/graphql_helpers.dart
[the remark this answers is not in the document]
Equartey: Thanks for the insight, I didn't realize dart operated like this. I want to keep the interface files because they feel like better code design, but to enable that I've converted this class to `ModelQueriesFactory` and added a `ModelQueries` class that provides static calls to the factory. Thanks for the feedback!

packages/amplify_api_plugin_interface/lib/amplify_api_graphql_helpers_interface.dart
```diff
@@ -0,0 +1,59 @@
...
+
+import 'src/types.dart';
+import 'src/GraphQL/GraphQLRequest.dart';
+import 'src/types/pagination/model_pagination.dart';
+import 'src/types/pagination/paginated_result.dart';
+
+abstract class ModelQueriesInterface {
+  // Get
+  GraphQLRequest<T> get<T extends Model>(ModelType<T> modelType, String id) {
+    throw UnimplementedError("get() has not been implemented.");
```
dnys1, line 27: For abstract classes, we don't want to provide a default impl which throws because subclasses which extend this will inherit that default behavior without warning. Leaving it as just the function signature, like below, means that subclasses will be forced to provide an implementation for all methods.

```dart
GraphQLRequest<T> get<T extends Model>(ModelType<T> modelType, String id);
```

packages/amplify_api/lib/model_queries.dart
```diff
@@ -0,0 +1,26 @@
...
+
+import 'package:amplify_api/amplify_api.dart';
+import 'package:amplify_api/graphql_helpers.dart';
+import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart';
+
+// This class provides static method calls to enable a simpler DX
+class ModelQueries {
+  static GraphQLRequest<T> get<T extends Model>(
+      ModelType<T> modelType, String id) {
+    return ModelQueriesFactory().get<T>(modelType, id);
```
dnys1, line 24: This is totally fine for now- we can work on how to improve this in the next go-around, since we don't want to be creating a new instance in every method or every time this method is called.

dnys1 (approved):
Looks good!

packages/amplify_api_plugin_interface/lib/amplify_api_graphql_helpers_interface.dart
Equartey: That makes sense. We're doing the same thing in the [plugin interface](https://github.com/aws-amplify/amplify-flutter/blob/c0ccdfdb535bf9679d442c9cb9ddccaf9149df4d/packages/amplify_api_plugin_interface/lib/amplify_api_plugin_interface.dart#L29). Does consistency between the two matter?

packages/amplify_api/lib/model_queries.dart
Equartey: Great, I'll be addressing this in my next PR!

packages/amplify_api_plugin_interface/lib/amplify_api_graphql_helpers_interface.dart
Equartey: Discussed this offline, we'll be removing the exception from these interfaces to avoid potential confusion when implementing subclasses. The plugin interface needs the extra context as a plugin could be used on an unsupported platform and the exception provides missing context.

PR #734, opened 2021-07-21
chore(amplify_storage_s3): add storage integration tests and update example app

ragingsquirrel3:
Adds integration tests for storage category. Unfortunately, I was unable to add a provision script here because it is not supported by the headless CLI, yet https://github.com/aws-amplify/amplify-cli/issues/7378 even though they have defined a schema for it https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-headless-interface/src/interface/storage/add.ts . Instead, I just documented how to provision the backend manually. The headless CLI does not support everything, so as we add coverage we may need to do more stuff like this.

*Description of changes:*
* some updates to storage example app to make it run with sound null safety
* an integration test for storage, should be complete: upload, list, delete, getUrl
* some documentation updates


packages/amplify_storage_s3/example/README.md
```diff
@@ -1,3 +1,54 @@
 # amplify_storage_s3_example
 
 Example app for the Amplify Flutter Storage plugin with AWS S3 provider.
+
```
ragingsquirrel3, line 4: view this on https://github.com/ragingsquirrel3/amplify-flutter/blob/chore-storage-integ-tests/packages/amplify_storage_s3/example/README.md to make it slightly easier to read.

packages/amplify_storage_s3/example/android/build.gradle
```diff
@@ -6,7 +6,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.5.0'
+        classpath 'com.android.tools.build:gradle:3.5.4'
```
ragingsquirrel3, line 9: Is it ok to just make this change here, or should we try to have the same version in all these files? IMO, ok to have this patch here as it's an example app, after all. This is needed here because of something related to new version of file_picker which has sound nulls safety https://github.com/miguelpruivo/flutter_file_picker/wiki/Troubleshooting#-issue.

packages/amplify_storage_s3/example/integration_test/main_test.dart
```diff
@@ -0,0 +1,121 @@
...
+    testWidgets('should list files and get expected attributes',
+        (WidgetTester tester) async {
+      if (lastUploadedKey == null) {
+        fail('No uploaded file to verify.');
+      }
+      final result = await Amplify.Storage.list(options: listOptions);
+      expect(result.items.length, greaterThan(0));
+      final uploadedStorageItem =
+          result.items.firstWhere((element) => element.key == lastUploadedKey);
+      expect(uploadedStorageItem.key, lastUploadedKey);
```
dnys1, line 97: This will always be true, by definition

packages/amplify_storage_s3/example/integration_test/main_test.dart
```diff
@@ -0,0 +1,121 @@
...
+      expect(uploadedStorageItem.lastModified?.day,
+          new DateTime.now().day); // was uploaded today
+      expect(uploadedStorageItem.eTag?.isNotEmpty, true);
+      expect(uploadedStorageItem.size, greaterThan(0));
```
dnys1, line 101: Would it be good to check the metadata items here as well? Like the `name` and `desc` fields set when uploading.

packages/amplify_storage_s3/example/integration_test/main_test.dart
```diff
@@ -0,0 +1,121 @@
...
+    });
+
+    testWidgets('should get the URL of an uploaded file',
+        (WidgetTester tester) async {
+      if (lastUploadedKey == null) {
+        fail('No uploaded file to verify.');
+      }
+      final result = await Amplify.Storage.getUrl(key: lastUploadedKey!);
+      expect(result.url.isNotEmpty, true);
```
dnys1, line 110: Maybe this is a bit nitpicky, but can this ever be empty? It seems like this test might just be checking to make sure that getUrl succeeds and does not throw, in which case there are some methods to express exactly that, which may help with how the test reads.

```dart
await expectLater(Amplify.Storage.getUrl(key: lastUploadedKey!), completes);
```

or, less wordy:
```dart
final getUrl = Amplify.Storage.getUrl(key: lastUploadedKey!);
await expectLater(getUrl, completes);
```

dnys1 (approved):
Looks good! Few minor things

packages/amplify_storage_s3/example/integration_test/main_test.dart
ragingsquirrel3: Yes, would be helpful, will add.

packages/amplify_storage_s3/example/integration_test/main_test.dart
ragingsquirrel3: Yeah I don't think assertion is quite right. I wanted to do **some** validation on the URL result here, though emptiness probably not good. Initially I tried to validate the key appears in the URL, but then I found it gets URI escaped only on iOS (seems like an issue?) in this case, bc key has special characters. So, assertion was not super straightforward. I guess I could try:
a) using a key with no special characters or otherwise asserting with URI escaping
b) validate another way, thought about looking for 's3'  or just asserting it's a valid URL
c) do what you suggested and just assert completion

I'm kinda leaning towards b now. If that's too cumbersome, might just do c like you suggested.

dnys1: Gotcha, I understand now. We could go all the way and perform a GET request with the returned url and compare the contents, if you want.

packages/amplify_storage_s3/example/integration_test/main_test.dart
ragingsquirrel3: Discussed offline, but the metadata is not available on StorageItem so not possible do this at the moment.

packages/amplify_storage_s3/example/integration_test/main_test.dart
ragingsquirrel3: updated assertion a la option b

Jordan-Nelson (approved):
These passed locally for me on iOS and Android. The tests look good 👍

PR #735, opened 2021-07-21
feat(amplify_auth_cognito): Auth Devices API

dnys1:
*Issue #, if available:*

*Description of changes:*

The following Devices API:
- `Amplify.Auth.rememberDevice`
- `Amplify.Auth.forgetDevice`
- `Amplify.Auth.fetchDevices`

This PR uses the escape hatch on Android to bring parity between the platforms. The long-term strategy is to change amplify-android for all users, so this marked as DRAFT in the meantime, although the changes needed will be very minor and the PR should be reviewed as-is.

### DIFF
--- a/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
+++ b/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
@@ -22,6 +22,7 @@ import android.os.Handler
 import android.os.Looper
 import androidx.annotation.NonNull
 import androidx.annotation.VisibleForTesting
+import com.amazonaws.amplify.amplify_auth_cognito.device.DeviceHandler
 import com.amazonaws.amplify.amplify_auth_cognito.types.FlutterSignUpResult
 import com.amazonaws.amplify.amplify_auth_cognito.types.FlutterSignInResult
 import com.amazonaws.amplify.amplify_auth_cognito.types.FlutterFetchCognitoAuthSessionResult
@@ -90,6 +91,11 @@ public class AuthCognito : FlutterPlugin, ActivityAware, MethodCallHandler, Plug
   var eventMessenger: BinaryMessenger? = null
   private lateinit var activityBinding: ActivityPluginBinding
 
+  /**
+   * Handles the Devices API.
+   */
+  private val deviceHandler: DeviceHandler = DeviceHandler(errorHandler)
+
   constructor() {
       authCognitoHubEventStreamHandler = AuthCognitoHubEventStreamHandler()
   }
@@ -102,7 +108,7 @@ public class AuthCognito : FlutterPlugin, ActivityAware, MethodCallHandler, Plug
 
   override fun onAttachedToEngine(@NonNull flutterPluginBinding: FlutterPlugin.FlutterPluginBinding) {
     channel = MethodChannel(flutterPluginBinding.getFlutterEngine().getDartExecutor(), "com.amazonaws.amplify/auth_cognito")
-    channel.setMethodCallHandler(this);
+    channel.setMethodCallHandler(this)
     context = flutterPluginBinding.applicationContext;
     eventMessenger = flutterPluginBinding.getBinaryMessenger();
     hubEventChannel = EventChannel(flutterPluginBinding.binaryMessenger,
@@ -153,6 +159,11 @@ public class AuthCognito : FlutterPlugin, ActivityAware, MethodCallHandler, Plug
       } catch (e: Exception) {
         handleAddPluginException("Auth", e, result)
       }
+      return
+    }
+
+    if (DeviceHandler.canHandle(call.method)) {
+      deviceHandler.onMethodCall(call, result)
       return
     }
 
--- a/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/base/AtomicResult.kt
+++ b/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/base/AtomicResult.kt
@@ -1,0 +1,47 @@
+package com.amazonaws.amplify.amplify_auth_cognito.base
+
+import io.flutter.plugin.common.MethodChannel
+import kotlinx.coroutines.*
+
+/**
+ * Thread-safe [MethodChannel.Result] wrapper which prevents multiple replies and automatically posts
+ * results to the main thread.
+ */
+class AtomicResult(private val result: MethodChannel.Result) : MethodChannel.Result {
+    private companion object {
+        /**
+         * Scope for performing result handling.
+         * Method channel results must be sent on the main (UI) thread.
+         */
+        val scope = MainScope()
+    }
+
+    /**
+     * Whether a response has been sent.
+     */
+    private var isSent = false
+
+    override fun success(value: Any?) {
+        scope.launch {
+            if (isSent) return@launch
+            result.success(value)
+            isSent = true
+        }
+    }
+
+    override fun error(errorCode: String?, errorMessage: String?, errorDetails: Any?) {
+        scope.launch {
+            if (isSent) return@launch
+            result.error(errorCode, errorMessage, errorDetails)
+            isSent = true
+        }
+    }
+
+    override fun notImplemented() {
+        scope.launch {
+            if (isSent) return@launch
+            result.notImplemented()
+            isSent = true
+        }
+    }
+}
--- a/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/base/AtomicResult.kt
+++ b/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/base/AtomicResult.kt
@@ -1,3 +1,17 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
 package com.amazonaws.amplify.amplify_auth_cognito.base
 
 import io.flutter.plugin.common.MethodChannel
--- a/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/base/AtomicResult.kt
+++ b/packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/base/AtomicResult.kt
@@ -14,14 +14,17 @@
  */
 package com.amazonaws.amplify.amplify_auth_cognito.base
 
+import io.flutter.Log
 import io.flutter.plugin.common.MethodChannel
 import kotlinx.coroutines.*
+import java.util.concurrent.atomic.AtomicBoolean
 
 /**
  * Thread-safe [MethodChannel.Result] wrapper which prevents multiple replies and automatically posts
  * results to the main thread.
  */
-class AtomicResult(private val result: MethodChannel.Result) : MethodChannel.Result {
+class AtomicResult(private val result: MethodChannel.Result, private val operation: String) :
+    MethodChannel.Result {
     private companion object {
         /**
          * Scope for performing result handling.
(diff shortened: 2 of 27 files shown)


packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/device/DeviceHandler.kt
```diff
@@ -0,0 +1,102 @@
...
+        /**
+         * Whether this class can handle [method].
+         */
+        fun canHandle(method: String): Boolean = methods.contains(method)
+    }
+
+    /**
+     * Scope for running asynchronous tasks.
+     */
+    private val scope = CoroutineScope(Dispatchers.IO) + CoroutineName("DeviceHandler")
```
haverchuck, line 36: Coroutines became stable in Kotlin version 1.3. Do we know if it's possible whether any existing customers may be using an older version of Kotlin? Would our library otherwise work if they were using an older version?

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/base/AtomicResult.kt
```diff
@@ -0,0 +1,47 @@
+package com.amazonaws.amplify.amplify_auth_cognito.base
+
+import io.flutter.plugin.common.MethodChannel
+import kotlinx.coroutines.*
+
+/**
+ * Thread-safe [MethodChannel.Result] wrapper which prevents multiple replies and automatically posts
+ * results to the main thread.
+ */
+class AtomicResult(private val result: MethodChannel.Result) : MethodChannel.Result {
```
haverchuck, line 10: This is great idea.

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
```diff
@@ -156,6 +162,11 @@ public class AuthCognito : FlutterPlugin, ActivityAware, MethodCallHandler, Plug
       return
     }
 
+    if (DeviceHandler.canHandle(call.method)) {
```
haverchuck, line 165: Is there a device-api-specific reason to create this additional handler, other than the AtomicResult and CoroutineScope stuff you are doing? Or is this just general separation of concerns?

haverchuck (commented):
Left a few comments.  We will also need to add some license headers to the new files.

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
dnys1: General separation and smaller class size, but also the argument handling right below this doesn't allow for `null` to be passed, which I wanted to do.

I also thought it could be cool as a potential refactor option, since if all the handlers were separated out into their own classes, we could keep this class pretty slim:

```kotlin
private val handlers = listOf(DeviceHandler(), ...)

override fun onMethodCall(@NonNull call: MethodCall, @NonNull result: Result) {
  for (handler in handlers) {
    if (handler.canHandle(call.method)) {
      handler.handle(call, result)
      return
    }
  }
  
  result.notImplemented()
}
```

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/device/DeviceHandler.kt
dnys1: I believe that apps and plugins can use different versions of Kotlin. But this would only be a potential issue for projects generated before 1/24/19, according to the [git history](https://github.com/flutter/flutter/commits/stable/packages/flutter_tools/templates/app/android-kotlin.tmpl/build.gradle).

dnys1:
> Left a few comments. We will also need to add some license headers to the new files.

Good catch. Fixed!

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/base/AtomicResult.kt
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * Whether a response has been sent.
+     */
+    private var isSent = false
+
+    override fun success(value: Any?) {
+        scope.launch {
+            if (isSent) return@launch
```
Jordan-Nelson, line 40: This just silently ignores everything other than the first response, correct? I don't know that we want to silently ignore them. If we are getting multiple responses, in most cases it there is probably an issue with our code or an issue in the native library. I think we should at least log something.

packages/amplify_auth_cognito/lib/src/CognitoDevice/cognito_device.dart
```diff
@@ -0,0 +1,127 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
+import 'package:collection/collection.dart';
+
+/// Helper functions for [AuthDevice].
+extension AuthDeviceX on AuthDevice {
```
Jordan-Nelson, line 20: This is interesting. I have wanted to put some time into how we can improve the DX for some of the APIs that have Cognito specific implementations. I had not thought of using extensions to at least make casting it easier. `AuthSession`/`CognitoAuthSession` is another place this would be useful. Probably a bunch more.

I still think there might be a larger change we can make to improve the DX of the Cognito specific APIs but this is definitely an improvement 👍

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
dnys1: We can change the pattern - I'm not at all attached. For this PR, I needed to separate it out for my own sanity haha. I agree that the use of classes is not totally necessary, although I do like it for the namespacing aspect. But they could all be top-level extension methods on a big class which could be separated out into files, as needed.

I'll leave the null-check up to you - I didn't want to touch previous code as much as possible.

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/device/DeviceHandler.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    @Suppress("UNCHECKED_CAST")
+    override fun onMethodCall(call: MethodCall, _result: MethodChannel.Result) {
+        val result = AtomicResult(_result)
+        when (call.method) {
+            "fetchDevices" -> fetchDevices(result)
+            "rememberDevice" -> rememberDevice(result)
+            "forgetDevice" -> {
+                val deviceJson = (call.arguments as? Map<*, *> ?: emptyMap<String, Any?>()) as Map<String, Any?>
```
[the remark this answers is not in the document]
dnys1: It's kind of arbitrary, I suppose. I prefer the `forgetDevice` method to take in an AuthDevice and not a Map or an Any.

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/base/AtomicResult.kt
dnys1: Right now we don't have a good solution for logging, although I totally agree it would be good to monitor. The most important thing, though, is that it doesn't submit multiple replies which will crash the app. It's pretty rare that I've seen it and usually the result of poorly written multithreaded code, so I don't expect it from the Amplify SDK or our own code. It's more a piece-of-mind for now. When I've finished the logging SDK (next sprint), I will make sure to capture that information here.

packages/amplify_auth_cognito/lib/src/CognitoDevice/cognito_device.dart
dnys1: Word 👍

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/base/AtomicResult.kt
haverchuck: We actually had to do a few fixes on the storage plugin which did something very similar, just not as elegantly as this.

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
haverchuck: I would lean towards leaving the previous code untouched, as we probably just need to refactor that anyway. I like the idea of separate classes for different handlers that this is establishing, so maybe continue to have it handle null in it's own way.  

Regarding the general idea of distinct class for handlers, how would you envision this being applied to the other method calls? What is the criteria that makes these new methods get handled by a single handler, other than the fact that they are being implemented together? Is it the fact that they are part of a sort of common workflow? If that's the case, would it makes sense to implement this pattern in a refactor and group APIs such as `signUp`, `confirmSignUp` and `resendSignUpCode` together?

dnys1: Yeah, like Jordan was saying, the lines are not so clear. It could be like `SignUpHandler` for the methods you mention, then `SignInHandler`, `SignOutHandler`, etc. Some may only have one method. Or it's just all top-level methods/extension methods on this class. The only problem with that is that we still have a long when/case which is hard to reason about.

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/base/AtomicResult.kt
dnys1: Good to know - yeah multi-threading is hard. Having these as a fallback is definitely peace of mind for me because I've written plenty of deadlocks and race conditions before.

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
haverchuck: Ah ok - sorry I missed @Jordan-Nelson 's comment.

haverchuck (approved):

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/base/AtomicResult.kt
Jordan-Nelson: > Right now we don't have a good solution for logging

Can we at least log this locally for now?

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/device/DeviceHandler.kt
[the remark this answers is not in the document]
Jordan-Nelson: If `forgetDevice` wasn't already coupled to the method channel invocation, I would agree that passing an AuthDevice makes far more sense, but since `forgetDevice` takes an argument of `MethodChannel.Result`, it makes sense to me to pass in all the arguments from that method channel call (call.arguments). It is a `nit` comment though so this definitely isn't blocking.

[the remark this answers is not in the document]
dnys1: Ah, I see your point, now. Thanks for clarifying. I think this is a matter of preference - I can see the argument for both options now, though. I do still prefer it this way and the reason is that I tend to think of result as being an interface which can be mocked or extended whereas the device argument is an ADT and by replacing AuthDevice with Map or Any, we lose the type hierarchy and any type safety in the method signature.

Again, seems like a personal preference but just wanted to expand too on my rationale.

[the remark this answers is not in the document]
dnys1: Ideally, we would have multiple layers to this architecture - one dedicated to decoding and one dedicated to working with decoded types. I think that's what is creating the disparity for us both, since we are dancing in the middle here.

packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/base/AtomicResult.kt
dnys1: Added!

Jordan-Nelson (approved):

PR #742, opened 2021-07-22
feat(amplify_api): GraphQL helper get query

Equartey:
*Issue #, if available:*

**Note: this is a branch of #729, thus the git history shows changes out of scope for this PR.**

*Description of changes:*
This is a very rough stab at implementing a `get` helper. It currently only supports get by id queries. I'd like to get  feedback as I look to divide and refactor code. 

Current items I'd like to discuss:
- Method availability (public vs internal)
- Code reusability - how to approach
- Pure functions as an approach


packages/amplify_api/lib/graphql_helpers.dart
```diff
@@ -0,0 +1,119 @@
...
+  static ModelSchema getSchema(ModelType modelType) {
+    ModelProviderInterface? provider = AmplifyAPI.instance.getModelProvider();
+
+    if (provider == null) {
+      throw ApiException('No modelProvider found',
+          recoverySuggestion:
+              'Pass in a modelProvider instance while instantiating APIPlugin');
+    }
+
+    return provider.modelSchemas.firstWhere(
```
ragingsquirrel3, line 75: question: what happens if the desired model type not in the provider?

packages/amplify_api/lib/graphql_helpers.dart
```diff
@@ -0,0 +1,119 @@
...
+      }
+    }
+
+    String doc =
+        '''$reqTypeStr $reqOperationStr$name${varInputUpperStr} { $reqOperationStr$name$varInputLowerStr { ${fields.join(' ')} } }''';
+
+    // TODO: create input map for variables & connect with current variableInput
+    var variables = {"id": id};
+
+    print(doc);
```
ragingsquirrel3, line 116: nit: should eventually not have print calls here committed, though I am guessing you were debugging with this.

packages/amplify_api/example/lib/graphql_api_view.dart
```diff
@@ -76,8 +79,9 @@ class _GraphQLApiViewState extends State<GraphQLApiView> {
       }
     }''';
 
-    var operation = await Amplify.API
-        .query<String>(request: GraphQLRequest(document: graphQLDocument));
+    var operation = await Amplify.API.query<String>(
+        request: ModelQueries.get(
+            Blog.classType, "f70d1142-12da-4564-a699-966a75f96db6"));
```
ragingsquirrel3, line 84: Fetching by a specific ID will not work for other folks in the example app, may need to rethink how these helpers are integrated in the example app. I think it will be nice to have this in example app at some point as a showcase, but it might make sense to exclude example app changes from this PR to keep it more focused.

packages/amplify_api/lib/graphql_helpers.dart
```diff
@@ -0,0 +1,119 @@
...
+class ModelQueries extends ModelQueriesInterface {
+  @override
+  static GraphQLRequest<T> get<T extends Model>(
+      ModelType modelType, String id) {
+    ModelSchema schema = GraphQLRequestFactory.getSchema(modelType);
+
+    String modelName = schema.name;
+    Map<String, ModelField?>? fieldsMap = schema.fields;
+
+    List<String> fieldsList = [];
```
ragingsquirrel3, line 30: suggestion: I think this might be  place you can break this out into a smaller utility method, something like `List<String> getFieldsFromModelType(ModelType modelType) {}`. This will accomplish a few things:
* make this method easier to read bc this whole block will be reduced to `final fieldsList = getFieldsFromModelType...`, pretty clear what it does
* easier to use this method somewhere else, like maybe a mutation/subscription helper
* can test the utility

packages/amplify_api/lib/graphql_helpers.dart
```diff
@@ -0,0 +1,119 @@
...
+      required String id,
+      required GraphQLRequestType requestType,
+      required GraphQLRequestOperation requestOperation}) {
+    String reqTypeStr = describeEnum(requestType.toString());
+    String reqOperationStr = describeEnum(requestOperation.toString());
+
+    var varInputUpperStr = '';
+    var varInputLowerStr = '';
+
+    if (variableInput!.isNotEmpty) {
```
ragingsquirrel3, line 103: suggestion/question: Would it make sense to make a separate utility method for populating the variableInput string?

packages/amplify_api/lib/graphql_helpers.dart
Equartey: Good catch, I've added a null check. It will now throw an exception.

packages/amplify_api/example/lib/graphql_api_view.dart
Equartey: For sure, I think we can maintain the original spec once we have the `list` method implemented. I've been using this for testing and its should be ripped out before production. I'll drop a comment saying as much too.

packages/amplify_api/lib/graphql_helpers.dart
Equartey: Great points. I've started to use this functional approach throughout the helper as it makes sense.

packages/amplify_api/lib/graphql_helpers.dart
Equartey: I've gone back and forth on this, I do think it will get refactored once I flush out how I want to build the document inputs, ie variables, predicates, etc.

packages/amplify_api/lib/graphql/graphql_helpers.dart
```diff
@@ -0,0 +1,131 @@
...
+        return "ID";
+      default:
+        return "Error: UnknownType!";
+    }
+  }
+
+  List<String> _getFieldsFromModelType(Map<String, ModelField?>? fieldsMap) {
+    List<String> fieldsList = [];
+    if (fieldsMap != null) {
+      fieldsMap.forEach((key, value) {
```
ragingsquirrel3, line 60: nit: rather than simple loop/add, could this use map function similar to https://stackoverflow.com/questions/49941361/dart-mapping-a-list-list-map? In JS I think map is better for these situations bc easier to read, dk if dart result is quite as nice.

packages/amplify_api/lib/graphql/graphql_helpers.dart
```diff
@@ -0,0 +1,131 @@
...
+    String name = modelName ?? 'No model name found';
+    String fields = _getFieldsFromModelType(schema.fields).join(' ');
+
+    String reqTypeStr = describeEnum(requestType.toString());
+    String reqOperationStr = describeEnum(requestOperation.toString());
+
+    var varInputUpperStr = '';
+    var varInputLowerStr = '';
+
+    if (variableInput!.isNotEmpty) {
```
ragingsquirrel3, line 109: What happens here if `variableInput` is null? Won't it be an error? Can `variableInput` just be non-nullable?

packages/amplify_api/lib/graphql/graphql_helpers.dart
```diff
@@ -0,0 +1,131 @@
...
+      List<String> variableInputUpper = [];
+      List<String> variableInputLower = [];
+
+      for (var key in variableInput.keys) {
+        String? modelInputType = variableInput[key];
+        variableInputUpper.add("\$$key: $modelInputType!");
```
ragingsquirrel3, line 115: Do we want to add `modelInputType` if null?

packages/amplify_api/lib/graphql/graphql_helpers.dart
```diff
@@ -0,0 +1,131 @@
...
+
+    if (requestOperation == GraphQLRequestOperation.list) {
+      fields = 'items { $fields } nextToken';
+    }
+
+    String doc =
+        '''$reqTypeStr $reqOperationStr$name${varInputUpperStr} { $reqOperationStr$name$varInputLowerStr { $fields } }''';
+
+    // TODO: create input map for variables & connect with current variableInput
+    Map<String, dynamic> variables = id != null ? {"id": id} : const {};
```
ragingsquirrel3, line 135: `GraphQLRequest` already has `{}` defined as default param in constructor, so I would not suggest repeating that default here. Instead, could `variables` just be null when no id so default constructor works as is?

packages/amplify_api/lib/graphql/graphql_helpers.dart
```diff
@@ -0,0 +1,131 @@
...
+      required GraphQLRequestOperation requestOperation}) {
+    ModelSchema schema = _getSchema(modelType);
+
+    String? modelName = requestOperation == GraphQLRequestOperation.list
+        ? schema.pluralName
+        : schema.name;
+    String name = modelName ?? 'No model name found';
+    String fields = _getFieldsFromModelType(schema.fields).join(' ');
+
+    String reqTypeStr = describeEnum(requestType.toString());
```
ragingsquirrel3, line 103: It's tough to know what these variables are for, and naming things is hard, so might suggest a small comment like `// e.g. "query"` for both these string variables to give reader some context.

packages/amplify_api/lib/graphql/graphql_helpers.dart
```diff
@@ -0,0 +1,131 @@
...
+      required Map<String, String>? variableInput,
+      required String? id,
+      required GraphQLRequestType requestType,
+      required GraphQLRequestOperation requestOperation}) {
+    ModelSchema schema = _getSchema(modelType);
+
+    String? modelName = requestOperation == GraphQLRequestOperation.list
+        ? schema.pluralName
+        : schema.name;
+    String name = modelName ?? 'No model name found';
```
ragingsquirrel3, line 100: If there is no model name, can the query be valid? Would it make sense just to throw error in this case? Why assign this name here when null?

packages/amplify_api/example/lib/graphql_api_view.dart
```diff
@@ -13,10 +13,13 @@
  * permissions and limitations under the License.
  */
 
+import 'package:amplify_api/graphql/model_queries.dart';
```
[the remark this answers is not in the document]
Equartey: Good catch, it actually is unneeded here as it's already being exported.

packages/amplify_api/lib/graphql/graphql_helpers.dart
```diff
@@ -0,0 +1,131 @@
...
+      case ModelFieldTypeEnum.int:
+        return "int";
+      case ModelFieldTypeEnum.model:
+        return "ID";
+      default:
+        return "Error: UnknownType!";
+    }
+  }
+
+  List<String> _getFieldsFromModelType(Map<String, ModelField?>? fieldsMap) {
```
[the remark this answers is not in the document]
Equartey: For sure, that's much cleaner.

packages/amplify_api/lib/graphql/graphql_helpers.dart
Equartey: Updated the conditional so the logic makes more sense. `variableInput` is an optional parameter so we need to make sure its exists and has value.

packages/amplify_api/lib/graphql/graphql_helpers.dart
Equartey: It really shouldn't be nullable since coming from a Map. I've updated the loop to include the value.

packages/amplify_api/lib/graphql/graphql_helpers.dart
Equartey: Since variables is being passed in regardless of value on the next line, the constructor won't accept a `null` value. It has to be at least be an empty Map. I did remove the unneeded `const`.

packages/amplify_api/lib/graphql/graphql_helpers.dart
Equartey: Add comments 👍

packages/amplify_api/lib/graphql/graphql_helpers.dart
Equartey: Realistically the name would only be invalid if the schema provided is, but since `schema.pluralName` is nullable I've added a null check.

packages/amplify_api/lib/graphql/graphql_helpers.dart
```diff
@@ -0,0 +1,155 @@
...
+  @override
+  GraphQLRequest<T> buildQuery<T extends Model>(
+      {required ModelType modelType,
+      required Map<String, String>? variableInput,
+      required String? id,
+      required GraphQLRequestType requestType,
+      required GraphQLRequestOperation requestOperation}) {
+    ModelSchema schema = _getSchema(modelType);
+
+    String? name = requestOperation == GraphQLRequestOperation.list
```
ragingsquirrel3, line 103: nit: I suggest declaring a boolean variable like `isListOperation` above and then using that variable in this if statement, as well as same condition repeated below. That way, this condition is only defined once. Another thing to think about is if all the list-specific conditions can be in the same place? Maybe not, but might make it clearer for reader to know what happens if list.

packages/amplify_api/lib/graphql/graphql_helpers.dart
```diff
@@ -0,0 +1,155 @@
...
+        ? schema.pluralName
+        : schema.name;
+
+    if (name == null) {
+      throw ApiException('No schema name found',
+          recoverySuggestion:
+              'Pass in a valid modelProvider instance while instantiating APIPlugin or provide a vaild ModelType');
```
ragingsquirrel3: nit: typo, "vaild" > "valid"

packages/amplify_api/lib/graphql/graphql_helpers.dart
```diff
@@ -0,0 +1,155 @@
...
+    }
+  }
+
+  String _getFieldsFromModelType(ModelSchema schema) {
+    Map<String, ModelField?>? fieldsMap = schema.fields;
+    return fieldsMap?.entries
+            .map((entry) => entry.value?.association == null ? entry.key : '')
+            .toList()
+            .join(' ') ??
+        '';
```
ragingsquirrel3, line 63: Is the empty string a valid case? If not, should this just throw error if null?

packages/amplify_api/lib/graphql/graphql_helpers.dart
```diff
@@ -0,0 +1,155 @@
...
+      throw ApiException('No modelProvider found',
+          recoverySuggestion:
+              'Pass in a modelProvider instance while instantiating APIPlugin');
+    }
+
+    ModelSchema schema = provider.modelSchemas.firstWhere(
+        (elem) => provider.getModelTypeByModelName(elem.name) == modelType);
+
+    if (schema == null) {
+      throw ApiException('No schema found for the ModelType provided',
```
ragingsquirrel3: It seems like the logic for asserting valid schema/model provider is a little scattered. Would it be possible to define a helper method like `validateModelProviderAndModelSchema` which can check all these things in one place? Then, following lines of code can precede assuming all relevant inputs from the model provider are not null? That method can also be easily checked with other/future error cases and message easily updated.

Equartey: Great idea. Let me throw something together.

packages/amplify_api/lib/graphql/graphql_helpers.dart
Equartey: Good question. The map's keys are nullable, so code-wise it's valid, but maybe not in practice? I imagine a model will always have fields.

packages/amplify_api/lib/graphql/graphql_helpers.dart
Equartey: Fair enough, I consolidated logic to avoid multiple conditions. Previously I was trying to have everything only defined once, but this seems like a better flow.

packages/amplify_api/lib/graphql/graphql_helpers.dart
ragingsquirrel3: I also think a valid model will have these fields, just suggest accounting for if that case is not true in a way caller will know why they are getting this error. If they just see empty string, will be hard to know that it changed and why if they were debugging.

One idea is to add this logic to the validate helper I mentioned, then can just assume this is not null here. Can just manually throw the error here, too.

packages/amplify_api/test/amplify_api_graphql_helpers_test.dart
```diff
@@ -0,0 +1,51 @@
...
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import 'package:amplify_api/amplify_api.dart';
+import 'package:amplify_api/graphql/model_queries.dart';
```
ragingsquirrel3, line 17: I played with this locally and it looks like lines 17-18 can be removed here. If so, would remove these lines so this tests only things that consumer of api plugin would have in scope. It will confirm exports work as expected.

packages/amplify_api/lib/src/graphql/helpers/model_queries.dart
```diff
@@ -14,13 +14,14 @@
  */
 
 import 'package:amplify_api/amplify_api.dart';
-import 'package:amplify_api/graphql_helpers.dart';
+import 'package:amplify_api/src/graphql/factories/model_queries_factory.dart';
 import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart';
 
-// This class provides static method calls to enable a simpler DX
+// This class provides static method calls to enable a simpler DX,
+// while preserving internal interfaces
```
ragingsquirrel3, line 21: This file itself can be in lib (not in src) I think as it's a public facing API and included as export. The internal dependencies it imports, like model_queries_factory, are correct to be in src.

packages/amplify_api_plugin_interface/lib/src/GraphQL/GraphQLHelpers.dart
```diff
@@ -16,44 +16,39 @@
...
+  GraphQLRequest<T> delete<T extends Model>(T model, {QueryPredicate? where});
 
   // DeleteById
-  GraphQLRequest<T> deleteById<T extends Model>(String Id) {
-    return throw UnimplementedError("delete() has not been implemented.");
-  }
+  GraphQLRequest<T> deleteById<T extends Model>(String Id);
+}
+
+abstract class GraphQLRequestFactoryInterface {
```
ragingsquirrel3, line 46: Can this class be removed from interface as it's an internal implementation detail?

packages/amplify_api/lib/src/graphql/helpers/graphql_helpers.dart
```diff
@@ -13,12 +13,4 @@
...
-import 'package:amplify_api_plugin_interface/amplify_api_plugin_interface.dart';
-import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart';
-
-class ModelQueriesFactory extends ModelQueriesInterface {
-  @override
-  GraphQLRequest<T> get<T extends Model>(ModelType<T> modelType, String id) {
-    throw UnimplementedError("get() has not been implemented.");
-  }
-}
+export 'model_queries.dart';
```
ragingsquirrel3, line 16: Is this file just to include things like "model_mutations" and "model_subscriptions" later? If so, can those things be directly exported from amplify_api.dart and reduce the need for this file?

Equartey: Yeah, that was the idea. We can forego it.

packages/amplify_api_plugin_interface/lib/src/GraphQL/GraphQLHelpers.dart
Equartey: Removed.

packages/amplify_api/lib/src/graphql/helpers/model_queries.dart
Equartey: Moved.

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    ModelSchema schema = provider.modelSchemas.firstWhere(
+        (elem) => provider.getModelTypeByModelName(elem.name) == modelType,
+        orElse: () => throw ApiException(
+            'No schema found for the ModelType provided',
+            recoverySuggestion:
+                'Pass in a valid modelProvider instance while instantiating APIPlugin or provide a valid ModelType'));
+
+    if (schema.fields == null) {
```
dnys1, line 59: This seems like bad DX, in my opinion. Not your code, Elijah, but the fact that `fields` and `pluralName` are nullable to begin with.

@ragingsquirrel3 thoughts? Is it worth just making these fields required on `ModelSchema`?

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
```diff
@@ -0,0 +1,135 @@
...
+      required GraphQLRequestOperation requestOperation}) {
+    ModelSchema schema = _getAndValidateSchema(modelType, requestOperation);
+
+    String name = schema.name;
+
+    // fields to retrieve, e.g. "id name createdAt"
+    String fields = _getFieldsFromModelType(schema);
+
+    // e.g. "query"
+    String requestTypeVal = describeEnum(requestType.toString());
```
dnys1, line 96: `.toString` is not needed on these calls. You can just pass the enum and it will handle the stringification.

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
```diff
@@ -0,0 +1,135 @@
...
+    // e.g. "get"
+    String requestOperationVal = describeEnum(requestOperation.toString());
+
+    // Upper document input: ($id: ID!)
+    var varInputUpperStr = '';
+    // Lower document input: (id: $id)
+    var varInputLowerStr = '';
+
+    if (variableInput != null && variableInput.isNotEmpty) {
```
dnys1, line 105: The input should not be used for building the query. Only the schema and the `_getModelType` function are required. For one, this variable input is not being validated against the schema and two, the values of the map cannot serve both as the type parameter and the value of the argument. For example, for a query op, `variableInput` should be `{'id': model.id}`, and the field name and type should be retrieved from the model schema.

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
```diff
@@ -0,0 +1,135 @@
...
+        return "int";
+      case ModelFieldTypeEnum.model:
+        return "ID";
+      default:
+        return "Error: UnknownType!";
+    }
+  }
+
+  String _getFieldsFromModelType(ModelSchema schema) {
+    Map<String, ModelField> fieldsMap = schema.fields!;
```
ragingsquirrel3, line 35: You might leave a little comment noting that fields is verified non-null in _getAndValidateSchema so it's safe to assert as non-null here.

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
```diff
@@ -0,0 +1,135 @@
...
+    if (requestOperation == GraphQLRequestOperation.list) {
+      name = schema.pluralName!;
+      fields = 'items { $fields } nextToken';
+    }
+
+    String doc =
+        '''$requestTypeVal $requestOperationVal$name${varInputUpperStr} { $requestOperationVal$name$varInputLowerStr { $fields } }''';
+
+    // TODO: create input map for variables & connect with current variableInput
+    Map<String, dynamic>? variables = id != null ? {"id": id} : {};
```
ragingsquirrel3, line 131: nit: I don't think this variable is nullable bc you are assigning it 1 of 2 things, so can be `Map<String, String>` I think, but also looks like you are gonna work on this soon, anyway.

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
```diff
@@ -0,0 +1,135 @@
...
+    var varInputUpperStr = '';
+    // Lower document input: (id: $id)
+    var varInputLowerStr = '';
+
+    if (variableInput != null && variableInput.isNotEmpty) {
+      List<String> variableInputUpper = [];
+      List<String> variableInputLower = [];
+
+      variableInput.forEach((key, value) {
+        variableInputUpper.add("\$$key: $value!");
```
dnys1, line 110: The `!` denotes a required field but not all fields will be required as part of the input. This information is available in the `ModelFieldDefinition.isRequired` field.

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
```diff
@@ -0,0 +1,135 @@
...
+import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart';
+import 'package:flutter/foundation.dart';
+
+class GraphQLRequestFactory {
+  String _getModelType(ModelFieldTypeEnum? val) {
+    switch (val) {
+      case ModelFieldTypeEnum.string:
+        return "String";
+      case ModelFieldTypeEnum.int:
+        return "int";
```
dnys1, line 26: Should be capital `Int`

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
```diff
@@ -0,0 +1,135 @@
...
+   *  buildQuery()
+   *
+   *  
+   *  Example: 
+   *    query getBlog($id: ID!) { getBlog(id: $id) { id name createdAt } }
+  */
+  GraphQLRequest<T> buildQuery<T extends Model>(
+      {required ModelType modelType,
+      required Map<String, String>? variableInput,
+      required String? id,
```
ragingsquirrel3, line 85: nit: could this not be required bc it's nullable? This is requiring someone to at least supply value of `null`, correct?

packages/amplify_api/test/amplify_api_graphql_helpers_test.dart
```diff
@@ -0,0 +1,48 @@
...
+
+void main() {
+  test("ModelQueries.get() should craft a valid request", () {
+    AmplifyAPI api = AmplifyAPI(modelProvider: ModelProvider.instance);
+
+    String id = UUID.getUUID();
+    String expected =
+        "query getBlog(\$id: ID!) { getBlog(id: \$id) { id name createdAt } }";
+
+    GraphQLRequest<Blog> req = ModelQueries.get<Blog>(Blog.classType, id);
```
ragingsquirrel3, line 30: I was noticing that if someone called a simpler version `final req = ModelQueries.get(Blog.classType, id);` (maybe what someone would actually do) then req is inferred to be of type `GraphQLRequest<Blog>` which is cool, and means it's working. Nice job! no change needed, just pointing out that it's cool.

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
```diff
@@ -0,0 +1,135 @@
...
+        schema.pluralName == null) {
+      throw ApiException('No schema name found',
+          recoverySuggestion:
+              'Pass in a valid modelProvider instance while instantiating APIPlugin or provide a valid ModelType');
+    }
+
+    return schema;
+  }
+
+  /*
```
ragingsquirrel3, line 75: nit: suggest changing this block comment to 
```
/**
   *  Example output `GraphQLRequest` with document string: 
   *    `"query getBlog($id: ID!) { getBlog(id: $id) { id name createdAt } }"`
  */
```
I believe the 2 stars make it a JSdoc style block comment, which also seems to work in dart so the comment will document the method properly. Later, might want to think about documenting the public APIs, but not needed her IMO>

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
dnys1: You can use `ModelFieldDefinition.idKey` for retrieving the primary key field, which may or may not be of type `ID!`. See [here](https://docs.amplify.aws/cli/graphql-transformer/key#example-get-customers-by-email) where `email: String!` is the primary key.

packages/amplify_api/lib/src/graphql/model_queries_factory.dart
```diff
@@ -0,0 +1,49 @@
...
+
+  static final ModelQueriesFactory _instance = ModelQueriesFactory._();
+
+  static ModelQueriesFactory get instance => _instance;
+
+  @override
+  GraphQLRequest<T> get<T extends Model>(ModelType<T> modelType, String id) {
+    Map<String, String> variableInput = {"id": "ID"};
+
+    return GraphQLRequestFactory().buildQuery<T>(
```
dnys1, line 33: Should also be a singleton class

packages/amplify_api/test/amplify_api_graphql_helpers_test.dart
```diff
@@ -0,0 +1,48 @@
...
+import 'resources/Blog.dart';
+import 'resources/ModelProvider.dart';
+
+void main() {
+  test("ModelQueries.get() should craft a valid request", () {
+    AmplifyAPI api = AmplifyAPI(modelProvider: ModelProvider.instance);
+
+    String id = UUID.getUUID();
+    String expected =
+        "query getBlog(\$id: ID!) { getBlog(id: \$id) { id name createdAt } }";
```
dnys1, line 28: Can use raw strings, if you want, for easier writing. This is the same as:

```dart
r"query getBlog($id: ID!) { getBlog(id: $id) { id name createdAt } }"
```

packages/amplify_api/test/amplify_api_graphql_helpers_test.dart
```diff
@@ -0,0 +1,48 @@
...
+
+    GraphQLRequest<Blog> req = ModelQueries.get<Blog>(Blog.classType, id);
+
+    expect(req.document, expected);
+    expect(req.variables.containsValue(id), true);
```
dnys1, line 33: Let's be more specific so we can make sure that `variables` contains everything that's necessary and nothing that's not. If map comparison is failing, try the `equals` matcher or use the `mapEquals` function from Flutter but this can only be used in integration tests. There is also `MapEquality` from package [collection](https://pub.dev/packages/collection) which can be used for very customizable Map comparison and can be used in unit tests.

```dart
expect(req.variables, equals({'id': id}));

// or

expect(mapEquals(req.variables, {'id': id}), isTrue);

// or

expect(const MapEquality().equals(req.variables, {'id': id}), isTrue);
```

packages/amplify_api/test/amplify_api_graphql_helpers_test.dart
```diff
@@ -0,0 +1,48 @@
...
+    AmplifyAPI api = AmplifyAPI();
+    try {
+      GraphQLRequest<Blog> req = ModelQueries.get<Blog>(Blog.classType, "");
+    } on ApiException catch (e) {
+      expect(e.message, "No modelProvider found");
+      expect(e.recoverySuggestion,
+          "Pass in a modelProvider instance while instantiating APIPlugin");
+      return;
+    }
+    throw new Exception("Expected an ApiException");
```
dnys1, line 46:
```dart
fail("Expected an ApiException");
```

packages/amplify_api_plugin_interface/lib/src/GraphQL/GraphQLHelpers.dart
```diff
@@ -16,44 +16,29 @@
...
-  GraphQLRequest<T> delete<T extends Model>(T model, {QueryPredicate? where}) {
-    return throw UnimplementedError("delete() has not been implemented.");
-  }
+  GraphQLRequest<T> delete<T extends Model>(T model, {QueryPredicate? where});
 
   // DeleteById
-  GraphQLRequest<T> deleteById<T extends Model>(String Id) {
-    return throw UnimplementedError("delete() has not been implemented.");
-  }
+  GraphQLRequest<T> deleteById<T extends Model>(String Id);
```
dnys1, line 43: Lowercase `id`, and you will probably need the model type as input, same as `get`, no?

dnys1 (changes requested):
Looks good overall! There are some changes needed in `buildQuery` which will help when adding the other request types. Keep up the great work!

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
ragingsquirrel3: I agree bad DX bc of weird nullable fields in schema class. I believe Elijah noted the same idea during design, that these fields should be made non-nullable in original definition. I don't think this PR is the place for it, though.

We were thinking it might make sense to let the implementation progress with existing defs so that a little down the road we might have a clearer list of all the things that need to change in model/provider definitions and/or any related codegen changes that might be needed so they can be made at once. Does that make sense? For now, I think these validation assertions are needed.

packages/amplify_api/test/amplify_api_graphql_helpers_test.dart
Equartey: Exciting stuff!

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
```diff
@@ -0,0 +1,135 @@
...
+    ModelProviderInterface? provider = AmplifyAPI.instance.getModelProvider();
+
+    if (provider == null) {
+      throw ApiException('No modelProvider found',
+          recoverySuggestion:
+              'Pass in a modelProvider instance while instantiating APIPlugin');
+    }
+
+    ModelSchema schema = provider.modelSchemas.firstWhere(
+        (elem) => provider.getModelTypeByModelName(elem.name) == modelType,
```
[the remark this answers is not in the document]
Equartey: Nice, easier to read too!

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
Equartey: Are we tracking items like these so we can follow up on them down the road?

packages/amplify_api/test/amplify_api_graphql_helpers_test.dart
Equartey: Oh fancy!

packages/amplify_api/test/amplify_api_graphql_helpers_test.dart
Equartey: Thanks for the tip! I suspected there was a better way to write assertions for these.

packages/amplify_api_plugin_interface/lib/src/GraphQL/GraphQLHelpers.dart
Equartey: Indeed, good catch.

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
```diff
@@ -0,0 +1,135 @@
...
+  String _getModelType(ModelFieldTypeEnum? val) {
+    switch (val) {
+      case ModelFieldTypeEnum.string:
+        return "String";
+      case ModelFieldTypeEnum.int:
+        return "int";
+      case ModelFieldTypeEnum.model:
+        return "ID";
+      default:
+        return "Error: UnknownType!";
```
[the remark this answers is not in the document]
Equartey: I believe it can be non-null, this will be used in the mutation requests so it's tricky to validate at the moment.

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
Equartey: I've refactored this section. It should now read clearer and be more generic so we can reuse down the road. Thanks for the insight 💯

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
dnys1: Great question. @ragingsquirrel3 what is the best approach to track this, do you think? Would it require a design review? If not, I can just make a PR for it.

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
```diff
@@ -0,0 +1,160 @@
...
+
+    // build inputs based on request operation
+    switch (operation) {
+      case GraphQLRequestOperation.get:
+        upperList.add(r"$id: ID!");
+        lowerList.add(r"id: $id");
+        lowerOutput = "(${lowerList.join(", ")})";
+        break;
+      case GraphQLRequestOperation.list:
+        break;
```
dnys1, line 105: You can remove `break` so it falls through to the default case. Don't need to fix now, just an FYI.

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
```diff
@@ -0,0 +1,160 @@
...
+      default:
+        schema.fields!.forEach((field, val) {
+          upperList.add("\$$field: ${_getModelType(val.type.fieldType)}");
+          lowerList.add("$field: \$$field");
+        });
+        lowerOutput = "(input: { ${lowerList.join(", ")} })";
+    }
+
+    if (upperList.length == 0 || lowerList.length == 0) {
+      return DocumentInputs("", "");
```
dnys1, line 115: This case will need to be handled in a slightly different way below, since selection sets require 1 or more values:
- https://spec.graphql.org/June2018/#sec-Selection-Sets
- https://spec.graphql.org/June2018/#sec-Grammar-Notation

So, the following is invalid GraphQL:

```graphql
query listItems {
  listItems { } 
}
```

Now, I'm not sure in what scenario this condition would ever hold true, since there will never be a model with no fields. Just thought I'd mention it, though - no need to fix right now.

ragingsquirrel3 (approved):
Nice job on this part! This is a good foundation, does a particularly great job of exposing public APIs properly while having some internal implementation details that can be reused later but not leaked to consumer. Validates input. Also, has foundation unit tests.

Some of this might have to change a little to support list queries with predicates or other request types, but that can be revisited then.

dnys1 (approved):

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
Equartey: Thanks for the feedback, sounds like I can clean up the list operations a little more. Currently how it sets up actually produces this document: 

```graphql
query listBlogs {
   listBlogs {
    items {
      id
      name
      createdAt
    }
    nextToken
  } 
}
```

Fields (id, name, createdAt) are added in another step. That being said, your comment highlights this can probably be consolidated so it's easier to follow/read. Will address in the next iteration 👍

PR #744, opened 2021-07-24
Datastore support nullable lists

fjnoyp:
*Issue #, if available:*

*Description of changes:*

Support nullable lists types.  The main changes will be in the Dart codegen models which will be generated differently to allow for lists of nullable types.  We are still working to finalize what the codegen models will look like. 

Testing on Android works.  iOS is not working currently. 

Also need to remove testing code on main.dart 

### DIFF
--- a/packages/amplify_datastore/example/lib/models/ModelProvider.dart
+++ b/packages/amplify_datastore/example/lib/models/ModelProvider.dart
@@ -26,6 +26,9 @@ import 'HasManyModel.dart';
 import 'HasOneModel.dart';
 import 'Post.dart';
 import 'PostAuthComplex.dart';
+import 'TestNullableListsEnum.dart';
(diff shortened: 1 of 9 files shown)


packages/amplify_datastore/example/lib/models/ModelProvider.dart
```diff
@@ -111,6 +120,21 @@ class ModelProvider implements ModelProviderInterface {
           return PostAuthComplex.classType;
         }
         break;
+      case "TestNullableListsEnum":
```
Jordan-Nelson, line 123: I have an open PR which adds a schema.graphql file to/amplify_datastore/example/tool/. It is used as part of the provisioning script for datastore.

The new models added in this PR should probably be added to that schema so that everyone can re-generate them.

PR #753, opened 2021-07-26
chore(datastore): add integration tests for datastore

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- add basic integration tests for the datastore save/query

### DIFF
--- a/packages/amplify_datastore/example/integration_test/main_test.dart
+++ b/packages/amplify_datastore/example/integration_test/main_test.dart
@@ -25,7 +25,7 @@ void main() async {
 
   group('amplify_datastore', () {
     setUpAll(() async {
-      await configureAuth();
+      await configureDataStore();
     });
 
     save_tests.main();
--- a/packages/amplify_datastore/example/integration_test/main_test.dart
+++ b/packages/amplify_datastore/example/integration_test/main_test.dart
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import 'package:amplify_datastore/amplify_datastore.dart';
+import 'package:amplify_datastore_example/models/ModelProvider.dart';
+import 'package:integration_test/integration_test.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:amplify_flutter/amplify.dart';
+import 'package:amplify_datastore_example/amplifyconfiguration.dart';
+
+import 'save_test.dart' as save_tests;
+import 'query_test.dart' as query_tests;
+
+void main() async {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  group('amplify_datastore', () {
+    setUpAll(() async {
+      final authPlugin =
+          AmplifyDataStore(modelProvider: ModelProvider.instance);
+      await Amplify.addPlugins([authPlugin]);
+      await Amplify.configure(amplifyconfig);
+    });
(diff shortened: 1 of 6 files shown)


packages/amplify_datastore/example/integration_test/main_test.dart
```diff
@@ -0,0 +1,40 @@
...
+
+import 'save_test.dart' as save_tests;
+import 'query_test.dart' as query_tests;
+
+void main() async {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  group('amplify_datastore', () {
+    setUpAll(() async {
+      final authPlugin =
```
ragingsquirrel3, line 31: nit: not an auth plugin?

packages/amplify_datastore/example/pubspec.yaml
```diff
@@ -23,6 +23,10 @@ dependencies:
   #   path: ../../amplify_api
   amplify_flutter:
     path: ../../amplify_flutter
+  flutter_driver:
```
ragingsquirrel3, line 26: suggestion: I would define the min flutter version as 2.2.0 in env block above like https://github.com/aws-amplify/amplify-flutter/blob/main/packages/amplify_api/example/pubspec.yaml#L10 . The integration test driver is only in 2.0.0 + and the versions between 2 and 2.2 have some null safety issue that sometimes involves adding an override dep, so I think stating for flutter 2.2.0 + technically the correct thing to do. In theory, could make CI integration easier later by avoiding cryptic error messages.

packages/amplify_datastore/example/integration_test/query_test.dart
```diff
@@ -0,0 +1,75 @@
...
+    testWidgets('should return all data by default',
+        (WidgetTester tester) async {
+      Blog testBlog1 = Blog(name: 'blog one');
+      Blog testBlog2 = Blog(name: 'blog two');
+      Blog testBlog3 = Blog(name: 'blog three');
+      await Amplify.DataStore.save(testBlog1);
+      await Amplify.DataStore.save(testBlog2);
+      await Amplify.DataStore.save(testBlog3);
+      var blogs = await Amplify.DataStore.query(Blog.classType);
+      expect(blogs.length, 3);
```
HuiSF, line 40: Should we also assert by comparing query result and the expected data?

Jordan-Nelson: Updated 👍

packages/amplify_datastore/example/pubspec.yaml
Jordan-Nelson: Updated 👍

packages/amplify_datastore/example/integration_test/main_test.dart
```diff
@@ -0,0 +1,40 @@
...
+import 'package:amplify_datastore_example/amplifyconfiguration.dart';
+
+import 'save_test.dart' as save_tests;
+import 'query_test.dart' as query_tests;
+
+void main() async {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  group('amplify_datastore', () {
+    setUpAll(() async {
```
ragingsquirrel3, line 30: Would it be possible to add this block (inside an `if (!Amplify.isConfigured) `) for the query test and save test as well? That way, the test files can be run individually.

ragingsquirrel3 (approved):
Thanks for adding these! It passed locally for me on ios and android. I left one small comment, but otherwise looks great.

HuiSF (approved):
Thanks!

packages/amplify_datastore/example/integration_test/main_test.dart
Jordan-Nelson: Good call. Updated.

packages/amplify_datastore/example/integration_test/utils/setup_utils.dart
```diff
@@ -0,0 +1,13 @@
+import 'package:amplify_datastore/amplify_datastore.dart';
+import 'package:amplify_datastore_example/amplifyconfiguration.dart';
+import 'package:amplify_datastore_example/models/ModelProvider.dart';
+import 'package:amplify_flutter/amplify.dart';
+
+Future<void> configureAuth() async {
```
ragingsquirrel3, line 6: nit: not auth, could change to datastore or similar

Jordan-Nelson: Good catch. Copy/paste is not my friend today 😅

PR #754, opened 2021-07-27
fix(datastore): cannot saving boolean as integer in SQLite

HuiSF:
*Issue #, if available:* #752

*Description of changes:*

Convert `bool` filed type to java `Boolean` object type instead of Kotlin primitive `Boolean` type. 

Reason: amplify-android uses Java `Boolean` class but not expecting Kotlin primitive boolean. This will let amplify-android correctly create boolean value column with `integer` type, and further to fix the issue described in #752

----


cshfang (approved):

cshfang (approved):
Thanks for updating and walking me through what was happening

fjnoyp (approved):
Looks good, thanks for doing this!  How will we notify users as this seems to be a possible major breaking change for them all.

PR #757, opened 2021-07-27
chore: set min flutter version to '>1.20.0' in packages where set to …

ragingsquirrel3:
*Description of changes:* Set min flutter version to '>1.20.0' in packages where set to something else. There were some warnings during publishing last release that in some places is set to things like ^1.20.0 which is incorrect bc has upper bounds.

Long-term, maybe should be changed to '>=2.0.0' bc I think these packages really only work on flutter 2+, but this PR at least just makes them consistent and does not have an upper bound.

### DIFF
--- a/packages/amplify_analytics_pinpoint/pubspec.yaml
+++ b/packages/amplify_analytics_pinpoint/pubspec.yaml
@@ -4,8 +4,8 @@ version: 0.2.1
 homepage: https://github.com/aws-amplify/amplify-flutter/tree/main/packages/amplify_analytics_pinpoint
 
 environment:
-  sdk: '>=2.12.0 <3.0.0'
-  flutter: ^1.20.0
+  sdk: ">=2.12.0 <3.0.0"
+  flutter: ">=1.20.0"
 
 dependencies:
   flutter:
--- a/packages/amplify_auth_cognito/example/pubspec.yaml
+++ b/packages/amplify_auth_cognito/example/pubspec.yaml
@@ -7,7 +7,7 @@ publish_to: "none" # Remove this line if you wish to publish to pub.dev
 
 environment:
   sdk: ">=2.12.0 <3.0.0"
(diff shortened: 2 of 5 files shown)


packages/amplify_analytics_pinpoint/pubspec.yaml
```diff
@@ -4,8 +4,8 @@ version: 0.2.1
 homepage: https://github.com/aws-amplify/amplify-flutter/tree/main/packages/amplify_analytics_pinpoint
 
 environment:
-  sdk: '>=2.12.0 <3.0.0'
-  flutter: ^1.20.0
+  sdk: ">=2.12.0 <3.0.0"
+  flutter: ">=1.20.0"
```
cshfang, line 8: Nit: Can we drop the quotes on non-ranges to be consistent with Flutter/Dart docs?

cshfang (approved):
Minor nit comment but lgtm

dnys1 (approved):
Thanks!

packages/amplify_analytics_pinpoint/pubspec.yaml
ragingsquirrel3: discussed offline, but we **think** this needs quotes based on how my editor was coloring it

packages/amplify_auth_cognito/example/pubspec.yaml
```diff
@@ -36,10 +36,6 @@ dev_dependencies:
 # For information on the generic Dart part of this file, see the
 # following page: https://dart.dev/tools/pub/pubspec
 
-# Needed to avoid conflict for integration tests until flutter driver has null safe version of crypto https://github.com/flutter/flutter/issues/77282
```
ragingsquirrel3, line 39: This was fixed with flutter 2.2.0, hence making it the min flutter version and removing this override.

cshfang (approved):

PR #758, opened 2021-07-27
feat(amplify_authenticator) confirmSignIn screen

israx:
*Issue #, if available:*

*Description of changes:*
 -Confirm sign in screen
 -Integration test for confirm sign in screen


packages/amplify_authenticator/lib/src/widgets/default_forms.dart
```diff
@@ -5,56 +5,80 @@ import 'package:amplify_authenticator/src/widgets/form_fields.dart';
...
+    );
+  }
+
+  static ConfirmSignInForm confirmSignInForm() {
+    return const ConfirmSignInForm(
+      formFields: FormFields(
+        children: [
+          ConfirmSignInFormField(
+              title: 'Code*',
+              hintText: "Enter your confirmation code",
```
dnys1, line 80: Interesting edge case I bumped into today on discord: so, if a user is created by an admin in Cognito, they issue a temporary password. When the user logs in for the first time with the temp password - the signInStep will be `CONFIRM_SIGN_IN_WITH_NEW_PASSWORD` and users should be prompted for a new password, after which calling  `Amplify.Auth.confirmSignIn(confirmationValue: newPassword);` saves the new password and logs them in.

Now, let's say that MFA is also enabled. The signInStep from *that* call will be `CONFIRM_SIGN_IN_WITH_SMS_MFA_CODE`. And finally after calling `confirmSignIn` again with the MFA code, the signInStep will be `DONE`. But they will need to be presented the confirm sign in screen twice. And on the first time, we probably want to show "Enter a new password" or something to that effect instead of "Enter your confirmation code"

dnys1 (commented):
Looks good. Posted an interesting edge case to consider!

packages/amplify_authenticator/lib/src/widgets/default_forms.dart
haverchuck: @israx -  @dnys1 's comment above is what I mentioned on Tuesday. You can handle it in a separate PR if you want.

packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -16,6 +16,9 @@
 library amplify_authenticator;
 
 import 'package:amplify_authenticator/src/keys.dart';
+
```
haverchuck, line 19: Do we need this extra lines?

packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -227,6 +298,8 @@ class _AuthenticatorState extends State<Authenticator> {
   @override
   void dispose() {
     _stateMachineBloc.dispose();
-      super.dispose();
+
```
haverchuck, line 301: Extra lines needed?

packages/amplify_authenticator/lib/src/widgets/buttons.dart
```diff
@@ -1,3 +1,6 @@
+
+import 'package:amplify_authenticator/src/views/confirm_signin_viewmodel.dart';
+
```
haverchuck, line 3: Same comment about extra lines.

haverchuck (approved):
Left a few minor comments, but otherwise LGTM.

packages/amplify_authenticator/lib/src/widgets/default_forms.dart
dnys1: Cool cool. Approving it now so @israx you can handle how you think best.

dnys1 (approved):
Well done!

Issue #444, opened 2021-03-15
Different query for android and iOS - Datastore


gauravdangi:
**Describe the bug**
I have a 1:n relation between a user and his books. Below is the graphql schema

**Graphql schema**
```
type User @model {
  id: ID!
  name: String!
  email: String!
  premiumTill: AWSDateTime!
  books: [Book] @connection(keyName: "byUser", fields: ["id"])
}

type Book @model
  @key(name: "byUser", fields: ["userID", "year"]) {
  id: ID!
  title: String!
  year: Int
  userID: ID!
  user: User @connection(fields: ["userID"])
  chapters: [Chapter] @connection(keyName: "byBook", fields: ["id"])
}
```

I noticed when I want to fetch books by userId, on iOS I have to write the following command:-
```
      this._books = await Amplify.DataStore.query(
        Book.classType,
        where: Book.USER.eq(AppSettings.userId),
        sortBy: [Book.YEAR.descending()],
      );
```
The above code runs fine when I am running my project on an iOS simulator. But when I ran my project on a real android device, it threw me the following error:-
<details>
  <summary>Click to expand!</summary>
```
E/SQLiteLog( 5157): (1) no such column: user in "SELECT `Book`.`id` AS `Book_id`, `Book`.`title` AS `Book_title`, `Book`.`year` AS `Book_year`, `Book`.`userID` AS `Book_userID`, `User`.`id` AS `User_id`, `User`.`email` AS `User_email
E/amplify:flutter:datastore( 5157): Query operation failed.
E/amplify:flutter:datastore( 5157): DataStoreException{message=Error in querying the model., cause=android.database.sqlite.SQLiteException: no such column: user (code 1 SQLITE_ERROR): , while compiling: SELECT `Book`.`id` AS `Book_id`, `Book`.`title` AS `Book_title`, `Book`.`year` AS `Book_year`, `Book`.`userID` AS `Book_userID`, `User`.`id` AS `User_id`, `User`.`email` AS `User_email`, `User`.`name` AS `User_name`, `User`.`premiumTill` AS `User_premiumTill` FROM `Book` LEFT JOIN `User` ON `Book`.`userID`=`User`.`id` WHERE user = ? ORDER BY `Book`.`year` DESC;, recoverySuggestion=See attached exception for details.}
[...]

PR #761, opened 2021-07-28
fix(datastore): Query nested model causes column not found sql error

HuiSF:
*Issue #, if available:* #444

*Description of changes:*

1. Passing query predicate corresponding `modelSchema` to `QueryPredicateBuilder.fromSerialziedMap`
2. If predicate field is a nested model, replace it with the actual model field 

Note: the [second commit](https://github.com/aws-amplify/amplify-flutter/commit/22057609c7816d08fab5a74b56f09a8e61bb6d3c) in this PR presents the actual code change

e.g. run below query 

```dart
final res = await Amplify.DataStore.query(Post.classType,
    where: Post.BLOG.eq("5eff14cf-9a67-4fe3-8457-11a30f5e3a86"));
print(res);
```

with schema

```graphql
type Post @model @key(name: "byBlog", fields: ["blogID"]) {
  id: ID!
  title: String!
  rating: Int!
  created: AWSDateTime
  blogID: ID!
  blog: Blog @connection(fields: ["blogID"])
  comments: [Comment] @connection(keyName: "byPost", fields: ["id"])
}
```

Before the change, query fails in Android due to SQL error: `no such a column: blog`, while the query is running correctly in iOS.

----

### DIFF
--- a/packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/query/QueryPredicateBuilder.kt
+++ b/packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/query/QueryPredicateBuilder.kt
@@ -18,6 +18,7 @@ package com.amazonaws.amplify.amplify_datastore.types.query
 import com.amazonaws.amplify.amplify_datastore.util.safeCastToList
 import com.amazonaws.amplify.amplify_datastore.util.safeCastToMap
 import com.amplifyframework.core.model.ModelSchema
+import com.amplifyframework.core.model.annotations.BelongsTo
 import com.amplifyframework.core.model.query.predicate.QueryField
 import com.amplifyframework.core.model.query.predicate.QueryPredicate
 import com.amplifyframework.core.model.query.predicate.QueryPredicateGroup
@@ -41,7 +42,11 @@ class QueryPredicateBuilder {
                 var field = queryPredicateOperationMap["field"] as String
 
                 if (modelSchema?.associations?.containsKey(field) == true) {
-                    field = modelSchema.associations.getValue(field).targetName
+                    val association = modelSchema.associations.getValue(field);
+
+                    if (BelongsTo::class.java.simpleName.equals(association.name)) {
+                        field = modelSchema.associations.getValue(field).targetName
+                    }
                 }
 
                 val queryField: QueryField = QueryField.field(field)
--- a/packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/query/QueryPredicateBuilder.kt
+++ b/packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/query/QueryPredicateBuilder.kt
@@ -32,10 +32,12 @@ class QueryPredicateBuilder {
             }
 
             if (serializedMap.containsKey("queryPredicateOperation")) {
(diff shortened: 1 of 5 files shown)


cshfang (approved):

cshfang (approved):
Looks better

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/query/QueryPredicateBuilder.kt
```diff
@@ -25,17 +26,27 @@ import com.amplifyframework.core.model.query.predicate.QueryPredicateOperation.n
...
             }
 
             if (serializedMap.containsKey("queryPredicateOperation")) {
-                val queryPredicateOperationMap: Map<String, Any> = serializedMap["queryPredicateOperation"].safeCastToMap()!!
-                val field = queryPredicateOperationMap["field"] as String
+                val queryPredicateOperationMap: Map<String, Any> =
+                    serializedMap["queryPredicateOperation"].safeCastToMap()!!
+                var field = queryPredicateOperationMap["field"] as String
+
+                if (modelSchema?.associations?.containsKey(field) == true) {
```
dnys1, line 43: Should this also check that `association.associatedType` is belongsTo?

HuiSF: Good catch, it should work only with `belongsTo`.

Though I found another issue within the generated model while testing `hasOne`. Use below schema as example:
```graphql
type Post @model{
  id: ID!
  title: String!
  content: String!
  reviewerId: ID
  reviewer: Person @connection(fields: [reviewerId])
}

type Person {
  id: ID!
  name: String!
}
```

Generated `Post` model has correct fields

```
class Post extends Model {
  ...
  final String? _reviewerId;
  final Person? _reviewer;
```

However, the generated model marks `reviewer` as a `QueryField`. 

```
  static final QueryField REVIEWERID = QueryField(fieldName: "reviewerId");
  static final QueryField REVIEWER = QueryField(
    fieldName: "reviewer",
    fieldType: ModelFieldType(ModelFieldTypeEnum.model, ofModelName: (Person).toString()));
```
This is incorrect, as with `hasOne` `reviewerId` should be the query field.

I will update the PR tomorrow, and open an issue for codegen to fix this newly found issue.

HuiSF: Correction: it should check `association.name` to determine relationship.

In addition, if the codegen issue gets fixed, theoretically the check of `association.name == "belongsTo"` could be omitted. Will add the guard regardless.

dnys1: Good call - I think the codegen issue could be similar to what Jordan noticed here: https://github.com/aws-amplify/amplify-ios/issues/1107#issuecomment-868538507. May want to consult with him, as this could be the expected codegen.

HuiSF: Thanks. That should be a different issue generating `hasOne` vs `belongsTo`. This one should be isolated as not generating `QueryField` when it's `hasOne`, I think.

dnys1 (approved):
Looks good. Thanks!

PR #762, opened 2021-07-29
chore(datastore): Remove redundant files for the datastore example

HuiSF:
*Issue #, if available:*

*Description of changes:*

Remove redundant model files (brought in by nullsafety branch merging) from the DataStore example app.
New files are generated with amplify-codegen with the schema provided in the example README.


cshfang (approved):

packages/amplify_datastore/example/lib/models/Blog.dart
```diff
@@ -37,7 +37,16 @@ class Blog extends Model {
   }
 
   String get name {
-    return _name!;
+    try {
```
Jordan-Nelson, line 40: *Question:* Do you know why some of the existing models were changed? I am assuming they were generated with one of the pre-release versions of null-safety codgen and were not updated?

Jordan-Nelson (approved):
LGTM. I left one question, but that is just out of curiosity.

packages/amplify_datastore/example/lib/models/Blog.dart
HuiSF: Thanks for double checking! I think the original files became inaccurate at all due to merging processes, and yes, the original files seemed generated with in-development codegen.

PR #763, opened 2021-07-29
feat(amplify_api): GraphQL Response Decoding

Equartey:
*Issue #, if available:*

*Problem Statement:*
Developers should be able to use types specified to the Model helpers when a response is returned by the API.

*Description of changes:*
This provides a type safe experience when using the Model based helpers. The Dart compiler will recognize and enforce Model types. For example: 

Before: 
```Dart
Amplify.API.query<String>(request:  ModelQueries.get<String>(Blog.classType, id));
```
After:
```Dart
Amplify.API.query<Blog>(request: ModelQueries.get<Blog>(Blog.classType, id));
```

### DIFF
--- a/packages/amplify_api/lib/method_channel_api.dart
+++ b/packages/amplify_api/lib/method_channel_api.dart
@@ -16,6 +16,7 @@
 import 'dart:async';
 import 'dart:typed_data';
 
+import 'package:amplify_api/src/graphql/graphql_response_decoder.dart';
 import 'package:amplify_core/types/index.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter/foundation.dart';
@@ -159,8 +160,8 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
             AmplifyExceptionMessages.nullReturnedFromMethodChannel);
       final errors = _deserializeGraphQLResponseErrors(result);
 
-      GraphQLResponse<T> response =
-          GraphQLResponse<T>(data: result['data'] ?? '', errors: errors);
+      GraphQLResponse<T> response = GraphQLResponseDecoder.instance
+          .decode<T>(request: request, data: result['data'], errors: errors);
 
       return response;
     } on PlatformException catch (e) {
--- a/packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
+++ b/packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
@@ -161,7 +161,7 @@ class GraphQLRequestFactory {
     String requestName = "$requestOperationVal$name";
     // e.g. query getBlog($id: ID!, $content: String) { getBlog(id: $id, content: $content) { id name createdAt } }
     String document =
-        '''$requestTypeVal $requestOperationVal$name${documentInputs.upper} { $requestOperationVal$name${documentInputs.lower} { $fields } }''';
+        '''$requestTypeVal $requestName${documentInputs.upper} { $requestName${documentInputs.lower} { $fields } }''';
 
     // TODO: convert model to variable input for non-get operations
     Map<String, dynamic> variables =
@@ -171,6 +171,6 @@ class GraphQLRequestFactory {
         document: document,
         variables: variables,
         modelType: modelType,
-        decodePath: "$requestName");
+        decodePath: requestName);
   }
 }
--- a/packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
+++ b/packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
@@ -32,6 +32,13 @@ class GraphQLRequestFactory {
 
   static GraphQLRequestFactory get instance => _instance;
 
+  String _getName(ModelSchema schema, GraphQLRequestOperation operation) {
+    // schema has been validated & schema.pluralName is non-nullable
+    return operation == GraphQLRequestOperation.list
+        ? schema.pluralName!
+        : schema.name;
+  }
+
   String _getModelType(ModelFieldTypeEnum val) {
     switch (val) {
       case ModelFieldTypeEnum.string:
@@ -45,13 +52,19 @@ class GraphQLRequestFactory {
     }
   }
 
-  String _getFieldsFromModelType(ModelSchema schema) {
+  String _getFieldsFromModelType(
+      ModelSchema schema, GraphQLRequestOperation operation) {
     // schema has been validated & schema.fields is non-nullable
-    Map<String, ModelField> fieldsMap = schema.fields!;
-    return fieldsMap.entries
+    String fields = schema.fields!.entries
         .map((entry) => entry.value.association == null ? entry.key : '')
         .toList()
         .join(' ');
+
+    if (operation == GraphQLRequestOperation.list) {
+      fields = 'items { $fields } nextToken';
+    }
+
+    return fields;
   }
 
   ModelSchema _getAndValidateSchema(
@@ -130,31 +143,34 @@ class GraphQLRequestFactory {
       String? id,
       required GraphQLRequestType requestType,
       required GraphQLRequestOperation requestOperation}) {
+    // retrieve schema from ModelType and validate required properties
     ModelSchema schema = _getAndValidateSchema(modelType, requestOperation);
 
-    // e.g. "Blog"
-    String name = schema.name;
-    // fields to retrieve, e.g. "id name createdAt"
-    String fields = _getFieldsFromModelType(schema);
+    // e.g. "Blog" or "Blogs"
+    String name = _getName(schema, requestOperation);
     // e.g. "query"
     String requestTypeVal = describeEnum(requestType);
     // e.g. "get"
     String requestOperationVal = describeEnum(requestOperation);
-    // {upper: "($id: ID!)", lower: "(id: $id)"}
-    DocumentInputs docInputs = _buildDocumentInputs(schema, requestOperation);
-
-    if (requestOperation == GraphQLRequestOperation.list) {
-      name = schema.pluralName!;
-      fields = 'items { $fields } nextToken';
-    }
-
-    String doc =
-        '''$requestTypeVal $requestOperationVal$name${docInputs.upper} { $requestOperationVal$name${docInputs.lower} { $fields } }''';
+    // e.g. {upper: "($id: ID!)", lower: "(id: $id)"}
+    DocumentInputs documentInputs =
+        _buildDocumentInputs(schema, requestOperation);
+    // e.g. "id name createdAt" - fields to retrieve
+    String fields = _getFieldsFromModelType(schema, requestOperation);
+    // e.g. "getBlog"
+    String requestName = "$requestOperationVal$name";
+    // e.g. query getBlog($id: ID!, $content: String) { getBlog(id: $id, content: $content) { id name createdAt } }
+    String document =
+        '''$requestTypeVal $requestOperationVal$name${documentInputs.upper} { $requestOperationVal$name${documentInputs.lower} { $fields } }''';
 
     // TODO: convert model to variable input for non-get operations
     Map<String, dynamic> variables =
         requestOperation == GraphQLRequestOperation.get ? {"id": id} : {};
 
-    return GraphQLRequest<T>(document: doc, variables: variables);
+    return GraphQLRequest<T>(
+        document: document,
+        variables: variables,
+        modelType: modelType,
+        decodePath: "$requestName");
   }
 }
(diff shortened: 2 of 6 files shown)


packages/amplify_api/lib/src/graphql/graphql_response_decoder.dart
```diff
@@ -0,0 +1,37 @@
...
+
+  static GraphQLResponseDecoder get instance => _instance;
+
+  GraphQLResponse<T> decode<T>(
+      {required GraphQLRequest request,
+      required String data,
+      required List<GraphQLResponseError> errors}) {
+    // if no ModelType fallback to default
+    if (request.modelType == null) {
+      return GraphQLResponse<T>(data: data as T, errors: errors);
```
dnys1, line 20: This cast will fail for everything but T == String. Probably better to write this piece as

```dart
if (request.modelType == null) {
  if (T == String) {
    return GraphQLResponse(data: data as T, errors: errors); // <T> is implied
  } else {
    // handle case (probably throw)
  }
}
```

packages/amplify_api/lib/src/graphql/graphql_response_decoder.dart
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    if (request.decodePath == null) {
+      throw ApiException('No decodePath found',
+          recoverySuggestion: 'Include decodePath when creating a request');
+    }
+
+    Map<String, dynamic> dataJson = json.decode(data);
```
dnys1, line 28: A little tidbit about Dart and something to watch out for. There is an implicit cast happening here, which can throw in certain situations, if you're not careful. `json.decode` returns a `dynamic` which, in Dart, means it could be any type, even `null`. This statement here is the same as writing:

```dart
dynamic decoded = json.decode(data);
Map<String, dynamic> data = decoded as Map<String, dynamic>;
```

So, you can see the problem. `decoded` can be *any* JSON value (null, String, Map, etc.) and if it's not a Map<String, dynamic>, force casting with `as` will throw. There is an analyzer flag in dart which warns you about this because it can really bite you, but we don't have it on currently in the repo. I would recommend turning it on every once in a while developing because it will help you write safer, more robust code. Although, as you'll see, it's very strict so it's probably best to just use it sparingly and when you're curious. To turn it on, just add the following to an `analysis_options.yaml` file in either the project root or package root:

```yaml
analyzer:
  strong-mode:
    implicit-dynamic: false
    implicit-casts: false
```

With that turned on, you'll be *gently reminded* about this haha

Screen Shot 2021-07-30 at 8 11 10 AM

packages/amplify_api/test/amplify_api_graphql_helpers_test.dart
```diff
@@ -15,15 +15,23 @@
...
-  test("ModelQueries.get() should craft a valid request", () {
-    AmplifyAPI api = AmplifyAPI(modelProvider: ModelProvider.instance);
+  AmplifyAPI? api;
+  setUp(() {
+    api = AmplifyAPI(modelProvider: ModelProvider.instance);
+  });
 
+  tearDown(() {
+    api = null;
+  });
```
dnys1, line 32: This block is not used by the tests (can be removed). If you do need an instance for these tests, I would just declare it once, since setup/teardown should be used more for like initialization/de-initialization logic which needs to be run before/after tests like if you're resetting properties on an object or if initializing an object is asynchronous. Here, the instance won't change between tests, though. And as you do later, you create a separate local instance for that specific test.

Another way you could do this is with groups, since you have a clear distinction of using `AmplifyAPI` with a model provider and without.

```dart
void main() {
  group('with ModelProvider', () {
    final api = AmplifyAPI(modelProvider: ModelProvider.instance);

    test('...', () {});
    test('...', () {});
    test('...', () {});
  });

  group('without ModelProvider', () {
    final api = AmplifyAPI();

    test('...', () {});
    test('...', () {});
    test('...', () {});
  });
}
```

packages/amplify_api/test/amplify_api_graphql_helpers_test.dart
```diff
@@ -46,4 +57,50 @@ void main() {
...
+            "createdAt": "2021-01-01T01:00:00.000000000Z",
+            "id": "$id",
+            "name": "TestAppBlog"
+        }
+    }''';
+
+    GraphQLResponse<Blog> response = GraphQLResponseDecoder.instance
+        .decode<Blog>(request: req, data: data, errors: errors);
+
+    expect(response.data.runtimeType, Blog);
```
dnys1, line 77:
```dart
expect(response.data, isA<Blog>());
```

This is guaranteed by the type contract, though, so it will always be true. What we want to test here is probably that is was decoded properly, so you can create the Blog object it's supposed to be and compare those, instead.

packages/amplify_api/test/amplify_api_graphql_helpers_test.dart
```diff
@@ -46,4 +57,50 @@ void main() {
...
+            "createdAt": "2021-01-01T01:00:00.000000000Z",
+            "id": "$id",
+            "name": "TestAppBlog"
+        }
+    }''';
+
+    GraphQLResponse<String> response = GraphQLResponseDecoder.instance
+        .decode<String>(request: req, data: data, errors: errors);
+
+    expect(response.data.runtimeType, String);
```
dnys1, line 104: Same as above. Better to compare the value than the type, since it's already guaranteed.

packages/amplify_api_plugin_interface/lib/src/GraphQL/GraphQLRequest.dart
```diff
@@ -13,14 +13,22 @@
...
+  String cancelToken = API_UUID.UUID.getUUID();
+  String? decodePath;
+  ModelType? modelType;
 
-  GraphQLRequest({required this.document, this.variables = const {}});
+  GraphQLRequest(
+      {required this.document,
+      this.variables = const {},
+      this.decodePath = null,
+      this.modelType = null}) {}
```
dnys1, line 31: Can leave out the `= null` as a default value, since it's implied.

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
```diff
@@ -130,31 +143,34 @@ class GraphQLRequestFactory {
...
     // TODO: convert model to variable input for non-get operations
     Map<String, dynamic> variables =
         requestOperation == GraphQLRequestOperation.get ? {"id": id} : {};
 
-    return GraphQLRequest<T>(document: doc, variables: variables);
+    return GraphQLRequest<T>(
+        document: document,
+        variables: variables,
+        modelType: modelType,
+        decodePath: "$requestName");
```
dnys1, line 174: Since `requestName` is already a String, you can just write `decodePath: requestName`.

dnys1 (approved):
Looks good! A few minor comments, but overall, the structure and logic is good 👍

packages/amplify_api/lib/src/graphql/graphql_response_decoder.dart
```diff
@@ -0,0 +1,37 @@
...
+      required String data,
+      required List<GraphQLResponseError> errors}) {
+    // if no ModelType fallback to default
+    if (request.modelType == null) {
+      return GraphQLResponse<T>(data: data as T, errors: errors);
+    }
+
+    if (request.decodePath == null) {
+      throw ApiException('No decodePath found',
+          recoverySuggestion: 'Include decodePath when creating a request');
```
ragingsquirrel3, line 25: nit: I suggest adding changing this recovery suggestion to 'Include decodePath when creating a request that includes a modelType.'

packages/amplify_api/lib/src/graphql/graphql_response_decoder.dart
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    Map<String, dynamic> dataJson = json.decode(data);
+    request.decodePath!.split(".").forEach((element) {
+      dataJson = dataJson[element];
```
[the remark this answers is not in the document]
ragingsquirrel3: +1 for adding some validation here on `decodePath` (beyond nullability of the string itself), see below comment about how decodePath and modelType are public parameters that might be supplied by a user without using the query helpers.

One suggestion might be just to check

```
if (dataJson[element] == null) {
  throw ApiException('decodePath did not match the structure of the JSON response',
          recoverySuggestion: 'Ensure the decodePath can be used to traverse the JSON response.');
```

packages/amplify_api_plugin_interface/lib/src/GraphQL/GraphQLRequest.dart
```diff
@@ -13,14 +13,22 @@
...
+import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart';
+
+import '../UUID.dart' as API_UUID;
 
 class GraphQLRequest<T> {
   String document;
   Map<String, dynamic> variables;
-  String cancelToken = UUID.getUUID();
+  String cancelToken = API_UUID.UUID.getUUID();
+  String? decodePath;
```
ragingsquirrel3, line 24: Something to think about is that `decodePath` and `modelType` are public-facing additions to the API. Even though they are designed to be supplied by the query helpers and users should not have to think about them, it is entirely possible for a user to manually populate a request and supply these without the helper methods. In that situation, we should at least understand how the code responds and aim to supply helpful error messages even if this is not something we encourage users to do.

No change needed in this file (although documenting these params is something to think about in the future). This is just something to consider during decoding.

packages/amplify_api/lib/method_channel_api.dart
```diff
@@ -159,8 +160,8 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
             AmplifyExceptionMessages.nullReturnedFromMethodChannel);
       final errors = _deserializeGraphQLResponseErrors(result);
 
-      GraphQLResponse<T> response =
-          GraphQLResponse<T>(data: result['data'] ?? '', errors: errors);
+      GraphQLResponse<T> response = GraphQLResponseDecoder.instance
```
ragingsquirrel3, line 163: I don't see any tests for the decoder's integration in the method channel. If you look at amplify_api_query_test.dart you will see testing for how the api plugin itself responds in certain situations (given that the JSON response [or error] from method channels is mocked). This would be a good place to add at least 1 test for this. You could pretty much copy/paste the first test, but produce the request from your helper, then test that the response is decoded, similar to existing test, just that this would ensure the methods work as expected in the plugin itself, rather than when calling internal helpers directly.

packages/amplify_api/test/amplify_api_graphql_helpers_test.dart
```diff
@@ -46,4 +57,50 @@ void main() {
...
+    List<GraphQLResponseError> errors = [];
+    String data = '''{
+        "getBlog": {
+            "createdAt": "2021-01-01T01:00:00.000000000Z",
+            "id": "$id",
+            "name": "TestAppBlog"
+        }
+    }''';
+
+    GraphQLResponse<String> response = GraphQLResponseDecoder.instance
```
ragingsquirrel3, line 101: Nice job testing this! More context to my above comment about adding the test in other file, this tests your internal helpers and would be good IMO to test this behavior on the plugin itself.

ragingsquirrel3 (changes requested):
Overall, this is very slick. Great job! This is reaping the rewards of good design and solid foundations. I would like to see at least 1 more test for the decoding behavior on the plugin itself as noted, as well as a little validation on the `decodePath`. All the other comments are non-actionable in this PR, just stuff to think about.

packages/amplify_api/lib/src/graphql/graphql_response_decoder.dart
Equartey: Added check and threw an error.

packages/amplify_api/lib/src/graphql/graphql_response_decoder.dart
Equartey: Very neat. I took it for a test drive and it lead me to create another null check. Thanks for the tip!

packages/amplify_api/test/amplify_api_graphql_helpers_test.dart
Equartey: Awesome, I love this.

packages/amplify_api/test/amplify_api_graphql_helpers_test.dart
Equartey: Thank you!! I could not find how to do this on google, much better 🥳

packages/amplify_api/lib/src/graphql/graphql_response_decoder.dart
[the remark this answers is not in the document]
Equartey: Great feedback. We really don't use this much yet so I considered not including it, but I think these changes make it viable for use in the future. Thanks!

ragingsquirrel3 (approved):
🚢  it (assuming remaining tests pass). Great work on that test in last commit.

PR #764, opened 2021-07-30
bug(analytics): Force session start

dnys1:
*Issue #, if available:*
- Internal ticket: P50379527
- Summary of issue:

The [AutoSessionTracker](https://github.com/aws-amplify/amplify-android/blob/main/aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AutoSessionTracker.java) in the Pinpoint plugin listens for lifecycle changes and starts and stops session tracking accordingly. It is registered during the call to `Amplify.configure`. In native Android apps, this call is made before launching the main activity and thus receives the initial onResume event, kicking off session tracking. However, in Flutter, the call to `Amplify.configure` is made sometime later. The initial onResume event is not received by the AutoSessionTracker, and no session tracking occurs.

The startSession/stopSession calls made by the AutoSessionTracker are not available via the escape hatch, the AWS SDK, or reflection, so an onPause/onResume cycle must be forced.

This method is invoked in the Flutter SDK just after `Amplify.configure`.

*Description of changes:*

- Force session start after configuration

### DIFF
--- a/packages/amplify_analytics_pinpoint/android/src/main/kotlin/com/amazonaws/amplify/amplify_analytics_pinpoint/AmplifyAnalyticsPinpointPlugin.kt
+++ b/packages/amplify_analytics_pinpoint/android/src/main/kotlin/com/amazonaws/amplify/amplify_analytics_pinpoint/AmplifyAnalyticsPinpointPlugin.kt
@@ -56,7 +56,9 @@ class AmplifyAnalyticsPinpointPlugin : FlutterPlugin, ActivityAware, MethodCallH
             "addPlugin" ->
                 AmplifyAnalyticsBridge.addPlugin(result, context)
             "startSession" -> {
-                // Hack: The AutoSessionTracker in the Pinpoint plugin listens for lifecycle changes and
+                // TODO: Update AutoSessionTracker logic to support Flutter.
+                //
+                // The AutoSessionTracker in the Pinpoint plugin listens for lifecycle changes and
                 // starts and stops session tracking accordingly. It is registered during the call
                 // to Amplify.configure. In native Android apps, this call is made before launching
                 // the main activity and thus receives the initial onResume event, kicking off session
--- a/packages/amplify_analytics_pinpoint/android/src/main/kotlin/com/amazonaws/amplify/amplify_analytics_pinpoint/AmplifyAnalyticsPinpointPlugin.kt
+++ b/packages/amplify_analytics_pinpoint/android/src/main/kotlin/com/amazonaws/amplify/amplify_analytics_pinpoint/AmplifyAnalyticsPinpointPlugin.kt
@@ -16,11 +16,9 @@
 package com.amazonaws.amplify.amplify_analytics_pinpoint
 
 import android.app.Activity
-import android.app.Application
 import android.content.Context
-import android.util.Log
+import android.content.Intent
 import androidx.annotation.NonNull
-import com.amplifyframework.analytics.pinpoint.AWSPinpointAnalyticsPlugin
 import com.amplifyframework.core.Amplify
 import io.flutter.embedding.engine.plugins.FlutterPlugin
 import io.flutter.embedding.engine.plugins.activity.ActivityAware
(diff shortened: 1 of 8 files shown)


packages/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
```diff
@@ -77,4 +77,11 @@ class AmplifyAnalyticsPinpoint extends AnalyticsPluginInterface {
       required AnalyticsUserProfile userProfile}) async {
     return _instance.identifyUser(userId: userId, userProfile: userProfile);
   }
+
+  // Internal
+
```
fjnoyp, line 82: is this new line needed?

packages/amplify_analytics_plugin_interface/lib/amplify_analytics_plugin_interface.dart
```diff
@@ -63,4 +63,8 @@ abstract class AnalyticsPluginInterface extends AmplifyPluginInterface {
       required AnalyticsUserProfile userProfile}) async {
     throw UnimplementedError('identifyUser() has not been implemented.');
   }
+
+  // Internal
+
+  Future<void> onConfigure() async {}
```
HuiSF, line 69: Would that be better to convert the `// internal` to the method document to explicitly call out that this API is internally used by amplify-flutter library?

packages/amplify_flutter/lib/method_channel_amplify.dart
```diff
@@ -22,13 +22,19 @@ class MethodChannelAmplify extends AmplifyClass {
...
+  Future<bool?> _configurePlatforms(
+      String version, String configuration) async {
+    final configured = await _channel.invokeMethod<bool>(
       'configure',
       <String, Object>{
         'version': version,
         'configuration': configuration,
       },
     );
+    if (configured ?? false) {
```
fjnoyp, line 34: how do we know this statement is true for analytics plugin case?

fjnoyp (commented):
looks good just one minor question

haverchuck (approved):

packages/amplify_analytics_pinpoint/android/src/main/kotlin/com/amazonaws/amplify/amplify_analytics_pinpoint/AmplifyAnalyticsPinpointPlugin.kt
```diff
@@ -67,6 +55,24 @@ public class AmplifyAnalyticsPinpointPlugin : FlutterPlugin, ActivityAware, Meth
         when (call.method) {
             "addPlugin" ->
                 AmplifyAnalyticsBridge.addPlugin(result, context)
+            "startSession" -> {
+                // Hack: The AutoSessionTracker in the Pinpoint plugin listens for lifecycle changes and
```
haverchuck, line 59: We shoud change this terminlogy to say with a 'TODO' what a better solution would entail

packages/amplify_flutter/lib/method_channel_amplify.dart
dnys1: It's the result of this call here. It will always be true, right now.

https://github.com/aws-amplify/amplify-flutter/blob/ed16e42a3ab802e20ecdaeb4eb325cae704ce60c/packages/amplify_flutter/android/src/main/kotlin/com/amazonaws/amplify/Amplify.kt#L121

packages/amplify_analytics_pinpoint/android/src/main/kotlin/com/amazonaws/amplify/amplify_analytics_pinpoint/AmplifyAnalyticsPinpointPlugin.kt
dnys1: Fixed

packages/amplify_analytics_plugin_interface/lib/amplify_analytics_plugin_interface.dart
dnys1: Marked as protected

HuiSF (approved):

PR #768, opened 2021-07-31
feat(amplify_authenticator): text customization

haverchuck:
*Description of changes:*
Allows localiztion for all static text.

### DIFF
--- a/packages/amplify_authenticator/lib/amplify_authenticator.dart
+++ b/packages/amplify_authenticator/lib/amplify_authenticator.dart
@@ -240,6 +240,7 @@ class _AuthenticatorState extends State<Authenticator> {
 
   @override
   Widget build(BuildContext context) {
+    widget.authStrings.setContext(context);
     return InheritedAuthBloc(
         key: const Key(keyInheritedAuthBloc),
         authBloc: _stateMachineBloc,
--- a/packages/amplify_authenticator/lib/amplify_authenticator.dart
+++ b/packages/amplify_authenticator/lib/amplify_authenticator.dart
@@ -17,6 +17,7 @@ library amplify_authenticator;
 
 import 'package:amplify_authenticator/src/keys.dart';
 import 'package:amplify_authenticator/src/screens/confirm_signin_screen.dart';
+import 'package:amplify_authenticator/src/strings/auth_strings.dart';
 import 'package:flutter/material.dart';
 
 //State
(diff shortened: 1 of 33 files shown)


packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -250,13 +249,12 @@ class _AuthenticatorState extends State<Authenticator> {
             signUpViewModel: SignUpViewModel(_stateMachineBloc),
             confirmSignUpViewModel: ConfirmSignUpViewModel(_stateMachineBloc),
             confirmSignInViewModel: ConfirmSignInViewModel(_stateMachineBloc),
-
             child: InheritedForms(
+                authStrings: widget.authStrings,
```
israx, line 254: Shouldn't this "authStrings" argument be part of another InheritedWidget?

It could be `InheritedAuthStrings(authStrings: widget.authStrings, child: InheritedForms)`

packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -84,20 +88,18 @@ class Authenticator extends StatefulWidget {
...
   Authenticator(
-
       {required this.child,
       SignInForm? signInForm,
       SignUpForm? signUpForm,
-      ConfirmSignInForm? confirmSignInForm}) {
-    this.signInForm = signInForm ?? DefaultForms.signInForm();
-    this.signUpForm = signUpForm ?? DefaultForms.signUpForm();
+      ConfirmSignInForm? confirmSignInForm,
+      AuthStringResolver? resolver}) {
```
israx, line 95: Wondering, would it be better if we have a default English resolver? So in that way in the future, we may want to store default resolvers in different languages.

packages/amplify_authenticator/lib/src/strings/navigation_strings.dart
```diff
@@ -0,0 +1,54 @@
+import 'package:amplify_authenticator/src/text_customization/auth_strings_resolver.dart';
+import 'package:amplify_authenticator/src/text_customization/button_resolver.dart';
+
+/// The class for displaying navigation related labels
+class NavigationStrings {
+  /// The optional class that customers can pass to create custom values
+  NavigationResolver? resolver;
+
+  static const _navigate_signin = "Sign In";
```
israx, line 23: As I said before, I think these values should come from a default English resolver.

packages/amplify_authenticator/lib/amplify_authenticator.dart
haverchuck: I don't think that is totally necessary - the customer can handle that through Flutter's OOTB internationaliztion functionality.

packages/amplify_authenticator/lib/amplify_authenticator.dart
haverchuck: The problem with doing this, from what I'm seeing, is the timing for instantiating the defaultForms vs the rest of the regular widgets.

packages/amplify_authenticator/lib/src/strings/navigation_strings.dart
haverchuck: As we discussed, we can hold off on this.

israx (approved):
Looks good to me.

PR #771, opened 2021-08-02
feat(amplify_api): GraphQL Query Helper - List 

Equartey:
*Issue #, if available:*

*Description of changes:*
This adds `ModelQueries.list()`, allows for proper response decoding of PaginatedResults, and relevant unit tests.

### DIFF
--- a/packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
+++ b/packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
@@ -101,19 +101,13 @@ class GraphQLRequestFactory {
   }
 
   DocumentInputs _buildDocumentInputs(
-      ModelSchema schema, GraphQLRequestOperation operation,
-      [Map<String, dynamic>? documentArgs]) {
+      ModelSchema schema, GraphQLRequestOperation operation) {
     String upperOutput = '';
     String lowerOutput = '';
     List<String> upperList = [];
     List<String> lowerList = [];
 
-    if (documentArgs != null) {
-      // e.g. "limit: 10"
-      documentArgs.forEach((key, value) {
-        lowerList.add("$key: $value");
-      });
-    }
+    String modelName = schema.name;
 
     // build inputs based on request operation
     switch (operation) {
@@ -123,9 +117,10 @@ class GraphQLRequestFactory {
         lowerOutput = "(${lowerList.join(", ")})";
         break;
       case GraphQLRequestOperation.list:
-        if (lowerList.length != 0) {
-          lowerOutput = "(${lowerList.join(", ")})";
-        }
+        upperOutput =
+            "(\$filter: Model${modelName}FilterInput, \$limit: Int, \$nextToken: String)";
+        lowerOutput =
+            r"(filter: $filter, limit: $limit, nextToken: $nextToken)";
         break;
       default:
         schema.fields!.forEach((field, val) {
@@ -149,10 +144,9 @@ class GraphQLRequestFactory {
   GraphQLRequest<T> buildQuery<T extends Model>(
       {required ModelType modelType,
       Model? model,
-      String? id,
       required GraphQLRequestType requestType,
       required GraphQLRequestOperation requestOperation,
-      Map<String, dynamic>? documentArgs}) {
+      required Map<String, dynamic> variables}) {
     // retrieve schema from ModelType and validate required properties
     ModelSchema schema = _getAndValidateSchema(modelType, requestOperation);
 
@@ -164,7 +158,7 @@ class GraphQLRequestFactory {
     String requestOperationVal = describeEnum(requestOperation);
     // e.g. {upper: "($id: ID!)", lower: "(id: $id)"}
     DocumentInputs documentInputs =
-        _buildDocumentInputs(schema, requestOperation, documentArgs);
+        _buildDocumentInputs(schema, requestOperation);
     // e.g. "id name createdAt" - fields to retrieve
     String fields = _getFieldsFromModelType(schema, requestOperation);
     // e.g. "getBlog"
@@ -174,10 +168,6 @@ class GraphQLRequestFactory {
         '''$requestTypeVal $requestName${documentInputs.upper} { $requestName${documentInputs.lower} { $fields } }''';
     // e.g "listBlogs"
     String decodePath = requestName;
-
-    // TODO: convert model to variable input for non-get operations
-    Map<String, dynamic> variables =
-        requestOperation == GraphQLRequestOperation.get ? {"id": id} : {};
 
     return GraphQLRequest<T>(
         document: document,
--- a/packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
+++ b/packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
@@ -101,11 +101,18 @@ class GraphQLRequestFactory {
   }
 
   DocumentInputs _buildDocumentInputs(
-      ModelSchema schema, GraphQLRequestOperation operation) {
+      ModelSchema schema, GraphQLRequestOperation operation,
+      [Map<String, dynamic>? documentArgs]) {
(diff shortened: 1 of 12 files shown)


packages/amplify_api/lib/src/graphql/paginated_modelType_impl.dart
```diff
@@ -0,0 +1,23 @@
...
+  PaginatedModelTypeImpl(this.modelType) : super(modelType);
+
+  @override
+  PaginatedResultImpl<T> fromJson(Map<String, dynamic> jsonData) {
+    List<dynamic>? dataList = List<dynamic>.from(jsonData['items']);
+
+    List<T> decodedData = dataList.map((e) => modelType.fromJson(e)).toList();
+
+    return PaginatedResultImpl<T>(decodedData, jsonData['nextToken']);
+  }
```
dnys1, line 17: There are a few implicit casts throughout here which we need to be careful of. Here's how I would do it to avoid null unwraps or bad casts:

```dart
  @override
  PaginatedResultImpl<T> fromJson(Map<String, dynamic> jsonData) {
    final itemsJson = jsonData['items'] as List?;

    if (itemsJson == null || itemsJson.isEmpty) {
      return const PaginatedResultImpl<T>([], null);
    }

    final items = itemsJson
        .cast<Map>()
        .map(
          (e) => modelType.fromJson(e.cast()),
        )
        .toList();

    return PaginatedResultImpl<T>(items, jsonData['nextToken'] as String?);
  }
```

packages/amplify_api/lib/src/graphql/paginated_result_impl.dart
```diff
@@ -0,0 +1,40 @@
...
+  @override
+  ModelType<Model> getInstanceType() {
+    ModelType<Model>? modelType = AmplifyAPI.instance.modelProvider
+        ?.getModelTypeByModelName(T.toString());
+    return PaginatedModelTypeImpl(modelType!);
+  }
+
+  @override
+  Map<String, dynamic> toJson() {
+    return {};
```
dnys1, line 38: While not required, let's add the JSON representation anyway

```dart
@override
Map<String, dynamic> toJson() => {
      'items': items,
      'nextToken': nextToken,
    };
```

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
```diff
@@ -101,12 +101,19 @@ class GraphQLRequestFactory {
...
     String upperOutput = '';
     String lowerOutput = '';
     List<String> upperList = [];
     List<String> lowerList = [];
 
+    if (documentArgs != null) {
+      documentArgs.forEach((key, value) {
+        lowerList.add("$key: $value");
+      });
+    }
```
dnys1, line 115: Let's remove this and use what you have for the `default` impl below. Ideally, the document we're building here needs no knowledge of the variables.

packages/amplify_api/lib/src/graphql/model_queries_factory.dart
```diff
@@ -37,10 +38,19 @@ class ModelQueriesFactory extends ModelQueriesInterface {
 
   @override
   GraphQLRequest<PaginatedResult<T>> list<T extends Model>(
-      ModelType<T> modelType,
-      {QueryPredicate? where,
-      ModelPagination? modelPagination}) {
-    // TODO: implement list
-    throw UnimplementedError();
+    ModelType<T> modelType, {
+    ModelPagination? modelPagination = const ModelPagination(),
```
dnys1, line 42: Actually, after considering it some more, I think the DX is good here 👍

packages/amplify_api/lib/src/graphql/model_queries_factory.dart
```diff
@@ -37,10 +38,19 @@ class ModelQueriesFactory extends ModelQueriesInterface {
...
+    QueryPredicate? where,
+  }) {
+    Map<String, dynamic> documentArgs = {};
+    if (modelPagination != null) {
+      documentArgs = {"limit": modelPagination.limit};
```
dnys1, line 47: Will need the `nextToken` as well, at some point. Maybe add a `// TODO` for now

dnys1 (commented):
Looking pretty snazzy! 😄 

General comments: 
- Let's make all constructors `const` and fields `final` when possible (a good rule of thumb)

I'm also going to recommend changing gears a bit and removing `documentArgs` from the document build step. Ideally, when building the document, we just look at the schema to see what it accepts for input. We're constructing a generic document, essentially, which can be used for any set of variable values (and this will help with caching later, too). The values for those inputs are passed separately in the GraphQL request, and then the server looks at both to do validation and build the actual query.

Example
---

A user requests a `listTodo` operation.

Schema:
```graphql
type Todo @model {
  id: ID!
  title: String!
  description: String
}

type Query {
  listTodos(filter: ModelTodoFilterInput, limit: Int, nextToken: String): ModelTodoConnection
}
```

Document we build:
```graphql
query listTodos($filter: ModelTodoFilterInput, $limit: Int, $nextToken: String) {
  listTodos(filter: $filter, limit: $limit, nextToken: $nextToken) {
    nextToken
    items {
      id
      title
      description
    }
  }
}
```

Variables (JSON) - nullable variables can be omitted or included as `null`:
```json
{
  "limit": 100,
  "nextToken": null
}
```

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
ragingsquirrel3: +1 . I'm a little confused how the `documentArgs` will support this (the pagination parameters) as well as query params and other variables. When we add the query params, it might make it a little easier to handle this param correctly. If you wanted, you could even remove the whole `documentArgs` parameter in this PR and just pass the limit as a primitive parameter very specific to list case, similar to the way you added id as primitive param in the get helper.

packages/amplify_api/lib/src/graphql/paginated_result_impl.dart
```diff
@@ -0,0 +1,40 @@
...
+  String getId() {
+    return '';
+  }
+
+  @override
+  ModelType<Model> getInstanceType() {
+    ModelType<Model>? modelType = AmplifyAPI.instance.modelProvider
+        ?.getModelTypeByModelName(T.toString());
+    return PaginatedModelTypeImpl(modelType!);
+  }
```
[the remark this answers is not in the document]
Equartey: Thanks for the suggestion. I added a check with your revisions.

packages/amplify_api/lib/src/graphql/paginated_modelType_impl.dart
Equartey: Slick! Thanks for the tip.

dnys1 (approved):
Booya 😎

ragingsquirrel3 (approved):
🚢  it ! Great job.

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
Equartey: Thanks for the clarification, this section has been refactored!

rahul-readwhere:
Hello ,
I am just curious to know, you mentioned ModelTodoFilterInput in your doc. But I am not able to get this in my schema for any model. Am I missing any update?

ragingsquirrel3:
@rahul-readwhere were you not able to find the ModelTodoFilterInput in your local schema or in the version you see in the AppSync Console? For a small example I have, I noticed that the ModelTodoFilterInput is in the the schema on AWS AppSync Console under AWS AppSync > [backendName] > Schema. For my case, this was created by Amplify when setting up the API resource in my app, which I created from a very basic schema `type Blog @model { id: ID!\n name: String!\n createdAt: AWSDate }`. More info about that can be found in https://docs.aws.amazon.com/appsync/latest/devguide/provision-from-schema.html

PR #775, opened 2021-08-03
feat(auth): add options to updateAttribute, updateAttributes, resendUserAttributeConfirmationCode

Jordan-Nelson:
*Issue #, if available:* N/A

*Description of changes:*
- Mark AuthUserAttributeRequest/AuthUserAttributeOptions as deprecated, and create FetchUserAttributesRequest/FetchUserAttributesOptions
- Add options/metadata to updateAttribute, updateAttributes, resendUserAttributeConfirmationCode


packages/amplify_auth_cognito/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/types/FlutterUpdateUserAttributeRequest.kt
```diff
@@ -21,18 +21,28 @@ import com.amazonaws.amplify.amplify_auth_cognito.utils.validateUserAttribute
...
+import com.amplifyframework.auth.cognito.options.AWSCognitoAuthUpdateUserAttributeOptions
 
 data class FlutterUpdateUserAttributeRequest(val map: HashMap<String, *>) {
 
-    val attribute: AuthUserAttribute = formatUpdateUserAttribute(map["attribute"] as HashMap<String, *>);
+    val attribute: AuthUserAttribute = createAttribute(map["attribute"] as HashMap<String, *>);
+    val options: AWSCognitoAuthUpdateUserAttributeOptions = createOptions(map["options"] as HashMap<String, Any>?)
 
-    private fun formatUpdateUserAttribute(@NonNull rawAttribute: HashMap<String, *>): AuthUserAttribute {
+    private fun createAttribute(@NonNull rawAttribute: HashMap<String, *>): AuthUserAttribute {
```
Jordan-Nelson, line 31: *Note:* This private method (and several others) were renamed for consistency.

packages/amplify_auth_plugin_interface/lib/src/Attribute/FetchUserAttributesOptions.dart
```diff
@@ -0,0 +1,30 @@
...
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+@Deprecated("Please use 'FetchUserAttributesOptions' instead")
+class AuthUserAttributeOptions extends FetchUserAttributesOptions {}
```
Jordan-Nelson, line 17: *Note:* This class was renamed for consistency. It is part of the public API. There is no Cognito implementation/extension of it. The only purpose it serves right now is to allow a customer that is creating a custom auth plugin (instead of using Cognito auth) to extend it to allow for additional options in their implementation of this fetchUserAttributes.

Dart 2.12 and lower doesn't have type aliasing, so this seemed to be the most straightforward way to deprecate it.

packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthCognitoPluginTest.kt
```diff
@@ -21,18 +21,13 @@ import com.amazonaws.amplify.amplify_core.exception.InvalidRequestException
...
 import com.amplifyframework.auth.cognito.AWSCognitoAuthSession
 import com.amplifyframework.auth.cognito.AWSCognitoUserPoolTokens
-import com.amplifyframework.auth.cognito.options.AWSCognitoAuthConfirmSignInOptions
-import com.amplifyframework.auth.cognito.options.AWSCognitoAuthConfirmSignUpOptions
-import com.amplifyframework.auth.cognito.options.AWSCognitoAuthSignInOptions
-import com.amplifyframework.auth.options.AuthConfirmSignInOptions
-import com.amplifyframework.auth.options.AuthConfirmSignUpOptions
-import com.amplifyframework.auth.options.AuthSignInOptions
+import com.amplifyframework.auth.cognito.options.*
+import com.amplifyframework.auth.options.*
```
haverchuck, line 30: I think we've been trying to avoid these star imports... I believe there is a setting in Android Studio that you can turn off to prevent it from happening automatically.

haverchuck (approved):
Left a couple of minor comments, but otherwise LGTM. One note on the renaming of `ConfirmUserAttributeResult.dart`: we should probably make clear in our docs that customers should not import and directly use our interface packages. If for what ever reason they are doing so and referencing this file, they would break.

Jordan-Nelson:
> we should probably make clear in our docs that customers should not import and directly use our interface packages

@haverchuck Since not importing directly from `src/` is an official dart recommendation and importing anything directly from `src/` will generate a lint warning with default dart lint setting, I am not sure that we really need to call that out.

There are some things not under `src/` that probably do belong there, but that class is under `src/`.

packages/amplify_auth_plugin_interface/lib/src/Attribute/AuthUserAttributeOptions.dart
```diff
@@ -12,10 +12,8 @@
...
-  const AuthUserAttributeOptions();
-  Map<String, dynamic> serializeAsMap() {
-    throw UnimplementedError(
-        'serializeAsMap() has not been implemented on AuthUserAttributeOptions.');
-  }
-}
+
+import 'FetchUserAttributesOptions.dart';
+
+@Deprecated("Please use 'FetchUserAttributesOptions' instead")
```
HuiSF, line 18: nit: should use backtick to wrap class name?

packages/amplify_auth_plugin_interface/lib/src/Attribute/FetchUserAttributesOptions.dart
```diff
@@ -13,16 +13,15 @@
...
 
-import './AuthUserAttributeOptions.dart';
+/// The shared fetch user attributes options among all Auth plugins
+class FetchUserAttributesOptions {
+  /// Default constructor
 
-class AuthUserAttributeRequest {
-  AuthUserAttributeOptions? options;
-  AuthUserAttributeRequest({this.options});
+  const FetchUserAttributesOptions();
```
HuiSF, line 20: nit: empty line 19 can be removed

packages/amplify_auth_plugin_interface/lib/src/Attribute/AuthUserAttributeOptions.dart
Jordan-Nelson: I mentioned this on another PR, but I am going to update these to follow what seems to be the pattern in the flutter/flutter repo, which is to not wrap functions/types in anything.

Jordan-Nelson:
```suggestion
@Deprecated('Please use FetchUserAttributesOptions instead')
```

HuiSF (approved):

PR #776, opened 2021-08-03
Amplify authenticator

israx:
*Issue #, if available:*

There is an overflow in some exceptions that exceed the screen size

*Description of changes:*
- Reset password screen.
- Confirm sign in with new password screen. 
- End user is able to see exceptions on the screen
- Username alias 


packages/amplify_authenticator/example/integration_test/finders/signin_finder.dart
```diff
@@ -3,12 +3,12 @@ import 'package:flutter_test/flutter_test.dart';
 import 'package:amplify_authenticator/src/keys.dart';
 
 final usernameSignInFormFieldFinder =
```
haverchuck, line 5: Let's add comments to explain the usage of these finders.

packages/amplify_authenticator/example/lib/main.dart
```diff
@@ -53,6 +52,7 @@ class _MyHomePageState extends State<MyHomePage> {
   @override
   Widget build(BuildContext context) {
     return Authenticator(
+      usernameAlias: Alias.username,
```
haverchuck, line 55: Can the customer accomplish this by overriding the default signin form?

packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -37,38 +42,43 @@ import 'package:amplify_authenticator/src/blocs/auth/auth_bloc.dart';
...
 class Authenticator extends StatefulWidget {
   ///# Amplify Authenticator
   ///
   /// A widget that allows customers to authenticate their apps.
   ///
-  ///This wrapper requires a child widget to be passed as an argument.
-  ///To work with custom forms there is the need to pass up to 3 additional
-  ///arguments, signInForm, signUpForm and confirmSignInForm with their
-  ///respective form fields.
+  ///The Authenticator widget requires two arguments, a child widget and an username alias
```
haverchuck, line 71: If we end up keeping usernameAlias, we should not make it required.

packages/amplify_authenticator/lib/src/enums/confirm_signin_types.dart
```diff
@@ -0,0 +1,26 @@
+enum ConfirmSignInType {
```
haverchuck, line 1: What are these used for? Please include documentation.

packages/amplify_authenticator/lib/src/enums/signin_step.dart
```diff
@@ -0,0 +1,5 @@
+enum SignInStep {
```
haverchuck, line 1: Please include documentation.

packages/amplify_authenticator/lib/src/screens/send_code_screen.dart
```diff
@@ -0,0 +1,14 @@
+import 'package:amplify_authenticator/src/state/inherited_forms.dart';
+import 'package:amplify_authenticator/src/widgets/containers.dart';
+import 'package:flutter/material.dart';
+
+class SendCodeScreen extends StatelessWidget {
```
haverchuck, line 5: Should this be called the ResetPasswordScreen?

packages/amplify_authenticator/lib/src/services/amplify_auth_service.dart
```diff
@@ -11,26 +10,29 @@ abstract class AuthService {
...
   Future<SignInResult> signIn(String username, String password) async {
-
     //making sure no user is logged in before logging in a new user
 
-    if (await isLoggedIn) {
-      await Amplify.Auth.signOut();
-    }
+    // if (await isLoggedIn) {
+    //   await Amplify.Auth.signOut();
+    // }
```
dnys1, line 35: Let's remove this

packages/amplify_authenticator/lib/src/views/confirm_signin_viewmodel.dart
```diff
@@ -151,7 +141,7 @@ class ConfirmSignInViewModel extends BaseViewModel {
 
     _authBloc.authEvent.add(AuthConfirmSignIn(confirm));
     await Future.any([
-      _authBloc.exceptions.first,
+      _authBloc.exceptions!.first,
       _authBloc.stream.first,
```
dnys1, line 145: We need to update this in all the VMs since each event can trigger multiple state changes now 

(swap `Authenticated` for whichever state we're waiting for)
```dart
_authBloc.stream.firstWhere((state) => state is Authenticated)
```

packages/amplify_authenticator/lib/src/views/confirm_signup_viewmodel.dart
```diff
@@ -12,44 +14,52 @@ class ConfirmSignUpViewModel extends BaseViewModel {
...
   }
 
   void setCode(String value) {
     _code = value;
   }
 
   Future<void> confirm() async {
-    if (!_formKey.currentState!.validate()) {
-      return;
-    }
```
dnys1, line 30: Thinking we probably still want this, no?

packages/amplify_authenticator/lib/src/views/confirm_signup_viewmodel.dart
```diff
@@ -12,44 +14,52 @@ class ConfirmSignUpViewModel extends BaseViewModel {
...
-      _authBloc.exceptions.first,
+      _authBloc.exceptions!.first,
       _authBloc.stream.first,
     ]);
     setBusy(false);
   }
 
+//This method is called directly from this view due to the fact that
+//its result doesn't form part of the authenticator state machine.
+  Future<void> resendSignUpCode() async {
```
dnys1, line 45: This is the best way to handle it, and we should be injecting AmplifyAuthService instead of calling it directly. Or use the auth service on the bloc.

packages/amplify_authenticator/lib/src/views/confirm_signup_viewmodel.dart
```diff
@@ -12,44 +14,52 @@ class ConfirmSignUpViewModel extends BaseViewModel {
...
+      print(e);
+      _authBloc.exceptionsSink!.add(AuthenticatorException(e.message));
+    }
+  }
+
   // Screens
 
   void goToSignIn() {
     clean();
+    _authBloc.exceptionsSink!.add(null);
```
dnys1, line 58: Prefer to create a method in `AuthBloc` called `clearException` or something which does this.

packages/amplify_authenticator/lib/src/views/signin_viewmodel.dart
```diff
@@ -46,13 +62,62 @@ class SignInViewModel extends BaseViewModel {
...
+    _authBloc.authEvent.add(AuthSendCode(sendCode));
+    await Future.any([
+      _authBloc.exceptions!.first,
+      _authBloc.stream.first,
+    ]);
+    setBusy(false);
+  }
+
+  Future<void> confirmPassword() async {
+    setBusy(true);
```
dnys1, line 81: We can still have basic validation for these - would recommend keeping 

```dart
if (!_formKey.currentState!.validate()) {
    return;
}
```

packages/amplify_authenticator/lib/src/widgets/auth_exceptions.dart
```diff
@@ -0,0 +1,59 @@
...
+        stream: _authenticatorExceptions.exceptions,
+        builder: (context, snapshot) {
+          final exception = snapshot.data;
+
+          if (exception != null) {
+            return Container(
+              margin: const EdgeInsets.only(bottom: 20),
+              width: double.infinity,
+              height: 53,
+              padding: const EdgeInsets.all(10),
```
dnys1, line 23:
```dart
padding: const EdgeInsets.all(10) + const EdgeInsets.only(bottom: 20)
```

or 

```dart
padding: const EdgeInsets.fromLTRB(10, 10, 10, 30)
```

packages/amplify_authenticator/lib/src/widgets/buttons.dart
```diff
@@ -152,3 +150,107 @@ class GoToSignInButton extends StatelessWidget {
...
+            )),
+        TextButton(
+          key: const Key(keyGoToSignInButton),
+          child: Text(
+            "Resend Code",
+            style: TextStyle(
+                fontSize: 13,
+                color: Theme.of(context).primaryColor != Colors.blue
+                    ? Theme.of(context).primaryColor
+                    : AuthenticatorColors.primary),
```
dnys1, line 176: Still not sure what this does 😅

dnys1 (commented):
Looking pretty good overall!

General comments would be to minimize the force unwrap `!` when possible - it's an anti-pattern which we want to avoid. If you find yourself doing this, see if there is a way to make the variable non-null or `late`, and if as a last resort you need to unwrap, centralize it to one place and use `asserts` with helpful error messages before unwrapping.

For example, instead of `InheritedForms.of(context)!` in a lot of places we could change `of` to make it return a non-null variable:

```dart
static InheritedForms of(BuildContext context) {
  final forms = context.dependOnInheritedWidgetOfExactType<InheritedForms>();
  assert(() {
    if (forms == null) {
      throw FlutterError.fromParts([
        ErrorSummary('No InheritedForms widget found.'),
        ErrorDescription(
            'Make sure your app is wrapped with an Authenticator widget.')
      ]);
    }
    return true;
  }());
  return forms!;
}
```

packages/amplify_authenticator/lib/src/state/inherited_forms.dart
```diff
@@ -19,12 +20,32 @@ class InheritedForms extends InheritedWidget {
...
+  final ResetPasswordForm resetPasswordForm;
+
+  final ConfirmSignInNewPasswordForm confirmSignInNewPasswordForm;
 
+  //final ResetPasswordForm resetPasswordForm;
 
   // ignore: public_member_api_docs
-  static InheritedForms? of(BuildContext context) =>
-      context.dependOnInheritedWidgetOfExactType<InheritedForms>();
+  static InheritedForms? of(BuildContext context) {
```
dnys1, line 34: Super close - now just make this return `InheritedForms` (non-null) and remove all the `!` from the `InheritedForms.of(context)!....` calls.

dnys1 (approved):
Great work!

haverchuck (approved):
LGTM (be sure to implement @dnys1's suggestion for the .of method)

Issue #398, opened 2021-02-19
Remote or Global Sign Out feature


fdarsot:
Firstly, great work with the package. appreciate all the work.

I was wondering can a feature be added where as soon a user Signs in via Amplify.Auth, the system logs the user out of all other active sessions? This way only a single user can be logged into a single device at a time.

Also, I noticed with the new release, the global signout option was deprecated.

Jordan-Nelson:
FYI - This issue will be used to track a global sign out feature, which is currently supported in the iOS and Android libraries. #418 will be used to track single device sign on.

Jordan-Nelson:
https://github.com/aws-amplify/amplify-flutter/pull/782 has been merged which adds global sign out. This will be included in the next release. I will update this issue once it is available.

miguelangelflores1993:
> Esta opción está disponible en la versión 0.2.5 y posteriores. La actualización de los documentos para esto aún está pendiente, por lo que mantendré este problema abierto. Hasta entonces, aquí hay un fragmento que muestra cómo utilizar esta opción:
> 
> ```dart
> try {
>   await Amplify.Auth.signOut(options: SignOutOptions(globalSignOut: true));
> } on AmplifyException catch (e) {
>   print(e.message);
> }
> ```

logout from cognito console but fetchAuthSession still fetching valid tokens
![image](https://github.com/aws-amplify/amplify-flutter/assets/116732288/cde3a73e-945d-42b4-9d04-6d9fc20a86cc)

PR #782, opened 2021-08-04
feat(auth): add global sign out

Jordan-Nelson:
*Issue #, if available:* #398

*Description of changes:*
- Add option for global sign out

*Testing Scenarios:*
- Global Sign Out on iOS device invalidates a session on an Android device
- Global Sign Out on Android device invalidates a session on an iOS device
- On iOS and Android, once a session has been invalidated, an attempt to perform a global sign out performs a local sign out
- On iOS and Android, once a session has been invalidated, local sign out still works as expected


packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthErrorHandlerTest.kt
```diff
@@ -83,7 +83,11 @@ class AuthErrorHandlerTest {
         doAnswer { invocation: InvocationOnMock ->
             errorHandler.handleAuthError(mockResult, exception)
             null as Void?
-        }.`when`(mockAuth).signOut(ArgumentMatchers.any<Action>(), ArgumentMatchers.any<Consumer<AuthException>>())
+        }.`when`(mockAuth).signOut(
+            ArgumentMatchers.any<AuthSignOutOptions>(),
```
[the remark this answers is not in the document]
haverchuck: I think we were just using signOut as a kind of hack to get back an arbitrary exception... feel free to change that if it makes sense to do so now.

packages/amplify_auth_plugin_interface/lib/src/SignOut/SignOutOptions.dart
```diff
@@ -13,10 +13,21 @@
  * permissions and limitations under the License.
  */
 
-abstract class SignOutOptions {
-  const SignOutOptions();
+/// The shared sign out options among all Auth plugins
+class SignOutOptions {
```
haverchuck, line 17: Should we make a CognitoSignOutOptions class which extends this at the plugin (as opposed to the category) level?

haverchuck (commented):
See a couple of inline comments/questions; otherwise, LGTM.

packages/amplify_auth_plugin_interface/lib/src/SignOut/SignOutOptions.dart
Jordan-Nelson: I wasn't sure if this would apply to all plugins either so I followed Amplify-Android and Amplify-iOS.

[Android has it on the non-cognito-specific class](https://github.com/aws-amplify/amplify-android/blob/main/core/src/main/java/com/amplifyframework/auth/options/AuthSignOutOptions.java#L25). iOS expects a [SignOutOptions from the SDK](https://github.com/aws-amplify/aws-sdk-ios/blob/2ce94a7fda175ed6ffd86c993b0ace219b926ac2/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileOptions.swift#L52), which has globalSignOut on it.

packages/amplify_auth_cognito/android/src/test/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AmplifyAuthErrorHandlerTest.kt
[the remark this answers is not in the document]
Jordan-Nelson: Yeah. I thought about updating them to methods that would be likely to throw the particular Exception being tested, but having them all be the same is probably easier, even if it is a little bit of a hack.

packages/amplify_auth_plugin_interface/lib/src/SignOut/SignOutRequest.dart
```diff
@@ -15,9 +15,17 @@
...
 
+/// Encapsulates parameters for a sign out request
 class SignOutRequest {
+  /// Advanced options for the request
+  ///
+  /// Can be used to specify global sign out
   SignOutOptions? options;
+
+  /// Default constructor
   SignOutRequest({this.options});
```
dnys1, line 26: Prefer a more descriptive doc comment for the constructor since this is what is shown when CX goes to create an instance:

Screen Shot 2021-09-22 at 1 19 53 PM

Also, can be final fields/const constructor while we're at it.

```suggestion
/// {@template sign_out_request}
/// Encapsulates parameters for a sign out request
/// {@endtemplate}
class SignOutRequest {
  /// Advanced options for the request
  ///
  /// Can be used to specify global sign out
  final SignOutOptions? options;

  /// {@macro sign_out_request}
  const SignOutRequest({this.options});
```

packages/amplify_auth_cognito/ios/Classes/FlutterSignOutRequest.swift
```diff
@@ -0,0 +1,32 @@
...
+  var options: AuthSignOutOperation.Request.Options?
+    
+  init(dict: NSMutableDictionary){
+    self.options = createOptions(options: dict["options"] as! Dictionary<String, Any>?)
+  }
+    
+  func createOptions(options: Dictionary<String, Any>?) -> AuthSignOutOperation.Request.Options {
+    let globalSignOut = options?["globalSignOut"] as? Bool ?? false
+    return AuthSignOutOperation.Request.Options(globalSignOut: globalSignOut)
+  }
```
dnys1, line 30:
```suggestion
    init(dict: NSMutableDictionary) {
        if let options = dict["options"] as? [String: Any] {
            self.options = createOptions(options: options)
        }
    }
    
    func createOptions(options: [String: Any]) -> AuthSignOutOperation.Request.Options {
        let globalSignOut = options["globalSignOut"] as? Bool ?? false
        return AuthSignOutOperation.Request.Options(globalSignOut: globalSignOut)
    }
```

dnys1 (approved):
LGTM. Thanks for the Dart doc comments 👍

packages/amplify_auth_plugin_interface/lib/src/SignOut/SignOutRequest.dart
Jordan-Nelson: Good call. Updated.

haverchuck (approved):

PR #786, opened 2021-08-06
fix(amplify_authenticator): text overflow & validation fixes 

israx:
*Issue #, if available:*

*Description of changes:*
- validation in confirm sign up form fields
- no null check in inherited forms
- overflow in buttons and exception widget
- loading state in sign out button


packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
```diff
@@ -44,10 +45,27 @@ class StateMachineBloc {
...
+      StreamController<AuthenticatorException?>.broadcast();
 
-  Stream<AuthenticatorException> get exceptions => _exceptionController.stream;
+  Stream<AuthenticatorException?> get exceptions => _exceptionController.stream;
+
+  void clearException() {
+    _exceptionController.add(null);
+  }
+
+  Future<void> resendSignUpCode(String username) async {
```
haverchuck, line 57: Could we move this down with the other service calls? Also, why is it public why the others are private (prefixed with "_")

packages/amplify_authenticator/lib/src/screens/confirm_signin_mfa_screen.dart
```diff
@@ -0,0 +1,16 @@
+import 'package:amplify_authenticator/src/state/inherited_forms.dart';
```
haverchuck, line 1: Didn't we already merge this?

packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
israx: They are public because they don't form part of the state machine which takes events and then triggers those private methods.

packages/amplify_authenticator/lib/src/widgets/buttons.dart
```diff
@@ -102,23 +102,30 @@ class GoToSignUpButton extends StatelessWidget {
...
+          ),
+          TextButton(
+            key: const Key(keyGoToSignUpButton),
+            onPressed: _authModelView.goToSignUp,
+            child: Text("Create account",
+                style: TextStyle(
+                    fontSize: AuthenticatorButtonConstants.fontSize,
+                    color: Theme.of(context).primaryColor != Colors.blue
+                        ? Theme.of(context).primaryColor
+                        : AuthenticatorColors.primary)),
```
dnys1, line 125: @haverchuck We will need to update this at some point but it's not a huge deal right now

dnys1 (approved):
Nice work!

haverchuck (approved):

PR #803, opened 2021-08-12
fix(amplify_api): GraphQL Helper Clean Up

Equartey:
*Issue #, if available:*

*Description of changes:*
This PR contains fixes for issues I've come across in testing.

### DIFF
--- a/packages/amplify_api/lib/method_channel_api.dart
+++ b/packages/amplify_api/lib/method_channel_api.dart
@@ -162,9 +162,8 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
 
       // Catch data while its non-nullable and provide context with error messages
       if (result['data'] == null) {
-        throw ApiException(
-          "App sync returned null for data, see errors: " + errors.toString(),
-        );
+        throw ApiException("App sync returned null for data, see errors: ",
+            underlyingException: errors.toString());
       }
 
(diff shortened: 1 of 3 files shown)


packages/amplify_api/lib/method_channel_api.dart
```diff
@@ -160,6 +160,13 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
             AmplifyExceptionMessages.nullReturnedFromMethodChannel);
       final errors = _deserializeGraphQLResponseErrors(result);
 
+      // Catch data while its non-nullable and provide error messages
+      if (result['data'] == null) {
```
dnys1, line 164: We may need to brainstorm another way to handle this case. There may not always be an error in `errors` if this is `null` - sometimes null is a valid response.

Equartey: This goes back to changing the GraphQLRequest to accept nullable data. Currently null responses will break workflows and I've found the error reporting for these errors is buried too deep. For now I think this will at least provide developers (mainly internal) with missing error context until a more permanent solution is reached.

dnys1: Oh that's right - sorry I have the memory of a goldfish!

dnys1 (approved):

Jordan-Nelson (approved):

PR #812, opened 2021-08-17
chore(api): Apply lints

dnys1:
*Issue #, if available:*

*Description of changes:*
- Apply lints to `amplify_api` and `amplify_api_plugin_interface`


Jordan-Nelson (commented):
I think the scope for `analyze` needs to be updated. LGTM otherwise.

melos.yaml
```diff
@@ -68,12 +68,13 @@ scripts:
...
   analyze:
     run: melos exec -c 1 --fail-fast -- \
-      flutter analyze --no-fatal-infos
+      flutter analyze
     description: >
       Analyzes all packages and fails if there are any errors.
     select-package:
       scope:
         - amplify_analytics_*
+        - amplify_api_*
```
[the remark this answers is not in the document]
dnys1: Whoops good catch. Fixed.

melos.yaml
```diff
@@ -68,12 +68,13 @@ scripts:
 
   analyze:
     run: melos exec -c 1 --fail-fast -- \
-      flutter analyze --no-fatal-infos
+      flutter analyze
```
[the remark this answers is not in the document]
dnys1: I like the idea and it would work fine for iOS/Flutter, but for Android it would need to build all the example apps before it's able to run the gradle script. I'll see if I can find a solution that doesn't involve building the apps.

Also, for iOS, the lint command is set to run as part of the Xcode build (without `--lint`) so that while you're developing, fixes are automatically applied. I'm not sure if Android has something similar (maybe with Gradle task deps), but that could be another solution as well.

Jordan-Nelson (approved):

packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/FlutterApiRequest.kt
```diff
@@ -15,132 +15,123 @@
...
+            throw AmplifyException(
+                "The apiName request argument was not passed as a String",
+                cause,
+                "The request should include the apiName as a String"
+            )
         }
+    }
 
-        fun getRestOptions(request: Map<String, Any>): RestOptions {
+    fun getRestApiName(request: Map<String, Any>): String? {
```
haverchuck, line 59: Are these additions a direct result of lint results, or are they more general refactoring? If the latter, then they might be better in a different PR since this one is already substantial.

packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/rest_api/FlutterRestApi.kt
```diff
@@ -28,239 +28,279 @@ typealias FunctionWithApiName = KFunction4<
...
-                }
-                return
-            } catch (e: Exception) {
-                handler.post {
-                    ExceptionUtil.postExceptionToFlutterChannel(flutterResult, "ApiException",
-                            ExceptionUtil.createSerializedUnrecognizedError(e))
-                }
-                return
-            }
+enum class RestOperationType(val value: String) {
```
haverchuck, line 31: Same question - is this general refactoring?

packages/amplify_api/ios/Classes/FlutterApiErrorHandler.swift
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
haverchuck, line 2: nit: I'm not sure we actually need to change these license years unless it's a brand new file?

packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/rest_api/FlutterRestApi.kt
dnys1: This was to address a lint warning triggered by the constants beforehand.

packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/FlutterApiRequest.kt
dnys1: This one is a bit of a catch-22. Fixing the lint errors actually caused some tests to start failing. It turned out that there was a latent logical error which was being masked. There is only one real change in this file to fix that, but it looks like a lot more.

packages/amplify_api/ios/Classes/FlutterApiErrorHandler.swift
dnys1: Ah gotcha - okay I'll change it back

dnys1: Actually, I think it's required as part of SwiftFormat to match the template, which has 2021. I can change it to match [amplify-ios](https://github.com/aws-amplify/amplify-ios/blob/main/.swiftformat) which doesn't have the date or just remove that from SwiftFormat altogether. Lemme know.

packages/amplify_api/lib/method_channel_api.dart
```diff
@@ -155,31 +153,34 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
...
 
   Future<GraphQLResponse<T>> _getMethodChannelResponse<T>({
-    required methodName,
+    required String methodName,
     required GraphQLRequest<T> request,
   }) async {
     try {
-      final Map<String, dynamic>? result =
-          (await (_channel.invokeMapMethod<String, dynamic>(
+      final result = await _channel.invokeMapMethod<String, dynamic>(
```
haverchuck, line 169: Is the removal of explicit type annotations part of the linting?

dnys1: Yep: https://dart-lang.github.io/linter/lints/omit_local_variable_types.html

packages/amplify_api/ios/Classes/FlutterApiErrorHandler.swift
haverchuck: Oh I see - I think it's ok. We might need to re-discuss before the end of year.

haverchuck (approved):

PR #838, opened 2021-08-25
feat(amplify_flutter): AmplifyConfig Dart class with getter and instantiation

haverchuck:
*Description of changes:*
Adds a Dart class to represent the amplify configuration. 

Note - will add documentation after API review.


packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -204,26 +205,42 @@ class _AuthenticatorState extends State<Authenticator> {
...
 
   late final StateMachineBloc _stateMachineBloc;
+  AmplifyConfig? _config;
 
   @override
   void initState() {
     super.initState();
     _stateMachineBloc = StateMachineBloc(_authService)
       ..authEvent.add(GetCurrentUser());
+    waitForConfiguration().then((value) {
```
dnys1, line 215: Do we need the added stream controller? 

```dart
Amplify.configCompleter.future
  .then((config) {
    if (mounted) {
      setState(() => _config = config);
    }
  })
  .catchError((e) => show error or something ?);
```

packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -204,26 +205,42 @@ class _AuthenticatorState extends State<Authenticator> {
...
+    });
+  }
+
+  Future<AmplifyConfig> waitForConfiguration() async {
+    return await _authService.waitForConfiguration();
   }
 
   @override
   Widget build(BuildContext context) {
+    _stateMachineBloc.configStream.listen((event) {
```
dnys1, line 226: Should move to `initState` so it's not called on every build (will add multiple listeners).

packages/amplify_flutter/lib/config/analytics/analytics_plugin.dart
```diff
@@ -0,0 +1,37 @@
...
+
+import 'package:amplify_flutter/config/analytics/pinpoint_analytics_config.dart';
+import 'package:amplify_flutter/config/analytics/pinpoint_targeting_config.dart';
+import 'package:json_annotation/json_annotation.dart';
+
+part 'analytics_plugin.g.dart';
+
+@JsonSerializable()
+class AnalyticsPlugin {
+  @JsonKey(name: 'pinpointAnalytics')
```
dnys1, line 24: No need to specify name if it matches the Dart field, only if recasing or renaming.

packages/amplify_flutter/lib/config/auth/login_mechanisms.dart
```diff
@@ -0,0 +1,24 @@
...
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+enum LoginMechanisms {
+  EMAIL,
```
dnys1, line 17: Prefer following Dart conventions for [enums](https://dart.dev/guides/language/effective-dart/style#prefer-using-lowercamelcase-for-constant-names) (camel case). Can adjust JSON value as I mentioned

```dart
enum LoginMechanisms {
  @JsonValue('EMAIL')
  email,

  // etc.
}
```

dnys1 (changes requested):

packages/amplify_authenticator/lib/amplify_authenticator.dart
haverchuck: I thought this could be useful in the future if we introduce the ability to call 'configure' multiple times, but no - we don't strictly need it. I'll remove for now.

packages/amplify_authenticator/lib/amplify_authenticator.dart
haverchuck: Done!

packages/amplify_flutter/lib/config/analytics/analytics_plugin.dart
haverchuck: Done.

packages/amplify_authenticator/lib/amplify_authenticator.dart
dnys1: Ah gotcha - yeah, we did discuss that. My bad! Fine either way, but yeah, I still prefer introducing complexity when needed.

packages/amplify_flutter/lib/config/auth/login_mechanisms.dart
haverchuck: Done!

packages/amplify_flutter/lib/config/analytics/analytics_config.dart
```diff
@@ -0,0 +1,39 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import 'package:amplify_flutter/config/analytics/analytics_plugin.dart';
+
+part 'analytics_config.g.dart';
+
+/// A Dart representation of an Amplify Analytics configuration
+class AnalyticsConfig {
```
dnys1, line 21: So sorry for the confusion - the `@JsonSerializable` annotation is needed on all class requiring codegen, but individual fields only need the `@JsonKey` annotation if they are overriding options from `JsonSerializable` or need to recase/rename the generated JSON key.

packages/amplify_flutter/lib/config/auth/mfa_configurations.dart
```diff
@@ -0,0 +1,16 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+enum MfaConfiguration { Optional, On, Off }
```
dnys1, line 16: Prefer camel case

packages/amplify_flutter/lib/config/auth/password_policy_characters.dart
```diff
@@ -0,0 +1,21 @@
...
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+enum PasswordPolicyCharacters {
+  RequiresLowercase,
```
dnys1, line 17: Prefer camel case

packages/amplify_flutter/lib/amplify.dart
```diff
@@ -158,15 +170,20 @@ class AmplifyClass extends PlatformInterface {
...
       _isConfigured = res ?? false;
+
+      configCompleter.complete(_config);
       if (!_isConfigured) {
         throw AmplifyException('Amplify failed to configure.',
             recoverySuggestion:
                 AmplifyExceptionMessages.missingRecoverySuggestion);
       }
     } on PlatformException catch (e) {
+      configCompleter.completeError(e);
```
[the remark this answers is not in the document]
dnys1: I believe that's also why the one test is failing in `amplify_flutter`.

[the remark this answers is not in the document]
haverchuck: Well, in my mind this completer could be something that took the place of using the `isConfigured` getter, in which case I think it would make sense to return an exception... but perhaps for now, for our use case, it would be simpler to do as you suggest. We could be clear about it in our docs.

packages/amplify_flutter/lib/amplify.dart
```diff
@@ -158,15 +170,20 @@ class AmplifyClass extends PlatformInterface {
           underlyingException: e.toString());
     }
     try {
+      _config = _createConfigObject(configuration);
```
Jordan-Nelson, line 173: This will throw an exception if a value that we expect to be non-null is null, correct?

Is there a spec or type definition for the config that you were able to reference to determine which values are optional/required?

How do we handle breaking changes? If the CLI changes a required property to be optional, would the CLI treat that as a breaking change and create a new major version? Would we then have to tell customers to not upgrade to a new version of the CLI until they are using version x.y.z of amplify-flutter?

I guess the iOS/Android/JS libraries are already dealing with this to an an extent, but nullability works differently in Java/Swift/TS. If something like `authenticationFlowType` hypothetically became optional in a future CLI version, an exception would be thrown trying to read the config since it is cast to a `String`, even though that value is never accessed in amplify-flutter or the authenticator. Is that true for  iOS/Android?

I know some of this was already discussed, so sorry if I am asking questions that have already been answered.

Jordan-Nelson: It would be good to provide meaningful exceptions in the case that there is an exception reading the config. Inevitably some customers are going to have issues with their config missing properties that we are expecting to be non-null (either because they are using an outdated version of the CLI, they are using a new version of the CLI with breaking changes, or because we simply marked something as non-nullable that should not have been).

The call to `_createConfigObject` could be wrapped in its own try catch, with a meaningful message.

We might event want to consider calling completeError and not throwing an exception if the config cannot be parsed successfully since it is only required for using the authenticator at the moment. In a scenario where a customer who is not using the authenticator has a config that iOS/Android can parse, but amplify-flutter cannot, it would be a poor experience to throw an exception here. If we have a spec/type definition for the config that we feel confident in, it is probably fine to throw. I just want to avoid impacting customers who aren't event using the authenticator.

packages/amplify_flutter/lib/config/analytics/analytics_config.dart
haverchuck: This should be resolved.

packages/amplify_flutter/lib/config/auth/login_mechanisms.dart
```diff
@@ -0,0 +1,24 @@
...
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+enum LoginMechanisms {
+  Email,
```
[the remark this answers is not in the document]
haverchuck: done

packages/amplify_flutter/lib/config/auth/password_policy_characters.dart
haverchuck: done!

packages/amplify_flutter/lib/config/auth/mfa_configurations.dart
haverchuck: done.

packages/amplify_flutter/lib/amplify.dart
```diff
@@ -60,6 +62,12 @@ class AmplifyClass extends PlatformInterface {
   // ignore: public_member_api_docs
   AmplifyHub Hub = AmplifyHub();
 
+  final configCompleter = Completer<AmplifyConfig>();
+
+  Future<AmplifyConfig> configNotification() {
```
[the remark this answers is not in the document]
dnys1: +1 though I typically prefer `onX` style for these getters 

```dart
Future<AmplifyConfig> get onConfigure => configCompleter.future;
```

packages/amplify_flutter/lib/amplify.dart
```diff
@@ -126,6 +134,10 @@ class AmplifyClass extends PlatformInterface {
     return _isConfigured;
   }
 
+  AmplifyConfig get config {
```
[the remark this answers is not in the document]
dnys1: +1

packages/amplify_flutter/lib/amplify.dart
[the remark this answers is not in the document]
haverchuck: Done!

packages/amplify_flutter/lib/amplify.dart
```diff
@@ -60,6 +62,12 @@ class AmplifyClass extends PlatformInterface {
   // ignore: public_member_api_docs
   AmplifyHub Hub = AmplifyHub();
 
+  final configCompleter = Completer<AmplifyConfig>();
```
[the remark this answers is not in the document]
haverchuck: Good idea... done!

packages/amplify_flutter/lib/amplify.dart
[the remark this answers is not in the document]
haverchuck: Removed!

packages/amplify_flutter/lib/amplify.dart
haverchuck: As we discussed, we are not going to use completeError, and are going to rely on a) native libraries catching major problems with the config, and b) making all config values nullable so that `_createConfigObject` should not fail.

haverchuck: I suppose we could wrap the call to  _createConfigObject in a try/catch, but at some point I think we would need an empty catch block because we don't want to propogate the exception all the way up?

packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -202,28 +207,54 @@ class Authenticator extends StatefulWidget {
...
   @override
   Widget build(BuildContext context) {
     var defaultForms = DefaultForms();
     AuthStringResolver resolver = widget.resolver ?? AuthStringResolver();
-
     defaultForms.context = context;
 
+    if (_configInitialized && _missingConfigValues.isNotEmpty) {
+      throw StateError(
+          'Encountered problem(s) building the Authenticator because your amplifyconfiguration.dart file is missing required values: ${_missingConfigValues.join('/n')}). You should correct your amplifyconfiguration.dart file and restart your app.');
```
dnys1, line 250: Prefer a multiline string here

packages/amplify_flutter/lib/amplify.dart
```diff
@@ -182,10 +194,52 @@ class AmplifyClass extends PlatformInterface {
...
-
     await DataStore.configure(configuration);
   }
 
+  Future<void> _parseConfigJson(String configuration) async {
+    try {
+      _config = _createConfigObject(configuration);
+      _configCompleter.complete(_config);
+    } catch (e) {
+      if (kDebugMode) {
```
dnys1, line 205: Prefer a utility function for this kind of stuff. Can add it to amplify_core, maybe.

packages/amplify_flutter/lib/amplify.dart
```diff
@@ -182,10 +194,52 @@ class AmplifyClass extends PlatformInterface {
             underlyingException: e.toString());
       }
     }
-
     await DataStore.configure(configuration);
   }
 
+  Future<void> _parseConfigJson(String configuration) async {
```
dnys1, line 200: Looks like this doesn't need to be async anymore

packages/amplify_flutter/lib/config/amplify_config.dart
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+import 'package:amplify_flutter/config/analytics/analytics_config.dart';
+import 'package:amplify_flutter/config/api/api_config.dart';
+import 'package:amplify_flutter/config/auth/auth_config.dart';
+import 'package:amplify_flutter/config/storage/storage_config.dart';
+
+part 'amplify_config.g.dart';
+
+class AmplifyConfig {
```
dnys1, line 23: I was thinking we leave the `JsonSerializable` annotations. Thoughts?

packages/amplify_flutter/lib/amplify.dart
```diff
@@ -182,10 +194,52 @@ class AmplifyClass extends PlatformInterface {
...
+    try {
+      _config = _createConfigObject(configuration);
+      _configCompleter.complete(_config);
+    } catch (e) {
```
dnys1, line 204: Let's add an `on` clause to this

packages/amplify_authenticator/lib/amplify_authenticator.dart
dnys1: Also, I feel like this might be TMI for the typical authenticator user. What about just saying "please re-run `amplify pull`"?

packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -202,28 +207,54 @@ class Authenticator extends StatefulWidget {
...
+        _configInitialized = true;
+        _missingConfigValues = missingConfigValues(config);
+      });
+    });
+  }
+
+  List<String> missingConfigValues(AmplifyConfig? config) {
+    List<String> missingValues = [];
+    var cognitoPlugin = config?.auth?.awsCognitoAuthPlugin?.auth?['Default'];
+    cognitoPlugin ?? missingValues.add('auth.plugins.Auth.Default');
```
dnys1, line 236: I actually think this is pretty cool, although I've never seen it done in Dart haha. It looks very JavaScript-y. Not sure yet, but the linter may not like this. I think standard `if (cognitoPlugin == null)` may also read a bit better but I could just be un-hip with the latest trends lol

dnys1 (commented):
Few general comments:

- I'm thinking we put the AmplifyConfig definitions inside the `src` directory so they cannot be easily imported by other packages. These APIs will change in the future and they're probably not ready for prime time.
- I think we should leave all the JSON annotations in place. It will be way too much work to add them back every time we need to run the generator. And the non-generated files should be the source of truth. It's typically recommended not to commit the generated files.

packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -202,28 +207,54 @@ class Authenticator extends StatefulWidget {
...
+        missingValues.add('auth.plugins.Auth.Default.signUpAttributes');
+    return missingValues;
   }
 
   @override
   Widget build(BuildContext context) {
     var defaultForms = DefaultForms();
     AuthStringResolver resolver = widget.resolver ?? AuthStringResolver();
-
     defaultForms.context = context;
```
dnys1, line 246: This is generally an anti-pattern. At some point, we'll want to rewrite this logic.

packages/amplify_authenticator/lib/amplify_authenticator.dart
Jordan-Nelson: +1 on including some info on how to correct it. I do think that including the missing config values could be helpful in some scenarios.

Maybe something like 

> Encountered problem(s) building the Authenticator because your amplifyconfiguration.dart file is missing required values. Please re-run `amplify pull` and restart your app.
> 
> Missing values are: values ...

packages/amplify_core/lib/amplify_core.dart
```diff
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
Jordan-Nelson, line 2: nit: 
```suggestion
 * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

Jordan-Nelson (approved):
I left a couple nit comments and do agree with a some of @dnys1 comments as well, but overall LGTM

packages/amplify_flutter/lib/amplify.dart
```diff
@@ -17,7 +17,9 @@ library amplify;
 
 import 'dart:async';
 import 'dart:convert';
+import 'dart:developer' as developer;
```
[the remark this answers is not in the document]
haverchuck: This was an artifact; removing it.

packages/amplify_core/lib/amplify_core.dart
haverchuck: Made this 2020 since this file was added in 2020. Not sure if that actually matters.

packages/amplify_authenticator/lib/amplify_authenticator.dart
haverchuck: I think we should keep CLI references out of the code, since customers are not required to use the CLI at all in order to use amplify-flutter.

packages/amplify_flutter/lib/amplify.dart
haverchuck: Writing a util for this, but keeping it in amplify_flutter since I don't expect it to be used by any other packages.

packages/amplify_flutter/lib/config/amplify_config.dart
haverchuck: Adding the annotations would have the benefit of easing re-running codegen when these configs get updated; conversely, it would pose the risk of adding friction for customers who are using conflicting versions of the codegen packages who would then have to handle dependency resolution. For now I think the risk outweighs the benefit, but if we find that these config values require frequent updates or we expand our usage of codegen (which we should!) then I would agree that the annotations should be committed.

packages/amplify_authenticator/lib/amplify_authenticator.dart
haverchuck: Yeah - I think I am just going to bite the bullet and widget-ify the default forms, which will obviate the need for this.

dnys1 (approved):

PR #831, opened 2021-08-23
test(DataStore): add local integration tests for datastore

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*

This PR has adds a subset of the overall planned suite of integ tests for datastore. Below are the tests that are planned. The ones that are checked off are included in this PR.

#### Save
- [x] Should save a single model
- [x] Should update an existing model, if a model already exists with a given ID
- [x] Should be able to save/update a model with any field type (String, Bool, Double, Int, Date, custom enum, List of any type)
- [x] Should be able to update a model of any relationship type (hasOne, HasMany, belongsTo)
#### Query
- [x] Should return a list of all models
- [x] Should return an empty list if no models match the query
- [x] Should return the correct model when queried by ID
- [ ] Should return the correct model(s) when any type is queried with each QueryPredicate
    - [x] String (eq, lt, gt, le, ge, ne, beginsWith, between, contains)
    - [x] Bool (eq, ne)
    - [x] Double (eq, lt, gt, le, ge, ne, beginsWith, between, contains)
    - [x] Int (eq, lt, gt, le, ge, ne, beginsWith, between, contains)
    - [ ] Date types (eq, lt, gt, le, ge, ne, beginsWith, between, contains) *(skipping due to issues with date types. See: https://github.com/aws-amplify/amplify-flutter/issues/837)*
    - [ ] custom enum (eq, ne)
- [ ] Should return the correct model(s) when compound query predicates are used
- [ ] Should return the models in the correct order when using sortBy (Ascending / Descending) for any field type (String, Bool, Double, Int, Date, custom enum)
- [ ] Should return the models in the correct order when using multiple query sorts
- [ ] Should return the first N models when using pagination (limit or page number)
- [x] Should return data for nested belongTo models
- [x] Should return data for a model with any field type (String, Bool, Double, Int, Date, custom enum, List of any type)
- [x] Should return data for a model of any relationship type (hasOne, HasMany, belongsTo)
#### Delete
- [x] Should delete an existing model
- [x] Should return data for a model of any relationship type (types: hasOne, HasMany, belongsTo)
- [x] Should delete data for a model with any field type (String, Bool, Double, Int, Date, custom enum, List of any type)
#### Clear
- [ ] Should clear all data in the local store
#### Observe
- [ ] Should broadcast events for the correct type (create, update, delete)
- [x] Should broadcast events with the new/update model
- [x] Should broadcast events with nested model data (for nested models)
[...]

PR #844, opened 2021-08-27
test(DataStore): add additional local integration tests for datastore

Jordan-Nelson:
*Issue #, if available:* n/a

*Description of changes:*

follow up PR to https://github.com/aws-amplify/amplify-flutter/pull/831 to add remaining tests. The added tests include:
- DataStore.clear()
- DataStore.observe()
- Query Predicate tests for enum types
- Compound Query Predicates
- Pagination
- Sort order


dnys1 (approved):
LGTM. Thanks!

packages/amplify_datastore/example/integration_test/query_test/pagination_test.dart
```diff
@@ -0,0 +1,112 @@
...
+      for (var model in models) {
+        await Amplify.DataStore.save(model);
+      }
+    });
+    testWidgets('should return the models for the given page and limit',
+        (WidgetTester tester) async {
+      // page 0
+      var pageZeroBlogs = await Amplify.DataStore.query(Blog.classType,
+          pagination: QueryPagination(page: 0, limit: 10));
+      var expectedPageZeroBlogs = models.getRange(0, 10).toList();
```
HuiSF, line 43: Should we add a sorting constrain to the query while comparing two lists using `ListEquality().equals` which the order of item matters for.

packages/amplify_datastore/example/integration_test/query_test/query_predicate_test/compound_query_predicate_test.dart
```diff
@@ -0,0 +1,375 @@
...
+
+        setUpAll(() async {
+          await configureDataStore();
+          await clearDataStore();
+          for (var model in models) {
+            await Amplify.DataStore.save(model);
+          }
+        });
+
+        testWidgets('eq() (and)', (WidgetTester tester) async {
```
HuiSF, line 301: nit: should it be `eq() & eq()` following other naming pattern above? Ditto below test descriptions

HuiSF (approved):
Thanks for adding the tests. Left a couple of nit-picking comments.

packages/amplify_datastore/example/integration_test/query_test/pagination_test.dart
```diff
@@ -0,0 +1,112 @@
...
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  group('pagination', () {
+    var models = List.generate(1000, (i) => Blog(name: 'blog $i'));
+    setUpAll(() async {
+      await configureDataStore();
+      await clearDataStore();
+      for (var model in models) {
+        await Amplify.DataStore.save(model);
```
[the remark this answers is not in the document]
Jordan-Nelson: I like this idea in general, but it becomes problematic with how these tests are currently.

The current tests query for a specific page, but do not apply a specific sort order (with the exception of one test that does use sort order). The current code will create the models in a specific order, which means they are also returned in that order if no sort order is applied. If the models are not created in a specific order, then there is no guarantee which blogs would be included for a specific page (without using a sort order).

Few options

1. Leave as is. Pros: Majority of tests are focused on pagination. Cons: Implicitly testing & relying on the order defaulting to date created.
2. Apply a sort order to every query. Pros/cons are basically the opposite of above.
3. Update expectations to only check the length of models returned. The tests would lose a lot of value. Probably not something to really consider but technically an option.

packages/amplify_datastore/example/integration_test/query_test/pagination_test.dart
Jordan-Nelson: This is a good point. I updated the tests to use unorderedEquals and orderedEquals. See my comment above which is sort of related as well

packages/amplify_datastore/example/integration_test/query_test/query_predicate_test/compound_query_predicate_test.dart
Jordan-Nelson: Since it is testing 4 predicates together, it would be `eq() & eq() & eq() & eq()` following the pattern, but I thought that was pretty verbose

packages/amplify_datastore/example/integration_test/query_test/pagination_test.dart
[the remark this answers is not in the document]
Jordan-Nelson: I think I am leaning towards leaving it, but if others had the preference for option 2, I think thats fine as well.

PR #852, opened 2021-09-02
feat(api): GraphQL Subscription Stream

dnys1:
*Issue #, if available:*

*Description of changes:*
- Changes GraphQL subscription interface to use `Stream`


packages/amplify_api_plugin_interface/lib/src/graphql/graphql_request.dart
```diff
@@ -30,7 +30,7 @@ class GraphQLRequest<T> {
   Map<String, dynamic> serializeAsMap() => <String, dynamic>{
         'document': document,
         'variables': variables,
-        'cancelToken': cancelToken,
+        'cancelToken': id,
```
Jordan-Nelson, line 33: *nit/question:* Would it make sense to update this to `'id': id,` (and then make the required updates on the platform side)?

packages/amplify_api/pubspec.yaml
```diff
@@ -10,6 +10,7 @@ environment:
 dependencies:
   amplify_api_plugin_interface: 0.2.3
   amplify_core: 0.2.3
+  async: ^2.6.0
```
Jordan-Nelson, line 13: Is this only used in tests? Should it be a dev dependency?

Jordan-Nelson (approved):

packages/amplify_api/pubspec.yaml
dnys1: Good catch - fixed!

packages/amplify_api_plugin_interface/lib/src/graphql/graphql_request.dart
dnys1: This would affect the REST API as well since they share some functions on the platform side. Happy to make the change, but initially held off because of that.

fjnoyp (approved):
Really awesome work!  Learned a bit reading over the PR so thanks :) 

When will we update the API Example app with these changes?

dnys1:
> Really awesome work! Learned a bit reading over the PR so thanks :)
> 
> When will we update the API Example app with these changes?

Sweet! Glad to hear it 🗡️ 

The changes were pretty small so I guess I snuck them in 😄
https://github.com/aws-amplify/amplify-flutter/pull/852/files#diff-5f1dcec94441f701ada71a5f00ca4571664225820b37db99e0aa2eb0317e91c3

PR #862, opened 2021-09-07
fix(api): OIDC/Lambda Fixes

dnys1:
*Issue #, if available:*

*Description of changes:*
- Fixes GraphQL error decoding
- Fixes OIDC/Lambda support for single-auth DataStore


melos.yaml
```diff
@@ -68,7 +68,7 @@ scripts:
 
   analyze:
     run: melos exec -c 1 --fail-fast -- \
-      flutter analyze
+      flutter analyze --no-fatal-infos
```
dnys1, line 71: @Jordan-Nelson The new analyzer bundled with Flutter 2.5 has changed a bit causing analysis to fail for missing public docs. I think `no-fatal-infos` is a good balance here - the analyzer will continue to show squiggly lines in the editor to remind you as you work, but not fail here if there are good reasons to leave them out. Same with the other "info" members.

And we can adjust rules to `warning` level if needed so that it does fail here.

haverchuck (approved):

Jordan-Nelson (approved):

PR #870, opened 2021-09-10
feat(authenticator): add support for verify user

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- initial draft of support for verify user


packages/amplify_authenticator/lib/src/widgets/form_fields.dart
```diff
@@ -715,3 +717,158 @@ class _ConfirmSignInFormFieldState extends State<ConfirmSignInFormField> {
...
+  late VerifyUserViewModel _verifyUserViewModel;
+  late ConfirmVerifyUserViewModel _confirmVerifyUserViewModel;
+  late List<String> unverifiedAttributeKeys;
+
+  // set value for the verify user & confirm verify user view model
+  void setUserAttributeKey(String? value) {
+    setState(() {
+      _value = value;
+      _verifyUserViewModel.setUserAttributeKey(value);
+      _confirmVerifyUserViewModel.setUserAttributeKey(value);
```
Jordan-Nelson, line 742: *Note:* I am not a huge fan of this pattern of setting state on two different view models at once. This is how username is shared between sign up and confirm sign up. It might be worth exploring another way to do share state.

packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
```diff
@@ -32,14 +32,15 @@ class StateMachineBloc {
   StreamSink<AuthEvent> get authEvent => _authEventController.sink;
 
   /// Outputs events into the event transformer.
-  late final Stream<AuthEvent> _authEventStream = _authEventController.stream;
+  Stream<AuthEvent> get authEventStream => _authEventController.stream;
```
Jordan-Nelson, line 35: *Note:* This was made public so that the Verify User View Model can listen to it and update state when certain events are emitted. This is a new pattern. The authenticator currently doesn't have a need to pass state to view models.

The state that is being passed is the list of unverified attributes. See `AuthSetUnverifiedAttributeKeys` farther down in this file as well as the `VerifyUserViewModel` when it is listened to.

dnys1: I would just create a new AuthUnverifiedUser state in place of AuthFlow.verifyUser and add the keys as data to that state object. This starts to get a little unwieldy since bloc are supposed to be a black box which translate events to states.

packages/amplify_authenticator/lib/src/services/amplify_auth_service.dart
```diff
@@ -109,4 +119,60 @@ class AmplifyAuthService implements AuthService {
...
+      confirmationCode: confirmationCode,
+    );
+  }
+
+  /// Get a list of the unverified attribute keys for the given user
+  ///
+  /// This is based off of `verifiedContact()` from amplify-js
+  /// https://github.com/aws-amplify/amplify-js/blob/6de9a1d743deef8de5205590bf7cf8134a5fb5f4/packages/auth/src/Auth.ts#L1199-L1224
+  @override
+  Future<List<String>> getUnverifiedAttributeKeys() async {
```
dnys1, line 147: This seems really verbose - what about:

```dart
List<AuthUserAttribute> userAttributes =
    await Amplify.Auth.fetchUserAttributes();

return userAttributes
    .map((attr) => attr.userAttributeKey)
    .where(
      (key) =>
          (key == 'email' || key == 'phone_number') &&
          userAttributes
                  .firstWhereOrNull(
                    (attr) => attr.userAttributeKey == '${key}_verified',
                  )
                  ?.value !=
              true,
    )
    .toList();
```

packages/amplify_authenticator/lib/src/widgets/form_fields.dart
dnys1: Yeah, I think just making the keys part of the bloc state solves all this.

dnys1 (commented):
Looking good- try the new state idea and see if that helps. I think we can minimize the complexity that way.

packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
Jordan-Nelson: That is how I implemented it initially actually. I would not be opposed to going back to that. I [have a branch](https://github.com/aws-amplify/amplify-flutter/compare/amplify-authenticator...Jordan-Nelson:feat/authenticator/verify-user) up with that approach (that branch is definitely in a less polished draft state that this branch fyi). 

I switched it because I felt like that approach was a little confusing with the amount of layers that the data had to be passed through (VerifyUserState -> build method of the AmplifyAuthenticator -> VerifiedUserScreen -> VerifiedUserForm -> set state on VerifyUserViewModel) but I could simplify that by having the VerifiedUserScreen update the VerifyUserViewModel with the initial data.

I probably won't get back to this PR for a couple days, but I might go with that approach when I get back to it.

packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
```diff
@@ -228,6 +235,32 @@ class StateMachineBloc {
...
+              unverifiedAttributeKeys: unverifiedAttributeKeys,
+            ),
+          ),
+        );
+        yield AuthFlow(screen: AuthScreen.verifyUser);
+      } else {
+        yield const Authenticated();
+      }
+    } on Exception catch (e) {
+      // TODO: How should exceptions be handled in this flow
```
haverchuck, line 255: It seems like the exceptions that would come back from this would not be useful for an end user, and should only happen if there is some sort of internal issue with our code (or maybe with a call to Cognito for attributes, if that's what's happening here)?

If that's the case, maybe we should just log these exceptions but not display them view the ExceptionController?

packages/amplify_authenticator/lib/src/text_customization/button_resolver.dart
```diff
@@ -35,6 +35,12 @@ class ButtonResolver {
   /// Label of button for sending a confirmation code
   StringResolver sendCode = (_) => "Send Code";
 
+  /// Label of button to verify a user after sign in
+  StringResolver verifyUser = (_) => "Verify";
```
haverchuck, line 39: I just realized we need to standardize the casing on these StringResolvers... some use camel, some use snake (general comment, not really applicable to this PR).

haverchuck (changes requested):
In general looks good, left a couple comments.

haverchuck (commented):
In general looks good, left a couple comments.

dnys1 (approved):

packages/amplify_authenticator/lib/src/services/amplify_auth_service.dart
Jordan-Nelson: Updated this to something similar 👍

haverchuck (approved):

PR #895, opened 2021-09-16
break(datastore): cannot saving boolean as integer in SQLite

haverchuck:
Issue #, if available: #752

Description of changes:

Convert bool filed type to java Boolean object type instead of Kotlin primitive Boolean type.

Reason: amplify-android uses Java Boolean class but not expecting Kotlin primitive boolean. This will let amplify-android correctly create boolean value column with integer type, and further to fix the issue described in #752

Note: This fix introduces breaking changes as the existing bug writes double and boolean type values as string type in Android. Directly applying this fix may cause type error while read/write local DB, to resolve would need to invoke Amplify.DataStore.clear() to recreate local DB tables.


HuiSF (approved):
Thanks for cleaning the RC branch!

fjnoyp (approved):

Issue #880, opened 2021-09-14
how to pass preferPrivateSession flag?


SuTran:
Hi team,

I found an update that helps user not to see a web view displayed when they sign out. 
https://github.com/aws-amplify/amplify-ios/releases/tag/v1.6.0

```
Amplify.Auth.signInWithWebUI(presentationAnchor: self.view.window!, 
                             options: .preferPrivateSession()) { ... }
```

It can help to resolve issue like https://github.com/aws-amplify/amplify-ios/issues/1016

Have we apply this new config into `amplify-flutter`? Thanks

fjnoyp:
Hi @SuTran thanks for reaching out we are looking into this issue.

fjnoyp:
Hi thanks for your patience.  It is currently not possible to pass that flag in amplify flutter.  I looked into adding possible support for that but unfortunately Android doesn't currently support this flag, and we don't want our amplify flutter api to be platform dependent.  

Thus, at the moment we will not be adding support for that flag.  We will keep you updated regarding any changes to that though.  

@SuTran

SuTran:
@fjnoyp thanks for your response. May I work on this too?

fjnoyp:
Sure @SuTran we'd be happy to have you contribute!  I'll respond tomorrow with what the API experience should look like and then you can submit a PR :)

SuTran:
Thanks @fjnoyp . I prepare this PR https://github.com/aws-amplify/amplify-flutter/pull/897 . Can you help to review if you have a chance.

fjnoyp:
Thanks @SuTran we'll continue the conversation on your PR.

dander11:
@fjnoyp is there any timeline for when this PR #897 will be released? It is currently blocking us from being able to release our app to production.

PR #897, opened 2021-09-17
feat(Auth) support preferPrivateSession flag

SuTran:
*Issue #880*

*Description of changes:*

### DIFF
--- a/packages/amplify_auth_cognito/ios/Classes/FlutterSignInWithWebUIRequest.swift
+++ b/packages/amplify_auth_cognito/ios/Classes/FlutterSignInWithWebUIRequest.swift
@@ -20,9 +20,13 @@ import amplify_core
 
 struct FlutterSignInWithWebUIRequest {
   var provider: AuthProvider?
+  var isPreferPrivateSession: Bool = false
 
   init(dict: NSMutableDictionary){
     self.provider = getAuthProvider(provider: dict["authProvider"] as! String?)
+    if (dict["isPreferPrivateSession"] != nil &&  dict["isPreferPrivateSession"] as! Bool) {
+        self.isPreferPrivateSession = dict["isPreferPrivateSession"] as! Bool
+    }
   }
 
   func getAuthProvider(provider: String?) -> AuthProvider? {
--- a/packages/amplify_auth_cognito/ios/Classes/FlutterSignInWithWebUIRequest.swift
+++ b/packages/amplify_auth_cognito/ios/Classes/FlutterSignInWithWebUIRequest.swift
@@ -24,8 +24,11 @@ struct FlutterSignInWithWebUIRequest {
 
   init(dict: NSMutableDictionary){
     self.provider = getAuthProvider(provider: dict["authProvider"] as! String?)
-    if (dict["isPreferPrivateSession"] != nil &&  dict["isPreferPrivateSession"] as! Bool) {
-        self.isPreferPrivateSession = dict["isPreferPrivateSession"] as! Bool
+    var options = dict["options"] as! Dictionary<String, Any>?
+    if (options != nil) {
+      if (options!["isPreferPrivateSession"] != nil && options!["isPreferPrivateSession"] as! Bool) {
+        self.isPreferPrivateSession = options!["isPreferPrivateSession"] as! Bool
+      }
     }
   }
 
--- a/packages/amplify_auth_cognito/ios/Classes/FlutterSignInWithWebUIRequest.swift
+++ b/packages/amplify_auth_cognito/ios/Classes/FlutterSignInWithWebUIRequest.swift
@@ -24,7 +24,6 @@ struct FlutterSignInWithWebUIRequest {
 
   init(dict: NSMutableDictionary){
     self.provider = getAuthProvider(provider: dict["authProvider"] as! String?)
-    var options = dict["options"] as! Dictionary<String, Any>?
     if let options = dict["options"] as? [String: Any] {
        self.isPreferPrivateSession = options["isPreferPrivateSession"] as! Bool
     }
--- a/packages/amplify_auth_cognito/ios/Classes/FlutterSignInWithWebUIRequest.swift
+++ b/packages/amplify_auth_cognito/ios/Classes/FlutterSignInWithWebUIRequest.swift
@@ -25,10 +25,8 @@ struct FlutterSignInWithWebUIRequest {
   init(dict: NSMutableDictionary){
     self.provider = getAuthProvider(provider: dict["authProvider"] as! String?)
     var options = dict["options"] as! Dictionary<String, Any>?
-    if (options != nil) {
-      if (options!["isPreferPrivateSession"] != nil && options!["isPreferPrivateSession"] as! Bool) {
-        self.isPreferPrivateSession = options!["isPreferPrivateSession"] as! Bool
-      }
+    if let options = dict["options"] as? [String: Any] {
+       self.isPreferPrivateSession = options["isPreferPrivateSession"] as! Bool
     }
   }
 
--- a/packages/amplify_auth_cognito/lib/src/CognitoSignIn/CognitoSignInWithWebUIOptions.dart
+++ b/packages/amplify_auth_cognito/lib/src/CognitoSignIn/CognitoSignInWithWebUIOptions.dart
@@ -16,13 +16,17 @@
 import 'package:amplify_auth_plugin_interface/amplify_auth_plugin_interface.dart';
 
 class CognitoSignInWithWebUIOptions extends SignInWithWebUIOptions {
-  bool? isPreferPrivateSession;
-  CognitoSignInWithWebUIOptions({this.isPreferPrivateSession}) : super();
-  Map<String, dynamic> serializeAsMap() {
-    final Map<String, dynamic> pendingRequest = {
-      'isPreferPrivateSession': isPreferPrivateSession
-    };
-    pendingRequest.removeWhere((_, v) => v == null);
-    return pendingRequest;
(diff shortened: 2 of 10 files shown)


packages/amplify_flutter/lib/categories/amplify_auth_category.dart
```diff
@@ -207,8 +207,10 @@ class AuthCategory {
         : throw _pluginNotAddedException("Auth");
   }
 
-  Future<SignInResult> signInWithWebUI({AuthProvider? provider}) {
-    var request = SignInWithWebUIRequest(provider: provider);
+  Future<SignInResult> signInWithWebUI(
```
fjnoyp, line 210: Could you update the category API to match this signature instead: 

Future<SignInResult> signInWithWebUI({AuthProvider? provider, SignInWithWebUIOptions options})

Then within amplify_auth_plugin_interface you could create the SignInWithWebUIOptions class. 

Then within amplify_auth_cognito, you could create CognitoSignInWIthWebUIOptions which would have the field "isPreferPrivateSession". 

We discussed internally about this best long term approach for this change, and believe that having this options object would give us the greatest long term flexibility while also avoiding cognito specific and platform specific implementations from leaking into the auth category file.

fjnoyp (changes requested):
Hi @SuTran thanks a ton for getting this PR out so fast!  We really appreciate your help with implementation here. 

I left a comment about the API changes.  We want to use an options parameter instead of directly passing the boolean.  More explanations in the comments.

packages/amplify_flutter/lib/categories/amplify_auth_category.dart
SuTran: Thanks @fjnoyp for your feedback.

I will modify the `amplify_auth_category.dart` to have `SignInWithWebUIOptions` options. It makes sense to me too.

I saw within `amplify_auth_plugin_interface.dart` there is a class called `SignInWithWebUIRequest`. If I created `SignInWithWebUIOptions`, so what is the key different between "Request" and "Options", I think design this way will lead to confusion later when we adding more params to support new features.

Can you share more thoughts on this?

fjnoyp: Hi.  Thanks!

The key difference is that "Request" is internal to our library and not exposed in the public API.  While "Options" would be a method parameter in the public API, `amplify_auth_category`

SuTran: Thanks. I have updated the code. Can you help to check

fjnoyp (approved):
Looks great @SuTran thanks again for submitting this PR. 

I'll contact some others on my team to get you more approvals for a merge.

packages/amplify_auth_cognito/ios/Classes/FlutterSignInWithWebUIRequest.swift
```diff
@@ -20,9 +20,16 @@ import amplify_core
...
+  var isPreferPrivateSession: Bool = false
 
   init(dict: NSMutableDictionary){
     self.provider = getAuthProvider(provider: dict["authProvider"] as! String?)
+    var options = dict["options"] as! Dictionary<String, Any>?
+    if (options != nil) {
+      if (options!["isPreferPrivateSession"] != nil && options!["isPreferPrivateSession"] as! Bool) {
+        self.isPreferPrivateSession = options!["isPreferPrivateSession"] as! Bool
+      }
+    }
```
dnys1, line 32:
```suggestion
    if let options = dict["options"] as? [String: Any] {
       self.isPreferPrivateSession = options["isPreferPrivateSession"] as! Bool
    }
```

packages/amplify_auth_cognito/lib/src/CognitoSignIn/CognitoSignInWithWebUIOptions.dart
```diff
@@ -0,0 +1,28 @@
...
+class CognitoSignInWithWebUIOptions extends SignInWithWebUIOptions {
+  bool? isPreferPrivateSession;
+  CognitoSignInWithWebUIOptions({this.isPreferPrivateSession}) : super();
+  Map<String, dynamic> serializeAsMap() {
+    final Map<String, dynamic> pendingRequest = {
+      'isPreferPrivateSession': isPreferPrivateSession
+    };
+    pendingRequest.removeWhere((_, v) => v == null);
+    return pendingRequest;
+  }
```
dnys1, line 27:
```suggestion
  @override
  Map<String, dynamic> serializeAsMap() => <String, dynamic>{
        'isPreferPrivateSession': isPreferPrivateSession,
      };
```

packages/amplify_auth_plugin_interface/lib/src/SignIn/SignInWithWebUIOptions.dart
```diff
@@ -0,0 +1,22 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+abstract class SignInWithWebUIOptions {
+  const SignInWithWebUIOptions();
+  Map<String, dynamic> serializeAsMap() {
+    throw UnimplementedError(
+        'serializeAsMap() has not been implemented on SignInWithWebUIOptions.');
+  }
```
dnys1, line 21: Prefer abstract method

```suggestion
  Map<String, dynamic> serializeAsMap();
```

packages/amplify_auth_cognito/lib/src/CognitoSignIn/CognitoSignInWithWebUIOptions.dart
```diff
@@ -0,0 +1,28 @@
...
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import 'package:amplify_auth_plugin_interface/amplify_auth_plugin_interface.dart';
+
+class CognitoSignInWithWebUIOptions extends SignInWithWebUIOptions {
+  bool? isPreferPrivateSession;
```
dnys1, line 19: Copied from iOS code:

```suggestion
  /// iOS-only: Starts the webUI signin in a private browser session, if supported by the current browser.
  ///
  /// Note that this value internally sets `prefersEphemeralWebBrowserSession` in ASWebAuthenticationSession.
  /// As per Apple documentation, Whether the request is honored depends on the user’s default web browser.
  /// Safari always honors the request.
  /// 
  /// Defaults to `false`.
  final bool isPreferPrivateSession;
```

packages/amplify_auth_cognito/lib/src/CognitoSignIn/CognitoSignInWithWebUIOptions.dart
```diff
@@ -0,0 +1,28 @@
...
+  CognitoSignInWithWebUIOptions({this.isPreferPrivateSession}) : super();
```
dnys1, line 20: Prefer default value here

```suggestion
  const CognitoSignInWithWebUIOptions({this.isPreferPrivateSession = false});
```

dnys1 (commented):
Looks good, thanks for the PR!

Needs some minor style things and doc comments, but otherwise it's good to go 👍

packages/amplify_auth_cognito/lib/src/CognitoSignIn/CognitoSignInWithWebUIOptions.dart
SuTran: Thank you!. I have updated the code

packages/amplify_auth_plugin_interface/lib/src/SignIn/SignInWithWebUIOptions.dart
SuTran: Thank you! I have updated the code

packages/amplify_auth_cognito/ios/Classes/FlutterSignInWithWebUIRequest.swift
SuTran: Thank you! I have updated the code

packages/amplify_auth_cognito/lib/src/CognitoSignIn/CognitoSignInWithWebUIOptions.dart
SuTran: Thank you! I have updated the code

packages/amplify_auth_cognito/lib/src/CognitoSignIn/CognitoSignInWithWebUIOptions.dart
SuTran: Thank you! I have updated the code

packages/amplify_auth_cognito/ios/Classes/FlutterSignInWithWebUIRequest.swift
```diff
@@ -20,9 +20,14 @@ import amplify_core
 
 struct FlutterSignInWithWebUIRequest {
   var provider: AuthProvider?
+  var isPreferPrivateSession: Bool = false
 
   init(dict: NSMutableDictionary){
     self.provider = getAuthProvider(provider: dict["authProvider"] as! String?)
+    var options = dict["options"] as! Dictionary<String, Any>?
```
dnys1, line 27: Please remove this line

dnys1 (approved):
Looks great! Minor change and it's good to go!

Thank you again 😄

PR #905, opened 2021-09-22
GraphQL Stream -> RC

dnys1:
*Issue #, if available:*

*Description of changes:*


haverchuck (approved):

Jordan-Nelson (approved):

packages/amplify_api/lib/graphql/graphql_subscription_event.dart
```diff
@@ -0,0 +1,93 @@
...
+  done,
+
+  /// Triggered when a platform error occurs for a subscription.
+  ///
+  /// Currently only valid in Dart.
+  error,
+}
+
+/// Helper functions for [GraphQLSubscriptionEventType].
+extension GraphQLSubscriptionEventTypeX on GraphQLSubscriptionEventType {
```
fjnoyp, line 34: Minor nit: naming seems strange here with "TypeX" 

Should name match function more like EventTypeStringParser or GraphQLSubscriptionEventTypeExtension?

fjnoyp (approved):
Looks great.  Just one minor non blocking nit comment regarding extension class naming

packages/amplify_api/lib/graphql/graphql_subscription_event.dart
dnys1: Hmm - could just be a personal preference thing. I've always liked `TypeX` for extensions. Typically, they aren't referred to by name, only when there are static methods associated with them like in this case. 

For a public type which is referred to by name, I agree there are probably more descriptive names/approaches. This one isn't exported, though, it's only an internal helper.

Issue #891, opened 2021-09-16
[DataStore] iOS and Android have different default behavior for pagination


Jordan-Nelson:
**Describe the bug**

**To Reproduce**
1. run an app with datastore enabled and any schema with at least one model
2. save a few hundred items to data store
3. run a query with no predicates or pagination
4. On Android, all models are returned
5. on iOS, only 100 items are returned

**Expected behavior**
With no pagination specified, all items should be returned.

**Additional context**

The [QueryPaginationBuilder](https://github.com/aws-amplify/amplify-flutter/blob/1959ea13fb6287a544987021e5225576059e10d6/packages/amplify_datastore/ios/Classes/types/query/QueryPaginationBuilder.swift) needs to be updated to return `nil` when the serialized data is nil and/or contains no page/limit.

PR #906, opened 2021-09-22
fix(datastore): remove default pagination behavior on iOS

Jordan-Nelson:
*Issue #, if available:* https://github.com/aws-amplify/amplify-flutter/issues/891

*Description of changes:*
- remove the default behavior in QueryPaginationBuilder on iOS to use a page of 0 and a limit of 100

Amplify-flutter, Amplify-iOS, and Amplify-Android all allow for null/nil page sizes in the `query()` API. Amplify-iOS and Amplify-iOS do not provide any default behavior if null/nil is provided. Currently in Amplify-flutter there are default values used on iOS, but not Android.

### DIFF
--- a/packages/amplify_datastore/ios/Classes/types/query/QueryPaginationBuilder.swift
+++ b/packages/amplify_datastore/ios/Classes/types/query/QueryPaginationBuilder.swift
@@ -18,7 +18,7 @@ import Amplify
 
 public class QueryPaginationBuilder {
    
-    static func fromSerializedMap(_ serializedMap: [String: Any]?) -> QueryPaginationInput {
+    static func fromSerializedMap(_ serializedMap: [String: Any]?) -> QueryPaginationInput? {
         var page: UInt = 0, limit: UInt = QueryPaginationInput.defaultLimit;
         if let data = serializedMap {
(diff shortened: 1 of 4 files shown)


packages/amplify_datastore/ios/Classes/types/query/QueryPaginationBuilder.swift
```diff
@@ -18,7 +18,7 @@ import Amplify
 
 public class QueryPaginationBuilder {
    
-    static func fromSerializedMap(_ serializedMap: [String: Any]?) -> QueryPaginationInput {
+    static func fromSerializedMap(_ serializedMap: [String: Any]?) -> QueryPaginationInput? {
         var page: UInt = 0, limit: UInt = QueryPaginationInput.defaultLimit;
```
HuiSF, line 22: nit: can probably move `var page: UInt = 0, limit: UInt = QueryPaginationInput.defaultLimit;` into the `if let data = serializedMap` block.

HuiSF (approved):
Thanks for fixing this discrepancy! One nit-picking comment.

dnys1 (approved):
LGTM

Issue #814, opened 2021-08-18
DataStore throws an exception parsing the value returned from type [AWSTimestamp]


Jordan-Nelson:
**Describe the bug**
DataStore throws an exception parsing the value returned from type `[AWSTimestamp]`. This is due to `TemporalTimestamp.fromSeconds` expecting an int, but receiving a double. This is only seen on Android.

example schema

```graphql
type MyModel @model {
  id: ID!
  timestampType: AWSTimestamp # <-- This works as expected
  timestampListType: [AWSTimestamp] # <-- This will cause an exception
}
```

**To Reproduce**
1. Create a model such as the one above in your schema
2. Attempt to save an object
```dart
var now = DateTime.now();
MyModel testModel = MyModel(
  timestampType: TemporalTimestamp(now),
  timestampListType: [TemporalTimestamp(now), TemporalTimestamp(now)],
);
await Amplify.DataStore.save(testModel);
```
3. Attempt to query that model
```dart
var models = await Amplify.DataStore.query(MyModel.classType);
```

**Expected behavior**
An exception should not be thrown

**Screenshots**
The following screen shot shows that AWSTimestamp is being returned as int, but a list of AWSTimestamp ([AWSTimestamp]) is being returned as double 

<img width="546" alt="Screen Shot 2021-08-18 at 10 07 59 AM" src="https://user-images.githubusercontent.com/20613561/129914609-fd773a2b-10c7-4029-9196-86626ea5afa8.png">

**Platform**
Amplify Flutter current supports iOS and Android.  This issue is reproducible in (check all that apply):
[x] Android
[] iOS

**NOTE**
Prior to this issue being marked as resolved, the DataStore integration tests that are skipped due to this issue should be updated. 

See: https://github.com/aws-amplify/amplify-flutter/blob/main/packages/amplify_datastore/example/integration_test/model_type_test.dart

Jordan-Nelson:
### Cause
The cause of this is that the values returned from the native side (on Android) are doubles. That can be seen in the screenshot below.

<img width="546" alt="Screen Shot 2021-08-18 at 10 07 59 AM" src="https://user-images.githubusercontent.com/20613561/129916877-c75d3c91-e4da-4979-9a7d-89bf8cfab373.png">

The code to parse the timestamp expects an int, which can be seen below.

[...]

PR #847, opened 2021-08-31
feat(datastore): Add CustomType functionality

HuiSF:
*Issue #, if available:* #260

*Description of changes:*

Implement `CustomType` functionality.

**Dependencies**
- PR [#1459](https://github.com/aws-amplify/amplify-android/pull/1459) in amplify-android repository (Android unit tests will break until integrate this PR)
- PR [#234](https://github.com/aws-amplify/amplify-codegen/pull/234) in amplify-codegen repository

Tasks
- [x] amplify-flutter Android implementation (commit: https://github.com/aws-amplify/amplify-flutter/commit/001fb3e5b764199a7edb8d0e17d440bbcf9f1372)
- [x] amplify-flutter Android code change unit tests (commit: https://github.com/aws-amplify/amplify-flutter/commit/001fb3e5b764199a7edb8d0e17d440bbcf9f1372)
- [x] amplify-flutter iOS implementation (commit: https://github.com/aws-amplify/amplify-flutter/pull/847/commits/92f9ea32a982a3c91cb2e48e46eb22db35b03895)
- [x] amplify-flutter iOS unit tests
- [x] amplify-flutter Dart implementation (commit: https://github.com/aws-amplify/amplify-flutter/commit/001fb3e5b764199a7edb8d0e17d440bbcf9f1372)
- [x] amplify-flutter Dart unit tests (commit: https://github.com/aws-amplify/amplify-flutter/commit/001fb3e5b764199a7edb8d0e17d440bbcf9f1372)
- [x] Upgrade amplify-android to the latest version
- [x] Add integration tests

----
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #920, opened 2021-09-27
feat(datastore): Add CustomType functionality (#847)

HuiSF:
Merge feature branch of CustomType into RC.

Update: Found an issue that `AWSTimestamp` typed field returned with `Double` type value instead of `Long`. The commit [`20f7085`](https://github.com/aws-amplify/amplify-flutter/pull/920/commits/20f7085def7608f2f9d3f8ac44b8c53d1fe7d58c) fixes this issue. This fix also resolves issue #814.


packages/amplify_datastore/CHANGELOG.md
```diff
@@ -9,6 +9,17 @@
 
   NOTE: Data stored in local database and not synced to cloud will be lost, as [local migration is not supported](https://docs.amplify.aws/lib/datastore/schema-updates/q/platform/flutter/#local-migrations).
 
+- Adding CustomType functionality introduces a breaking change to the amplify-codegen generated `ModelProvider.dart`, as a new class member `customTypeSchemas` has been added to `ModelProviderInterface.dart`.
+
+  **How to migrate?**
```
Jordan-Nelson, line 14: This is necessary even if they aren't using custom types, correct? I just wanted to confirm.

packages/amplify_datastore/example/lib/models/AllScalarTypesListCustomType.dart
```diff
@@ -0,0 +1,383 @@
...
+// ignore_for_file: public_member_api_docs
+
+import 'ModelProvider.dart';
+import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart';
+import 'package:collection/collection.dart';
+import 'package:flutter/foundation.dart';
+
+/** This is an auto generated class representing the AllScalarTypesListCustomType type in your schema. */
+@immutable
+class AllScalarTypesListCustomType {
```
Jordan-Nelson, line 25: Can you add the new models to https://github.com/aws-amplify/amplify-flutter/blob/main/packages/amplify_datastore/example/tool/schema.graphql?

That schema is used in the provisioning script for DataStore integ tests. It helps to add them there so that the model files can be re-generated. I also think it will be more important once we have integ tests that sync to the cloud.

packages/amplify_datastore/example/lib/models/ModelProvider.dart
```diff
@@ -68,8 +74,10 @@ export 'IntListTypeModel.dart';
 export 'IntTypeModel.dart';
 export 'JSONListTypeModel.dart';
 export 'JSONTypeModel.dart';
+export 'ModelNestedCustomType.dart';
 export 'MultiTypeModel.dart';
 export 'Post.dart';
+export 'SimpleCustomType.dart';
```
Jordan-Nelson, line 80: I think this is okay for now, but I think there is a limit of 33 model types in order for cloud sync to work, and we are either near or at that limit.

Jordan-Nelson: We could probably combine some of the types that I created for integ tests. I created a model for each type just to make the tests a little bit more independent, but it isn't 100% necessary. We could also just split these into multiple apps at some point, which I think we want to do eventually anyway to test different configs.

packages/amplify_datastore/CHANGELOG.md
HuiSF: Yes that's correct.

packages/amplify_datastore/example/lib/models/AllScalarTypesListCustomType.dart
HuiSF: Oops I think I forgot to commit the change of schema.graphql 😅

HuiSF: Added the missing changes.

cshfang (approved):
I think this lgtm since this cherry picking things we've already reviewed. Let's just make sure to test it

Jordan-Nelson (approved):

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterCustomTypeSchema.kt
```diff
@@ -0,0 +1,33 @@
...
+
+package com.amazonaws.amplify.amplify_datastore.types.model
+
+import com.amplifyframework.core.model.CustomTypeSchema
+
+data class FlutterCustomTypeSchema(val map: Map<String, Any>) {
+    val name: String = map["name"] as String
+    private val pluralName: String? = map["pluralName"] as String?
+    private val fields: Map<String, FlutterCustomTypeField> = (map["fields"] as Map<String, Any>)
+        .mapValues { FlutterCustomTypeField(it.value as Map<String, Any>) }
```
cshfang, line 24: Nit:
> the latter looks better
```suggestion
    private val fields: Map<String, FlutterCustomTypeField> = (map["fields"] as Map<String, Map<String, Any>>)
        .mapValues { FlutterCustomTypeField(it.value) }
```

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -512,15 +510,146 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
...
+        return null
+    }
+
+    private fun registerSchemas(request: Map<String, Any>) {
+        val modelSchemas: List<Map<String, Any>> = request["modelSchemas"].safeCastToList()!!
+        val customTypeSchemas: List<Map<String, Any>> = request["customTypeSchemas"].safeCastToList() ?: emptyList()
+        modelProvider.setVersion(request["modelProviderVersion"] as String)
+
+        val flutterModelSchemaList =
+            modelSchemas.map { modelSchema -> FlutterModelSchema(modelSchema) }
```
cshfang, line 632: Nit: Could use `it` here and for the customTypeSchema.map below 🤷

cshfang (approved):

cshfang (approved):

PR #924, opened 2021-09-28
update graphql feature branch with commits from main

ragingsquirrel3:
The graphql helper branch (intern project) is a few months behind main. This PR merges the original helper branch into a new branch which was made off main. Note that most of this code is Elijah's code which was already reviewed, along with conflict resolution and updates to pass analysis, which was added in main for the api category but not for these files, yet.

Note: I ran the API integration tests with this code and it passed.


packages/amplify_api/pubspec.yaml
```diff
@@ -14,6 +14,9 @@ dependencies:
     sdk: flutter
   plugin_platform_interface: ^2.0.0
 
+  # TODO: Datastore dependencies temporarily added in API. Eventually they should be moved to core or otherwise reconciled to avoid duplication.
+  amplify_datastore_plugin_interface: 0.2.0
```
ragingsquirrel3, line 18: even though this is temp, should prob be latest version here, will update

packages/amplify_api_plugin_interface/lib/src/graphql/graphql_request.dart
```diff
@@ -13,19 +13,27 @@
...
-  final String document;
-  final Map<String, dynamic> variables;
-  final String cancelToken = UUID.getUUID();
 
-  GraphQLRequest({
-    this.apiName,
-    required this.document,
-    this.variables = const <String, dynamic>{},
-  });
+  String document;
```
ragingsquirrel3, line 25: should add final keyword here I suppose, my bad, messed this up during conflict resolution

packages/amplify_api_plugin_interface/lib/src/types/pagination/model_pagination.dart
```diff
@@ -0,0 +1,21 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+class ModelPagination {
```
ragingsquirrel3, line 16: callout: Do we still want this class? It makes the customer-facing API a little verbose just to add a limit, also adds files/cost/weight/electricity.

packages/amplify_api_plugin_interface/pubspec.yaml
```diff
@@ -15,6 +15,9 @@ dependencies:
   meta: ^1.3.0
   uuid: ^3.0.1
 
+  # TODO: Datastore dependencies temporarily added in API. Eventually they should be moved to core or otherwise reconciled to avoid duplication.
+  amplify_datastore_plugin_interface: 0.2.0
```
ragingsquirrel3, line 19: same, will update to latest for temp use

packages/amplify_api_plugin_interface/lib/src/types/pagination/model_pagination.dart
dnys1: It should also have a `nextToken` field, so I think that might be why it was originally a class maybe?

dnys1 (approved):
Looks good. Sorry for the delay!

packages/amplify_api_plugin_interface/lib/src/types/pagination/model_pagination.dart
ragingsquirrel3: There was some discussion about that. For now, I think I will leave this as-is and we can have the discussion about this and/or exposing the nextToken param in the PR for the pagination helpers as it will be easier to have that discussion with the context that the PR will provide.

packages/amplify_api/example/integration_test/resources/Blog.dart
```diff
@@ -0,0 +1,142 @@
+/*
```
fjnoyp, line 1: Our codegen models are being increasingly scattered across multiple packages. 

1) Datastore Plugin Interface: test/testData 
2) Datastore: example/lib/models 
3) Datastore: test/test_models 

4) API: integration_test/resources/... 
5) API: test/resources/....

We should create a separate package that contains all Dart Codegen models so that any future updates to these models are properly propagated across all 5 of these cases.  At this point there is an ever increasing chance that we'll miss a set of models to update :/

fjnoyp (approved):
Looks good.  Made callout about the scattering of our Dart codegen models.  Probably don't need to address in this PR but we need to resolve it sometime down the line.

PR #946, opened 2021-10-05
Graphql helpers query pred updated

ragingsquirrel3:
This PR adds additional pagination helpers to the graphql helpers branch. Namely, adds `hasNextResult` and `getRequestForNextResult` to paginated result. Users can call these on a graphql response to get requests for additional pages of data.

### DIFF
--- a/packages/amplify_api/lib/src/graphql/graphql_response_decoder.dart
+++ b/packages/amplify_api/lib/src/graphql/graphql_response_decoder.dart
@@ -54,9 +54,11 @@ class GraphQLResponseDecoder {
           recoverySuggestion:
               "Current GraphQLResponse is non-nullable, please ensure item exists before fetching");
     }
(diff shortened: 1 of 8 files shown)


packages/amplify_api/lib/src/graphql/graphql_response_decoder.dart
```diff
@@ -54,9 +54,11 @@ class GraphQLResponseDecoder {
           recoverySuggestion:
               "Current GraphQLResponse is non-nullable, please ensure item exists before fetching");
     }
+    if (request.variables['limit'] != null) {
```
ragingsquirrel3, line 57: callout here: Throwing the limit in the JSON was a lazy way to pass the limit to `PaginatedResult`, which will they use it to make next request. Would it be better to `PaginatedResult` more explicitly somehow? Maybe it could have a setter?

PR #947, opened 2021-10-05
Graphql helpers pagination helpers

ragingsquirrel3:
This PR has additional pagination helpers `getRequestForNextResult()` and `hasNextResult` on `PaginatedResult`. When the user gets a paginated response from API graphql helpers, they can use these helpers to get a pre-populated request for the next page of data.

### DIFF
--- a/packages/amplify_api/lib/src/graphql/graphql_response_decoder.dart
+++ b/packages/amplify_api/lib/src/graphql/graphql_response_decoder.dart
@@ -54,9 +54,11 @@ class GraphQLResponseDecoder {
           recoverySuggestion:
               "Current GraphQLResponse is non-nullable, please ensure item exists before fetching");
     }
+    if (request.variables['limit'] != null) {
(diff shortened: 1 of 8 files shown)


packages/amplify_api/lib/src/graphql/graphql_response_decoder.dart
```diff
@@ -54,9 +54,11 @@ class GraphQLResponseDecoder {
           recoverySuggestion:
               "Current GraphQLResponse is non-nullable, please ensure item exists before fetching");
     }
+    if (request.variables['limit'] != null) {
```
ragingsquirrel3, line 57: Callout here: throwing the limit in the json was a lazy way to pass to paginated result, though I think it needs another solution. I suppose there could be an error if the data JSON already has limit defined for some reason? A more explicit way would be I guess to provide a setter for it as an integer on the paginated result, though that will make things more complicated.

Issue #936, opened 2021-10-04
DataStore not syncing


calvinmuller:
**Describe the bug**
It seems when using @auth() the subscriptions fail to connect and then send the subsequent updates.

**To Reproduce**
1. `schema.json`
```
type Todo @model @auth(rules: [{allow: owner, operations: [create, delete, update, read]}]) {
  id: ID!
  name: String!
  description: String
}

type Boards @model @auth(rules: [{ allow: owner }]) {
  id: ID!
  name: String!
  description: String
  length: String
  width: String
  manufacturer: String
  volume: String
  images: [String]
}
```
2.  amplify models codegen
3. Start up app (already logged in) 

**Expected behavior**
It should sync and not show odd errors in the console

**Platform**
Amplify Flutter current supports iOS and Android.  This issue is reproducible in (check all that apply):
[] Android
[x] iOS

** Flutter **
```
[✓] Flutter (Channel stable, 2.5.1, on macOS 11.2.3 20D91 darwin-x64, locale en-ZA)
    • Flutter version 2.5.1 at /usr/local/Caskroom/flutter/2.2.3/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision ffb2ecea52 (2 weeks ago), 2021-09-17 15:26:33 -0400
    • Engine revision b3af521a05
    • Dart version 2.14.2

```

**Smartphone (please complete the following information):**
 - Device: iPhone 11
 - Os: iOS14.6

**Additional context**
I get this when launching my app and being signed in.
```

[StarscreamAdapter] socket.write - {"id":"3FB782F7-08D2-4F56-BAA1-5B69C38423E4","payload":{"data":"{\"query\":\"subscription OnCreateTodo($owner: String!) {\\n  onCreateTodo(owner: $owner) {\\n    id\\n    description\\n    name\\n    __typename\\n    _version\\n    _deleted\\n    _lastChangedAt\\n    owner\\n  }\\n}\",\"variables\":{\"owner\":\"\"}}","extensions":{"authorization":{"Authorization":"MYJWTTOKEN.IS.HERE","host":"xxx.appsync-api.eu-west-1.amazonaws.com"}}},"type":"start"}

[...]

Issue #935, opened 2021-10-02
Cannot receive subscription event from DataStore.observe with owner auth rule setting 


csnguyen-gmail:
**Describe the bug**
After I change model auth from private to **owner** like below
```dart
type Todo @model @auth(rules: [{ allow: owner }]) {
  id: ID!
  title: String!
  lastUpdated: AWSDateTime!
}
```
Then no longer receive the event from DataStore.observe.
Note: I can receive subscription events normally with the private auth rule.

**To Reproduce**
1. Sign in with **email-1** on device-1
2. Sign in with **email-1** on device-2
3. Create/Update/Delete some Todo items in device-1
4. Todo list on device-2 is unchanged due to not receiving subscription update events.

**Expected behavior**
In case of sign-in with the same user, once modify on one device then another device should receive update events also.

**Platform**
Amplify Flutter current supports iOS and Android.  This issue is reproducible in (check all that apply):
[x] Android
[x] iOS

<details>

  <summary>Output of <code>flutter doctor -v</code></summary>
  
  ```[✓] Flutter (Channel stable, 2.5.1, on macOS 11.6 20G165 darwin-arm, locale en-VN)
    • Flutter version 2.5.1 at /Users/csnguyen/Documents/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision ffb2ecea52 (2 weeks ago), 2021-09-17 15:26:33 -0400
    • Engine revision b3af521a05
    • Dart version 2.14.2

[✓] Android toolchain - develop for Android devices (Android SDK version 31.0.0)
    • Android SDK at /Users/csnguyen/Library/Android/sdk
    • Platform android-31, build-tools 31.0.0
    • Java binary at: /Applications/Android Studio.app/Contents/jre/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.10+0-b96-7281165)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 13.0, Build version 13A233
    • CocoaPods version 1.11.1

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2020.3)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
[...]

PR #955, opened 2021-10-08
fix(datastore): Add auth rule provider info to generate schema

HuiSF:
**Please do not merge for now, this fix requires corresponding amplify-codegen fix.**

*Issue #, if available:*

- #935 
- #936 

*Description of changes:*

Add `provider` field to `AuthRule` to allow codegen generated model schema obtain model auth rule provider information to  mitigate amplify-ios introduced breaking change.

This is a breaking change, which requries:

- amplify-codegen to insert model auth rule provider information into generated dart models
- amplify-flutter consumers need to upgrade both amplify-codegen and amplify-flutter to regenerate models

----


packages/amplify_datastore/android/src/test/kotlin/com/amazonaws/amplify/amplify_datastore/SchemaData.kt
```diff
@@ -170,6 +170,79 @@ val postAuthComplexSchema = ModelSchema.builder()
...
+                        .isRequired(false)
+                        .isArray(false)
+                        .build()
+
+        )
+    )
+    .authRules(
+        listOf(
+            AuthRule.builder()
+                .authProvider(AuthStrategy.Provider.API_KEY)
```
dnys1, line 237: Not 100% what this is testing but API key + owner Auth is not allowed by the CLI

HuiSF: Ahh OK, I can update the test data.

dnys1 (approved):

PR #993, opened 2021-10-21
chore: 0.2.6 release

dnys1:
### Fixes

- fix(datastore): Re-emit events on hot restart

### Features

- feat(datastore): Add observeQuery API
- feat(storage): Upload/download progress listener


HuiSF:
Looks good but we may need to update this file as well
https://github.com/aws-amplify/amplify-flutter/blob/a6ca33cd3c5fc6b15e6db6af991857f69ec57f37/example/pubspec.yaml#L28-L31

dnys1:
> Thanks, LGTM, assuming not updating lints package stuff w every release.

Yep, that will have it's own versioning since it has no deps and will not change frequently.

PR #1011, opened 2021-10-28
chore(test): Add test pkgs and integ tests for core

dnys1:
*Issue #, if available:*

*Description of changes:*
- Added integ tests for Amplify core to test configuration loading scenarios (must be run with Xcode)
- Added `amplify_test` package for common test utilities


packages/amplify_flutter/example/pubspec.yaml
```diff
@@ -21,9 +21,18 @@ dependencies:
     sdk: flutter
 
 dev_dependencies:
+  amplify_analytics_pinpoint: ^0.2.4
```
ragingsquirrel3, line 24: Should these versions be changed to match RC version?

packages/amplify_flutter/example/integration_test/config_test.dart
```diff
@@ -0,0 +1,282 @@
+import 'dart:io';
```
ragingsquirrel3, line 1: Should this (and all new dart files in this PR) have the big header with copyright info?

packages/amplify_flutter/example/pubspec.yaml
Jordan-Nelson: I think relative paths would be consistent with the other example projects. Not sure if there is a reason for this to be different?

dnys1: Fixed

packages/amplify_flutter/example/integration_test/config_test.dart
dnys1: Fixed

Jordan-Nelson (approved):

ragingsquirrel3 (approved):

PR #1017, opened 2021-11-02
chore: refactor model mutation processing in QuerySnapshot

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*

There are two goals of this PR:
1. Improve code readability for QuerySnapshot.withSubscriptionEvent().
2. Improve efficiency of processing mutation events, specifically in cases where the mutation doesn't result in a new QuerySnapshot (due to a query predicate). This PR removes a call to `list.indexWhere()` that was running on each mutation. The purpose of this was to lookup a model in the cached data by ID. In place of this, a map of models is cached and updated, which reflects the models in the list. Lookups on the map are O(1), opposed to O(n) for lookups in the list. With the updates, processing mutations that do not result in a new snapshot should no longer depend on the size of the snapshot. The time to process mutations that _do_ result in a new snapshot will likely increase due to the need to clone/update the map each time. 

*Perf Testing*

The theoretical performance impacts were validated via performance testing on an iPhone 11 in profile mode. A data set of ~7,000 Posts were synced to the device, with a query predicate that would result in about half of the 7,000 models resulting in a new QuerySnapshot. This is intended to simulate a high TPS scenario in which many of the mutations do not result in new QuerySnapshots.

Below are the details. With the changes, the amount of time it takes to process a mutation that doesn't result in a new snapshot is constant (does not depend on snapshot size). The amount of time it takes to process a snapshot that _does_ result in a new QuerySnapshot is increased as expected.

Avg time below is the average amount of time it took to process one model mutation event.

| Snapshot size 	| Mutation Result 	| Avg time (main branch) 	| Avg time (w/ PR changes) 	|
|---------------	|-----------------	|------------------------	|--------------------------	|
| ~100 items    	| No new snapshot 	| ~0.010 ms              	| ~0.010 ms                	|
| 5,000+ items  	| No new snapshot 	| **~0.250 ms**              	| **~0.010 ms**                	|
| ~100 items    	| New Snapshot    	| ~0.015 ms              	| ~0.020 ms                	|
| 5,000+ items  	| New Snapshot    	| ~0.250 ms              	| ~0.500 ms                	|

### DIFF
--- a/packages/amplify_datastore_plugin_interface/lib/src/types/models/query_snapshot.dart
+++ b/packages/amplify_datastore_plugin_interface/lib/src/types/models/query_snapshot.dart
@@ -31,7 +31,8 @@ class QuerySnapshot<T extends Model> {
   // A list of models sorted according to the value provided for sortBy
   final SortedList<T> _sortedList;
 
-  // A Set of modelIds, used to
+  // A map of models in the sorted list, used to effeciently perform
+  // lookups on the list by model id
   final Map<String, T> _models;
 
   /// A list of models from the local store at the time that the snapshot was generated
--- a/packages/amplify_datastore_plugin_interface/lib/src/types/models/query_snapshot.dart
+++ b/packages/amplify_datastore_plugin_interface/lib/src/types/models/query_snapshot.dart
@@ -31,6 +31,9 @@ class QuerySnapshot<T extends Model> {
   // A list of models sorted according to the value provided for sortBy
   final SortedList<T> _sortedList;
 
+  // A Set of modelIds, used to
+  final Map<String, T> _models;
+
   /// A list of models from the local store at the time that the snapshot was generated
   List<T> get items => _sortedList.toList();
 
@@ -45,10 +48,12 @@ class QuerySnapshot<T extends Model> {
 
   const QuerySnapshot._({
     required SortedList<T> sortedList,
+    required Map<String, T> models,
     required this.isSynced,
     this.where,
     this.sortBy,
-  }) : _sortedList = sortedList;
+  })  : _sortedList = sortedList,
+        _models = models;
 
   /// {@macro query_snapshot}
   factory QuerySnapshot({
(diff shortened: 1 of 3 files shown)


BramDriesen:
Seems like a great performance improvement 👍

packages/amplify_datastore_plugin_interface/lib/src/types/models/query_snapshot.dart
```diff
@@ -89,44 +110,60 @@ class QuerySnapshot<T extends Model> {
...
+    T item = event.item;
+    String id = item.getId();
+    bool itemMatchesPredicate = where == null || where!.evaluate(item);
+    bool snapshotContainsItem = _models.containsKey(id);
+
+    switch (event.eventType) {
+      case EventType.create:
+        // if the newly created item matches the predicate and the item is
+        // not already in the list, return a new snapshot with the new item
+        if (itemMatchesPredicate && !snapshotContainsItem) {
```
HuiSF, line 122: `!snapshotContainsItem` may cause ignoring pushing cloud synced model into the snapshot.
See [this issue](https://github.com/aws-amplify/amplify-flutter/issues/1355#issuecomment-1036669491).

packages/amplify_datastore_plugin_interface/lib/src/types/models/query_snapshot.dart
```diff
@@ -89,44 +110,60 @@ class QuerySnapshot<T extends Model> {
...
+            models: Map.from(_models)..updateItem(id, item),
+          );
+        }
+        break;
+      case EventType.update:
+        // if the updated item is in the list, update or remove based
+        // on the the query predicate
+        if (snapshotContainsItem) {
+          T currentItem = _models[id]!;
+          if (itemMatchesPredicate && currentItem != item) {
```
HuiSF, line 134: `currentItem != item` may cause ignoring pushing cloud synced model into the snapshot.
See https://github.com/aws-amplify/amplify-flutter/issues/1355#issuecomment-1036669491.

dnys1:
@Jordan-Nelson can this be closed?

PR #1020, opened 2021-11-03
Amplify authenticator/formfield refactor

haverchuck:
*Description of changes:*
1. Refactors form fields to allow for multiple input types (date, phone, radio, text).
2. Eliminates aliases and adds ability for end user to pick username attribute type when both are available.


packages/amplify_authenticator/lib/src/mixins/authenticator_date_field.dart
```diff
@@ -0,0 +1,71 @@
...
+import 'package:amplify_authenticator/src/widgets/form_field.dart';
+
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+
+mixin AuthenticatorDateField<FieldType,
+        T extends AuthenticatorFormField<FieldType, String, T>>
+    on AuthenticatorFormFieldState<FieldType, String, T> {
+  String _convertToDateString(DateTime dt) {
+    return "${dt.year.toString()}-${dt.month.toString().padLeft(2, '0')}-${dt.day.toString().padLeft(2, '0')}";
```
Jordan-Nelson, line 11: The `intl` package has a date formatter. You should be able to do something like 
```dart
import 'package:intl/intl.dart';

final DateFormat formatter = DateFormat('yyyy-MM-dd');

mixin AuthenticatorDateField ... {
  String _convertToDateString(DateTime dt) {
    return formatter.format(dt);
  }
```
}

packages/amplify_authenticator/lib/src/mixins/authenticator_date_field.dart
```diff
@@ -0,0 +1,71 @@
...
+        tooltip: 'Select birthdate',
+        onPressed: _pickTime,
+      ),
+      Expanded(
+          child: TextFormField(
+        inputFormatters: [LengthLimitingTextInputFormatter(10)],
+        style: enabled
+            ? null
+            : const TextStyle(
+                color: AmplifyColors.black20,
```
Jordan-Nelson, line 49: Do we want to hard code this color? I don't think the consumer will be able to modify this. I am not sure what our plan is for theming.

packages/amplify_authenticator/lib/src/mixins/authenticator_date_field.dart
```diff
@@ -0,0 +1,71 @@
...
+  Widget buildFormField(BuildContext context) {
+    final inputResolver = stringResolver.inputs;
+    final hintText = widget.hintText == null
+        ? inputResolver.resolve(context, widget.hintTextKey!)
+        : widget.hintText!;
+
+    final TextEditingController _controller =
+        TextEditingController(text: initialValue);
+    void _pickTime() {
+      showDatePicker(
```
Jordan-Nelson, line 24: *nit:*

```dart
      DateTime now = DateTime.now();
      showDatePicker(
              context: context,
              initialDate: now,
              firstDate: DateTime(now.year - 110),
              lastDate: now)
```

Jordan-Nelson (commented):
I am about halfway through. I will finish reviewing it later today. I have seen one bug so far, which is that the email is being used as the username even though I have a standard username/password config.

packages/amplify_authenticator/lib/src/enums/user_name_attribute.dart
```diff
@@ -0,0 +1,16 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+enum UsernameAttribute { email, phoneNumber }
```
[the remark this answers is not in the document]
haverchuck: Given that Cognito's usernameAttribute concept can sometimes create a bit of confusion, I wanted to keep this similar to [Cognito's definition](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cognito-userpool.html#cfn-cognito-userpool-usernameattributes) (so as not to create additional confusion)

packages/amplify_authenticator/lib/src/enums/verified_attributes.dart
```diff
@@ -0,0 +1,17 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+/// Uses snake case to conform with cognito service
```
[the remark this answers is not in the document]
haverchuck: I had this originally, but changed it because the two enums are used for very different purposes. I can see the value in consolidating though - I'll change these back.

packages/amplify_authenticator/lib/src/widgets/form_field.dart
```diff
@@ -95,73 +109,58 @@ abstract class AuthenticatorFormField<FieldType,
...
   /// Validates inputs of this form field.
-  FormFieldValidator<String>? get validator => null;
+  FormFieldValidator<FieldValue>? get validator => null;
 
   /// Whether to hide input.
   bool get obscureText => false;
 
   /// The keyboard type to use when this field is focused.
-  TextInputType get keyboardType => TextInputType.text;
+  TextInputType? get keyboardType => TextInputType.text;
```
dnys1, line 160: Doesn't need to be nullable - some field states just won't use it.

packages/amplify_authenticator/lib/src/widgets/form_field.dart
```diff
@@ -175,13 +174,16 @@ abstract class _AuthenticatorFormFieldState<FieldType,
   /// Maximum number of lines to use for error text.
   int get errorMaxLines => 1;
 
+  /// Value used by stateful widgets which require state for selected values
+  FieldValue? selectionValue;
```
dnys1, line 178: Only needed in specific form fields - I would just keep it there. Or better yet, in a radio group like before. Most form fields do not control their own selection - they belong to a set of form fields which together are told their selection by a parent.

packages/amplify_authenticator/lib/src/widgets/form_fields/confirm_sign_in_form_field.dart
```diff
@@ -205,47 +233,29 @@ class _ConfirmSignInFormFieldState extends _AuthenticatorFormFieldState<
...
 
   @override
-  Widget? get suffixIcon {
-    switch (widget.field) {
-      case ConfirmSignInField.password:
-        return visibilityToggle;
-      default:
-        return null;
-    }
+  Widget buildForm(BuildContext context) {
```
dnys1, line 257: Can delete this I think

packages/amplify_authenticator/example/lib/main.dart
```diff
@@ -78,15 +78,15 @@ class _MyAppState extends State<MyApp> {
...
-      signUpForm: SignUpForm.custom(
-        fields: [
-          SignUpFormField.username(
-            validator: _validateUsername,
-          ),
-          const SignUpFormField.password(),
-          const SignUpFormField.passwordConfirmation(),
-        ],
-      ),
+      // signUpForm: SignUpForm.custom(
```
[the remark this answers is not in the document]
haverchuck: The latter - this should get fixed today.

packages/amplify_authenticator/lib/src/l10n/generated/input_localizations_en.dart
```diff
@@ -127,4 +127,13 @@ class AuthenticatorInputLocalizationsEn
...
   String get rememberDevice => 'Remember Device?';
+
+  @override
+  String get usernameType => 'Select what you want to use as a username';
+
+  @override
+  String get usernameTypeEmail => 'Email';
+
+  @override
+  String get usernameTypePhoneNumber => 'Phone Number';
```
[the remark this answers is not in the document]
haverchuck: @dnys1 re: your comment about lint warnings - it did throw lint warnings which is why I left a comment, but I've moved to an extension.

packages/amplify_authenticator/lib/src/mixins/authenticator_phone_field.dart
```diff
@@ -0,0 +1,75 @@
...
+  @override
+  void initState() {
+    selectionValue = '1';
+    super.initState();
+  }
+
+  @override
+  Widget buildFormField(BuildContext context) {
+    final inputResolver = stringResolver.inputs;
+    List<int> countryCodeStrings = countryCodes
```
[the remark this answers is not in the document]
haverchuck: Moved this out of the buildFormField method ; we definitely need something better for this list in the long run. I'm still thinking it over.

packages/amplify_authenticator/lib/src/mixins/authenticator_radio_field.dart
```diff
@@ -0,0 +1,44 @@
+import 'package:amplify_authenticator/src/widgets/form_field.dart';
+
+import 'package:flutter/material.dart';
+
+mixin AuthenticatorRadioField<FieldType, FieldValue,
+        T extends AuthenticatorFormField<FieldType, FieldValue, T>>
+    on AuthenticatorFormFieldState<FieldType, FieldValue, T> {
+  void initState() {
+    selectionValue = widget.inputconfig!.selections[0].value as FieldValue;
```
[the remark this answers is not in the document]
haverchuck: I personally think it makes sense to preselect, since it could potentially save the end user an action and also obviates the need for a validator. But I'm fine with moving to null as an initial value if we agree that's better.

packages/amplify_authenticator/lib/src/mixins/authenticator_phone_field.dart
```diff
@@ -0,0 +1,75 @@
...
+import 'package:amplify_authenticator/src/widgets/form_field.dart';
+
+import 'package:flutter/material.dart';
+
+mixin AuthenticatorPhoneField<FieldType,
+        T extends AuthenticatorFormField<FieldType, String, T>>
+    on AuthenticatorFormFieldState<FieldType, String, T> {
+  @override
+  void initState() {
+    selectionValue = '1';
```
Jordan-Nelson, line 12: *nit:* Took me a minute to realize the significance of "1". If `countryCodeStrings` is taken out of the build method, maybe this can be `countryCodeStrings.first`

packages/amplify_authenticator/lib/src/widgets/form.dart
```diff
@@ -171,53 +171,75 @@ class SignUpForm extends AuthenticatorForm<SignUpForm> {
...
+          signUpFields.add(SignUpFormField.email());
+        } else if (usernameAttributes
+            .contains(AwsCognitoUsernameAttributes.phoneNumber)) {
+          signUpFields.add(SignUpFormField.phoneNumber());
+        }
+      } else if (usernameAttributes.length == 2) {
+        signUpFields.add(SignUpFormField.email());
+        signUpFields.add(SignUpFormField.phoneNumber());
+        signUpFields.add(SignUpFormField.selectedUserNameType());
+      }
```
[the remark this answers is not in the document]
haverchuck: Yeah - this is part of the code I'm reworking today.

packages/amplify_authenticator/lib/src/mixins/authenticator_radio_field.dart
```diff
@@ -0,0 +1,44 @@
...
+    super.initState();
+  }
+
+  @override
+  Widget buildFormField(BuildContext context) {
+    final inputResolver = stringResolver.inputs;
+    onChanged!(selectionValue!);
+    // return Text('Hi');
+    return Column(
+      children: <Widget>[
```
Jordan-Nelson, line 19: *nit:* I guess this is really just preference, but this is easier to understand imo

```dart
children: widget.inputconfig!.selections.map((selection) {
  return ListTile(...)
})
```

packages/amplify_authenticator/lib/src/mixins/authenticator_radio_field.dart
```diff
@@ -0,0 +1,44 @@
...
+  void initState() {
+    selectionValue = widget.inputconfig!.selections[0].value as FieldValue;
+    super.initState();
+  }
+
+  @override
+  Widget buildFormField(BuildContext context) {
+    final inputResolver = stringResolver.inputs;
+    onChanged!(selectionValue!);
+    // return Text('Hi');
```
Jordan-Nelson, line 17: I assume this can be removed

packages/amplify_authenticator/lib/src/widgets/field_config.dart
```diff
@@ -0,0 +1,31 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+*/
+
+import 'package:amplify_authenticator/amplify_authenticator.dart';
+
+/// Configuration for defining content of radio and switch types in [AuthenticatorFormField].
+class AuthenticatorInputConfig {
+  final List<InputSelection> selections;
```
Jordan-Nelson, line 20: This is implicitly `List<InputSelection<dynamic>>`. I think you want this to be `List<InputSelection<T>>`, and to make `AuthenticatorInputConfig` generic, `AuthenticatorInputConfig<T>`.

These is another location in the PR where inputconfig!.selections[0].value has to be cast to FieldValue. This is the reason that cast is needed.

packages/amplify_authenticator/lib/src/widgets/form_fields/sign_up_form_field.dart
```diff
@@ -19,298 +19,343 @@ part of authenticator.form_field;
...
+      );
+
+  static SignUpFormField selectedUserNameType(
+          {Key? key, FormFieldValidator<Enum>? validator}) =>
+      _SignUpRadioField(
+          key: key ?? keySelectedUsernameSignUpFormField,
+          titleKey: InputResolverKey.usernameType,
+          field: SignUpField.selectedUsername,
+          inputConfig: AuthenticatorInputConfig(selections: [
+            InputSelection<Enum>(
```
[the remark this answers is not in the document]
Jordan-Nelson: Actually, I think this really should be `InputSelection<SignUpField>`, and then `value` should be `SignUpField.email`.

The fact that only email and phone are listed will prevent anything else from being selected. I think `UsernameAttribute` could possibly be removed altogether. I think having both might be part of the reason that `Enum` was using in some locations.

This is also related to another comment I made that `AuthenticatorInputConfig` should be generic. I started making these changes locally, and it results in a cascade of related type changes that are probably easiest if they are all made at the same time.

The full code snippet of just this section I think should be something like

```dart
static SignUpFormField selectedUserNameType(
        {Key? key, FormFieldValidator<SignUpField>? validator}) =>
    _SignUpRadioField(
        key: key ?? keySelectedUsernameSignUpFormField,
        titleKey: InputResolverKey.usernameType,
        field: SignUpField.selectedUsername,
        inputConfig: AuthenticatorInputConfig<SignUpField>(selections: [
          InputSelection<SignUpField>(
              label: InputResolverKey.usernameTypeEmail,
              value: SignUpField.email),
          InputSelection<SignUpField>(
              label: InputResolverKey.usernameTypePhoneNumber,
              value: SignUpField.phoneNumber)
        ]),
        validator: validator);
```

Jordan-Nelson (commented):
Still reviewing, but wanted to provide some info on the type changes that I think are needed. I left a couple more comments.

packages/amplify_authenticator/lib/src/mixins/authenticator_radio_field.dart
haverchuck: Lol, correct.

packages/amplify_authenticator/lib/src/state/auth_viewmodel.dart
```diff
@@ -356,21 +398,28 @@ class AuthViewModel extends ChangeNotifier {
...
     _username = '';
     _password = '';
     _passwordConfirmation = '';
     _confirmationCode = '';
     _newUsername = '';
     _newPassword = '';
-
+    _selectedUsername = UsernameAttribute.email;
     _authAttributes.clear();
     _formKey = GlobalKey<FormState>();
```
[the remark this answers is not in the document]
haverchuck: Changed `Enum` to `UsernameAttribute`

packages/amplify_authenticator/lib/src/enums/verified_attributes.dart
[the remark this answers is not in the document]
Jordan-Nelson: I think another option would be to make one enum `CognitoUserAttribute` to replace `UsernameAttribute` and `VerifiedAttributes`. It might even replace the need for `SignUpField`. `UsernameAttribute`, `VerifiedAttributes`, and `SignUpField` are all just subsets of a more comprehensive enum that includes all user attributes (`CognitoUserAttribute`). I see the potential advantage of having a more specific enums, but right now it seems like they are used somewhat interchangeably.

I might look into the option of combining them.

packages/amplify_authenticator/lib/src/mixins/authenticator_date_field.dart
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  @override
+  Widget buildFormField(BuildContext context) {
+    final inputResolver = stringResolver.inputs;
+    final hintText = widget.hintText == null
+        ? inputResolver.resolve(context, widget.hintTextKey!)
+        : widget.hintText!;
+
+    final TextEditingController _controller =
```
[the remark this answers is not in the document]
haverchuck: The issue here was that initState is getting called when befire viewModel is initialized. I'm looking into it more.

packages/amplify_authenticator/lib/src/l10n/generated/input_localizations_en.dart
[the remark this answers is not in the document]
dnys1: For this comment, though, can we remove these extra localizations and just use the existing ones for email and phone number?

packages/amplify_authenticator/lib/src/mixins/authenticator_phone_field.dart
dnys1: Agreed. Can you repurpose `initialValue` for this?

Also, `super.initState` should always be the first call made when overriding `initState`

packages/amplify_authenticator/lib/src/mixins/authenticator_radio_field.dart
dnys1: Actually gonna disagree on this one, I think list comprehensions are nicer. In your example, you gotta add a `.toList()` at the end. Always seemed a bit clunky to me, personally.

packages/amplify_authenticator/lib/src/widgets/form_field.dart
```diff
@@ -84,8 +96,10 @@ abstract class AuthenticatorFormField<FieldType,
   /// The field this form field controls.
   final FieldType field;
 
+  final AuthenticatorInputConfig? inputconfig;
```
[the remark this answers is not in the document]
haverchuck: Fixed!

packages/amplify_authenticator/lib/src/mixins/authenticator_phone_field.dart
haverchuck: Yeah - the whole country code data structure needs re-work.

packages/amplify_authenticator/lib/src/widgets/field_config.dart
haverchuck: I changed AuthenticatorInputConfig to an interface which the mixins each implement, which forces each class that uses the mixin to implement the correct config.

haverchuck: Also, the name is now `SelectableConfig`.

packages/amplify_authenticator/lib/src/state/auth_viewmodel.dart
```diff
@@ -102,6 +114,12 @@ class AuthViewModel extends ChangeNotifier {
   bool _rememberDevice = false;
   bool get rememberDevice => _rememberDevice;
 
+  Enum _attributeKeyToVerify = UsernameAttribute.email;
```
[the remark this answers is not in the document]
haverchuck: Good catch!! Fixed.

packages/amplify_authenticator/lib/src/enums/verified_attributes.dart
```diff
@@ -0,0 +1,17 @@
...
+enum VerifiedAttributes { email, phone_number }
```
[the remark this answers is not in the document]
haverchuck: Changed.

packages/amplify_authenticator/lib/src/mixins/authenticator_date_field.dart
haverchuck: I think we are punting on this for now?

packages/amplify_authenticator/lib/src/widgets/form_fields/sign_up_form_field.dart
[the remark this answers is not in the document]
haverchuck: As noted elsewhere, this is now handled differently.

packages/amplify_authenticator/lib/src/l10n/generated/input_localizations_en.dart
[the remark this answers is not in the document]
haverchuck: Yeah I can just use the existing ones for email; apologies, I thought the existing ones had additional content.

packages/amplify_authenticator/example/lib/main.dart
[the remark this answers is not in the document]
dnys1: This will be in next PR then?

packages/amplify_authenticator/lib/src/mixins/authenticator_date_field.dart
```diff
@@ -0,0 +1,77 @@
+import 'package:amplify_authenticator/src/theme/amplify_theme.dart';
+import 'package:amplify_authenticator/src/widgets/form_field.dart';
+import 'package:intl/intl.dart';
+
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+
+final DateFormat formatter = DateFormat('yyyy-MM-dd');
```
dnys1, line 8: Make private, or static on AuthenticatorDateField, or rename to usDateFormatter or something and put in file for intl/formatters.

packages/amplify_authenticator/lib/src/mixins/authenticator_date_field.dart
```diff
@@ -0,0 +1,77 @@
...
+        : widget.hintText!;
+
+    DateTime now = DateTime.now();
+    Future<void> _pickTime() async {
+      final date = await showDatePicker(
+          context: context,
+          initialDate: now,
+          firstDate: DateTime(DateTime.now().year - 110),
+          lastDate: now);
+      _controller.text = _convertToDateString(date!);
```
dnys1, line 39: `date` can be null here, I'm pretty sure, if they cancel without selecting a date. Need a null check.

packages/amplify_authenticator/lib/src/mixins/authenticator_phone_field.dart
```diff
@@ -0,0 +1,78 @@
...
+mixin AuthenticatorPhoneField<FieldType,
+        T extends AuthenticatorFormField<FieldType, String, T>>
+    on AuthenticatorFormFieldState<FieldType, String, T>
+    implements SelectableConfig<String> {
+  static late List<int> countryCodeStrings = countryCodes
+      .map((Country country) {
+        return int.parse(country.value);
+      })
+      .toSet()
+      .toList();
```
dnys1, line 17: `late final` and `..sort()` should probably go here, instead of initState.

packages/amplify_authenticator/lib/src/mixins/authenticator_date_field.dart
Jordan-Nelson: Thats fine with me

packages/amplify_authenticator/lib/src/state/auth_viewmodel.dart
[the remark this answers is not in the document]
dnys1: No, the formKey. Since it's already being reset in resetFormKey.

packages/amplify_authenticator/lib/src/utils/country_code.dart
```diff
@@ -0,0 +1,243 @@
...
+  final String countryCode;
+  final String display;
+  final String value;
+  const Country(this.countryCode, this.value, this.display);
+}
+
+/// List of country information
+/// TODO: use 3rd party package or change this data to allow for default selection
+/// based on device locale
+var countryCodes = const [
```
dnys1, line 26: `const countryCodes = [`

packages/amplify_authenticator/lib/src/widgets/authenticator_input_config.dart
```diff
@@ -0,0 +1,32 @@
...
+ * permissions and limitations under the License.
+*/
+
+import 'package:amplify_authenticator/amplify_authenticator.dart';
+
+/// Configuration for defining content of radio and switch types in [AuthenticatorFormField].
+class SelectableConfig<T> {
+  final List<InputSelection> selections;
+  late T? selectionValue;
+  SelectableConfig({required this.selections});
```
dnys1, line 22: `const SelectableConfig`

packages/amplify_authenticator/lib/src/widgets/authenticator_input_config.dart
```diff
@@ -0,0 +1,32 @@
...
+}
+
+/// Allows for configuration of fields with multiple selections, such as dropdowns, switches and radio buttons
+class InputSelection<T> {
+  final InputResolverKey label;
+
+  final T value;
+
+  InputSelection({required this.label, required this.value});
```
dnys1, line 31: `const InputSelection`

packages/amplify_authenticator/lib/src/mixins/authenticator_radio_field.dart
```diff
@@ -0,0 +1,47 @@
...
+    super.initState();
+  }
+
+  @override
+  Widget buildFormField(BuildContext context) {
+    final inputResolver = stringResolver.inputs;
+    onChanged?.call(selectionValue!);
+    return Column(
+      children: <Widget>[
+        if (selections.isNotEmpty)
```
Jordan-Nelson, line 22: *nit:* Is this needed? Wouldn't this just be an empty list if selections was empty?

packages/amplify_authenticator/lib/src/services/amplify_auth_service.dart
```diff
@@ -176,7 +176,7 @@ class AmplifyAuthService implements AuthService {
     String code,
     String newPassword,
   ) {
-    return Amplify.Auth.confirmPassword(
+    return Amplify.Auth.confirmResetPassword(
```
Jordan-Nelson, line 179: 💯

packages/amplify_authenticator/lib/src/widgets/form.dart
```diff
@@ -153,11 +147,11 @@ class SignUpForm extends AuthenticatorForm<SignUpForm> {
...
 
   /// Whether to include the attributes specified by your Amplify configuration.
-  final bool includeDefaultAttributes;
+  final bool useAttributesFromConfig;
 
   @override
-  List<AuthenticatorFormField> additionalFields(BuildContext context) {
-    if (!includeDefaultAttributes) {
+  List<SignUpFormField> runtimeFields(BuildContext context) {
+    if (!useAttributesFromConfig || fields.isNotEmpty) {
```
dnys1, line 154: If we're not gonna allow both custom fields and auto-attribute fields, then we should remove the option from the constructor, since it doesn't matter.

packages/amplify_authenticator/lib/src/widgets/form.dart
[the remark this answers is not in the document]
dnys1: Next PR?

packages/amplify_authenticator/lib/src/widgets/form_field.dart
```diff
@@ -95,64 +106,47 @@ abstract class AuthenticatorFormField<FieldType,
...
-    return (String value) {
-      for (var op in ops) {
-        op(value);
-      }
-    };
-  }
-
   /// Callback for when `onChanged` is triggered on the [FormField].
-  ValueChanged<String> get onChanged => (_) {};
+  ValueChanged<FieldValue>? get onChanged => (_) {};
```
dnys1, line 146: Doesn't need to be nullable. That's why there is a default value of a no-op.

packages/amplify_authenticator/lib/src/widgets/form_field.dart
```diff
@@ -16,28 +16,39 @@
...
+
+import 'package:flutter/services.dart';
 
 part 'form_fields/sign_in_form_field.dart';
 part 'form_fields/sign_up_form_field.dart';
 part 'form_fields/confirm_sign_up_form_field.dart';
 part 'form_fields/confirm_sign_in_form_field.dart';
 part 'form_fields/verify_user_form_field.dart';
-part 'form_fields/confirm_verify_user_form_field.dart';
+// part 'form_fields/confirm_verify_user_form_field.dart';
```
Jordan-Nelson, line 51: Should this be uncommented?

Jordan-Nelson (approved):
Left a couple small nit comments but LGTM

packages/amplify_authenticator/lib/src/state/auth_viewmodel.dart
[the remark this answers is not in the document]
haverchuck: Oh, I see - sure.

packages/amplify_authenticator/lib/src/mixins/authenticator_date_field.dart
```diff
@@ -0,0 +1,77 @@
...
+  String _convertToDateString(DateTime dt) {
+    return formatter.format(dt);
+  }
+
+  late TextEditingController _controller;
+
+  @override
+  void initState() {
+    _controller = TextEditingController(text: '');
+    super.initState();
```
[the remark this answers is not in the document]
haverchuck: Good to know, thanks - fixed in a couple other place.

packages/amplify_authenticator/lib/src/mixins/authenticator_date_field.dart
haverchuck: Done

packages/amplify_authenticator/lib/src/mixins/authenticator_date_field.dart
haverchuck: Thanks, good catch.

packages/amplify_authenticator/lib/src/widgets/form_field.dart
haverchuck: This should be deleted, actually. We combined the verification form fields as part of the refactor.

packages/amplify_authenticator/lib/src/mixins/authenticator_radio_field.dart
```diff
@@ -0,0 +1,47 @@
...
+    implements SelectableConfig<FieldValue> {
+  @override
+  void initState() {
+    super.initState();
+  }
+
+  @override
+  Widget buildFormField(BuildContext context) {
+    final inputResolver = stringResolver.inputs;
+    onChanged?.call(selectionValue!);
```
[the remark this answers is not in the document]
haverchuck: This is going to be changed when I rework the radio button mixin.

packages/amplify_authenticator/lib/src/widgets/form_field.dart
haverchuck: This is due to the fact that each input-specific extension of a field type (ex: `class _SignInTextFieldState extends _SignInFormFieldState`) contains switch cases on the field itself, which include EVERY field. For example, the switch case for `_SignInTextFieldState` will need to handle cases for fields which use the radio button mixin - unless of course we revise this design again. 

Currently, I just return null for those cases, but that's probably not the best way because we have to do `onChanged?.call` elsewhere. Maybe throw some sort of exception?

packages/amplify_authenticator/lib/src/mixins/authenticator_radio_field.dart
haverchuck: I'm hoping to avoid this once I iterate on the radio buttons, but currently it's in place due the fact that we generate these at runtime.

packages/amplify_authenticator/lib/src/widgets/authenticator_input_config.dart
haverchuck: Can you do that without make selectionValue final?

packages/amplify_authenticator/lib/src/widgets/form.dart
haverchuck: We could actually include both... the tricky thing would be visually ordering the fields in a way that makes sense. I'll give it a try, and we could refine as needed. Potentially some sort of sort order value on the fields?

haverchuck: Hmm... yeah, I'm going to remove the option for now until we figure out a good API for creating a union of custom and config values. I think an experience where you pass a custom form, and all you see are the values you pass, is fairly simple and makes sense.

PR #1024, opened 2021-11-03
Datastore support read only

fjnoyp:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify_datastore_plugin_interface/test/testData/Person.dart
+++ b/packages/amplify_datastore_plugin_interface/test/testData/Person.dart
@@ -13,29 +13,20 @@
 * permissions and limitations under the License.
 */
 
-// ignore_for_file: public_member_api_docs
-
-import 'ModelProvider.dart';
 import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart';
 import 'package:collection/collection.dart';
 import 'package:flutter/foundation.dart';
 
+// ignore_for_file: public_member_api_docs
+
+import 'ModelProvider.dart';
+
 /** This is an auto generated class representing the Person type in your schema. */
 @immutable
-class Person extends Model {
-  static const classType = const _PersonModelType();
-  final String id;
+class Person {
   final String? _name;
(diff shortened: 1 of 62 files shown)


packages/amplify_api/lib/src/method_channel_api.dart
```diff
@@ -251,6 +251,9 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
     inputsMap['restOptions'] = restOptions.serializeAsMap();
     inputsMap['cancelToken'] = cancelToken;
 
+    // Refresh auth tokens
+    await _authProviderRefresher();
```
HuiSF, line 255: This change has been reverted on main, which should not be brought into RC. Please remove corresponding commit.

HuiSF (changes requested):

packages/amplify_api/lib/src/method_channel_api.dart
fjnoyp: Ah thank you.  Reverted the corresponding commit.

HuiSF (approved):

packages/amplify_datastore_plugin_interface/test/testData/Person.dart
```diff
@@ -13,30 +13,21 @@
...
+// ignore_for_file: public_member_api_docs
+
+import 'ModelProvider.dart';
+
 /** This is an auto generated class representing the Person type in your schema. */
 @immutable
-class Person extends Model {
-  static const classType = const _PersonModelType();
-  final String id;
+class Person {
```
dnys1, line 26: What's up with this class

dnys1 (approved):

packages/amplify_datastore_plugin_interface/test/testData/Person.dart
HuiSF: Oops missed this one, seems the model generated by 

```
type Person {
  name: String!
  propertiesAddresses: [Address]
  contact: Contact!
}
```

which should be with `@model`.

Issue #837, opened 2021-08-25
Most QueryPredicates do not work with AWSDate, AWSDateTime, AWSTime, and AWSTimestamp


Jordan-Nelson:
**Describe the bug**
The majority of query predicates (eq, ne, lt, gt, etc.) do not work with AWSDate, AWSDateTime, AWSTime, and AWSTimestamp.

The examples in this issue will focus in AWSDate, but similar issues exist with all the date/time related types.

Example schema:
```graphql
type DateTypeModel @model {
  id: ID!
  value: AWSDate
}
```

Using the schema above, this integration test for eq() will fail. This test just shows eq() not working, but the test can easily be modified for the other operators.

```dart
testWidgets('eq()', (WidgetTester tester) async {
  var models = [
    DateTypeModel(value: TemporalDate(DateTime(2020, 1, 1))),
    DateTypeModel(value: TemporalDate(DateTime(2020, 1, 2))),
    DateTypeModel(value: TemporalDate(DateTime(2020, 2, 1))),
  ];
  for (var model in models) {
    await Amplify.DataStore.save(model);
  }

  var result1 = await Amplify.DataStore.query(
    DateTypeModel.classType,
    where: DateTypeModel.VALUE.eq(DateTime(2020, 1, 1)), // note: using TemporalDate will throw an exception
  );

  // this test will fail since the query returns no results
  expect(result1.length, 1);
  expect(result1.contains(models[0]), isTrue);
});
```

**To Reproduce**
See above.

**Expected behavior**
Query Predicates should work for Date and Time related types. eq(), ne(), lt(), le(), gt(), ge(), and between() should all work. I am not sure what the expected behavior for contains() and beginsWith() would be. They may not really apply to Date and Time related types.

Jordan-Nelson:
This was resolved in 0.2.7

PR #1027, opened 2021-11-04
fix(datastore): fix temporal date/time query predicates

Jordan-Nelson:
*Issue #, if available:* https://github.com/aws-amplify/amplify-flutter/issues/837

*Description of changes:*
- serialize temporal date/time types in query predicates
- add unit test for the new serialization
- add integ tests for all temporal date/time types

### DIFF
--- a/packages/amplify_datastore_plugin_interface/lib/src/types/query/query_field.dart
+++ b/packages/amplify_datastore_plugin_interface/lib/src/types/query/query_field.dart
@@ -17,6 +17,7 @@ library query_field;
 
 import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart';
 import 'package:amplify_datastore_plugin_interface/src/types/models/model_field_type.dart';
+import 'package:flutter/foundation.dart';
 import '../temporal/datetime_parse.dart';
 import '../utils/parsers.dart';
 
--- a/packages/amplify_datastore_plugin_interface/lib/src/types/query/query_field_operators.dart
+++ b/packages/amplify_datastore_plugin_interface/lib/src/types/query/query_field_operators.dart
@@ -55,7 +55,10 @@ abstract class QueryFieldOperator<T> {
 
   /// check the type of [value] and invoke corresponding serialize method
   dynamic serializeDynamicValue(dynamic value) {
-    if (value is TemporalDate) {
+    // DateTime is deprecated and will be removed in the next major version
+    if (value is DateTime) {
+      return value.toDateTimeIso8601String();
+    } else if (value is TemporalDate) {
       return value.format();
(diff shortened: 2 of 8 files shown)


packages/amplify_datastore_plugin_interface/lib/src/types/query/query_field_operators.dart
```diff
@@ -75,7 +87,8 @@ class EqualQueryOperator<T> extends QueryFieldOperator<T> {
 
   @override
   bool evaluate(T? other) {
-    return other == value;
+    var serializedValue = serializeDynamicValue(value);
```
Jordan-Nelson, line 90: A few callouts here:
- Because `QueryPredicateOperation.evaluate()` (see code below) converts the model to json, `other` will be serialized if it is not a primitive type. `value` needs to be serialized as well to compare non-primitive (i.e. Temporal) types.
- There was a [discussion on the initial observeQuery PR](https://github.com/aws-amplify/amplify-flutter/pull/892/files#r724667952) about alternatives to serializing the entire model to json, which I think is still worth considering. However, it will require codegen changes, and will then be a breaking change.
- While this method is only used for observeQuery(). Query Predicates for query() are evaluated on the native side, and all values are serialized with serializeDynamicValue() prior to being sent to native. This is consistent with how non-primitive types are handled in query().

**QueryPredicateOperation.evaluate()**
https://github.com/aws-amplify/amplify-flutter/blob/8e9b7c5676bc5ac9ae77d102c56d862bbba2a749/packages/amplify_datastore_plugin_interface/lib/src/types/query/query_predicate.dart#L56-L60

fjnoyp (approved):
Looks great!  Thanks for the helpful contextual comments.

packages/amplify_datastore_plugin_interface/lib/src/types/query/query_field.dart
```diff
@@ -17,6 +17,7 @@ library query_field;
 
 import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart';
 import 'package:amplify_datastore_plugin_interface/src/types/models/model_field_type.dart';
+import 'package:flutter/foundation.dart';
```
HuiSF, line 20: Nit: is this needed as no extra changes made within this file.

HuiSF (approved):

packages/amplify_datastore_plugin_interface/lib/src/types/query/query_field.dart
Jordan-Nelson: Yeah, this is sort of confusing, but it is needed. This was added due to the changes in query_field_operators.dart. 

query_field_operators.dart is part of this file (part of 'query_field.dart';).

PR #1029, opened 2021-11-04
fix: resend signup code

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- do not check form validation when requesting a new code
- do not set the form to busy when requesting a new code


packages/amplify_authenticator/lib/src/state/auth_viewmodel.dart
```diff
@@ -337,15 +337,10 @@ class AuthViewModel extends ChangeNotifier {
   }
 
   Future<void> resendSignUpCode() async {
-    if (!_formKey.currentState!.validate()) {
```
Jordan-Nelson, line 340: Since resending the code doesn't submit the form, we shouldn't care if the form is valid.

packages/amplify_authenticator/lib/src/state/auth_viewmodel.dart
```diff
@@ -337,15 +337,10 @@ class AuthViewModel extends ChangeNotifier {
...
-      return;
-    }
-    setBusy(true);
```
Jordan-Nelson, line 343: Similarly to above, `setBusy` sets the form to busy. The form's primary button goes to a loading state. Resending the code doesn't submit the form though. This makes it look like the form was submitted, but it wasn't.

haverchuck (approved):

dnys1 (approved):

dnys1 (approved):

haverchuck (approved):

PR #1032, opened 2021-11-05
feat(datastore): Custom Error Handler

fjnoyp:
*Issue #, if available:*

*Description of changes:*

Custom ErrorHandler support 

I haven't written integration tests for this.  Should be a small code change but a bit tricky to figure out as I don't know how to trigger the Datastore Exceptions that would in turn trigger the custom error handler (I've just manually triggered it from native atm) 

Update November 8: 

We should include integration tests for this code change in a separate PR.  Writing integration tests will require enabling cloud sync and thus reworking all existing integration tests in Datastore as well.  Given ongoing talks with native teams, I might also need to enable Auth as well to trigger the edge case that would get the custom error callback to be called in the first place.  As this could require significantly more time and is outside scope of original task I plan to tackle this in a separate line of work.  There are already unit tests for this feature though. 

### DIFF
--- a/packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
+++ b/packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
@@ -15,6 +15,7 @@
 
 package com.amazonaws.amplify.amplify_datastore
 
+import android.database.DefaultDatabaseErrorHandler
 import android.os.Handler
 import android.os.Looper
 import androidx.annotation.NonNull
@@ -39,6 +40,7 @@ import com.amplifyframework.core.model.query.QueryOptions
 import com.amplifyframework.core.model.query.predicate.QueryPredicates
 import com.amplifyframework.datastore.AWSDataStorePlugin
 import com.amplifyframework.datastore.DataStoreConfiguration
+import com.amplifyframework.datastore.DataStoreErrorHandler
 import com.amplifyframework.datastore.DataStoreException
 import io.flutter.embedding.engine.plugins.FlutterPlugin
 import io.flutter.plugin.common.EventChannel
@@ -162,6 +164,18 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
         }
 
         val dataStoreConfigurationBuilder = DataStoreConfiguration.builder()
+
+        if(request["hasCustomErrorHandler"] as Boolean) {
+            var handler = DataStoreErrorHandler {
+                val args = hashMapOf(
+                        "errorCode" to "DataStoreException",
+                        "errorMessage" to ExceptionMessages.defaultFallbackExceptionMessage,
+                        "details" to createSerializedError(it)
+                )
+                channel.invokeMethod("errorHandler", args)
+            }
+            dataStoreConfigurationBuilder.errorHandler(handler!!)
+        }
 
         modelProvider.setVersion(request["modelProviderVersion"] as String)
         val defaultDataStoreConfiguration = DataStoreConfiguration.defaults()
--- a/packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
+++ b/packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
@@ -165,7 +165,7 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
 
         val dataStoreConfigurationBuilder = DataStoreConfiguration.builder()
 
-        if(request["hasCustomErrorHandler"] as Boolean) {
+        if( (request["hasErrorHandler"] as? Boolean? == true) ) {
             var handler = DataStoreErrorHandler {
                 val args = hashMapOf(
                         "errorCode" to "DataStoreException",
--- a/packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
+++ b/packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
@@ -15,7 +15,6 @@
 
 package com.amazonaws.amplify.amplify_datastore
 
-import android.database.DefaultDatabaseErrorHandler
 import android.os.Handler
 import android.os.Looper
 import androidx.annotation.NonNull
@@ -165,8 +164,9 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
 
         val dataStoreConfigurationBuilder = DataStoreConfiguration.builder()
 
-        if( (request["hasErrorHandler"] as? Boolean? == true) ) {
-            var handler = DataStoreErrorHandler {
+        var errorHandler : DataStoreErrorHandler;
+        errorHandler = if( (request["hasErrorHandler"] as? Boolean? == true) ) {
+            DataStoreErrorHandler {
                 val args = hashMapOf(
                         "errorCode" to "DataStoreException",
                         "errorMessage" to ExceptionMessages.defaultFallbackExceptionMessage,
@@ -174,8 +174,13 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
                 )
                 channel.invokeMethod("errorHandler", args)
             }
-            dataStoreConfigurationBuilder.errorHandler(handler!!)
         }
+        else {
+            DataStoreErrorHandler {
+                LOG.error(it.toString())
+            }
+        }
+        dataStoreConfigurationBuilder.errorHandler(errorHandler)
 
         modelProvider.setVersion(request["modelProviderVersion"] as String)
         val defaultDataStoreConfiguration = DataStoreConfiguration.defaults()
--- a/packages/amplify_datastore/example/lib/main.dart
+++ b/packages/amplify_datastore/example/lib/main.dart
@@ -88,7 +88,7 @@ class _MyAppState extends State<MyApp> {
       datastorePlugin = AmplifyDataStore(
         modelProvider: ModelProvider.instance,
         errorHandler: ((error) =>
-            {print("Custom ErrorHandler result: " + error.toString())}),
+            {print("Custom ErrorHandler received: " + error.toString())}),
       );
       await Amplify.addPlugin(datastorePlugin);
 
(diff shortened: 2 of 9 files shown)


packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -163,6 +165,18 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
 
         val dataStoreConfigurationBuilder = DataStoreConfiguration.builder()
 
+        if(request["hasCustomErrorHandler"] as Boolean) {
```
fjnoyp, line 168: I think this naming `hasCustomErrorHandler` is inconsistent and should just be `hasErrorHandler`

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -15,6 +15,7 @@
 
 package com.amazonaws.amplify.amplify_datastore
 
+import android.database.DefaultDatabaseErrorHandler
```
HuiSF, line 18: Accidentally imported?

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -163,6 +165,18 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
...
+        if( (request["hasErrorHandler"] as? Boolean? == true) ) {
+            var handler = DataStoreErrorHandler {
+                val args = hashMapOf(
+                        "errorCode" to "DataStoreException",
+                        "errorMessage" to ExceptionMessages.defaultFallbackExceptionMessage,
+                        "details" to createSerializedError(it)
+                )
+                channel.invokeMethod("errorHandler", args)
+            }
+            dataStoreConfigurationBuilder.errorHandler(handler!!)
```
HuiSF, line 177: `!!` is probably unnecessary.

packages/amplify_datastore/example/lib/main.dart
```diff
@@ -85,7 +85,11 @@ class _MyAppState extends State<MyApp> {
   // Platform messages are asynchronous, so we initialize in an async method.
   Future<void> initPlatformState() async {
     try {
-      datastorePlugin = AmplifyDataStore(modelProvider: ModelProvider.instance);
+      datastorePlugin = AmplifyDataStore(
+        modelProvider: ModelProvider.instance,
+        errorHandler: ((error) =>
+            {print("Custom ErrorHandler result: " + error.toString())}),
```
HuiSF, line 91: mega nit: "result" sounds like we are expecting some thing to be returned intentionally, could be "Custom ErrorHandler received error" :D

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
```diff
@@ -121,10 +121,30 @@ public class SwiftAmplifyDataStorePlugin: NSObject, FlutterPlugin {
 
             flutterModelRegistration.version = modelProviderVersion
             let syncExpressions: [DataStoreSyncExpression] = try createSyncExpressions(syncExpressionList: syncExpressionList)
-
+            
             self.dataStoreHubEventStreamHandler?.registerModelsForHub(flutterModelRegistration: flutterModelRegistration)
+            
+
+            var errorHandler: DataStoreErrorHandler
+            if( (args["hasErrorHandler"] as? Bool) == true ){
```
HuiSF, line 129:
```suggestion
            if((args["hasErrorHandler"] as? Bool) == true) {
```

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -163,6 +165,18 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
 
         val dataStoreConfigurationBuilder = DataStoreConfiguration.builder()
 
+        if( (request["hasErrorHandler"] as? Boolean? == true) ) {
```
HuiSF, line 168: I like the idea that defining a default error handler like what you did in this PR for the iOS implementation, can we adapt it to the Android implementation as well?

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
```diff
@@ -121,10 +121,30 @@ public class SwiftAmplifyDataStorePlugin: NSObject, FlutterPlugin {
...
+                errorHandler = { error in
+                    let map : [String:Any] = [
+                        "errorCode" : "DataStoreException",
+                        "errorMesage" : ErrorMessages.defaultFallbackErrorMessage,
+                        "details" : FlutterDataStoreErrorHandler.createSerializedError(error: error)
+                    ]
+                    self.channel!.invokeMethod("errorHandler", arguments: args)
+                }
+            }
+            else{
```
HuiSF, line 139:
```suggestion
            } else {
```

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
```diff
@@ -133,7 +153,7 @@ public class SwiftAmplifyDataStorePlugin: NSObject, FlutterPlugin {
             
             Amplify.Logging.logLevel = .info
             print("Amplify configured with DataStore plugin")
-            result(true)
+            result(true)
```
HuiSF, line 156:
```suggestion
            result(true)
```

packages/amplify_datastore/lib/method_channel_datastore.dart
```diff
@@ -231,6 +252,23 @@ class AmplifyDataStoreMethodChannel extends AmplifyDataStore {
     return serializedItem["modelName"] as String;
   }
 
+  AmplifyException _deserializeExceptionFromMap(Map<String, dynamic> e) {
+    if (e['errorCode'] == 'DataStoreException') {
+      return DataStoreException.fromMap(Map<String, String>.from(e['details']));
+    } else if (e['errorCode'] == 'AmplifyAlreadyConfiguredException') {
+      return AmplifyAlreadyConfiguredException.fromMap(
+          Map<String, String>.from(e['details']));
+    } else {
```
HuiSF, line 261: It looks like the `else` covers use case that `e['errorCode'].isEmpty !== true`, with which there is an error code, but not handled by this `_deserializeExceptionFromMap` method, for this case categorizing them as "unhandled" could be more accurate than "missingExceptionMessage"?

packages/amplify_datastore_plugin_interface/lib/amplify_datastore_plugin_interface.dart
```diff
@@ -80,15 +84,18 @@ abstract class DataStorePluginInterface extends AmplifyPluginInterface {
   }
 
   /// Configure AmplifyDataStore plugin with mandatory [modelProvider]
-  /// and optional datastore configuration properties including
+  /// and optional DataStore configuration properties including
+  ///
+  /// [errorHandler]: custom error handler code executed when DataStore encounters an unhandled error during its background operations
```
HuiSF, line 89: Maybe this could be a bit clearer that we are expecting from the error handler. 

```suggestion
  /// [errorHandler]: Custom error handler function that may receive an [AmplifyException] object when DataStore encounters an unhandled error during its background operations
```

HuiSF (commented):
Thanks for adding the error handler support to DataStore.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
fjnoyp: thanks - looks like it .. :)

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
fjnoyp: thanks removed

packages/amplify_datastore/example/lib/main.dart
fjnoyp: nice good point!  updated

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
fjnoyp: thanks updated

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
fjnoyp: Good point !  Updated android: 

  ```
      var errorHandler : DataStoreErrorHandler;
        errorHandler = if( (request["hasErrorHandler"] as? Boolean? == true) ) {
            DataStoreErrorHandler {
                val args = hashMapOf(
                        "errorCode" to "DataStoreException",
                        "errorMessage" to ExceptionMessages.defaultFallbackExceptionMessage,
                        "details" to createSerializedError(it)
                )
                channel.invokeMethod("errorHandler", args)
            }
        }
        else {
            DataStoreErrorHandler {
                LOG.error(it.toString())
            }
        }
        dataStoreConfigurationBuilder.errorHandler(errorHandler)

```

File: packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift (hunk 7 above)
Comment (fjnoyp) [reply packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift:139]: thanks updated

File: packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift (hunk 8 above)
Comment (fjnoyp) [reply packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift:156]: my bad ... removed

File: packages/amplify_datastore/lib/method_channel_datastore.dart (hunk 9 above)
Comment (fjnoyp) [reply packages/amplify_datastore/lib/method_channel_datastore.dart:261]: Hmm makes sense.  However when looking at: 

```
  static const missingExceptionMessage =
      'An unknown exception has happened. ' + _createGithubIssueString;

```

It appears the existing error message returned is sufficient?  Do you think we should be saying "An unhandled exception has happened" instead? 

@HuiSF

packages/amplify_datastore_plugin_interface/lib/amplify_datastore_plugin_interface.dart
fjnoyp: Thanks updated with your suggestions!

ragingsquirrel3 (approved):

HuiSF (approved):
👍

PR #1035, opened 2021-11-05
chore(auth): CognitoUserAttributes type

dnys1:
*Issue #, if available:*

*Description of changes:*
- Introduces `UserAttributeKey` type for auth plugin
- Re-land `CognitoUserAttributes` type which implements `UserAttributeKey` for Cognito plugin


packages/amplify_flutter/lib/src/categories/amplify_auth_category.dart
```diff
@@ -247,7 +247,7 @@ class AuthCategory {
   ///
   /// Accepts plugin-specific, advanced [options] for the request
   Future<UpdateUserAttributeResult> updateUserAttribute({
-    required String userAttributeKey,
+    required Object userAttributeKey,
```
Jordan-Nelson, line 250: I think `Object` is sort of a confusing public API. If the goal is to avoid putting a cognito-specific class (CognitoUserAttribute) in the category interface, I think we could make a generic UserAttribute class and then either create a cognito-specific implementation of that class, or add the cognito specific pieces via an extension.

Example of how `CognitoUserAttribute` could be split into a generic `UserAttribute` and cognito-specific extension

```dart
// generic class for the category interface
class UserAttribute {
  const UserAttribute._(this.key, {this.readOnly = false});

  /// The JSON key for this attribute.
  final String key;

  /// Whether this attribute is read only
  final bool readOnly;

  /// Creates a custom attribute.
  const UserAttribute.custom(String key)
      : key = 'custom:$key',
        readOnly = false;

  // other immutable methods ...
}

// cognito implementation
extension CognitoUserAttribute on UserAttribute {
  static const address = UserAttribute._('address');

  // other attributes ...

  /// Parses the given Cognito attribute key.
  static parse(String key) {
    return values.firstWhere(
      (attr) => attr.key == key,
      orElse: () => UserAttribute._(key),
    );
  }

  static const values = [
    address,
    // other attributes ...
  ];
}
```

packages/amplify_auth_cognito/lib/src/CognitoSignUp/cognito_user_attributes.dart
```diff
@@ -1,24 +1,185 @@
...
-  static const nickname = 'nickname';
-  static const phoneNumber = 'phone_number';
-  static const picture = 'picture';
-  static const preferredUsername = 'preferred_username';
-  static const profile = 'profile';
-  static const updatedAt = 'updated_at';
-  static const website = 'website';
-  static const zoneinfo = 'zoneinfo';
+@immutable
+class CognitoUserAttributes extends UserAttributeKey {
```
Jordan-Nelson, line 11: This should be singular, `CognitoUserAttribute`.

Preferably, it should be `CognitoUserAttributeKey`

This definitely isn't a well definied pattern we have established, but in general the Cognito specific classes have the same name as the abstract class they extend, with a prefix of `Cognito`. I think this is a nice pattern to follow. Since the public API docs for methods like `updateUserAttribute()` will show the abstract class `UserAttributeKey`, users have to do a bit of guess work (or reference the docs) to know what the actual class name is. That guess work is a little easier if we consistently prefix the abstract class with `Cognito`.

packages/amplify_auth_cognito/lib/src/CognitoSignUp/CognitoSignUpOptions.dart
```diff
@@ -36,10 +38,10 @@ class CognitoSignUpOptions extends SignUpOptions {
   ///
   /// For example:
   /// ```dart
-  /// var userAttributes = <String, String>{
+  /// var userAttributes = <CognitoUserAttributes, String>{
```
Jordan-Nelson, line 41: 💯

packages/amplify_auth_cognito/lib/src/CognitoSignUp/cognito_user_attributes.dart
```diff
@@ -1,24 +1,185 @@
...
+      : key = '$_customPrefix$key',
+        readOnly = false;
+
+  /// Parses the given Cognito attribute key.
+  factory CognitoUserAttributes.parse(String key) {
+    key = key.toLowerCase();
+    return values.firstWhere(
+      (attr) => attr.key == key,
+      orElse: () {
+        if (!key.startsWith(_customPrefix)) {
```
Jordan-Nelson, line 35: This is throwing when `fetchUserAttributes()` is called because there is an attribute returned with a key of `sub`.

packages/amplify_auth_cognito/example/.gitignore
```diff
@@ -43,7 +43,7 @@ app.*.map.json
 # Exceptions to above rules.
 !/packages/flutter_tools/test/data/dart_dependencies_test/**/.packages
 
-#amplify
+#amplify-do-not-edit-begin
```
haverchuck, line 46: Is this what the current version of the CLI is outputting now?

packages/amplify_auth_cognito/lib/src/CognitoSignUp/cognito_user_attributes.dart
```diff
@@ -1,24 +1,185 @@
...
-  static const profile = 'profile';
-  static const updatedAt = 'updated_at';
-  static const website = 'website';
-  static const zoneinfo = 'zoneinfo';
+@immutable
+class CognitoUserAttributes extends UserAttributeKey {
+  const CognitoUserAttributes._(this.key, {this.readOnly = false});
+
+  /// Prefix for custom Cognito attributes.
+  static const _customPrefix = 'custom:';
```
haverchuck, line 15: Did you test custom attributes on both Android and iOS with this code? Both the native libraries handle this prefix a bit differently; we are accomodating this difference in our platform code, and I just want to make sure everything still works as expected.

haverchuck (commented):
See inline questions.

packages/amplify_auth_cognito/lib/src/CognitoSignUp/cognito_user_attributes.dart
haverchuck: Good catch @Jordan-Nelson

packages/amplify_auth_cognito/example/.gitignore
dnys1: Yep

Jordan-Nelson (approved):

haverchuck (approved):

PR #1037, opened 2021-11-05
chore: update amplify pods to 1.15.5

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- update amplify-ios to 1.15.5 to bring in https://github.com/aws-amplify/amplify-ios/pull/1494
- add melos script intended to be run after updating amplify pod versions

### DIFF
--- a/melos.yaml
+++ b/melos.yaml
@@ -131,6 +131,18 @@ scripts:
     melos exec -- \
       flutter analyze --no-fatal-infos --no-fatal-warnings &>/dev/null || true
 
+  pod:repo-update:
+    run: melos exec -c 8 --scope="*example*,sample_app" "(cd ios && pod repo update)"
+    description: runs "pod repo update" in all example apps
+
+  pod:update-amplify:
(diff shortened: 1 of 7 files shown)


melos.yaml
```diff
@@ -131,6 +131,18 @@ scripts:
     melos exec -- \
       flutter analyze --no-fatal-infos --no-fatal-warnings &>/dev/null || true
 
+  pod:repo-update:
```
Jordan-Nelson, line 134: I added these just to make running the example apps after a version update a bit less painful. There may be a better way to do this, but this is pretty much what I have been doing (just manually) after updating the version.

The issue is that after updating the version, if you run any example app you will likely see [this issue](https://docs.amplify.aws/lib/troubleshooting/upgrading/q/platform/flutter/).

Our clean script deletes podfile.lock files, so most/some of the time running a `melos clean && melos bootstrap` fixes it. But, that is still a pain, and will not resolve it if the new version isn't in your local pod spec.

Running `melos run pod:update` _should_ resolve it all scenarios and isn't as annoying as having to clean/bootstrap.

HuiSF (approved):

melos.yaml
ragingsquirrel3: thanks, I've noticed the same thing.

ragingsquirrel3 (approved):
thanks!

PR #1039, opened 2021-11-05
chore(amplify_authenticator): Improves phone and date input mixins

haverchuck:
*Description of changes:*
Makes the phone number and date selection mixins visually consistent with other inputs.


packages/amplify_authenticator/lib/src/utils/country_code.dart
```diff
@@ -13,231 +13,234 @@
...
-  Country('VE', '58', 'Venezuela (+58)'),
-  Country('VN', '84', 'Vietnam (+84)'),
-  Country('VG', '84', 'Virgin Islands - British (+1284)'),
-  Country('VI', '84', 'Virgin Islands - US (+1340)'),
-  Country('WF', '681', 'Wallis and Futuna (+681)'),
-  Country('YE', '969', 'Yemen (North) (+969)'),
-  Country('YE', '967', 'Yemen (South) (+967)'),
-  Country('ZM', '260', 'Zambia (+260)'),
-  Country('ZW', '263', 'Zimbabwe (+263)')
+  Country('US', '1', CountryResolverKey.unitedStates),
```
Jordan-Nelson, line 30: If we iterate over CountryResolverKey.values, can this be a way that the consumer can control which countries show up in the country code list?

Jordan-Nelson (commented):
Seems like a reasonable way to handle localization with country codes 👍

packages/amplify_authenticator/lib/src/mixins/authenticator_phone_field.dart
```diff
@@ -31,50 +20,60 @@ mixin AuthenticatorPhoneField<FieldType,
...
+                width: _prefixWidth,
+                child: PopupMenuButton(
+                  initialValue: selectionValue,
+                  child: Center(child: Text('+$selectionValue')),
+                  itemBuilder: (context) {
+                    return countryCodes.map((Country country) {
+                      return PopupMenuItem<String>(
+                        value: country.value,
+                        child: Text(
+                            '${countryResolver.resolve(context, country.key)} (+${country.value})',
```
Jordan-Nelson, line 54: Probably out of scope, but I think some customers may want to localize `value` as well. I assume cognito requires 0-9, but some other languages use other numerals and customers may want to display those.

I would say this is out of scope because we probably need more info on requirements around this if it is something we want to support it. Adding some sort of concept of "value" (what is sent to cognito) versus a localized "displayValue" (what displays on the screen) would probably meet these use cases, but I am not totally sure.

packages/amplify_authenticator/lib/src/utils/country_code.dart
haverchuck: Yeah - haven't implemented it yet, but we can have them pass this to the constructor for phone_number.

packages/amplify_authenticator/lib/src/mixins/authenticator_phone_field.dart
haverchuck: Yeah, maybe we can just display `${countryResolver.resolve(context, country.key)}`, and then our defauly en_arb can include the '+$value'?

packages/amplify_authenticator/lib/src/widgets/authenticator_input_config.dart
```diff
@@ -24,10 +24,12 @@ abstract class SelectableConfig<T> {
...
+class InputSelection<Label, Value> {
+  /// Display text for the selection
+  final L label;
 
-  final T value;
+  /// Value for the selection
+  final V value;
 
   const InputSelection({required this.label, required this.value});
 }
```
dnys1, line 35: Does this compile? Should be:

```dart
class InputSelection<Label, Value> {
  /// Display text for the selection
  final Label label;

  /// Value for the selection
  final Value value;

  const InputSelection({required this.label, required this.value});
}
```

dnys1 (approved):
Rebase on Jordan's work

packages/amplify_authenticator/lib/src/utils/country_code.dart
```diff
@@ -13,231 +13,234 @@
...
+  Country('GB', '44', CountryResolverKey.unitedKingdom)
+  // Country('DZ', '213', 'Algeria (+213)'),
```
Jordan-Nelson, line 32: I assume the rest of these will be added back before the PR is complete?

haverchuck: I was going to do this as a separate PR, since it will be a ton of repetition.

Jordan-Nelson (approved):

PR #1050, opened 2021-11-08
fix(amplify_authenticator): sends username for password recovery

haverchuck:
*Description of changes:*
Sends the username for forgot password flow, and removes unneeded fields.

### DIFF
--- a/packages/amplify_authenticator/lib/src/state/auth_viewmodel.dart
+++ b/packages/amplify_authenticator/lib/src/state/auth_viewmodel.dart
@@ -30,7 +30,6 @@ class AuthViewModel extends ChangeNotifier {
 
     /// When auth flow is complete, reset entirety of [ViewModel] state
     _authBloc.stream
-        // .where((event) => event is Authenticated)
         .distinct()
         .listen((event) {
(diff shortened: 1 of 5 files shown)


packages/amplify_authenticator/lib/src/state/auth_viewmodel.dart
```diff
@@ -27,20 +27,17 @@ class AuthViewModel extends ChangeNotifier {
...
-        .where((event) => event is AuthFlow)
-        .distinct()
-        .listen((event) {
-      resetFormKey();
-    });
 
     /// When auth flow is complete, reset entirety of [ViewModel] state
     _authBloc.stream
-        .where((event) => event is Authenticated)
+        // .where((event) => event is Authenticated)
```
Jordan-Nelson, line 33: Should this be removed?

Jordan-Nelson:
```suggestion
```

Jordan-Nelson (approved):
There is a comment line of code that I think needs to be removed but other than that lgtm

PR #1054, opened 2021-11-08
fix: create newPassword field for confirm sign up

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- create a newPassword field for confirm sign up
- separate MFA confirm sign up and force password reset confirm sign up
- update the language for the hint text

### DIFF
--- a/packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
+++ b/packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
@@ -152,7 +152,7 @@ class StateMachineBloc {
   ) async* {
     try {
       var result = await _authService.confirmSignIn(
-        code: data.code,
+        confirmationValue: data.confirmationValue,
         attributes: data.attributes,
       );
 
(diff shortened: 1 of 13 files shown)


packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
```diff
@@ -152,7 +152,7 @@ class StateMachineBloc {
   ) async* {
     try {
       var result = await _authService.confirmSignIn(
-        code: data.code,
+        confirmationValue: data.confirmationValue,
```
Jordan-Nelson, line 155: This can be a code (for MFA) or a new password (for force password reset). confirmationValue is consistent with the name that is used on the public API for `confirmSignIn()`

packages/amplify_authenticator/lib/src/l10n/generated/input_localizations_en.dart
```diff
@@ -29,6 +29,9 @@ class AuthenticatorInputLocalizationsEn
   @override
   String get password => 'Password';
 
+  @override
+  String get newPassword => 'New Password';
```
Jordan-Nelson, line 33: Previously we were re-using the content for password, so the hint text was "Enter your password", which is confusing because the form was forcing a new password.

dnys1 (approved):

haverchuck (approved):

PR #1055, opened 2021-11-08
feature(amplify_api): Graphql helpers, include parent and decode nested/parent values

ragingsquirrel3:
This PR contains additional functionality for child/parent models in API graphql helpers.
* for all model helpers in models with a parent, include the full parent's selection set in the selection set (up 1 level)
* In the response decoder, decode recursively nested parents/children in the event a custom request is nested that way


packages/amplify_api/test/graphql_helpers_test.dart
```diff
@@ -496,6 +497,153 @@ void main() {
...
+          },
+          'blog': {'id': 'abc123', 'title': 'blog about life'}
+        };
+        final expectedOutput = <String, dynamic>{
+          'id': 'xyz456',
+          'title': 'Lorem Ipsum',
+          'rating': 0,
+          'comments': [
+            {
+              'serializedData': {
```
fjnoyp, line 575: Thanks for writing all of the code to translate between JSON formats.  Do you think it would have been better if we changed our own JSON instead to ensure total uniformity of JSON.

fjnoyp (approved):
Everything looks good thanks for making these changes.  Just left one question.

packages/amplify_api/test/graphql_helpers_test.dart
ragingsquirrel3: long-term, aligning would be better, just don't have time right now to reconcile the datastore part and add extra codegen logic. This is the safest/easiest thing to do for now.

dnys1 (approved):

HuiSF (approved):

PR #1057, opened 2021-11-09
fix(authenticator): remove the label from verify user screen and display email first

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- removes the title from the verify user screen
- updates the ordering so that email always displays before phone
- properly sets the default attribute

**Before**
Screen Shot 2021-11-09 at 9 55 00 AM

**After**
Screen Shot 2021-11-09 at 10 18 22 AM

### DIFF
--- a/packages/amplify_authenticator/lib/src/l10n/input_resolver.dart
+++ b/packages/amplify_authenticator/lib/src/l10n/input_resolver.dart
@@ -327,6 +327,9 @@ class InputResolverKey {
     InputResolverKeyType.title,
     field: InputField.usernameType,
   );
+
+  String resolve(BuildContext context, InputResolver inputResolver) =>
+      inputResolver.resolve(context, this);
 }
 
 class InputResolver extends Resolver<InputResolverKey> {
--- a/packages/amplify_authenticator/lib/src/widgets/form_field.dart
+++ b/packages/amplify_authenticator/lib/src/widgets/form_field.dart
@@ -70,11 +70,7 @@ abstract class AuthenticatorFormField<FieldType, FieldValue,
     this.title,
     this.hintText,
     FormFieldValidator<FieldValue>? validator,
-  })  : assert(
-          titleKey != null || title != null,
-          'Either title or titleKey must be provided',
-        ),
-        validatorOverride = validator,
+  })  : validatorOverride = validator,
         super(key: key);
 
   /// Resolver key for the title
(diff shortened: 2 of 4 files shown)


packages/amplify_authenticator/lib/src/widgets/form_fields/verify_user_form_field.dart
```diff
@@ -210,5 +197,5 @@ class _VerifyAttributeFieldState
   }
 
   @override
-  UsernameAttribute? selectionValue = _default;
```
Jordan-Nelson, line 213: With the current implementation, `_default` is initialized to email, and then is updated on line 191. Line 191 is not running before selectionValue is set (unless the user goes logs out and goes through the flow a second time), causing selectionValue to always be email.

packages/amplify_authenticator/lib/src/widgets/form_field.dart
```diff
@@ -172,15 +168,15 @@ abstract class AuthenticatorFormFieldState<FieldType, FieldValue,
...
+    final String? title =
+        widget.title ?? widget.titleKey?.resolve(context, inputResolver);
+
     return Container(
       margin: FormFieldConstants.marginBottom,
       child: Column(
         crossAxisAlignment: CrossAxisAlignment.start,
         children: <Widget>[
-          Text(title),
+          if (title is String) Text(title),
```
dnys1, line 179: Works but I think I prefer `if (title != null)` and unwrapping it - uglier, but I think it reads more clearly... idk, maybe there's enough context to know that title is a nullable string, and this is a null check. Will let you decide.

packages/amplify_authenticator/lib/src/l10n/input_resolver.dart
```diff
@@ -327,6 +327,9 @@ class InputResolverKey {
     InputResolverKeyType.title,
     field: InputField.usernameType,
   );
+
+  String resolve(BuildContext context, InputResolver inputResolver) =>
```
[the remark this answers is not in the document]
dnys1: Nice!

dnys1 (approved):
LGTM. Thanks!

haverchuck (approved):

packages/amplify_authenticator/lib/src/widgets/form_field.dart
Jordan-Nelson: Yeah, the intention is more clean with `!=null`. Will update.

PR #1060, opened 2021-11-09
chore: remove DateTime.now() from tests

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- remove calls to DateTime.now() from tests to prevent flaky tests

Note: I left a call to DateTime.now() in the storage integ tests that check the DateTime of file uploads. I think the use there is reasonable and I don't think there is an easy alternative.

### DIFF
--- a/packages/amplify_datastore_plugin_interface/test/amplify_temporal_datetime_test.dart
+++ b/packages/amplify_datastore_plugin_interface/test/amplify_temporal_datetime_test.dart
@@ -18,27 +18,27 @@ import 'package:flutter_test/flutter_test.dart';
 
 void main() {
   test('AWSDateTime from DateTime success', () async {
-    DateTime now = DateTime.now();
+    DateTime dateTime = DateTime.parse("2021-11-09T18:53:12.183540Z");
 
-    TemporalDateTime time = TemporalDateTime(now);
-    now = now.toUtc();
+    TemporalDateTime time = TemporalDateTime(dateTime);
+    dateTime = dateTime.toUtc();
 
     expect(time.getOffset(), Duration());
-    expect(time.getDateTimeInUtc(), now);
-    expect(time.format(), now.toIso8601String().substring(0, 26) + "000Z");
+    expect(time.getDateTimeInUtc(), dateTime);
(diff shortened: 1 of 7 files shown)


HuiSF (approved):

packages/amplify_datastore_plugin_interface/test/amplify_temporal_datetime_test.dart
```diff
@@ -134,16 +134,16 @@ void main() {
   });
 
   test('AWSDateTime from offset with single digit duration', () async {
-    DateTime now = DateTime.now();
+    DateTime dateTime = DateTime.parse("2021-11-09T18:53:12.183540Z");
```
Jordan-Nelson, line 137: This is the test that I observed fail in CI. Initially I thought there might be a bug that was only causing failures for certain times, but I think this the code is working as intended and the test is the issue.

I was able to repeat that failure by testing with a DateTime with 0 milliseconds / microseconds. This fails the test because the format() method doesn't include milliseconds / microseconds if the value is 0, and the test is written in a way that assumes the string will include milliseconds / microseconds.

Here is code in the format() method that omits milliseconds / microseconds from the string if the value is 0. 
https://github.com/aws-amplify/amplify-flutter/blob/f658f9487ad94aface04ff9b0c361741ab3f9623/packages/amplify_datastore_plugin_interface/lib/src/types/temporal/temporal_datetime.dart#L144-L146

Here is an example of what the format() method will produce with and without milliseconds / microseconds
```dart
DateTime dateTime1 = DateTime.parse("2021-11-09T18:53:12.183540Z");
dateTime1.format(); // result is "2021-11-09T18:53:12.183540000Z"

DateTime dateTime2 = DateTime.parse("2021-11-09T18:53:12.000000Z");
dateTime2.format(); // result is "2021-11-09T18:53:12Z" **NOT** "2021-11-09T18:53:12.000000Z"
```

I am not sure if there is a specific reason to exclude ".000000", but it seems intentional and as far as I can tell isn't causing any actual issues with DS.

fjnoyp (approved):
Great call on this .. my bad for using .Now() in the first place.  Really appreciate the fix here.

Issue #1059, opened 2021-11-09
Amplify.Storage.list() crashes App on iOS


Ozak93:
**Describe the bug**
When i navigate to a page that calls Amplify's Storage function List, the page freezes and throws the following exception in Xcode's console:

`amplify_storage_s3/FlutterListRequest.swift:37: Fatal error: Unexpectedly found nil while unwrapping an Optional value
2021-11-08 06:34:07.841725-0800 Runner[4310:26946] amplify_storage_s3/FlutterListRequest.swift:37: Fatal error: Unexpectedly found nil while unwrapping an Optional value`

**To Reproduce**
1. Call the following function `await Amplify.Storage.list(path: path);` in iOS Flutter app. 

**Expected behavior**
returns a list of objects in the provided bucket path.

**Screenshots**
![image](https://user-images.githubusercontent.com/31257994/140968189-a5764ae6-9de8-4a9a-bec3-b682fc95a09b.png)

**Platform**
Amplify Flutter current supports iOS and Android.  This issue is reproducible in (check all that apply):
[] Android
[x] iOS

<details>

  <summary>Output of <code>flutter doctor -v</code></summary>
  
  ```[✓] Flutter (Channel stable, 2.5.3, on macOS 11.5.1 20G80 darwin-x64, locale en-US)
    • Flutter version 2.5.3 at /Users/zutch/Desktop/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 18116933e7 (4 weeks ago), 2021-10-15 10:46:35 -0700
    • Engine revision d3ea636dc5
    • Dart version 2.14.4

[✓] Android toolchain - develop for Android devices (Android SDK version 31.0.0)
    • Android SDK at /Users/zutch/Library/Android/sdk
    • Platform android-31, build-tools 31.0.0
    • Java binary at: /Applications/Android Studio.app/Contents/jre/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.10+0-b96-7281165)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 12.5.1, Build version 12E507
    • CocoaPods version 1.11.2

[✗] Chrome - develop for the web (Cannot find Chrome executable at /Applications/Google
    Chrome.app/Contents/MacOS/Google Chrome)
[...]

PR #1061, opened 2021-11-09
fix(storage): Storage.list crash on null "options"

dnys1:
*Issue #, if available:*
- https://github.com/aws-amplify/amplify-flutter/issues/1059

Issue: Omitting the `options` cause a crash on iOS due to improper null checks

*Description of changes:*
- Fix crash on null "options" dict
- Remove validation which did not do anything


haverchuck (approved):

fjnoyp (approved):
Looks good - thanks for doing this!

PR #1072, opened 2021-11-10
fix(amplify_authenticator) uses session for _authLoad

haverchuck:
*Description of changes:*
With this PR we no longer use getCurrentUser when deciding to display the SignIn form or the customer app, as this method has inconsistent behavior between platforms (on main) and returns an AuthUser even when the session has expired (RC). We now use fetchAuthSession.

### DIFF
--- a/packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
+++ b/packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
@@ -124,7 +124,7 @@ class StateMachineBloc {
     yield const AuthLoading();
     final config = await Amplify.asyncConfig;
     yield AuthLoaded(config);
-    yield* _getCurrentUser();
+    yield* _isValidSession();
   }
 
   Stream<AuthState> _getCurrentUser() async* {
@@ -139,6 +139,26 @@ class StateMachineBloc {
     } on AmplifyException catch (e) {
       _exceptionController.add(AuthenticatorException(e.message));
     } on Exception catch (e) {
+      _exceptionController.add(AuthenticatorException(e.toString()));
+    }
+  }
+
+  Stream<AuthState> _isValidSession() async* {
+    try {
+      bool isValidSession = await _authService.isValidSession;
+      if (isValidSession) {
+        yield const Authenticated();
+      } else {
+        yield AuthFlow.signin;
+      }
+    } on AmplifyException catch (e) {
+      /// On all other AmplifyExceptions other than SignedOutException or SessionExpired, we provide a message to the user
+      /// and surface the SignIn form.
+      _exceptionController.add(AuthenticatorException(e.message));
+      yield AuthFlow.signin;
+    } on Exception catch (e) {
+      /// On all other Exceptions we display an an exception and do not
+      /// load the Auththenticator, as the app may be in an unusable state
       _exceptionController.add(AuthenticatorException(e.toString()));
     }
   }
--- a/packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
+++ b/packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
@@ -152,13 +152,13 @@ class StateMachineBloc {
         yield AuthFlow.signin;
       }
 
-      /// The [isValidSession] method catches [SignedOutException] and [SessionExpiredException]
-      /// All other exceptions are re-thrown and caught here.
-    } on Exception {
+      /// The [isValidSession] method will only surface a SessonExpiredException,
+      /// as we want to display a banner to alert users. All other exceptions are swallowed.
+    } on SessionExpiredException {
       /// On all Exceptions other than SignedOutException or SessionExpired,
       /// we provide a generic message to the user and surface the SignIn form.
       _exceptionController.add(const AuthenticatorException(
-          'An unknown exception has occurred while parsing the session.'));
+          'Your session has expired. Please sign in.'));
       yield AuthFlow.signin;
     }
   }
--- a/packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
+++ b/packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
@@ -167,10 +167,9 @@ class StateMachineBloc {
       _exceptionController
           .add(AuthenticatorException(e.message, showBanner: false));
       yield AuthFlow.signin;
-    } on Exception {
-      _exceptionController.add(const AuthenticatorException(
-          'An unknown exception has occurred when fetching the session',
-          showBanner: false));
+    } on Exception catch (e) {
+      _exceptionController
+          .add(AuthenticatorException(e.toString(), showBanner: false));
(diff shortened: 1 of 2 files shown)


packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
```diff
@@ -146,6 +146,26 @@ class StateMachineBloc {
...
+  Stream<AuthState> _isValidSession() async* {
+    try {
+      bool isValidSession = await _authService.isValidSession;
+      if (isValidSession) {
+        yield const Authenticated();
+      } else {
+        yield AuthFlow.signin;
+      }
+    } on AmplifyException catch (e) {
+      /// On all AmplifyExceptions other than SignedOutException or SessionExpired,
```
Jordan-Nelson, line 158: *nit:* It would be nice to note that isValidSession is catching SignedOutException and SessionExpired. I was a little confused what this comment meant since I didn't see those exceptions handled differently.

packages/amplify_authenticator/lib/src/services/amplify_auth_service.dart
```diff
@@ -167,6 +170,21 @@ class AmplifyAuthService implements AuthService {
     }
   }
 
+  @override
+  Future<bool> get isValidSession async {
```
Jordan-Nelson, line 174: It probably doesn't make sense for this to be a getter. I generally would not expect a getter to have any side effects or throw exceptions.
```suggestion
  Future<bool> isValidSession() async {
```

packages/amplify_authenticator/lib/src/services/amplify_auth_service.dart
```diff
@@ -38,6 +39,8 @@ abstract class AuthService {
 
   Future<AuthUser?> get currentUser;
 
+  Future<bool> get isValidSession;
```
Jordan-Nelson, line 42: See comment below. Making this a getter is a little confusing.
```suggestion
  Future<bool> isValidSession();
```

packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
```diff
@@ -146,6 +146,26 @@ class StateMachineBloc {
     }
   }
 
+  Stream<AuthState> _isValidSession() async* {
+    try {
+      bool isValidSession = await _authService.isValidSession;
```
Jordan-Nelson, line 151: See comments in amplify_auth_service.dart. Making this a getter is a little confusing.
```suggestion
      bool isValidSession = await _authService.isValidSession();
```

packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
```diff
@@ -146,6 +146,26 @@ class StateMachineBloc {
...
+      }
+    } on AmplifyException catch (e) {
+      /// On all AmplifyExceptions other than SignedOutException or SessionExpired,
+      /// we provide a message to the user and surface the SignIn form.
+      _exceptionController.add(AuthenticatorException(e.message));
+      yield AuthFlow.signin;
+    } on Exception catch (e) {
+      /// On all other Exceptions we display an an exception and do not
+      /// load the Auththenticator, as the app may be in an unusable state
+      _exceptionController.add(AuthenticatorException(e.toString()));
```
Jordan-Nelson, line 165: This results in the end user seeing a CircularProgressIndicator and a banner with the error message. If an end user actually encountered this, that isn't great. It might be better to just show the the sign in screen.

packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
```diff
@@ -146,6 +146,26 @@ class StateMachineBloc {
...
+      bool isValidSession = await _authService.isValidSession;
+      if (isValidSession) {
+        yield const Authenticated();
+      } else {
+        yield AuthFlow.signin;
+      }
+    } on AmplifyException catch (e) {
+      /// On all AmplifyExceptions other than SignedOutException or SessionExpired,
+      /// we provide a message to the user and surface the SignIn form.
+      _exceptionController.add(AuthenticatorException(e.message));
```
Jordan-Nelson, line 160: Are there known exceptions that this can throw other than SignedOutException or SessionExpired?

I think this is fine for now, but I am wondering how useful this message will be to an end user, and how we will handle this when we localize error messages (assuming we will do that). When/if we do that, we will have to map amplify exceptions to specific messages. For anything that we don't have mapped we will either have to display a generic error message (ex: "An unknown issue occurred." or "We could not validate your session. Please login.") or suppress the error if we don't think it is useful to an end user.

packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
haverchuck: Done.

packages/amplify_authenticator/lib/src/services/amplify_auth_service.dart
haverchuck: Yeah good call.

packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
haverchuck: I went back and forth on this, and felt that maybe not loading anything would be better. But... I agree that an infinite spinner is really maddening.  We could potentially red-screen the app, but I don't think that's much better. For now let's just show SignIn.

packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
haverchuck: You're right about localization... let's go with a single catch with a generic message..

Jordan-Nelson (approved):

packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
```diff
@@ -143,6 +143,38 @@ class StateMachineBloc {
...
+
+      /// On [AmplifyException] and [Exception], update exception controller, do not show banner
+      /// and go to sign in screen.
+    } on AmplifyException catch (e) {
+      _exceptionController
+          .add(AuthenticatorException(e.message, showBanner: false));
+      yield AuthFlow.signin;
+    } on Exception {
+      _exceptionController.add(const AuthenticatorException(
+          'An unknown exception has occurred when fetching the session',
```
dnys1, line 172: Probably just pass back the exception, no?

`AuthenticatorException(e.toString())`

Jordan-Nelson (approved):

packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
haverchuck: @dnys1 Makes sense - done.

dnys1 (approved):

PR #1074, opened 2021-11-11
chore(authenticator): Re-theme

dnys1:
*Issue #, if available:*

*Description of changes:*
- Update theme to match latest Amplify UI specs


packages/amplify_authenticator/assets/Inter/OFL.txt
```diff
@@ -0,0 +1,93 @@
+Copyright 2020 The Inter Project Authors (https://github.com/rsms/inter)
```
haverchuck, line 1: nit: The license in their GH has this line: 
`"Inter" is trademark of Rasmus Andersson.`
Not sure if that's important?

haverchuck (approved):
Left one comment about the license; otherwise, I'm looking forward to seeing this!

packages/amplify_authenticator/assets/Inter/OFL.txt
dnys1: It's the license file they provide

packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -89,8 +89,8 @@ class Authenticator extends StatefulWidget {
     ConfirmSignInMFAForm? confirmSignInMFAForm,
     this.stringResolver = const AuthStringResolver(),
     required this.child,
-  })  : useAmplifyTheme = false,
-        super(key: key) {
+    this.useAmplifyTheme = true,
```
Jordan-Nelson, line 92: IMO this should still default to false. The standard material theme is going to feel _much_ more familiar to customers than the amplify theme. Also, if customers drop this into an existing MaterialApp that has the `theme` attribute set, the authenticator will pick up their theme automatically if this is false, which would be pretty nice.

On a related note, the tabs on the authenticator don't look very good if the user sets this to false. 

Screen Shot 2021-11-11 at 12 39 56 PM

PR #1077, opened 2021-11-11
chore(authenticator): Update username attributes

dnys1:
*Issue #, if available:*

*Description of changes:*
- Move `CognitoUserAttributeKey` to the plugin interface. Use it in `amplify_flutter` config parsing.
- Move username logic to separate mixin which all username fields can use.
- Add toggle for email or phone number option.


packages/amplify_authenticator/lib/src/enums/username_input.dart
```diff
@@ -13,17 +13,14 @@
  * permissions and limitations under the License.
  */
 
-enum UsernameAttribute { email, phoneNumber }
+enum UsernameType { username, email, phoneNumber }
```
haverchuck, line 16: We named this UsernameAttribute and exclused 'username' to tie it to the Cognito configuration, which is often confusing. I thought that creating any abstraction around this might add to confusion down the road. But I agree that what you have here is cleaner code and makes more sense within the Authenticator. Could you maybe just add some documentation here to reference the cognito usernameAttribute value?

packages/amplify_authenticator/lib/src/mixins/username_field.dart
```diff
@@ -0,0 +1,224 @@
...
+import 'package:flutter/material.dart';
+
+enum _UsernameType {
+  username,
+  email,
+  phoneNumber,
+  emailOrPhoneNumber,
+}
+
+mixin AuthenticatorUsernameField<FieldType,
```
haverchuck, line 29: Ideally we would make this a generic switch mixin which can be used for other fields. It's not a priority though.

Jordan-Nelson (approved):

haverchuck (approved):
Left a suggestion about adding documentation, but otherwise LGTM.

PR #1081, opened 2021-11-11
feat(amplify_authenticator): adds search bar in dial code lookup, and additional country localizations

haverchuck:
*Description of changes:*
Adds localizations for Algeria - Cook Islands (UK and US still at top of list as they were my test cases, but we can change that).

Adds search field to filter the countries using their *localized* values:
image


packages/amplify_authenticator/lib/src/l10n/country_resolver.dart
```diff
@@ -19,36 +19,373 @@ import 'package:flutter/material.dart';
...
 class CountryResolver extends Resolver<CountryResolverKey> {
   const CountryResolver();
 
-  /// The label for the United States.
   String selectDialCode(BuildContext context) {
     return AuthenticatorLocalizations.countriesOf(context).selectDialCode;
   }
 
-  /// The label for the United States.
+  String noDialCodeSearchResults(BuildContext context) {
```
HuiSF, line 78: It would be nice if all these classes, fields, methods and getters can be automatically generated based on a `json` file or something similar.

HuiSF (approved):

packages/amplify_authenticator/lib/src/l10n/country_resolver.dart
haverchuck: @HuiSF Made some progress on this, but I don't know when it will be production ready.

haverchuck: @HuiSF We now (somewhat clunky) code that generates all the necessart country information.

packages/amplify_authenticator/example/lib/localized_country_resolver.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:amplify_authenticator/amplify_authenticator.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_gen/gen_l10n/app_localizations.dart';
+
+/// Our custom button string resolver which calls into our localizations class.
```
Jordan-Nelson, line 5: Nit: "Our custom country code resolver"

Jordan-Nelson (approved):

PR #1094, opened 2021-11-15
feat(amplify_api): subscription model-based graphql helpers

ragingsquirrel3:
This PR contains logic to to add subscription request generators for model-based graphql helpers. It has the request generation logic, subscription event decoding, as well as unit/integration tests for that stuff.

This PR also adds private constructors to the helper classes so they cannot be instantiated (as they only have static methods). I noticed this makes code lens a little easier to use as constructors no longer there.


packages/amplify_api/example/integration_test/graphql_tests.dart
```diff
@@ -112,10 +115,11 @@ void main() {
       await deleteAllBlogsAndPosts();
     });
 
-    testWidgets('should fetch', (WidgetTester tester) async {
-      const listBlogs = 'listBlogs';
-      const items = 'items';
-      String graphQLDocument = '''query MyQuery {
+    group('queries', () {
```
ragingsquirrel3, line 118: adding these groups (and indentation) caused the diff to explode. Sorry about that. Only subscriptions really changed here (and added a field to create helper selection set).

dnys1 (approved):
Nice one!

HuiSF (approved):
🚀

PR #1096, opened 2021-11-15
chore(authenticator): Custom fields

dnys1:
*Issue #, if available:*

*Description of changes:*
- Merge custom fields with default and required fields
- Add marker for required fields (currently an asterisk)
- Clean up form fields

### DIFF
--- a/packages/amplify_authenticator/lib/src/widgets/form.dart
+++ b/packages/amplify_authenticator/lib/src/widgets/form.dart
@@ -129,7 +129,7 @@ class AuthenticatorFormState<T extends AuthenticatorForm<T>>
     return fields.toList(growable: false)
       ..sort((a, b) {
         // Sort larger priorities first.
-        return -a.priority.compareTo(b.priority);
+        return -a.displayPriority.compareTo(b.displayPriority);
       });
   }
 
--- a/packages/amplify_authenticator/lib/src/widgets/form.dart
+++ b/packages/amplify_authenticator/lib/src/widgets/form.dart
@@ -143,12 +143,13 @@ class AuthenticatorFormState<T extends AuthenticatorForm<T>>
           ..._allFields,
           Column(
             children: [
+              if (widget.actions.isNotEmpty) const SizedBox(height: 4),
               ...widget.actions,
               if (_runtimeActions.isNotEmpty) ...[
                 const Divider(),
                 ..._runtimeActions,
               ]
-            ].spacedBy(const SizedBox(height: 10)),
+            ].spacedBy(const SizedBox(height: 12)),
           ),
         ],
       ),
@@ -286,6 +287,7 @@ class SignInForm extends AuthenticatorForm<SignInForm> {
           ],
           actions: const [
             SignInButton(),
+            ForgotPasswordButton(),
           ],
         );
 
@@ -468,8 +470,8 @@ class SendCodeForm extends AuthenticatorForm<SendCodeForm> {
             SignInFormField.username(),
           ],
           actions: const [
-            BackToSignInButton(),
             SendCodeButton(),
+            BackToSignInButton(),
           ],
         );
 
--- a/packages/amplify_authenticator/lib/src/widgets/form.dart
+++ b/packages/amplify_authenticator/lib/src/widgets/form.dart
@@ -109,7 +109,7 @@ class AuthenticatorFormState<T extends AuthenticatorForm<T>>
     );
   }
 
-  List<AuthenticatorFormField> get _allFields {
+  List<AuthenticatorFormField> get allFields {
     final fields = HashSet<AuthenticatorFormField>(
       equals: (a, b) {
         return a.runtimeType == b.runtimeType && a.field == b.field;
@@ -140,7 +140,7 @@ class AuthenticatorFormState<T extends AuthenticatorForm<T>>
       key: viewModel.formKey,
       child: Column(
         children: [
-          ..._allFields,
+          ...allFields,
           Column(
             children: [
               if (widget.actions.isNotEmpty) const SizedBox(height: 4),
--- a/packages/amplify_authenticator/lib/src/widgets/form.dart
+++ b/packages/amplify_authenticator/lib/src/widgets/form.dart
@@ -15,6 +15,8 @@
 
 library authenticator.form;
 
+import 'dart:collection';
+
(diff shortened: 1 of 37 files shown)


packages/amplify_authenticator/lib/src/widgets/form.dart
```diff
@@ -102,15 +109,38 @@ class AuthenticatorFormState<T extends AuthenticatorForm<T>>
...
+    fields.addAll(runtimeFields(context));
+
+    // Add widget's fields last, since they will contain user's custom fields
+    // and will override the default fields included by runtimeFields.
+    fields.addAll(widget.customFields.where(
+      (el) => el.usernameType != selectedUsernameType,
+    ));
+
+    return fields.toList(growable: false)
+      ..sort((a, b) {
```
Jordan-Nelson, line 130: This takes away the developer's ability to specify the order of the fields with `SignUpForm.custom`.

Maybe we can take `requiredFields`, and remove any values that are also in `requiredFields`, and then display the left over `requiredFields` first, followed by the  `customFields`? That way by default `requiredFields` are first, followed by `customFields` in the order that the customer specifies. If the customer wants to move any of the `requiredFields` down lower in the list, they can just pass them in as a customField in the order they want.

Jordan-Nelson: Maybe we still leave sorting in for the runtime fields if we think it's needed. I understand the use case for the run time fields. The require fields we already control the order of and seem like they should be first, and the custom fields the customer controls the order of, and I think should go last since they can always just include the required/runtime fields in as custom fields if they want to fully control the order.

packages/amplify_authenticator/lib/src/widgets/form_fields/sign_up_form_field.dart
```diff
@@ -237,10 +262,84 @@ abstract class SignUpFormField<FieldValue> extends AuthenticatorFormField<
...
         attributeKey: attributeKey,
+        required: required,
       );
 
   /// Custom Cognito attribute key.
   final CognitoUserAttributeKey? _customAttributeKey;
+
+  @override
+  int get priority {
+    switch (field) {
```
haverchuck, line 273: Is priority descriptive enough? Maybe something like `displayPriority`?  If you feel strongly I'm fine with this as is.

dnys1: I agree - I prefer `displayPriority`. Updated.

packages/amplify_authenticator/lib/src/widgets/form.dart
dnys1: May need to sync on this

packages/amplify_authenticator/lib/src/widgets/form.dart
```diff
@@ -102,15 +109,38 @@ class AuthenticatorFormState<T extends AuthenticatorForm<T>>
...
+        return a.runtimeType == b.runtimeType && a.field == b.field;
+      },
+      hashCode: (field) => hashValues(field.runtimeType, field.field),
+    );
+
+    fields.addAll(widget.requiredFields);
+    fields.addAll(runtimeFields(context));
+
+    // Add widget's fields last, since they will contain user's custom fields
+    // and will override the default fields included by runtimeFields.
```
Jordan-Nelson, line 124: `.add()` and `.addAll()` will not change the set if an equal value is already in the set, so by adding customFields last they won't override a value that is already there.

packages/amplify_authenticator/lib/src/widgets/form.dart
Jordan-Nelson: With the code as is, if a user provides the following signUpForm with a default username/password config, the fields will be displayed in the order username, password, confirm password, email, address. The order they are displayed in doesn't match the order that the customer provided them in. The customer has no ability to control the order.

```dart
  fields: [
    SignUpFormField.username(),
    SignUpFormField.email(),
    SignUpFormField.address(),
    SignUpFormField.password(),
    SignUpFormField.passwordConfirmation(),
  ],
),
```

We need a logical default order that works for most use cases, and then let the customer easily change the order as they want.

I think a reasonable default order would be required fields -> runtime fields -> custom fields, with the following handling of duplicates. If a runtime field is already in the list of required fields, it should be ignored. If a required field or runtime field is in the list of custom fields, it should be ignored. I can see some logic in providing an order for the runtime fields, since we don't know what order the CLI will list them in, but as long as the user has the ability to control it, I don't think that is really needed. 

I think this provides a pretty reasonable default ordering of fields in most case, and still gives the developer full control of the order if they want to change it. If they provide the `signUpForm` from the example above, the fields are displayed in the order they provided them in (username, email, address, password, confirm password).

Here is an example of what that implementation might look like

```dart
List<AuthenticatorFormField> get _allFields {
  List<AuthenticatorFormField> fields = [];

  // add all required fields
  fields.addAll(widget.requiredFields);

  // add all runtime fields, as long as they are not already a required field
  for (var field in runtimeFields(context)) {
    if (!fields.contains(field)) {
      fields.add(field);
    }
  }

  // add all custom fields, replacing any existing fields of the same type
  for (var field in widget.customFields) {
    // remove will have no impact if the field is not present
    fields.remove(field);
    fields.add(field);
  }

  return fields;
}
```

For that implementation, you also need an equality override on `AuthenticatorFormField`, but the same could be achieved without it.

```dart
  @override
  bool operator ==(Object other) {
    if (identical(other, this)) return true;
    return other is AuthenticatorFormField &&
        other.runtimeType == runtimeType &&
        other.field == field;
  }
```

packages/amplify_authenticator/lib/src/widgets/form.dart
dnys1: Oh yeah, I was thinking about it backwards

packages/amplify_authenticator/lib/src/widgets/form.dart
dnys1: Okay, I think we almost achieve this with my implementation. Maybe can leverage LinkedHashSet to improve the ordering. Will work it out in the morning.

Jordan-Nelson (approved):
I think we could do away with `displayPriority`, but I won't hold up the PR over it

PR #1100, opened 2021-11-16
chore(authenticator): Feedback fixes

dnys1:
*Issue #, if available:*

*Description of changes:*
- Use username field for confirm sign up
- No validation for optional fields
- Fix validation for phone number field
- Fix `obscureText`
- Fix some required fields showing as optional

### DIFF
--- a/packages/amplify_authenticator/lib/src/mixins/authenticator_phone_field.dart
+++ b/packages/amplify_authenticator/lib/src/mixins/authenticator_phone_field.dart
@@ -15,6 +15,7 @@
 
 import 'package:amplify_authenticator/src/l10n/country_resolver.dart';
 import 'package:amplify_authenticator/src/utils/country_code.dart';
+import 'package:amplify_authenticator/src/utils/validators.dart';
 import 'package:amplify_authenticator/src/widgets/authenticator_input_config.dart';
 import 'package:amplify_authenticator/src/widgets/form_field.dart';
 import 'package:flutter/material.dart';
@@ -48,6 +49,14 @@ mixin AuthenticatorPhoneFieldMixin<FieldType,
             .contains(_searchVal.toLowerCase()),
       )
       .toList();
+
+  @override
+  FormFieldValidator<String> get validator {
+    if (isOptional) {
+      return super.validator;
+    }
+    return validatePhoneNumber;
+  }
 
   String formatPhoneNumber(String phoneNumber) {
     return phoneNumber.ensureStartsWith('+${_selectedCountry.value}');
--- a/packages/amplify_authenticator/lib/src/mixins/authenticator_phone_field.dart
+++ b/packages/amplify_authenticator/lib/src/mixins/authenticator_phone_field.dart
@@ -52,12 +52,9 @@ mixin AuthenticatorPhoneFieldMixin<FieldType,
 
   @override
   FormFieldValidator<String> get validator {
-    if (isOptional) {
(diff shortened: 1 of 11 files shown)


packages/amplify_authenticator/lib/src/mixins/authenticator_phone_field.dart
```diff
@@ -49,8 +50,19 @@ mixin AuthenticatorPhoneFieldMixin<FieldType,
       )
       .toList();
 
-  String formatPhoneNumber(String phoneNumber) {
-    return phoneNumber.ensureStartsWith('+${_selectedCountry.value}');
+  @override
+  FormFieldValidator<String> get validator {
+    if (isOptional) {
+      return super.validator;
```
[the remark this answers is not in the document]
Jordan-Nelson: This might apply to all fields that have a specific validator (phone, email, date, etc.)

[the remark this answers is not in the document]
haverchuck: Seems like for some of these, we need a validator that requires a value if the field is required, but applies additional validation only if a value is present?

packages/amplify_authenticator/lib/src/utils/validators.dart
```diff
@@ -117,6 +117,9 @@ String? validatePhoneNumber(String? phoneNumber) {
   if (phoneNumber == null || phoneNumber.isEmpty) {
     return 'Phone number cannot be empty';
   }
+  if (!phoneNumberRegex.hasMatch(phoneNumber)) {
+    return 'Invalid phone number format';
```
haverchuck, line 121: We need to go through these validators and add localization. I can take that in a separate PR.

packages/amplify_authenticator/lib/src/mixins/authenticator_phone_field.dart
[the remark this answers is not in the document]
haverchuck: Agree with @Jordan-Nelson 's comment about performing validation but allowing null on optional fields.

packages/amplify_authenticator/lib/src/utils/validators.dart
```diff
@@ -130,8 +151,14 @@ String? validateEmail(String? email) {
   return null;
 }
 
-String? validateCode(String? code) {
+String? validateCode(
+  String? code, {
+  required bool isOptional,
```
haverchuck, line 156: Would code ever be optional

packages/amplify_authenticator/lib/src/widgets/form_fields/sign_in_form_field.dart
```diff
@@ -90,16 +90,18 @@ abstract class _SignInFormFieldState<FieldValue>
...
+      case SignInField.password:
+        return true;
     }
   }
 
   @override
   TextInputType get keyboardType {
     switch (widget.field) {
       case SignInField.username:
+        return TextInputType.text;
```
Jordan-Nelson, line 104: 👍

packages/amplify_authenticator/lib/src/utils/validators.dart
dnys1: Hmm I guess it wouldn't. Updated.

haverchuck (approved):

Jordan-Nelson (approved):

PR #1103, opened 2021-11-16
fix(authenticator): update theme _isDark for material theme

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- update _isDark for material theme

### DIFF
--- a/packages/amplify_authenticator/lib/src/theme/amplify_theme.dart
+++ b/packages/amplify_authenticator/lib/src/theme/amplify_theme.dart
@@ -647,7 +647,7 @@ class _AmplifyThemeWrapper {
     if (!useAmplifyTheme) {
       return ElevatedButtonTheme.of(context);
     }
-    return _isDark
+    return isDark
         ? AmplifyTheme.elevatedButtonThemeDark
         : AmplifyTheme.elevatedButtonTheme;
   }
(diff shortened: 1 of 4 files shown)


packages/amplify_authenticator/lib/src/theme/amplify_theme.dart
```diff
@@ -880,6 +880,7 @@ class _AmplifyThemeWrapper {
   Color? get transparent =>
       useAmplifyTheme ? AmplifyColors.transparent : Colors.transparent;
 
-  bool get _isDark =>
-      MediaQuery.platformBrightnessOf(context) == Brightness.dark;
+  bool get _isDark => useAmplifyTheme
```
Jordan-Nelson, line 883: Most of the places `_isDark` only applies when `useAmplifyTheme == true`, however there are a few locations where it is still used when `useAmplifyTheme == false`. That was causing the system brightness to be used, instead of the brightness from MaterialTheme. It resulted in the text on the tabs being the same color as the background if the system brightness and MaterialTheme brightness did not match.

Screen Shot 2021-11-16 at 4 19 05 PM

haverchuck (approved):

dnys1 (approved):
Would add `darkTheme` so that it responds to system brightness fully

dnys1 (approved):

PR #1110, opened 2021-11-17
feat: support email & phone in confirm sign in

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- support the ability to customize the confirm sign in screen
- add email and phone attributes to the confirm sign in screen

With these changes, the following code snippet will produce the screenshot below it.

```dart
final authenticator = Authenticator(
  confirmSignInNewPasswordForm: ConfirmSignInNewPasswordForm.custom(
    fields: [
      ConfirmSignInFormField.email(required: true),
      ConfirmSignInFormField.custom(
          required: true,
          attributeKey: CognitoUserAttributeKey.name,
          title: 'Name',
          hintText: 'Enter your name'),
    ],
  ),
  child: const SignedInScreen(),
);
```

Screen Shot 2021-11-17 at 10 26 08 AM

### DIFF
--- a/packages/amplify_authenticator/lib/amplify_authenticator.dart
+++ b/packages/amplify_authenticator/lib/amplify_authenticator.dart
@@ -100,7 +100,7 @@ class Authenticator extends StatefulWidget {
     Key? key,
     SignInForm? signInForm,
     SignUpForm? signUpForm,
-    ConfirmSignInMFAForm? confirmSignInMFAForm,
+    ConfirmSignInNewPasswordForm? confirmSignInNewPasswordForm,
     this.stringResolver = const AuthStringResolver(),
     required this.child,
     this.useAmplifyTheme = false,
@@ -109,7 +109,8 @@ class Authenticator extends StatefulWidget {
   }) : super(key: key) {
     this.signInForm = signInForm ?? SignInForm();
     this.signUpForm = signUpForm ?? SignUpForm();
-    this.confirmSignInMFAForm = confirmSignInMFAForm ?? ConfirmSignInMFAForm();
+    this.confirmSignInNewPasswordForm =
+        confirmSignInNewPasswordForm ?? ConfirmSignInNewPasswordForm();
   }
 
   /// Whether to use Amplify colors and styles in the Authenticator,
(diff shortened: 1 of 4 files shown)


packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -100,7 +100,7 @@ class Authenticator extends StatefulWidget {
     Key? key,
     SignInForm? signInForm,
     SignUpForm? signUpForm,
-    ConfirmSignInMFAForm? confirmSignInMFAForm,
```
Jordan-Nelson, line 103: I am not sure if we need to support a customizable MFA screen, but it wasn't really supported since `ConfirmSignInMFAForm` only has a default constructor. I think this was added when `ConfirmSignInMFAForm` and `ConfirmSignInNewPasswordForm` were one class, but now they are separate.

packages/amplify_authenticator/lib/src/widgets/form_fields/confirm_sign_in_form_field.dart
```diff
@@ -217,6 +251,55 @@ abstract class _ConfirmSignInFormFieldState<FieldValue>
   }
 }
 
+class _ConfirmSignInPhoneField extends ConfirmSignInFormField<String> {
```
Jordan-Nelson, line 254: copy/paste from SignUp. Will look into reducing duplicated code in a follow up PR.

packages/amplify_authenticator/lib/src/mixins/authenticator_text_field.dart
```diff
@@ -26,7 +26,7 @@ mixin AuthenticatorTextField<FieldType,
   Widget buildFormField(BuildContext context) {
     final inputResolver = stringResolver.inputs;
     final hintText = widget.hintText == null
-        ? inputResolver.resolve(context, widget.hintTextKey!)
+        ? widget.hintTextKey?.resolve(context, inputResolver)
```
Jordan-Nelson, line 29: Not directly related to the PR, but I happened to try to add a custom field without hint text, and this threw.

haverchuck: Which form field type? I think hintText is required in `static SignUpFormField custom`... maybe we should do that elsewhere?

haverchuck (approved):
I left some comments/questions, but overall LGTM.

dnys1 (approved):
Looks good - thanks!

packages/amplify_authenticator/lib/src/mixins/authenticator_text_field.dart
Jordan-Nelson: I think all of the other form fields have it as optional, although none of them were really being used until this PR. We should make it the same. I would be fine with either. Since hintText is optional on `TextFormField`, I would lean towards making it optional on our form fields.

I can try to remember to update this as well in the follow up

PR #1111, opened 2021-11-17
chore(authenticator): Add sign-up-with-username test

dnys1:
*Issue #, if available:*

*Description of changes:*
- Add first integration test to follow Amplify UI feature: https://github.com/aws-amplify/amplify-ui/blob/main/packages/e2e/features/ui/components/authenticator/sign-up-with-username.feature

### DIFF
--- a/packages/amplify_authenticator/example/integration_test/config.dart
+++ b/packages/amplify_authenticator/example/integration_test/config.dart
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
+import 'package:amplify_flutter/amplify.dart';
+
+import 'envs/auth_with_username.dart' as auth_with_username;
+
+const environmentsByConfiguration = {
+  // From: https://github.com/aws-amplify/amplify-ui/blob/main/examples/angular/src/pages/ui/components/authenticator/sign-up-with-username/sign-up-with-username.component.ts
+  'ui/components/authenticator/sign-up-with-username': 'auth-with-username',
+};
+
+const environments = {
+  'auth-with-username': auth_with_username.amplifyconfig,
+};
+
+Future<void> loadConfiguration(String configurationName) async {
+  final envName = environmentsByConfiguration[configurationName]!;
+  final envConfig = environments[envName]!;
+  await Amplify.addPlugin(AmplifyAuthCognito());
+  await Amplify.configure(envConfig);
+}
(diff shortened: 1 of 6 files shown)


packages/amplify_authenticator/example/integration_test/config.dart
```diff
@@ -0,0 +1,35 @@
...
+ * permissions and limitations under the License.
+ */
+
+import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
+import 'package:amplify_flutter/amplify.dart';
+
+import 'envs/auth_with_username.dart' as auth_with_username;
+
+const environmentsByConfiguration = {
+  // From: https://github.com/aws-amplify/amplify-ui/blob/main/examples/angular/src/pages/ui/components/authenticator/sign-up-with-username/sign-up-with-username.component.ts
```
haverchuck, line 22: 👍

packages/amplify_authenticator/example/integration_test/sign_up_with_username_test.dart
```diff
@@ -0,0 +1,181 @@
...
+        ),
+        findsOneWidget,
+      );
+    });
+
+    // Scenario: "Preferred Username" is included from `aws_cognito_signup_attributes`
+    testWidgets(
+        '"Preferred Username" is included from `aws_cognito_signup_attributes`',
+        (tester) async {
+      // TODO: Clarify requirements
```
haverchuck, line 66: Is this TODO from JS? I think there may have been some confusion from the `alias` change, but in this case it is valid to just include this as a regular signup attribute if it's in the config.

dnys1: No, it's for us to clarify - I agree, they probably just need to update their tests. But with the zero-config auth change, preferred username would not be shown in addition to the username field if the user selects `Username` as their login mechanism.

haverchuck: If that's the case right now, it may be a bug. I think that preferred_username is a valid attribute to collect in addition to username/phone/email. The use case would be a *mutable* display name in a social media app, or something of that nature. I'll clarify this.

haverchuck (approved):

packages/amplify_authenticator/example/integration_test/sign_up_with_username_test.dart
dnys1: Sounds good, thanks

packages/amplify_authenticator/example/integration_test/config.dart
```diff
@@ -0,0 +1,35 @@
...
+  'ui/components/authenticator/sign-up-with-username': 'auth-with-username',
+};
+
+const environments = {
+  'auth-with-username': auth_with_username.amplifyconfig,
+};
+
+Future<void> loadConfiguration(String configurationName) async {
```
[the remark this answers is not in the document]
dnys1: `configureEnv` is only partly what it does - first it retrieves the environment name from the configurations table. Maybe `setupConfiguration`?

packages/amplify_authenticator/example/integration_test/sign_up_with_username_test.dart
```diff
@@ -0,0 +1,193 @@
...
+
+extension on WidgetTester {
+  /// Loads the given [authenticator] widget and waits for it to render.
+  Future<void> loadAuthenticator(Widget authenticator) async {
+    await pumpWidget(authenticator);
+    await pumpAndSettle();
+  }
+
+  /// Navigate to the "Sign Up" screen.
+  Future<void> navigateToSignUp() async {
```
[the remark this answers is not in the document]
dnys1: Yeah, didn't want to prematurely decide on a structure for these until we had done more cases.

packages/amplify_authenticator/example/integration_test/sign_up_with_username_test.dart
```diff
@@ -0,0 +1,193 @@
...
+      // with fixture "sign-up-with-username"
+      // TODO: Clarify requirements
+
+      // When I type a new "username"
+      await tester.enterText(
+        find.byKey(keyUsernameSignUpFormField),
+        generateUsername(),
+      );
+
+      // And I type my password
```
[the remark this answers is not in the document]
dnys1: Same comment as above - this will likely be what happens. Just didn't seem like the right PR for it.

Jordan-Nelson (approved):

PR #1112, opened 2021-11-17
chore(amplify_api): graphql model helpers with embedded (custom) types

ragingsquirrel3:
This PR contains logic to make API model-based graphql helpers work with embedded types (aka custom types). It adds them to the selection set, allows putting them in mutations, and decodes them in responses. There are also tests for that, including changes to integration test provisioning. 


packages/amplify_api/example/integration_test/graphql_tests.dart
```diff
@@ -285,6 +285,23 @@ void main() {
         expect(data?.id, equals(blog.id));
       });
 
+      testWidgets('should CREATE a blog with Model helper and embedded type',
```
HuiSF, line 288: nit: I think we should use custom type for consistency... non-model, embedded different names could be confusing :D custom type is the formal name referred by Admin UI

HuiSF (approved):
Thanks for testing and supporting this addtion.

packages/amplify_api/example/integration_test/graphql_tests.dart
ragingsquirrel3: ok, will update, good call

dnys1 (approved):

PR #1125, opened 2021-11-19
chore: implement page object pattern

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- refactor tests to use the [Page Object Pattern](https://www.protractortest.org/#/page-objects)


packages/amplify_authenticator/example/integration_test/sign_up_with_username_test.dart
```diff
@@ -45,106 +45,70 @@ void main() {
...
-      await tester.loadAuthenticator(authenticator);
-      await tester.navigateToSignUp();
-
-      // Then I see "Username" as an input field
-      expect(find.byKey(keyPreferredUsernameSignUpFormField), findsOneWidget);
+      SignUpPage signUpPage = SignUpPage(tester: tester);
+      SignInPage signInPage = SignInPage(tester: tester);
+      await loadAuthenticator(tester: tester, authenticator: authenticator);
+      await signInPage.navigateToSignUp();
+      signUpPage.expectPreferredUserNameIsPresent();
```
Jordan-Nelson, line 63: This step of the test seems to be flaky. This is the case in the amplify-authenticator branch as well. I am having trouble figuring out why, but since it is unrelated I am going to fix it in a follow up PR.

packages/amplify_authenticator/example/integration_test/sign_up_with_username_test.dart
```diff
@@ -45,106 +45,70 @@ void main() {
...
-      // And I type a new "preferred username"
-      await tester.enterText(
-        find.byKey(keyPreferredUsernameSignUpFormField),
-        generateUsername(),
-      );
+      await signUpPage.enterUsername(username);
+      await signUpPage.enterPassword(password);
+      await signUpPage.enterPasswordConfirmation(password);
+      await signUpPage.enterEmail(email);
+      await signUpPage.enterPreferredUsername(username);
```
Jordan-Nelson, line 107: Similar to the note above, this seems to be flaky. Will address in a follow up PR.

packages/amplify_authenticator/example/integration_test/sign_up_with_username_test.dart
```diff
@@ -45,106 +48,69 @@ void main() {
...
+      SignUpPage signUpPage = SignUpPage(tester: tester);
+      SignInPage signInPage = SignInPage(tester: tester);
+      await loadAuthenticator(tester: tester, authenticator: authenticator);
+      await signInPage.navigateToSignUp();
+      signUpPage.expectUserNameIsPresent();
     });
 
     // Scenario: "Preferred Username" is included from `aws_cognito_signup_attributes`
     testWidgets(
-        '"Preferred Username" is included from `aws_cognito_signup_attributes`',
```
Jordan-Nelson, line 65: The back ticks were causing the following error when running the test individually from VS code.

```
No tests match regular expression "^sign-in-with-username "Preferred Username" is included from ( \(variant: .*\))?$".
```

packages/amplify_authenticator/example/integration_test/sign_up_with_username_test.dart
Jordan-Nelson: Addressed with https://github.com/aws-amplify/amplify-flutter/pull/1125/commits/e5b8ffffe16c69fc2b6d222c547a6a32c3957fca

packages/amplify_authenticator/example/integration_test/sign_up_with_username_test.dart
Jordan-Nelson: Addressed with https://github.com/aws-amplify/amplify-flutter/pull/1125/commits/e5b8ffffe16c69fc2b6d222c547a6a32c3957fca

packages/amplify_authenticator/example/integration_test/sign_up_with_username_test.dart
```diff
@@ -17,19 +17,22 @@
 // https://github.com/aws-amplify/amplify-ui/blob/main/packages/e2e/features/ui/components/authenticator/sign-up-with-username.feature
```
haverchuck, line 17: nit: Personally I would move this to after the import statements so it doesn't blend in so much with the license comment block.

haverchuck (approved):
I left one minor nit, but otherwise LGTM.

packages/amplify_authenticator/example/integration_test/sign_up_with_username_test.dart
```diff
@@ -45,106 +48,69 @@ void main() {
...
-      // And I type a new "preferred username"
-      await tester.enterText(
-        find.byKey(keyPreferredUsernameSignUpFormField),
-        generateUsername(),
-      );
+      await signUpPage.enterUsername(username);
+      await signUpPage.enterPassword(password);
+      await signUpPage.enterPasswordConfirmation(password);
+      await signUpPage.enterEmail(email);
+      await signUpPage.enterPreferredUsername(username);
```
dnys1, line 109: Not sure it matters, but I would probably make this a separate username than above

dnys1 (approved):

PR #1134, opened 2021-11-23
feat(amplify_authenticator): sets country selection default based on …

haverchuck:
*Description of changes:*
Sets the default dial code selection based on the user's device location.

### DIFF
--- a/packages/amplify_authenticator/lib/src/utils/country_code.dart
+++ b/packages/amplify_authenticator/lib/src/utils/country_code.dart
@@ -14,6 +14,7 @@
  */
 
 import 'package:amplify_authenticator/src/l10n/country_resolver.dart';
+import 'package:flutter/widgets.dart';
 
 class Country {
   final String countryCode;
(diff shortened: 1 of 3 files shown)


packages/amplify_authenticator/lib/src/utils/country_code.dart
```diff
@@ -270,3 +271,13 @@ const countryCodes = [
   Country('ZM', '260', CountryResolverKey.zm$),
   Country('ZW', '263', CountryResolverKey.zw$)
 ];
+
+Country? findDeviceLocationFromList() {
+  try {
+    return countryCodes.firstWhere((country) =>
+        country.countryCode ==
+        WidgetsBinding.instance?.window.locale.countryCode);
```
Jordan-Nelson, line 279: Should we use `Localizations.localeOf(context).countryCode`? I think `WidgetsBinding.instance?.window.locale.countryCode` always grabs from their device, even if they have specified a different value for the app. I would assume  `Localizations.localeOf(context).countryCode` falls back to the device setting if a value wasn't explicitly set, although I am not positive.

PR #1135, opened 2021-11-23
feat: add remaining fields to confirm sign in

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- add remaining fields to confirm sign in


packages/amplify_authenticator/lib/src/widgets/form_fields/sign_up_form_field.dart
```diff
@@ -251,7 +251,7 @@ abstract class SignUpFormField<FieldValue> extends AuthenticatorFormField<
     Key? key,
     required String title,
     required CognitoUserAttributeKey attributeKey,
-    required String hintText,
```
Jordan-Nelson, line 254: I updated this for consistency with Confirm Sign In. I don't think hint text needs to be required. It isn't required for TextFormField / InputDecoration, and it makes sense to try to match to APIs.

dnys1 (approved):
Looks good - thanks!

packages/amplify_authenticator/lib/src/widgets/form_fields/sign_up_form_field.dart
haverchuck: What happens if the customer creates a custom form field and doesn't supply hint text?

Jordan-Nelson: It just won't display any hint text then

Screen Shot 2021-11-23 at 12 51 31 PM
y

haverchuck (approved):

PR #1144, opened 2021-11-24
chore(core): Amplify configuration revamp

dnys1:
*Issue #, if available:*

*Description of changes:*
- Fully tested, documented, and public-facing plugin configuration


packages/amplify_flutter/lib/src/config/amplify_config.dart
```diff
@@ -1,47 +1,69 @@
...
- * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
- * express or implied. See the License for the specific language governing
- * permissions and limitations under the License.
- */
-
-import 'package:amplify_flutter/src/config/analytics/analytics_config.dart';
-import 'package:amplify_flutter/src/config/api/api_config.dart';
-import 'package:amplify_flutter/src/config/auth/auth_config.dart';
-import 'package:amplify_flutter/src/config/storage/storage_config.dart';
+//
```
haverchuck, line 1: Is there any particular reason we are switching from ` /* comment */` to `/// comment`?  Asking out of curiosity.

packages/amplify_flutter/lib/src/config/auth/cognito/auth.dart
```diff
@@ -0,0 +1,81 @@
...
+  final List<MfaType>? mfaTypes;
+  final List<CognitoUserAttributeKey>? verificationMechanisms;
+
+  @override
+  List<Object?> get props => [
+        oAuth,
+        authenticationFlowType,
+        socialProviders,
+        loginMechanism,
+        loginMechanisms,
```
haverchuck, line 67: I haven't gotten through the whole PR... I am not sure what the purpose of these two singular/plural variants are?

packages/amplify_flutter/lib/src/config/auth/cognito/auth.g.dart
```diff
@@ -0,0 +1,138 @@
...
+      mfaConfiguration: $enumDecodeNullable(
+          _$MfaConfigurationEnumMap, json['mfaConfiguration']),
+      mfaTypes: (json['mfaTypes'] as List<dynamic>?)
+          ?.map((e) => $enumDecode(_$MfaTypeEnumMap, e))
+          .toList(),
+      verificationMechanisms: (json['verificationMechanisms'] as List<dynamic>?)
+          ?.map((e) =>
+              const CognitoUserAttributeKeyConverter().fromJson(e as String))
+          .toList(),
+      loginMechanism: (json['loginMechanism'] as List<dynamic>?)
```
haverchuck, line 56: See above.

packages/amplify_flutter/lib/src/config/auth/cognito/mfa.dart
```diff
@@ -0,0 +1,35 @@
...
+  @JsonValue('TOTP')
+  totp,
+}
+
+enum MfaConfiguration {
+  @JsonValue('OPTIONAL')
+  optional,
+
+  @JsonValue('ON')
+  on,
```
haverchuck, line 31: Any issues with using a reserved word here?

packages/amplify_flutter/pubspec.yaml
```diff
@@ -16,19 +16,48 @@ dependencies:
   amplify_storage_plugin_interface: 0.3.0-rc.2
   flutter:
     sdk: flutter
+  json_annotation: ^4.3.0
   meta: ^1.7.0
   plugin_platform_interface: ^2.0.0
 
+# dependency_overrides:
```
haverchuck, line 23: If we intend to check out commented-out code, can we document why? Or is this an artifact?

packages/amplify_flutter/test/config/testdata/src/yarn.lock
```diff
@@ -0,0 +1,364 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
+amplify-frontend-flutter@^1.3.0:
```
haverchuck, line 5: Why are we checking in a yarn.lock?

haverchuck (commented):
Looks good - left a few comments/questions.

packages/amplify_flutter/lib/src/config/amplify_config.dart
dnys1: I mean, it might be. I used Google's tool to add them automatically: https://github.com/google/addlicense. Google's code use `//` from what I've seen.

packages/amplify_flutter/lib/src/config/auth/cognito/auth.dart
dnys1: Yeah, just different cases I've seen in old CLI generated files.

packages/amplify_flutter/lib/src/config/auth/cognito/mfa.dart
dnys1: Nope, not sure why not - maybe because `on` is not a valid keyword within a class/enum scope, only in the mixin declartion scope.

packages/amplify_flutter/pubspec.yaml
dnys1: Added a comment.

Running builders with melos is impossible right now (you may have run into this). The only way to run `flutter pub run build_runner build` successfully is to use dependency overrides so that you can run `flutter pub get`.

packages/amplify_flutter/test/config/testdata/src/yarn.lock
dnys1: Thought this was common practice in JS land?

packages/amplify_flutter/lib/src/config/amplify_config.dart
haverchuck: Ok, let's go with it then.

packages/amplify_flutter/lib/src/config/auth/cognito/mfa.dart
haverchuck: Alright, as long as we've tested.

packages/amplify_flutter/test/config/testdata/src/yarn.lock
haverchuck: It is common practice; however, it can cause some headaches when all contributers aren't the same version of yarn or are using npm instead. Given the very limited scope of our usage I think it's ok. We just need to be very cognizant of changes to this file.

haverchuck (approved):

packages/amplify_flutter/lib/src/config/auth/cognito/mfa.dart
dnys1: Yeah, some keywords do not cause issues, but I think it depends on the scope still. Like you can use `factory` as a variable name no problem: https://github.com/dnys1/amplify-flutter/blob/authenticator/common-config/packages/amplify_flutter/lib/src/config/amplify_plugin_registry.dart#L91

packages/amplify_flutter/lib/src/config/auth/cognito/auth.dart
Jordan-Nelson: Should we attempt to support older configs? I though the authenticator requires a config that was generated with a recent CLI version anyway.

If we do leave it, it might be nice to mark it as deprecated or at least comment somewhere what the purpose is (if you have already done that and I just missed it, my apologies)

dnys1: Yep, it's deprecated via the constructor. I think it's worth the potential DX improvement, but I'm not attached either way.

packages/amplify_flutter/lib/src/utils/equatable.dart
```diff
@@ -0,0 +1,32 @@
...
+/// hashing.
+mixin AWSEquatable<T extends AWSEquatable<T>> on Object {
+  /// The object's properties which should be equated and hashed.
+  List<Object?> get props;
+
+  @override
+  bool operator ==(Object other) =>
+      identical(this, other) ||
+      other is T &&
+          const DeepCollectionEquality.unordered().equals(props, other.props);
```
Jordan-Nelson, line 28: Shouldn't this be ordered? With it unordered, the following evaluates to true

```dart
AmplifyConfig(userAgent: 'foo', version: 'bar') == AmplifyConfig(userAgent: 'bar', version: 'foo') 
```

A unit test on this would be nice.

packages/amplify_flutter/lib/src/utils/equatable.dart
```diff
@@ -0,0 +1,32 @@
...
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+//
+
+import 'package:collection/collection.dart';
+
+/// Provides overrides for [==] and [hashCode] for consistent equality and
+/// hashing.
+mixin AWSEquatable<T extends AWSEquatable<T>> on Object {
```
Jordan-Nelson, line 20: Should this be in core? Right now it is only used in amplify_flutter, but it seems useful elsewhere. I guess we can always move it later so I don't think it is a big deal either way.

packages/amplify_flutter/test/config/testdata/auth/old/auth_with_email.json
```diff
@@ -0,0 +1,61 @@
+{
```
Jordan-Nelson, line 1: What is the significance of the folder name "old" here? Is this an output from an older CLI version? Do we want to name to folder something like `cli-6.1.0` to clarify what version we are testing against? The same goes for "zero_config"

packages/amplify_flutter/test/config/testdata/src/generate.js
```diff
@@ -0,0 +1,179 @@
+/*
```
Jordan-Nelson, line 1: A readme file in this dir would be nice

packages/amplify_flutter/test/config/testdata/src/generate.js
```diff
@@ -0,0 +1,179 @@
...
+    //
+    //  http://aws.amazon.com/apache2.0
+    //
+    // or in the "license" file accompanying this file. This file is distributed
+    // on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+    // express or implied. See the License for the specific language governing
+    // permissions and limitations under the License.
+    //
+    
+    // GENERATED FILE. DO NOT MODIFY BY HAND.
```
Jordan-Nelson, line 146:
```suggestion
    // GENERATED FILE. DO NOT MODIFY BY HAND. SEE `./src/generate.js`.
```

Would be nice to point someone to the source of the generated code.

packages/amplify_flutter/lib/src/utils/equatable.dart
dnys1: No, unordered only applies to lists/iterables of things. It just means `['a', 'b', 'c'] == ['b', 'c', 'a']`.

packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -39,14 +38,13 @@ import 'package:amplify_authenticator/src/theme/amplify_theme.dart';
...
-import 'package:amplify_flutter/src/config/amplify_config.dart';
+import 'package:amplify_flutter/amplify.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 
 export 'package:amplify_auth_cognito/amplify_auth_cognito.dart'
     show AuthProvider;
-export 'package:amplify_flutter/src/config/auth/password_policy_characters.dart';
-export 'package:amplify_flutter/src/config/auth/password_protection_settings.dart';
+export 'package:amplify_flutter/amplify.dart';
```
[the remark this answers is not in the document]
dnys1: Yeah, I think so. Why should they need to import two packages?

packages/amplify_flutter/lib/src/utils/equatable.dart
dnys1: It will be in `aws_common` eventually. That is where I copied it from. Since I believe that will be published before the end of the year, I just put it here for now.

packages/amplify_flutter/test/config/testdata/auth/old/auth_with_email.json
dnys1: I suppose - zero config is a CLI milestone, old means before that

packages/amplify_flutter/lib/src/utils/equatable.dart
dnys1: Oh I see what you're saying - yeah, I can add some tests.

packages/amplify_authenticator/lib/amplify_authenticator.dart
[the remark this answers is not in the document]
Jordan-Nelson: Okay. I think it is fine. Mainly wanted to make sure it was intentional.

You could make that same argument with amplify_auth_cognito as well. Right now only AuthProvider is exported here from amplify_auth_cognito. But I guess that is a little different since you don't have to use anything from amplify_auth_cognito to use the authenticator.

packages/amplify_flutter/test/config/testdata/auth/old/auth_with_email.json
Jordan-Nelson: What CLI version is zero config? What is the oldest version we aim to support? If these tests need to be updated, how would someone know what version of the CLI to use to regenerate the configs?

"zero config" is sort of a buzz word we are using now, but in a year from now will someone know what "old" and "zero config" means?

packages/amplify_flutter/lib/src/utils/equatable.dart
dnys1: Good catch - should not be unordered. Fixed!

packages/amplify_flutter/test/config/testdata/src/generate.js
dnys1: Fixed

packages/amplify_flutter/test/config/testdata/src/generate.js
dnys1: Fixed

Jordan-Nelson (approved):

PR #1148, opened 2021-11-29
fix: reset password

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*


packages/amplify_authenticator/lib/src/state/auth_viewmodel.dart
```diff
@@ -287,7 +287,7 @@ class AuthViewModel extends ChangeNotifier {
     AuthConfirmPasswordData confirmPassword = AuthConfirmPasswordData(
       username: _username.trim(),
       confirmationCode: _confirmationCode.trim(),
-      newPassword: _newPassword.trim(),
+      newPassword: _password.trim(),
```
Jordan-Nelson, line 290: I could have alternatively updated `onChanged` to update _newPassword instead of updating _password, but using _password here creates a nice UX because the password is pre-populated when the user gets redirected to sign in after the reset password flow.

haverchuck: I'm a little on the fence about pre-populating a password field. Have you seen other apps that do this?

dnys1 (approved):

haverchuck (approved):
Left one comment that should be considered; but otherwise LGTM.

PR #1153, opened 2021-11-30
fix(api): 0.3.0 fixes

dnys1:
*Issue #, if available:*
- API auth providers triggering false positives internally (no OIDC/Lambda provider but native SDK thinks there is)
- Need default auth rule provider on Android to prevent null exception crash

*Description of changes:*
- Pass registered auth providers to `addPlugin`
- Fallback to default auth provider for rule strategy

### DIFF
--- a/packages/amplify_api/ios/Classes/Auth/FlutterAuthProvider.swift
+++ b/packages/amplify_api/ios/Classes/Auth/FlutterAuthProvider.swift
@@ -77,12 +77,24 @@ class FlutterAuthProviders: APIAuthProviderFactory {
 
         return token ?? unknownError
     }
+    
+    private let authProviders: Set<AWSAuthorizationType>
+    
+    init(_ authProviders: [AWSAuthorizationType]) {
+        self.authProviders = Set(authProviders)
+    }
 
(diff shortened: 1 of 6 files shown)


packages/amplify_api/ios/Classes/Auth/FlutterAuthProvider.swift
```diff
@@ -78,11 +78,23 @@ class FlutterAuthProviders: APIAuthProviderFactory {
...
+
+    init(_ authProviders: [AWSAuthorizationType]) {
+        self.authProviders = Set(authProviders)
+    }
+
     override func oidcAuthProvider() -> AmplifyOIDCAuthProvider? {
+        guard authProviders.contains(.openIDConnect) else {
+            return nil
+        }
         return FlutterAuthProvider(type: .openIDConnect)
```
dnys1, line 91: iOS gets tripped up if these return anything but nil, they believe we've registered an auth provider and this triggers different flows within DataStore.

haverchuck (approved):

HuiSF (approved):

PR #1168, opened 2021-12-02
chore(authenticator): Force new password test

dnys1:
*Issue #, if available:*

*Description of changes:*
- Minor l10n spacing fix
- Add confirm new password field
- Add sign-in-force-new-password test

### DIFF
--- a/packages/amplify_authenticator/example/integration_test/pages/authenticator_page.dart
+++ b/packages/amplify_authenticator/example/integration_test/pages/authenticator_page.dart
@@ -13,7 +13,10 @@
  * permissions and limitations under the License.
  */
 
+import 'package:amplify_authenticator/amplify_authenticator.dart';
 import 'package:amplify_authenticator/src/keys.dart';
+import 'package:amplify_authenticator/src/screens/authenticator_screen.dart';
+import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart';
 
 class AuthenticatorPage {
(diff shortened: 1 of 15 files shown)


packages/amplify_authenticator/example/integration_test/sign_in_force_new_password_test.dart
```diff
@@ -0,0 +1,112 @@
...
+      'FORCE_CHANGE_PASSWORD state',
+      (WidgetTester tester) async {
+        final po = SignInPage(tester: tester);
+        await loadAuthenticator(tester: tester, authenticator: authenticator);
+
+        // When I select my country code with status "FORCE_CHANGE_PASSWORD"
+        await po.selectCountryCode();
+
+        // And I type my "phone number" with status "FORCE_CHANGE_PASSWORD"
+        await po.enterUsername(username.substring(2));
```
Jordan-Nelson, line 58: *nit:* It took me a second to understand what `substring(2)` was doing. 
```dart
String phoneWithoutCountryCode = username.substring(2)
await po.enterUsername(phoneWithoutCountryCode);
``` 

might be a little better.

This is pretty picky, so maybe not worth the change.

I think we discussed at some point separating the country code from the rest of the phone number for the purpose of creating a more valuable validator, which would imply that we create some sort of PhoneNumber class which stores them separately. At that point, generatePhoneNumber could probably just return a new PhoneNumber and then it would already be stored separately. So, maybe it could just be changed then.

Jordan-Nelson (approved):

packages/amplify_authenticator/example/integration_test/sign_in_force_new_password_test.dart
```diff
@@ -0,0 +1,112 @@
...
+
+    tearDownAll(() async {
+      await deleteUser(username);
+    });
+
+    // Scenario: Sign in using a valid phone number and password and user is in
+    // a FORCE_CHANGE_PASSWORD state
+    testWidgets(
+      'Sign in using a valid phone number and password and user is in a '
+      'FORCE_CHANGE_PASSWORD state',
```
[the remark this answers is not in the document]
dnys1: Lol, yeah to maintain the 80-char line length

packages/amplify_authenticator/example/integration_test/sign_in_force_new_password_test.dart
dnys1: Not nit-picky at all. It should be more clear. Updated - thanks!

And yeah, that would be a good change to have 👍

packages/amplify_authenticator/example/integration_test/pages/authenticator_page.dart
```diff
@@ -26,24 +29,48 @@ class AuthenticatorPage {
...
+      find.descendant(
+        of: finder,
+        matching: find.textContaining(errorText),
+      ),
+      findsOneWidget,
+    );
+  }
+
+  /// Expects an error banner containing [errorText].
+  void expectError(String errorText) {
```
haverchuck, line 55: 👍

packages/amplify_authenticator/example/integration_test/sign_in_force_new_password_test.dart
```diff
@@ -0,0 +1,113 @@
...
+
+  group('Sign In with Force New Password flow', () {
+    final username = generatePhone();
+    final phoneNumber = username.substring(2); // without '+1'
+    final password = generatePassword();
+
+    // Background
+    setUpAll(() async {
+      // Given I'm running the example
+      // "ui/components/authenticator/sign-in-with-phone"
```
haverchuck, line 34: I think if we wanted to reference JS we might make this ui/components/authenticator/sign-in-force-new-password (even if it's using the same env)

haverchuck (approved):

packages/amplify_authenticator/example/integration_test/sign_in_force_new_password_test.dart
dnys1: Added a reference at the top - but this is to match the feature definition

Issue #1164, opened 2021-12-02
AuthUserAttribute always attempts to convert String to Int


look1n:
**Describe the bug**
Custom user attributes forced conversion to int

**To Reproduce**
1. Create custom user attribute with String type
2. Save custom user attribute with string value that contains valid number i.e "012345"
3. Fetch user attributes
4. See that created custom attribute value has int type and value is 12345

**Expected behavior**
Custom attribute value should have same type specified on custom attribute creation form.

**Platform**
[] Android
[] iOS

**Additional context**
Obvious reason of such behavior is AuthUserAttribute init function that forcefully cast any string value that holds a number into int
 ```
AuthUserAttribute.init({required this.userAttributeKey, required this.value}) {
    this.userAttributeKey = userAttributeKey;
    if (userAttributeKey != 'phone_number') {
      try {
        this.value = int.parse(value);
      } on FormatException {}
    }
}
```

Jordan-Nelson:
Hello @look1n - Thanks for opening the issue. 

Some context - I believe the underlying API that is being used sends all attributes back as strings, without the information about the type specified in Cognito. The code you referenced above is (incorrectly) attempting to convert the attribute to the correct type.

The int conversion code leads to unpredictable behavior, so I think this is clearly a bug and can be removed. Unfortunately this will result in all attributes being returned as Strings, but at least the behavior will be predictable so I think this is an improvement. This would be a breaking change since some applications may be relying on that automatic conversion. I can take this back and see when we would be able to implement this.

As you suggested in the expected behavior, the better fix would be to convert the attribute to the type that was specified in Cognito. I think this will require some more investigation on our part to see how we could obtain the type info from Cognito.

Jordan-Nelson:
@look1n - The change to remove `int.parse()` was included in the latest release candidate for the next major release, `0.3.0-rc.3`. It will be included in the first stable version of `0.3.0` when it is released.

Jordan-Nelson:
@look1n - v0.3.0 has been released, which includes this fix. Please note that v0.3.0 does include several breaking changes (this being one of them). Please refer to the change log for details - https://pub.dev/packages/amplify_flutter/changelog.

As mentioned above, all attributes will now be returned as a String, regardless of the type defined in Cognito (String or Bool). I opened a [separate issue](https://github.com/aws-amplify/amplify-flutter/issues/1288) to track the feature request to have the types match the definition in Cognito console. I am going to close this issue out.

PR #1169, opened 2021-12-02
fix(amplify_auth_cognito): remove int.parse from AuthUserAttribute

Jordan-Nelson:
*Issue #, if available:* #1164

*Description of changes:*
- removes the attempt to convert the attribute value to an int.

### DIFF
--- a/packages/amplify_auth_cognito/example/lib/Widgets/ViewUserAttributes.dart
+++ b/packages/amplify_auth_cognito/example/lib/Widgets/ViewUserAttributes.dart
@@ -41,7 +41,7 @@ class _ViewUserAttributesState extends State<ViewUserAttributes> {
         SnackBar(backgroundColor: Colors.red[900], content: Text(message)));
   }
 
-  Future _fetchAttributes({bool isRefresh = false}) {
+  Future<void> _fetchAttributes({bool isRefresh = false}) {
     return Amplify.Auth.fetchUserAttributes().then((attributes) {
       setState(() => _userAttributes = attributes
         ..sort((a, b) => a.userAttributeKey.compareTo(b.userAttributeKey)));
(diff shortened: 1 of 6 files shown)


packages/amplify_auth_cognito/example/lib/Widgets/ViewUserAttributes.dart
```diff
@@ -106,14 +106,14 @@ class _ViewUserAttributesState extends State<ViewUserAttributes> {
...
-                  var key = _userAttributes[index].userAttributeKey
-                      as CognitoUserAttributeKey;
-                  var value = _userAttributes[index].value;
-                  var stringValue = value.toString();
+                  AuthUserAttribute atrribute = _userAttributes[index];
+                  CognitoUserAttributeKey userAttributeKey =
+                      atrribute.userAttributeKey as CognitoUserAttributeKey;
+                  String value = atrribute.value;
                   return ListTile(
-                    title: Text(key.key),
```
Jordan-Nelson, line 114: I renamed `key` to `userAttributeKey` because `key.key` was confusing me 🙂 

This is well out of scope of this change, but `CognitoUserAttributeKey` is new in 0.3.0. The definition is below. I think `key` should be renamed to `name` or `keyName` because the entire thing is the key, and this field is the name of that key.

If we believe this API makes more sense, it would be good to update now, because it will be a breaking change if done after 0.3.0. I could include this.

```dart
class CognitoUserAttributeKey extends UserAttributeKey {
  const CognitoUserAttributeKey._(this.key, {this.readOnly = false});

  /// The JSON key for this attribute.
  final String key;

  /// Whether this attribute can only be read from Cognito.
  final bool readOnly;

  // remaining fields
}
```

dnys1: I like `name`. I think it's worth it.

Jordan-Nelson: Just fyi - To use `name` we do have to update the static fields to const constructors since we have a static field right now called `name`

This:

```dart
static const name = CognitoUserAttributeKey._('name');
```

Needs to be:

```dart
const CognitoUserAttributeKey.name()  : name = 'name', readOnly = false;
```

The update is easy to make. There could be some confusion between the field `name` and constructor `name` but I think it make sense.

Jordan-Nelson: Discussed with @dnys1 offline. `UserAttributeKey` / `CognitoUserAttributeKey` has a `.toString()` method, which is what we should push customers to use instead of `.key`. I updated the sample (including this location) to use `userAttributeKey.toString()`.

dnys1 (approved):

HuiSF (approved):

PR #1176, opened 2021-12-03
test: confirm sign in tests

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- add confirm sign in tests

### DIFF
--- a/packages/amplify_authenticator/example/integration_test/pages/authenticator_page.dart
+++ b/packages/amplify_authenticator/example/integration_test/pages/authenticator_page.dart
@@ -13,6 +13,8 @@
  * permissions and limitations under the License.
  */
 
+import 'dart:io';
+
 import 'package:amplify_authenticator/amplify_authenticator.dart';
 import 'package:amplify_authenticator/src/keys.dart';
 import 'package:amplify_authenticator/src/screens/authenticator_screen.dart';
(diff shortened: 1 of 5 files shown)


packages/amplify_authenticator/example/integration_test/pages/authenticator_page.dart
```diff
@@ -105,6 +107,17 @@ abstract class AuthenticatorPage {
     expect(expectCombinationNotFound, findsOneWidget);
   }
 
+  /// Then I see Invalid verification code
+  Future<void> expectInvalidVerificationCode() async {
+    if (Platform.isAndroid) {
```
Jordan-Nelson, line 112: It surprised me that the messages were different on iOS and Android. Seems like this is a bug. Not specific to the authenticator though.

packages/amplify_authenticator/example/integration_test/utils/data_utils.dart
```diff
@@ -98,3 +102,62 @@ Future<AdminCreateUserResponse?> adminCreateUser(
...
+          currentCode
+        }
+      }''';
+
+  Completer<void> subscriptionCompleter = Completer<void>();
+
+  final Stream<GraphQLResponse<String>> operation = Amplify.API.subscribe(
+    GraphQLRequest<String>(document: subscriptionDocument),
+    onEstablished: () {
+      subscriptionCompleter.complete();
```
haverchuck, line 130: I assume you tested and this worked?  When I tried this approach it hung.

haverchuck (approved):
One question, but otherwise LGTM

packages/amplify_authenticator/example/integration_test/pages/authenticator_page.dart
Jordan-Nelson: Here is where the message on Android is coming from - https://github.com/aws-amplify/amplify-android/blob/main/core/src/main/java/com/amplifyframework/auth/AuthException.java#L317

packages/amplify_authenticator/example/integration_test/utils/data_utils.dart
Jordan-Nelson: It hung some (most?) of the time but did seem to work some of the time. I have `await subscriptionCompleter.future` commented out and instead have `await Future.delayed(const Duration(seconds: 1), () {});` as a temp solution.

haverchuck: Alrighty, thanks

PR #1183, opened 2021-12-06
chore(authenticator): Docs clean up

dnys1:
*Issue #, if available:*

*Description of changes:*
- Clean up doc comments, exports, and example app

### DIFF
--- a/packages/amplify_authenticator/lib/amplify_authenticator.dart
+++ b/packages/amplify_authenticator/lib/amplify_authenticator.dart
@@ -144,7 +144,18 @@ export 'src/widgets/form_field.dart'
 /// using a custom form component. For example, to collect your user's address
 /// information on the sign up form, use the [SignUpForm.custom] constructor:
 ///
-/// {@macro amplify_authenticator.custom_sign_up_form}
+/// {@template amplify_authenticator.custom_sign_up_form}
+/// ```dart
+/// Authenticator(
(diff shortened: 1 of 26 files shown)


packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -42,59 +45,175 @@ import 'package:amplify_flutter/amplify.dart';
...
+/// However, you can also use the a hand-crafted Amplify theme by setting
+/// [useAmplifyTheme] to `true`.
 ///
-/// This authenticator accepts the following custom forms, sign in, sign up and confirm sign in.
+/// ### Forms
 ///
-/// Note that working with custom forms is optional. Thus, if no additional arguments
-/// are passed, the authenticator will defined the following default forms with their
-/// respective form fields:
+/// The Authenticator uses your app's Cognito configuration to determine which
```
haverchuck, line 143: I would say something like 'your app's Cognito configuration as defined in the amplifyconfiguration.dart file', or even just 'your app's amplifyconfiguration.dart file.' We want to make this clear for customers who are using existing resources and writing their own config.

Jordan-Nelson (approved):

packages/amplify_authenticator/lib/amplify_authenticator.dart
dnys1: Fixed - thanks! Good call.

PR #1186, opened 2021-12-06
chore(amplify_authenticator): documentation for input and country resolver classes

haverchuck:
*Description of changes:*
* documentation for input and country resolver classes
* includes codegen change

### DIFF
--- a/packages/amplify_authenticator/lib/src/l10n/input_resolver.dart
+++ b/packages/amplify_authenticator/lib/src/l10n/input_resolver.dart
@@ -53,7 +53,7 @@ enum InputResolverKeyType {
   empty,
   passwordRequirements,
   format,
-  mistmatch
+  mismatch
 }
 
 class InputResolverKey {
@@ -112,7 +112,7 @@ class InputResolverKey {
     field: InputField.passwordConfirmation,
   );
   static const passwordsDoNotMatch = InputResolverKey._(
-    InputResolverKeyType.mistmatch,
+    InputResolverKeyType.mismatch,
     field: InputField.passwordConfirmation,
   );
   const InputResolverKey.passwordRequirementsUnmet(
@@ -372,10 +372,10 @@ class InputResolverKey {
 class InputResolver extends Resolver<InputResolverKey> {
   const InputResolver();
 
-  /// Returns the [String] that is displayed as the input title.
+  /// Returns the label that is displayed as the input title.
   ///
   /// In addition to displaying the input title, this method is
-  /// used by the default implmentations of the [hint, empty and format] methods
+  /// used by the default implmentations of the [hint], [empty] and [format] methods
   /// to resolve the base name of the field.
   String title(BuildContext context, InputField field) {
     switch (field) {
@@ -433,7 +433,7 @@ class InputResolver extends Resolver<InputResolverKey> {
     }
   }
 
-  /// Returns the [String] displayed as the input hint.
+  /// Returns the label displayed as the input hint.
   String hint(BuildContext context, InputField field) {
     final fieldName = title(context, field);
     final lowercasedFieldName = fieldName.toLowerCase();
@@ -441,7 +441,8 @@ class InputResolver extends Resolver<InputResolverKey> {
         .promptFill(lowercasedFieldName);
   }
 
-  /// Returns the [String] used as the hint for the password confirmation field.
+  /// Returns the hint text used for confirmation fields where the
+  /// user is asked to re-enter information prior to form submission.
   String confirmHint(BuildContext context, InputField field) {
     final fieldName = AuthenticatorLocalizations.inputsOf(context).password;
     final lowercasedFieldName = fieldName.toLowerCase();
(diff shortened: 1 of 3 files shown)


packages/amplify_authenticator/lib/src/l10n/input_resolver.dart
```diff
@@ -364,9 +364,19 @@ class InputResolverKey {
...
 
+/// The resolver class for input widgets.
+///
+/// This class is responsible for resolving text to display
+/// in input field titles, as well as hints, warnings and other
+/// guidance for the end-user.
 class InputResolver extends Resolver<InputResolverKey> {
   const InputResolver();
 
+  /// Returns the [String] that is displayed as the input title.
```
dnys1, line 375: General comment: no need to highlight types like this. The function signature will already have a link for all types. We generally want to highlight items which are not in the function declaration but may be in the surrounding scope, or related items.

packages/amplify_authenticator/lib/src/l10n/input_resolver.dart
```diff
@@ -364,9 +364,19 @@ class InputResolverKey {
...
+  ///
+  /// In addition to displaying the input title, this method is
+  /// used by the default implmentations of the [hint, empty and format] methods
```
dnys1, line 378:
```suggestion
  /// used by the default implmentations of the [hint], [empty], and [format] methods
```

packages/amplify_authenticator/lib/src/l10n/input_resolver.dart
```diff
@@ -423,30 +433,37 @@ class InputResolver extends Resolver<InputResolverKey> {
...
 
+  /// Returns the [String] displayed as the input hint.
   String hint(BuildContext context, InputField field) {
     final fieldName = title(context, field);
     final lowercasedFieldName = fieldName.toLowerCase();
     return AuthenticatorLocalizations.inputsOf(context)
         .promptFill(lowercasedFieldName);
   }
 
+  /// Returns the [String] used as the hint for the password confirmation field.
```
dnys1, line 444: This is a general confirmation text, not specific to password

packages/amplify_authenticator/lib/src/l10n/input_resolver.dart
```diff
@@ -423,30 +433,37 @@ class InputResolver extends Resolver<InputResolverKey> {
...
 
+  /// Returns the [String] displayed when a field fails a format validation check,
+  /// such as an invalid email format, an invalid confirmation code length, etc.
   String format(BuildContext context, InputField field) {
     return AuthenticatorLocalizations.inputsOf(context)
         .warnInvalidFormat(title(context, field).toLowerCase());
   }
 
+  /// Returns the [String] displayed when a password input does match the password requirements
+  /// defined in the amplifyconfiguration.dart file.
```
dnys1, line 466:
```suggestion
  /// defined in the configuration.
```

Not necessarily in the file if they are building configs by hand.

packages/amplify_authenticator/tool/generate_country_localization.dart
```diff
@@ -90,6 +90,7 @@ enum CountryResolverKey {
 
   for (var element in _countries) {
     output.write('''
+  /// The [String] displayed as a label for ${element.displayName}.
```
dnys1, line 93:
```suggestion
  /// The label for ${element.displayName}.
```

dnys1 (approved):
Looks good - would just adjust those few things

packages/amplify_authenticator/lib/src/l10n/input_resolver.dart
haverchuck: It could be used generally, true, but right now the fieldName is being set to `AuthenticatorLocalizations.inputsOf(context).password`, and the only place it's used is password confirmation:
```dart
  static const passwordConfirmationHint = InputResolverKey._(
    InputResolverKeyType.confirmHint,
    field: InputField.passwordConfirmation,
  );
  ```

File: packages/amplify_authenticator/lib/src/l10n/input_resolver.dart (hunk 2 above)
Comment (haverchuck) [reply packages/amplify_authenticator/lib/src/l10n/input_resolver.dart:378]: Yeah - wasn't sure on the preferred style here. Thanks!

File: packages/amplify_authenticator/lib/src/l10n/input_resolver.dart (hunk 3 above)
Comment (dnys1) [reply packages/amplify_authenticator/lib/src/l10n/input_resolver.dart:444]: Right, but the function signature takes in an `InputField`. When you say password confirmation here, it's confusing how to implement the method for a generic field. This method just produces `Re-enter ${field}` so any overrides should do the same.

Comment (Jordan-Nelson) [reply packages/amplify_authenticator/lib/src/l10n/input_resolver.dart:444]: You can probably make the doc a little more generic without losing the info about it being used for password confirmation. Something like
```suggestion
  /// Returns the [String] used as the hint for a confirmation field, for example, a password confirmation field.
```

Jordan-Nelson (approved):

packages/amplify_authenticator/lib/src/l10n/input_resolver.dart
haverchuck: Ok, fair enough.

PR #1188, opened 2021-12-06
chore(amplify_api): REST integration tests and change API provision script to use init

ragingsquirrel3:
1) Add basic integration tests for REST API
2) Because these tests add requirements for the backend resource which CLI headless add cannot do, this PR also makes the API backend provision script use `amplify init` with a specific subset of the /amplify dir designed for this purpose.

2 is part of potentially larger migration to make all the integ tests use these backend files so it's easier to reproduce the stack, source control things like lambda code, and generally do the things headless CLI cannot do. amplify-ui uses [this strategy](https://github.com/aws-amplify/amplify-ui/tree/main/environments). See more documentation https://aws.amazon.com/blogs/mobile/amplify-cli-adds-scaffolding-support-for-amplify-apps-and-authoring-plugins/.


packages/amplify_api/example/integration_test/rest_tests.dart
```diff
@@ -0,0 +1,94 @@
...
+    path: path,
+  );
+  final restOptionsWithBody =
+      RestOptions(path: path, body: ascii.encode('{"name":"Mow the lawn"}'));
+
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  group('REST', () {
+    setUpAll(() async {
+      if (!Amplify.isConfigured) {
```
dnys1, line 36: Would remove this - probably want the tests to fail if configure doesn't execute.

packages/amplify_api/example/amplify/backend/function/flutterapiintegtest17270866/src/index.js
```diff
@@ -0,0 +1,15 @@
...
+exports.handler = async (event) => {
+    // TODO implement
+    const response = {
+        statusCode: 200,
+    //  Uncomment below to enable CORS requests
+    //  headers: {
+    //      "Access-Control-Allow-Origin": "*",
+    //      "Access-Control-Allow-Headers": "*"
+    //  }, 
+        body: JSON.stringify('Hello from Lambda!'),
```
dnys1, line 12: Can you return just the string? I was a bit confused why it was quoted in the test case - think a string makes more sense than a "JSON string"

packages/amplify_api/example/integration_test/rest_tests.dart
```diff
@@ -0,0 +1,94 @@
...
+import 'dart:convert';
+
+import 'package:amplify_api/amplify_api.dart';
+import 'package:amplify_api_example/amplifyconfiguration.dart';
+import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
+import 'package:amplify_flutter/amplify.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:integration_test/integration_test.dart';
+
+void main() {
```
dnys1, line 24: Can we add an error case? Could be 404 so there's no backend work to do.

dnys1 (approved):
Awesome! Thanks for doing this!

packages/amplify_api/example/integration_test/rest_tests.dart
ragingsquirrel3: Want to keep to support running the file individually (rather than from main test file).

packages/amplify_api/example/amplify/backend/function/flutterapiintegtest17270866/src/index.js
ragingsquirrel3: updated. Good suggestion. This is how the CLI does the simple "hello world" lambda by default but agree it's confusing.

packages/amplify_api/example/integration_test/rest_tests.dart
ragingsquirrel3: Good suggestion, added (although backend will actually give you 403 if you try some random route as I guess that's how API gateway is configured in this case).

packages/amplify_api/example/integration_test/rest_tests.dart
dnys1: Just meant to remove the if condition - for example if it's already configured then the first test will fail but maybe not for the reasons we expect. We'd rather have a failure in the setup method if there are issues configuring.

packages/amplify_api/example/integration_test/rest_tests.dart
dnys1: Perfect, works just as well. Thanks!

packages/amplify_api/example/integration_test/rest_tests.dart
ragingsquirrel3: discussed offline, keeping as-is here

haverchuck (approved):
LGTM - just to confirm, we aren't checking in any api keys, right?

ragingsquirrel3:
> LGTM - just to confirm, we aren't checking in any api keys, right?

No.

ragingsquirrel3:
this CLI issue needs to be fixed before merging this or provision script won't work https://github.com/aws-amplify/amplify-cli/issues/9778

ragingsquirrel3:
This issue https://github.com/aws-amplify/amplify-cli/issues/9508 is blocking merge as `amplify init` won't work until that's fixed.

PR #1192, opened 2021-12-07
chore(amplify_datastore): comment for FlutterSerializedModel.swift

haverchuck:
*Description of changes:*
Adds a note specifying that we are using duplicated code for FlutterSerializedModel.swift in amplify-ios integ tests.

### DIFF
--- a/packages/amplify_datastore/ios/Classes/types/model/FlutterSerializedModel.swift
+++ b/packages/amplify_datastore/ios/Classes/types/model/FlutterSerializedModel.swift
@@ -17,6 +17,12 @@ import Flutter
 import Foundation
 import Amplify
 
+/*
+    Helper class to assist with handling serialized data
+    Note: This class is used in amplify-ios serialized model integration tests.
+    Until this code can be properly shared between repositories, this class should be updated
+    in the ios integration tests as needed.
+*/
 struct FlutterSerializedModel: Model, JSONValueHolder {
     public let id: String
 


packages/amplify_datastore/ios/Classes/types/model/FlutterSerializedModel.swift
```diff
@@ -17,6 +17,12 @@ import Flutter
 import Foundation
 import Amplify
 
+/*
+    Helper class to assist with handling serialized data
+    Note: This class is used in amplify-ios serialized model integration tests.
+    Until this code can be properly shared between repositories, this class should be updated
+    in the ios integration tests as needed.
+*/
```
HuiSF, line 25: Thanks for adding the descriptive comment! I'm a bit confused by reading the PR description:
> duplicated code for FlutterSerializedModel.swift
> this class should be updated

What or which part is duplicate here? And what should be updated, would that be helpful to create a backlog issue with details and attached to this comment?

haverchuck: @HuiSF In amplify-ios integration tests focusing on serialized model, we use the FlutterSerializedModel class as a helper. Since there is no way to actually shared code between the two repositories right now, I just copied the whole thing over. Amplify-ios requested we put this comment in our codebase.

HuiSF: Ahh, understood. Thanks for clarifying.
How about we alter the comment a bit, what do you think?

```
/*
    Helper class to assist with handling serialized data
    Note: This class has a copy located in amplify-ios repo and is used in amplify-ios integration tests.
    Until this code can be properly shared between repositories, any changes made to this class
    should be copied over to the duplicated in amplify-ios repo (link).
*/
```

PR #1195, opened 2021-12-08
feat(amplify_authenticator): update username selector to use toggle buttons

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Update the UI for the username selection to use toggle buttons

#### Amplify Theme

Screen Shot 2021-12-07 at 5 14 24 PM

#### Material

Screen Shot 2021-12-07 at 5 17 22 PM

#### Custom Material

Screen Shot 2021-12-07 at 5 17 39 PM


packages/amplify_authenticator/lib/src/widgets/form_field.dart
```diff
@@ -216,11 +216,14 @@ abstract class AuthenticatorFormFieldState<FieldType, FieldValue,
...
 
   /// Margin below the form field
-  EdgeInsetsGeometry? get marginBottom => FormFieldConstants.marginBottom;
+  double? get marginBottom => FormFieldConstants.marginBottom;
 
   /// Gap between the label and the form field
   double? get labelGap => FormFieldConstants.gap;
 
+  /// Widget to show above the label.
+  Widget? get surlabel => null;
```
Jordan-Nelson, line 225: Open to other names here. The username selection widget was previously set with `suffix`, but since it has a new placement, I added this.

This widget is above the label. "sur" as a prefix I think typically means "above" in more of a figurative sense (surcharge), but can also be in the literal sense (a surtitle is a like a subtitle but is placed at the top)

packages/amplify_authenticator/test/ui/tab_view_test.dart
```diff
@@ -67,191 +80,202 @@ void main() {
...
-      // Tests the default Material theme (dark mode).
-      testWidgets('(default)', (WidgetTester tester) async {
-        await tester.pumpWidget(defaultMaterial);
-        await tester.pumpAndSettle();
-
-        await expectGoldenMatches(
-          tabViewFinder,
-          'tab_view_dark_default.png',
-        );
+    for (var config in configs.entries) {
```
Jordan-Nelson, line 152: This diff here appears large, but it is mostly whitespace. See https://github.com/aws-amplify/amplify-flutter/pull/1195/files?diff=split&w=1 to ignore whitespace changes.

packages/amplify_authenticator/lib/src/mixins/authenticator_username_field.dart
```diff
@@ -109,51 +109,79 @@ mixin AuthenticatorUsernameField<FieldType,
...
+                          '';
+                  viewModel.setUsername(newUsername);
+                },
+              );
+            }),
+            SizedBox(height: marginBottom),
+          ],
         );
+      default:
+        null;
```
dnys1, line 184: `return null` or can remove

packages/amplify_authenticator/lib/src/widgets/form_field.dart
dnys1: Sounds good - sur means above/on in French, e.g. *sur la table*

dnys1 (approved):
Looks good, thanks 👍

haverchuck (approved):

PR #1197, opened 2021-12-08
fix: consistently use brightness from context

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- consistently use brightness from context

### DIFF
--- a/packages/amplify_authenticator/lib/amplify_authenticator.dart
+++ b/packages/amplify_authenticator/lib/amplify_authenticator.dart
@@ -491,7 +491,7 @@ class _AuthenticatorBody extends StatelessWidget {
     final stateMachineBloc = InheritedAuthBloc.of(context);
     final useAmplifyTheme = InheritedConfig.of(context).useAmplifyTheme;
     final userAppTheme = Theme.of(context);
-    final isDark = MediaQuery.platformBrightnessOf(context) == Brightness.dark;
+    final bool isDark = AmplifyTheme.of(context).isDark;
     return Theme(
       data: useAmplifyTheme
           ? (isDark ? AmplifyTheme.dark : AmplifyTheme.light)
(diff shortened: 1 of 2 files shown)


packages/amplify_authenticator/lib/src/widgets/authenticator_banner.dart
```diff
@@ -27,7 +27,8 @@ MaterialBanner createMaterialBanner(
   required List<Widget> actions,
   required bool useAmplifyTheme,
 }) {
-  final bool isDark = AmplifyTheme.of(context).isDark;
```
Jordan-Nelson, line 30: This was throwing because AmplifyTheme.of() calls InheritedConfig.of(), and InheritedConfig is a defendant of the widget where these subscriptions are set up. `Theme.of(context).brightness == Brightness.dark` is the same thing as `AmplifyTheme.of(context).isDark`.

packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -491,7 +491,7 @@ class _AuthenticatorBody extends StatelessWidget {
     final stateMachineBloc = InheritedAuthBloc.of(context);
     final useAmplifyTheme = InheritedConfig.of(context).useAmplifyTheme;
     final userAppTheme = Theme.of(context);
-    final isDark = MediaQuery.platformBrightnessOf(context) == Brightness.dark;
+    final bool isDark = AmplifyTheme.of(context).isDark;
```
dnys1, line 494: This needs to stay the same - AmplifyTheme.of(context).isDark will look at the closest ThemeData ancestor and check its brightness. Here - that would be the user's theme's brightness. However, we are more interested what the platform has to say because we use this to show the dark theme or not.

dnys1 (approved):
Approved (with minor change). Thanks for fixing this!

haverchuck (approved):

packages/amplify_authenticator/lib/amplify_authenticator.dart
Jordan-Nelson: Discussed offline. Will keep as is.

Issue #1198, opened 2021-12-08
 Cascade delete on belongsTo and hasMany is not working in iOS


HuiSF:
**Describe the bug**
The dependencies of amplify-flutter: amplify-ios and amplify-android supports cascade delete for certain model relationships.

|                                          | belongsTo                                          | hasOne                    | hasMany                                                  |
| ---------------------------------------- | -------------------------------------------------- | ------------------------- | -------------------------------------------------------- |
| Example                                  | model A belongs to model B                         | model A has a model B     | model A has many model B                                 |
| supported cascade delete use case in iOS | Delete model B, associated model A will be deleted | No cascade delete support | Delete model A, all associated model B’s will be deleted |
| supported cascade delete use case in Android | Delete model B, associated model A will be deleted | No cascade delete support | Delete model A, all associated model B’s will be deleted |

However, cascade delete is not happening in iOS for both `belongsTo` and `hasMany` use cases, and cascade is working as expected in Android.

**NOTE**: cascade delete is happening partially in iOS that the record in SQLite db would be deleted, but the records in DynamoDB were not updated.

(Cascade delete for hasOne is a requested feature, see https://github.com/aws-amplify/amplify-flutter/issues/818)

**To Reproduce**
1. Create iOS App with DataStore, and create models with belongsTo and hasMany
2. Delete models

**Expected behavior**
Cascade delete should work in iOS for both `belongsTo` and `hasMany` use cases.

**Screenshots**

**Platform**
Amplify Flutter current supports iOS and Android.  This issue is reproducible in (check all that apply):
[] Android
[x] iOS

<details>

  <summary>Output of <code>flutter doctor -v</code></summary>
  
  ```Paste your output of "flutter doctor -v" here```

</details>

<details>

  <summary>Dependencies (<code>pubspec.lock</code>)</summary>
  
  ```
  Paste the contents of your "pubspec.lock" file here
  ```

</details>

**Smartphone (please complete the following information):**

**Additional context**

HuiSF:
Fix PR merged.

HuiSF:
This unexpected behavior now is fixed with [version 0.3.0](https://github.com/aws-amplify/amplify-flutter/releases/tag/v0.3.0).

PR #1199, opened 2021-12-08
fix(datastore): cascade delete not working in iOS

HuiSF:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-flutter/issues/1198

*Description of changes:*

Make the registered model decoder capable to handle decoding json string that presents a single object besides an array of objects.

- Updated relevant integration tests to test the use cases.

----

### DIFF
--- a/packages/amplify_datastore/example/integration_test/relationship_type_test.dart
+++ b/packages/amplify_datastore/example/integration_test/relationship_type_test.dart
@@ -110,8 +110,9 @@ void main() {
     });
 
     group('BelongsTo', () {
-      var child = ChildModel(name: 'child');
-      var parent = BelongsToModel(name: 'HasOne', child: child);
+      var relationshipParent = ChildModel(name: 'child');
+      var relationshipChild =
+          BelongsToModel(name: 'BelongsTo', child: relationshipParent);
       late Future<SubscriptionEvent<ChildModel>> belongsToEvent;
       late Future<SubscriptionEvent<BelongsToModel>> hasOneEvent;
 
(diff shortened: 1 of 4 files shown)


packages/amplify_datastore/example/integration_test/relationship_type_test.dart
```diff
@@ -110,8 +110,9 @@ void main() {
     });
 
     group('BelongsTo', () {
-      var child = ChildModel(name: 'child');
-      var parent = BelongsToModel(name: 'HasOne', child: child);
+      var relationshipParent = ChildModel(name: 'child');
```
Jordan-Nelson, line 113: If we want to make these tests easier to understand, I think we need to update the schema. `ChildModel` was probably a bad name choice (by me 😅 ) to begin with. `ChildModel` is currently related to `BelongsToModel` and `HasOneModel` in the schema. It might make sense to replace `ChildModel` with two separate models, like below. The schema below is using the v2 directives, but that could be modified.

```graphql
type HasOneModel @model @auth(rules: [{allow: public}]) {
  id: ID!
  name: String!
  childID: ID!
  child: HasOneChildModel @hasOne(fields: ["childID"])
}

type HasOneChildModel @model @auth(rules: [{allow: public}]) {
  id: ID!
  name: String!
}

type BelongsToModel @model @auth(rules: [{allow: public}]) {
  id: ID!
  name: String!
  parent: BelongsToParentModel! @hasOne
}

type BelongsToParentModel @model @auth(rules: [{allow: public}]) {
  id: ID!
  name: String!
}
```

Another option would be to remove all use of "parent" and "child" from the schema, but this would be a much larger change.

HuiSF: As GraphQL transformer v1 is still being supported as legacy feature (not recommended for new projects), I think we can leave the original integ tests suite as is. When adding the tests suite for v2, we can use better naming. Thoughts?

Jordan-Nelson: How long is v1 being supported? I think it is fine to consider it out of scope for this PR, but if we will be supporting v1 for a while it might be nice to update the v1 schema at some point too.

HuiSF: I'm not aware of a clear deprecation plan. We can improve while implementing the test suites for v2.

dnys1: Here's is what I used for testing hasOne/belongsTo (agree that current naming is suuuper confusing)

**v1**

```graphql
type ParentModel @model {
  id: ID!
  name: String!
  childID: ID
  child: ChildModel @connection(fields: ["childID"]) @hasOne
}

type ChildModel @model {
  id: ID!
  name: String!
  parent: ParentModel! @connection # @belongsTo
}
```

**v2**

```graphql
type ParentModel @model {
  id: ID!
  name: String!
  childID: ID
  child: ChildModel @hasOne(fields: ["childID"])
}

type ChildModel @model {
  id: ID!
  name: String!
  parent: ParentModel! @belongsTo
}
```

dnys1 (approved):
Looks good - would just adjust naming

Jordan-Nelson (approved):

PR #1201, opened 2021-12-09
fix(datastore): Hub memory usage

dnys1:
*Issue #, if available:*
- Hub history tracking is consuming too much memory

*Description of changes:*
- Only store necessary hub events
- Disable for release mode

### DIFF
--- a/packages/amplify_datastore/ios/Classes/DataStoreHubEventStreamHandler.swift
+++ b/packages/amplify_datastore/ios/Classes/DataStoreHubEventStreamHandler.swift
@@ -53,6 +53,7 @@ public class DataStoreHubEventStreamHandler: NSObject, FlutterStreamHandler {
         // Replay events. On hot restart, `onListen` is called again with a new listener. However,
         // DataStore will not re-emit events such as ready and modelSynced. As a result, this info
         // is lost on the Flutter side unless we replay the history prior to the hot restart.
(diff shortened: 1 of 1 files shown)


packages/amplify_datastore/ios/Classes/DataStoreHubEventStreamHandler.swift
```diff
@@ -53,19 +53,23 @@ public class DataStoreHubEventStreamHandler: NSObject, FlutterStreamHandler {
...
             for payload in eventHistory {
-                if replayableEvents.contains(payload.eventName) {
-                    sendPayload(payload)
-                }
+                sendPayload(payload)
             }
         }
+        #endif
         token = Amplify.Hub.listen(to: .dataStore) { [unowned self] (payload) in
+            #if DEBUG
```
HuiSF, line 66: Should we apply similar strategy in Android implement as well?

haverchuck (approved):

packages/amplify_datastore/ios/Classes/DataStoreHubEventStreamHandler.swift
dnys1: Yeah, we should probably do some profiling there as well

Jordan-Nelson (approved):

Issue #1208, opened 2021-12-13
Support remote logging and dynamic config


xingzhan1:
Hi team,

This is a feature request for supporting remote logging and dynamic config:

Remote logging: send logs and metrics from mobile app to cloud for debugging and monitoring, possibly cloudwatch as the solution.

Dynamic config: periodically pulling config from cloud, can be used for runtime config, A/B testing, etc. Possibly AppConfig as the solution.

Thanks,
Xing

HuiSF:
Hi @xingzhan1 thanks for the request.

If you could more information about the use cases that would be very helpful for use to understand the scope.

> supporting remote logging

Do you mean you need an API to allow sending any type of logs to CloudWatch any time and any where in your App?

The Amplify library functionalities are implemented on top of AWS services, such as AppSync, Cognito, Pinpoint etc, each of them should have service level logging mechanism. Take DataStore, you can enabled [logging and monitoring](https://docs.aws.amazon.com/appsync/latest/devguide/monitoring.html) within AppSync console (the App associated with your Amplify project), the logs will be generated in CloudWatch and you can query or generate dashboards. You can also learn more about logging and monitoring from [Amplify document](https://docs.aws.amazon.com/amplify/latest/userguide/access-logs.html).

For metrics (if you meant use cases such as collecting data of how your App users use the App), the [Analytics category](https://docs.amplify.aws/lib/analytics/getting-started/q/platform/flutter/#set-up-analytics-backend) should be able to help.

>dynamic config

Currently Amplify libraries common set up API `configure` allows to be invoked only once, reconfigure at runtime is currently not supported. I think you could set up an infrastructure to serve Amplify configurations, but it would requires your App user to quit and restart the App to apply a new configuration. 

>A/B testing

I would like to learn more about this as well, do you have any specific example that would required to change Amplify configurations?

xingzhan1:
>Do you mean you need an API to allow sending any type of logs to CloudWatch any time and any where in your App?

Yes, I am more referring to the forntend log and metric in our mobile app, such as clickstream, response time, etc. 

> The Amplify library functionalities are implemented on top of AWS services, such as AppSync, Cognito, Pinpoint etc, each of them should have service level logging mechanism. 

We are using the amplify logging already, but we would like to collect frontend log and metric as well

> For metrics (if you meant use cases such as collecting data of how your App users use the App), the Analytics category should be able to help.

[...]

Issue #1115, opened 2021-11-18
Require a better convention for using existing AWS resources like API Gateway resources in amplifyconfiguration.dart


mithunkavullakandiyil:
**Describe the bug**
As per Amplify Flutter docs: [Use existing AWS resources](https://docs.amplify.aws/lib/restapi/existing-resources/q/platform/flutter/), to use existing AWS resources like AWS API Gateway resources, we need to modify  amplifyconfiguration.json file. But in Amplify Flutter project, no such file exists, instead a amplifyconfiguration.dart gets created. 

As per the [amplifyconfiguration.json](https://docs.amplify.aws/cli/reference/files/#amplifyconfigurationjson) docs, this file is not intended for manual edit and version control. Also, both these filenames are in [.gitignore](https://docs.amplify.aws/cli/reference/files/#gitignore). 

**Expected behavior**
Need a solution to modify and version control existing AWS resources meta info without the sensitive information such as Auth provider, API key, etc.

Propose to have this configuration in [backend-config.json](https://docs.amplify.aws/cli/reference/files/#backend-configjson) so that, it can be version controlled. Also, a CLI command to add an existing AWS resources (similar to CDK and CloudFormation) would be good. It will be good if the same can be modified via Admin UI as well.

dnys1:
Thank you for your ticket @mithunkavullakandiyil. We will investigate ways to improve the experience of both working with these configuration files and being able to version control the non-sensitive bits.

Jordan-Nelson:
Related to https://github.com/aws-amplify/amplify-flutter/issues/132

Issue #1090, opened 2021-11-13
API with AuthMode API_KEY giving error "No client information for API"


rahul-insight:
**Describe the bug**
We are primarily using AMAZON_COGNITO_USER_POOLS as auth mode for Amplify APIs. For an API to be made public, we have added API_KEY auth mode. The new API with API_KEY works correctly on AWS AppSync console and returns the right response.

After following the steps over [here](https://docs.amplify.aws/lib/graphqlapi/authz/q/platform/flutter/#use-amplify-cli-to-configure-authorization-modes) the new API_KEY configuration is appearing in the "auth" section of amplifyconfiguration.dart instead of the "api" section as mentioned in the document [here](https://docs.amplify.aws/lib/graphqlapi/authz/q/platform/flutter/#configure-multiple-authorization-modes)

"auth": {
        "plugins": {
            "awsCognitoAuthPlugin": {
               "AppSync": {
                    "[MY_FRIENDLY-NAME-API-WITH-API-KEY]": {
                    "endpointType": "GraphQL",
                    "endpoint": "[GRAPHQL-ENDPOINT]",
                    "region": "[REGION]",
                    "authorizationType": "API_KEY",
                    "apiKey": "[API_KEY]"
                },
},
},
}

Because of this calling the public API with API_KEY auth mode fails with the below error:

`ApiException{message=No client information for API named [MY_FRIENDLY-NAME-API-WITH-API-KEY] cause=null, recoverySuggestion=Check your amplify configuration to make sure there is a correctly configured section for [MY_FRIENDLY-NAME-API-WITH-API-KEY]}`

**To Reproduce**
1. Configure additional auth mode API_KEY as described [here](https://docs.amplify.aws/lib/graphqlapi/authz/q/platform/flutter/#use-amplify-cli-to-configure-authorization-modes)
2. Check that the newly added API_KEY configuration with [MY_FRIENDLY-NAME-API-WITH-API-KEY] is available in amplifyconfiguration.dart
3. Call the API as described in the documentation here 

```
var operation = Amplify.API.string<String>(
    request: GraphQLRequest(
        document: graphQLDocumentString,
        apiName: '[MY_FRIENDLY-NAME-API-WITH-API-KEY]',
    ),
);
var response = await operation.response;
var data = response.data;
```

4. See error:
[...]

Issue #132, opened 2020-10-03
Simplify amplifyconfiguration.dart


rmelian:
Hello guys, thank you for putting together amplify for flutter.

After playing with the project I noticed if you want to use another tool like CDK to generate aws resources is a little cumbersome to edit the amplifyconfiguration.dart file. I have used before the amplify Auth lib and I don't remember it been that complicated. 

In simple words the file amplifyconfiguration.dart is not easy to edit by hand, so having a simplified structure would be great for users having the aws resources created already. 

Keep the awesome work 😃

jamesonwilliams:
Thanks @rmelian for the feedback. We'll take this into account as we plan future revisions to the configuration file structure.

dnys1:
We have released the [`AmplifyConfig`](https://pub.dev/documentation/amplify_flutter/latest/amplify_flutter/AmplifyConfig-class.html) class which can be used to parse and interact with the `amplifyconfiguration` in a fully-typed way, e.g.

```dart
import 'dart:convert';

import 'package:amplify_flutter/amplify_flutter.dart';

import 'amplifyconfiguration.dart';

final parsedConfig = AmplifyConfig.fromJson(json.decode(amplifyconfig));

// Get all API endpoints
final endpoints = config.api!.awsPlugin!.endpoints;
print(endpoints);

// {
//     "myapi": {
//         "endpointType": "GraphQL",
//         "endpoint": "<ENDPOINT>",
//         "region": "us-west-2",
//         "authorizationType": "API_KEY",
//         "apiKey": "<API_KEY>"
//     }
// }


// Add an endpoint configuration
final myApi = endpoints['myapi'];
endpoints['myapi_IAM'] = myApi.copyWith(
  authorizationType: APIAuthorizationType.iam,
);

// Configure Amplify
await Amplify.configure(json.encode(parsedConfig));
```

We plan to iterate on this concept to address the issues of #1115 and make creating a configuration from scratch a lot easier.

PR #1215, opened 2021-12-15
chore: remove support for loginMechanisms from Authenticator

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Remove loginMechanism and loginMechanisms support from Authenticator. Currently the Authenticator partially supports these (breaks in some cases). This PR removes all support and throws a clear error indicating what is supported/required.

### DIFF
--- a/packages/amplify_authenticator/lib/amplify_authenticator.dart
+++ b/packages/amplify_authenticator/lib/amplify_authenticator.dart
@@ -442,13 +442,15 @@ class _AuthenticatorState extends State<Authenticator> {
     if (_configInitialized && _missingConfigValues.isNotEmpty) {
       throw FlutterError.fromParts([
         ErrorSummary(
-            'Encountered problem(s) building the Authenticator due to an invalid config. See below for more info.'),
+          'Encountered problem(s) building the Authenticator due to an invalid config. See below for more info.',
+        ),
         ErrorDescription(
           '\nYour amplifyconfiguration.dart file is missing the following required attributes:'
           '\n - ${_missingConfigValues.join('\n - ')}',
         ),
         ErrorDescription(
-          '\nThis can occur if your project was not generated with the Amplify CLI, or if the project was generated with the Amplify CLI prior to version 6.5.0.', // TODO: Confirm 6.5.0
+          '\nThis can occur if your project was not generated with the Amplify CLI, '
+          'or if the project was generated with the Amplify CLI prior to version 6.5.0.', // TODO: Confirm 6.5.0
         ),
         ErrorDescription(
           '\nPlease refer to the amplify flutter documentation for more info on how to resolve this and the full list of required attributes.',
--- a/packages/amplify_authenticator/lib/amplify_authenticator.dart
+++ b/packages/amplify_authenticator/lib/amplify_authenticator.dart
@@ -450,7 +450,7 @@ class _AuthenticatorState extends State<Authenticator> {
         ),
         ErrorDescription(
           '\nThis can occur if your project was not generated with the Amplify CLI, '
-          'or if the project was generated with the Amplify CLI prior to version 6.5.0.', // TODO: Confirm 6.5.0
+          'or if the project was generated with the Amplify CLI prior to version 6.4.0.',
         ),
         ErrorDescription(
           '\nPlease refer to the amplify flutter documentation for more info on how to resolve this and the full list of required attributes.',
--- a/packages/amplify_authenticator/lib/amplify_authenticator.dart
+++ b/packages/amplify_authenticator/lib/amplify_authenticator.dart
@@ -426,13 +426,13 @@ class _AuthenticatorState extends State<Authenticator> {
       return const ['auth.plugins.Auth.Default'];
     }
     if (cognitoPlugin.usernameAttributes == null) {
-      missingValues.add('\n - usernameAttributes');
+      missingValues.add('usernameAttributes');
     }
     if (cognitoPlugin.signupAttributes == null) {
-      missingValues.add('\n - signUpAttributes');
+      missingValues.add('signUpAttributes');
     }
     if (cognitoPlugin.passwordProtectionSettings == null) {
-      missingValues.add('\n - passwordProtectionSettings');
+      missingValues.add('passwordProtectionSettings');
     }
     return missingValues;
   }
@@ -440,18 +440,23 @@ class _AuthenticatorState extends State<Authenticator> {
   @override
   Widget build(BuildContext context) {
     if (_configInitialized && _missingConfigValues.isNotEmpty) {
-      throw StateError(
-        'Encountered problem(s) building the Authenticator.'
-        '\n\n'
-        'Your amplifyconfiguration.dart file is missing the following required attributes: ${_missingConfigValues.join()}'
-        '\n\n'
-        'This can occur if you are not using the Amplify CLI, or if the project was generated with the Amplify CLI prior to version X.Y.Z.' // TODO: Update version
-        '\n\n'
-        'Please refer to the amplify flutter documentation for more info on how to address this.'
-        '\n\n'
-        'Once you have added the missing values to your amplifyconfiguration.dart file, you will need to restart your app.'
-        '\n',
-      );
+      throw FlutterError.fromParts([
+        ErrorSummary(
+            'Encountered problem(s) building the Authenticator due to an invalid config. See below for more info.'),
+        ErrorDescription(
+          '\nYour amplifyconfiguration.dart file is missing the following required attributes:'
+          '\n - ${_missingConfigValues.join('\n - ')}',
+        ),
+        ErrorDescription(
+          '\nThis can occur if your project was not generated with the Amplify CLI, or if the project was generated with the Amplify CLI prior to version 6.5.0.', // TODO: Confirm 6.5.0
+        ),
+        ErrorDescription(
+          '\nPlease refer to the amplify flutter documentation for more info on how to resolve this and the full list of required attributes.',
+        ),
+        ErrorDescription(
+          '\nOnce you have added the missing values to your amplifyconfiguration.dart file, you will need to restart your app.',
+        ),
+      ]);
     }
 
     return Localizations.override(
--- a/packages/amplify_authenticator/lib/amplify_authenticator.dart
+++ b/packages/amplify_authenticator/lib/amplify_authenticator.dart
@@ -425,11 +425,14 @@ class _AuthenticatorState extends State<Authenticator> {
     if (cognitoPlugin == null) {
       return const ['auth.plugins.Auth.Default'];
     }
+    if (cognitoPlugin.usernameAttributes == null) {
+      missingValues.add('\n - usernameAttributes');
+    }
     if (cognitoPlugin.signupAttributes == null) {
-      missingValues.add('auth.plugins.Auth.Default.signUpAttributes');
+      missingValues.add('\n - signUpAttributes');
     }
     if (cognitoPlugin.passwordProtectionSettings == null) {
-      missingValues.add('auth.plugins.Auth.Default.passwordProtectionSettings');
+      missingValues.add('\n - passwordProtectionSettings');
     }
     return missingValues;
   }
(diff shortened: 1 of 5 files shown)


packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -425,21 +425,33 @@ class _AuthenticatorState extends State<Authenticator> {
     if (cognitoPlugin == null) {
       return const ['auth.plugins.Auth.Default'];
     }
+    if (cognitoPlugin.usernameAttributes == null) {
+      missingValues.add('\n - usernameAttributes');
+    }
     if (cognitoPlugin.signupAttributes == null) {
-      missingValues.add('auth.plugins.Auth.Default.signUpAttributes');
+      missingValues.add('\n - signUpAttributes');
```
Jordan-Nelson, line 432: The formatting updates here (and some below) are to make this easier to read. See before/after below.

### Before

Screen Shot 2021-12-15 at 5 10 11 PM
Screen Shot 2021-12-15 at 5 10 24 PM

### After

Screen Shot 2021-12-15 at 5 03 50 PM
Screen Shot 2021-12-15 at 5 03 34 PM

packages/amplify_flutter/test/config/auth_config_test.dart
```diff
@@ -39,12 +39,6 @@ void main() {
             final expectedConfig = expected[name]!;
             final cognitoConfig = config.auth!.awsPlugin!.auth!.default$!;
             expect(cognitoConfig, equals(expectedConfig));
-            expect(
```
Jordan-Nelson, line 42: This was essentially testing that deserializing and then serializing the config results in the original config. That is no longer the case for older versions of the config with `loginMechanism` and `loginMechanisms` essentially being ignored.

We could have a modified version of this test where we only check a subset of the values, but I am not sure how much value that adds. I think the expectation on line 41 is sufficient.

packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -425,21 +425,33 @@ class _AuthenticatorState extends State<Authenticator> {
...
     return missingValues;
   }
 
   @override
   Widget build(BuildContext context) {
     if (_configInitialized && _missingConfigValues.isNotEmpty) {
-      throw StateError('''Encountered problem(s) building the Authenticator.
-          Your amplifyconfiguration.dart file is missing required values: ${_missingConfigValues.join('\n')}). 
-          You should correct your amplifyconfiguration.dart file and restart your app.''');
+      throw StateError(
```
dnys1, line 443: Can also consider using `FlutterError` which handles this for you. See the inherited widgets for examples.

packages/amplify_flutter/test/config/auth_config_test.dart
dnys1: This may not be true for old versions but **must** be true for all new ones if we drop support for old versions. We need to just the files not passing and leave this expectation.

packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -425,21 +425,33 @@ class _AuthenticatorState extends State<Authenticator> {
...
+        'Encountered problem(s) building the Authenticator.'
+        '\n\n'
+        'Your amplifyconfiguration.dart file is missing the following required attributes: ${_missingConfigValues.join()}'
+        '\n\n'
+        'This can occur if you are not using the Amplify CLI, or if the project was generated with the Amplify CLI prior to version X.Y.Z.' // TODO: Update version
```
haverchuck, line 448: Can probably just use the version that was live when Authenticator was released?

haverchuck (commented):
Left one minor comment re: the TODO. Once that's resolved I'll approve.

packages/amplify_authenticator/lib/amplify_authenticator.dart
dnys1: Yep, was 6.5.0 I believe, when zero-config was released

packages/amplify_flutter/test/config/auth_config_test.dart
Jordan-Nelson: > must be true for all new ones
Do you mind elaborating on that a little? I might be missing some context on this. I don't see `AmplifyConfig.toJson()` used outside of tests. I am fine with removing the tests under "old" and adding this expectation back, but would like to understand the value of this expectation and how you envision these tests working moving forward.

The latest CLI version is 7.6.3. 7.6.3 might not produce the same auth config as 6.5.0, but let's just say hypothetically that it does and these tests are effectively covering all versions of the CLI we support for the Authenticator. What happens when 8.0.0 introduces a new property. Let's say this new property is used for something outside the authenticator. We don't necessarily care about this property, but we should update `CognitoAuthConfig` to add the new property. That will make the tests for 6.5.0 fail. Won't we still want tests for 6.5.0-7.X.X in addition to tests for 8.0.0? I am envisioning that we maintain tests for every major version that we support (or every major version that produces a unique config at least).

FWIW I left the tests under "old" because I thought it would be good to have tests that validate that old configs still produce a valid `CognitoAuthConfig`. Right now we only use `CognitoAuthConfig`/`AmplifyConfig` for the authenticator, which will support 6.5.0 and up. However, if we were to start using the config for anything else, we may want to support older versions, so knowing that they produce an expected result would be valuable. With this PR, we are dropping support for anything older than 6.5.0 for the Authenticator, but I am not necessarily saying that across the board amplify_flutter does not support older versions on the config... We can always add tests for older configs if/when we use the config for other needs though so I am fine to remove the pre 6.5.0 tests.

packages/amplify_authenticator/lib/amplify_authenticator.dart
Jordan-Nelson: Thank you for that info. I knew it was around that version but wasn't sure off the top of my head. I will generate some configs with 6.4.0 and 6.5.0 just to confirm that is the cut off and then update this.

packages/amplify_authenticator/lib/amplify_authenticator.dart
Jordan-Nelson: Thanks. I will look into FlutterError and update this.

packages/amplify_flutter/test/config/auth_config_test.dart
dnys1: The utility of the `AmplifyConfig` class is two-fold for the end-user: to be able to create Amplify configurations at runtime or to be able to update Amplify configurations at runtime. In both cases, they will call `toJson` and pass the results to the `Amplify.configure`. This JSON **must** exactly match the JSON that would have been created by the CLI for the same configuration - no information can be lost or incorrectly manipulated by our decode/encode process.

That's also why support for CLI versions is all or nothing - we cannot decode an old config, then re-encode it with missing values (like loginMechanisms) - that completely defeats the purpose. Maybe not for the Authenticator, but for the utility of this class, which is public.

dnys1: If you want to keep support for decoding old configs for the purpose of the Authenticator only, then those tests should be moved to the Authenticator package, as they are no longer testing the config class.

dnys1: But I also believe we should keep login mechanisms for these reasons also - it's a very lightweight addition to the class which allows for complete backwards compatibility and a fully correct `AmplifyConfig` implementation.

Jordan-Nelson: > The utility of the AmplifyConfig class is two-fold for the end-user: to be able to create Amplify configurations at runtime or to be able to update Amplify configurations at runtime. 

Why would a customer that is using `AmplifyConfig` to create or update Amplify configurations at runtime want to add a property that the amplify libraries do not use?

> In both cases, they will call toJson and pass the results to the Amplify.configure. This JSON must exactly match the JSON that would have been created by the CLI for the same configuration

For what versions of the CLI must this match? 

What about versions of the CLI between 6.5.0 and 7.6.3? Do we know that none of those versions temporarily added an attribute only for it to be removed later? Would we need to retain those? 

What about hand edited configs? If a developer adds `"foo": "bar"` to their config, does that need to be retained? 

Why would either of these hypothetical situations be any different that `loginMechanisms`? `loginMechanisms` is unused. It was never used by a stable version of any of the amplify flutter libraries (correct me if that is wrong). Why is it different than `foo`?

I know these are somewhat ridiculous hypotheticals, but I don't understand why `AmplifyConfig.fromJson(config).toJson()` has to equal the original config, and I am also not sure it is always is true right now even with `loginMechanisms`. The only way it would always equal the original config is if we maintained a copy of the original config that was passed in.

dnys1: You can make no assumptions about how the iOS or Android libraries use those config files. Can you say with 100% certainty that `{"loginMechanisms": null}`, `{"loginMechanisms": []}`,  and `{}` will all be interpreted the same way like they are in Authenticator? Now multiply that by all the values, and appreciate the fact that null de-references can and do crash programs. If there's an empty array in the JSON we parse, it should be there when it leaves - it's that simple.

I don't think customers' extraneous K-V pairs is the same thing as what I'm describing.

dnys1: By the way - this is *not* straightforward to implement and required fine-tuning the json_serializable settings. Hence the need for the unit test - I was getting null dereferences!!

Jordan-Nelson: > You can make no assumptions about how the iOS or Android libraries use those config files

100% agreed. Which is why we shouldn't pass the value of AmplifyConfig.toJson() to iOS and Android. Are we doing that now? If so, I think that is an issue. We should always pass the unaltered version. If we pass the value of AmplifyConfig.toJson() to iOS and Android, we will have to really closely monitor changes in the CLI library and Amplify-iOS and Amplify-Android, regardless of wether or not we make the changes in this PR. I don't feel comfortable  passing that to iOS and Android unless we have integration tests that call configure (through to iOS and Android) with every major version of the CLI we intend to support.

dnys1: This class is not for Authenticator. It's a commitment to providing a fully-typed CLI config to customers which they can create from scratch, modify, and send to iOS/Android. This is a feature request we continuously get! The fact that Authenticator uses this is irrelevant.

Explain the difference between my unit test verifying decode/encode and an integration test calling configure and I'll be happy to implement it.

dnys1: I've already scaffolded the functionality to stay up to date with the CLI - the Node script uses the same package used internally to generate the configuration.

Jordan-Nelson: Can you let me know if the value from AmplifyConfig.toJson() is currently being passed to Amplify-iOS or Amplify-Android?

dnys1: No - but again, you've missed the point. I wrote this class specifically so that you can call `.toJson` on it and pass it to iOS/Android, because that is what customers want. They don't want a half-baked AmplifyConfig object with a bunch of caveats like you can only read values from it.

Jordan-Nelson: I very clearly am missing the point still 🙂 

Is there any use case (current or future) for calling `AmplifyConfig.fromJson(myJsonConfig).toJson()`?

Jordan-Nelson: If we want to allow customers to call `AmplifyConfig({prop1: value1, ... })` and then call `.toJson()` and pass it to iOS and Android, that is one thing. Taking a JSON config, converting it to an AmplifyConfig object, and then converting it back to JSON to pass to iOS and Android is something different. If we take a JSON object, convert it to an AmplifyConfig object, and then back to JSON, there is no way to know that we aren't dropping values unless we test every possible combination (every possible output from every possible version of the CLI we want to support). If a customer provides a JSON config, we should just pass that straight to native. No need to serialize/deserialize.

dnys1: > Is there any use case (current or future) for calling AmplifyConfig.fromJson(myJsonConfig).toJson()?

Yes.

- https://github.com/aws-amplify/amplify-flutter/issues/1115
- https://github.com/aws-amplify/amplify-flutter/issues/132
- https://github.com/aws-amplify/amplify-flutter/issues/1090
- https://github.com/aws-amplify/amplify-flutter/issues/1208

Customers should be able to do the following:

```dart
final json = jsonDecode(amplifyconfig) as Map<String, Object?>;
final config = AmplifyConfig.fromJson(json).copyWith(...);
await Amplify.configure(jsonEncode(config));
```

> there is no way to know that we aren't dropping values unless we test every possible combination

That's not correct - you only need to check the different types (string, num, list, etc) and null/non-null/empty - which is like 15 combos - plus the AWS-specific stuff, which follows patterns is already covered at 100%. I would argue that we have close to 100% confidence with the current version range tested (CLI v5-7) that any given config will serialize correctly. 

This is a finite space we're dealing with here. I've read the iOS mobile client's auth config parsing code and looked at all the code paths to determine this info - I'm not just pulling it out of my rear.

Jordan-Nelson: > I've read the iOS mobile client's auth config parsing code
Are you going to read through it every time we bump versions of amplify-ios? That isn't really a testing strategy. I'll talk to you about this one offline.

packages/amplify_authenticator/lib/amplify_authenticator.dart
Jordan-Nelson: Updated to 6.4.0. 6.4.0 includes usernameAttributes and omits loginMechanisms. 6.3.0 was the last version to use `loginMechanisms`.

dnys1 (approved):

haverchuck (approved):

Issue #1229, opened 2021-12-22
Query predicates do not properly support lists for observeQuery()


Abdelazeem777:
**Describe the bug**
I am using `observeQuery` to get a stream for my changes on (Chat) table, every chat object contains `usersInfo` item which is a list of JSON.

So when any changes happen to this table I am getting this exception.

`Unhandled Exception: type 'CastList<Object?, String>' is not a subtype of type 'String?' of 'other'`

I think the problem is with your map casting executed with every chat item.

**To Reproduce**
1. Just listen to any changes that happen to a table containing items with a list of AWSJSON using `observeQuery`.

2. then you will get this exception:
`Unhandled Exception: type 'CastList<Object?, String>' is not a subtype of type 'String?' of 'other'`

**Expected behavior**
Should return all changed objects without any exceptions.

**Screenshots**
![Screenshot from 2021-12-22 12-29-59](https://user-images.githubusercontent.com/30933932/147080584-5544991e-66a0-481d-a996-05175584c594.png)

**Platform**
Amplify Flutter current supports iOS and Android.  This issue is reproducible in (check all that apply):
[x] Android
[x] iOS

<details>

  <summary>Output of <code>flutter doctor -v</code></summary>
  
  ```[✓] Flutter (Channel stable, 2.8.0, on Ubuntu 21.10 5.13.0-22-generic, locale en_US.UTF-8)
    • Flutter version 2.8.0 at /home/abdelazeem/flutter
    • Upstream repository https://github.com/flutter/flutter
    • Framework revision cf44000065 (2 weeks ago), 2021-12-08 14:06:50 -0800
    • Engine revision 40a99c5951
    • Dart version 2.15.0

[✓] Android toolchain - develop for Android devices (Android SDK version 31.0.0-rc5)
    • Android SDK at /home/abdelazeem/Android/Sdk
    • Platform android-31, build-tools 31.0.0-rc5
    • Java binary at: /snap/android-studio/current/android-studio/jre/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.10+0-b96-7249189)
    • All Android licenses accepted.

[✓] Chrome - develop for the web
    • Chrome at google-chrome

[✓] Android Studio (version 2020.3)
    • Android Studio at /snap/android-studio/115/android-studio
    • Flutter plugin version 63.0.1
    • Dart plugin version 203.8452
    • Java version OpenJDK Runtime Environment (build 11.0.10+0-b96-7249189)

[...]

PR #1233, opened 2021-12-28
fix: support lists for .contains query predicate in observeQuery

Jordan-Nelson:
*Issue #, if available:* https://github.com/aws-amplify/amplify-flutter/issues/1229

*Description of changes:*

- Update ContainsQueryOperator to work for `List<String>`
- Add tests for observeQuery() contains predicate with lists
- Add integration tests query() contains predicate with lists


packages/amplify_datastore/example/integration_test/query_test/query_predicate_test/list_query_predicate_test.dart
```diff
@@ -0,0 +1,77 @@
...
+import 'package:flutter_test/flutter_test.dart';
+import 'package:amplify_flutter/amplify_flutter.dart';
+
+import '../../utils/query_predicate_utils.dart';
+import '../../utils/setup_utils.dart';
+
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  group('type List<String>', () {
```
Jordan-Nelson, line 27: *Note:* These integ tests are for `.query()`, which this PR doesn't change the functionality for. I added them to ensure I wasn't introducing breaking changes for query.

dnys1 (approved):

packages/amplify_core/lib/src/types/query/query_field_operators.dart
```diff
@@ -180,11 +180,19 @@ class ContainsQueryOperator extends QueryFieldOperatorSingleValue<String> {
       : super(value, QueryFieldOperatorType.contains);
 
   @override
-  bool evaluate(String? other) {
+  bool evaluate(dynamic other) {
     if (other == null) {
```
HuiSF, line 184: We are excluding `contains(null)` is because `[String]` and `[String!]` in model schema are both generated as `List<String>` in flutter correct?

HuiSF (approved):

packages/amplify_core/lib/src/types/query/query_field_operators.dart
Jordan-Nelson: The null check here isn't excluding `other.contains(null)` from being called. It is preventing `null.contains(value)` from being called. `other` is the field on the model being compared, and `value` is the value passed into the predicate.

However, `other.contains(null)` will not be called. This is because `value` is a non-nullable `String`. That makes sense for string comparisons. `'foo'.contains(null)` would throw an exception. However, `['foo', null].contains(null)` is valid.

I can test what happens if value is type `String?`, and a null value is provided. I a not sure if it will work if codegen is typing `[String]` as `List<String>` as you said. I am also not sure if this currently works for .query(), although it seems like it should.

On a similar topic, I originally changed `value` from `String` to `Comparable` to support calling contains() for any `List<T>` where `T` is Comparable. This would allow the contains operator to work for not only `[String]`, but also `[Int]`, `[Float]`, `[AWSDateTime]`, etc. I left it typed as `String` because .query() does not seem to support this. On the native side (not sure if it is in Amplify-Flutter or Amplify-iOS) the value is getting cast to a String and so it throws an exception.

That seems like a bug to me, but was outside of scope of this PR. For this PR I just want to get `.observeQuery()` on par with `.query()`. Would you agree that is a bug? It seems like .contains() should work for any `List<T>` where `T` is Comparable.

Jordan-Nelson: To make things more complicated ...

`ModelWithAppsyncScalarTypes.LISTOFINTVALUE.contains('1')` will actually work for .query() even though LISTOFINTVALUE is `List<int>` and `'1'` is a `String`. This will not work for .observeQuery() (with or without this PR). 

`ModelWithAppsyncScalarTypes.INTVALUE.contains('1')` will also work for .query(), but not for .observeQuery().

I don't know if it is intentional that these both work for .query(). It might just be a result of the SQL implementation (probably using a LIKE operator).

We could make both of these work for .observeQuery() as well by doing the following:

```dart
bool evaluate(dynamic other) {
  if (other == null) {
    return false;
  } else if (other is List<String>) {
    return other.contains(value);
  } else if (other is List) {
     // support non-string Lists 
    return other.map((dynamic e) => e.toString()).contains(value);
  } else if (other is String) {
    return other.contains(value);
  } else {
    // support non String values
    return other.toString().contains(value);
  }
}
```

This seems hacky and error prone. I would prefer to leave the implementation in the PR, even though it means there will still be a difference in support between .query() and .observeQuery() for the .contains() predicate. We can always revisit this if it turns out this functionality is intentional.

I am going to add some of this context in the issue I open.

HuiSF: Thanks for the digging and testing!

>I manually changed the generated models to allow this, and it caused the app to crash.

I think this is expected as we've started some work enabling `[String?]` some time last year but never finished, there is a draft PR for this...

>This would allow the contains operator to work for not only [String], but also [Int], [Float], [AWSDateTime]

I'm not sure if it's necessary to enable `contains` on other types other than `String`, as AppSync `contains` filter supports only `String` and `ID` (which is `String` in client) types. 

My comments are not blocking merging this PR.

PR #1236, opened 2021-12-30
feat(amplify_auth_cognito): deleteUser API

haverchuck:
**Note: Do NOT merge/release this pull request until the accompanying release of deleteUser has been completed in amplify-ios.**

*Description of changes:*
Adds the deleteUser API for Amplify.Auth"

This PR also moves some integ test helpers to the amplify_test package.


packages/amplify_auth_cognito/example/integration_test/delete_user_test.dart
```diff
@@ -0,0 +1,79 @@
...
+  group('deleteUser', () {
+    setUpAll(() async {
+      await configureAuth(additionalConfigs: [
+        AmplifyAPI(),
+      ]);
+      await signOutUser();
+    });
+
+    if (Platform.isIOS) {
+      testWidgets('should delete a confirmed user on iOS',
```
dnys1, line 38: Prefer using `skip` argument to `testWidgets`

packages/amplify_auth_cognito/example/lib/main.dart
```diff
@@ -137,6 +115,20 @@ class _MyAppState extends State<MyApp> {
...
+      await Amplify.Auth.deleteUser();
+      changeDisplay('SHOW_SIGN_IN');
+    } on AmplifyException catch (e) {
+      setState(() {
+        _error = e;
+      });
+      print(e);
+    } on UnimplementedError catch (e) {
+      print(e.message);
+    }
```
dnys1, line 129: Catching errors is an anti-pattern, especially here. The correct way to handle this is from a UX perspective is to not even show the delete user button on Android. And then have a check at the start of this function making it a no-op on Android. The goal is that the `UnimplementedError` is never thrown in the first place.

packages/amplify_test/lib/src/integration_test_utils/auth_cognito/integration_test_auth_utils.dart
```diff
@@ -15,15 +15,15 @@
...
 import 'package:flutter_test/flutter_test.dart';
+
 import 'package:stream_transform/stream_transform.dart';
 
-import 'types/admin_create_user_response.dart';
-import 'types/confirm_sign_up_response.dart';
-import 'types/delete_user_response.dart';
+import './types/admin_create_user_response.dart';
+import './types/confirm_sign_up_response.dart';
+import './types/delete_user_response.dart';
```
dnys1, line 26: Remove leading `./`

packages/amplify_auth_cognito/ios/Classes/AuthCognitoBridge.swift
```diff
@@ -270,4 +270,16 @@ class AuthCognitoBridge {
             }
         }
     }
+    
+    func onDeleteUser(flutterResult: @escaping FlutterResult) {
+        Amplify.Auth.deleteUser() { response in
+            switch response {
+            case .success:
+                let emptyMap: Dictionary<String, Any> = [:]
+                flutterResult(emptyMap)
```
dnys1, line 279:
```suggestion
                flutterResult(nil)
```

packages/amplify_auth_cognito/ios/Classes/SwiftAuthCognito.swift
```diff
@@ -26,6 +26,7 @@ public class SwiftAuthCognito: NSObject, FlutterPlugin {
     private var token: UnsubscribeToken?
     private let cognito: AuthCognitoBridge
     private let authCognitoHubEventStreamHandler: AuthCognitoHubEventStreamHandler?
+    private let nullaryMethods = ["deleteUser"]
```
dnys1, line 29:
```suggestion
    private let nullaryMethods: Set<String> = ["deleteUser"]
```

packages/amplify_auth_cognito/lib/method_channel_auth_cognito.dart
```diff
@@ -549,4 +551,19 @@ class AmplifyAuthCognitoMethodChannel extends AmplifyAuthCognito {
...
     }
   }
+
+  @override
+  Future<void> deleteUser() async {
+    if (!Platform.isIOS) {
+      throw UnimplementedError(
+        'The deleteUser API is currently available on the iOS platform only.',
+      );
+    } else {
```
dnys1, line 561: No need for `else` here

packages/amplify_auth_cognito/lib/amplify_auth_cognito_stream_controller.dart
```diff
@@ -52,6 +52,11 @@ _onListen() {
             _authStreamController.add(AuthHubEvent(event["eventName"]));
           }
           break;
+        case "USER_DELETED":
+          {
+            _authStreamController.add(AuthHubEvent(event["eventName"]));
+          }
+          break;
```
dnys1, line 59:
```suggestion
        case "USER_DELETED":
          _authStreamController.add(AuthHubEvent(event["eventName"]));
          break;
```

packages/amplify_auth_plugin_interface/lib/amplify_auth_plugin_interface.dart
```diff
@@ -130,4 +130,9 @@ abstract class AuthPluginInterface extends AmplifyPluginInterface {
   Future<List<AuthDevice>> fetchDevices() {
     throw UnimplementedError('fetchDevices() has not been implemented.');
   }
+
+  /// Deletes the authenticated [AuthUser].
```
dnys1, line 134: This is where we need good docs. In big, bold letters: iOS only, cannot be called from Android.

packages/amplify_auth_cognito/example/lib/main.dart
haverchuck: I had this here so that the Android behavior could be tested in the example app; but since we have integ tests for this, I'm fine with removing it.

dnys1: Okay. My point is that we need to model how best to use this API here.

packages/amplify_auth_cognito/example/integration_test/delete_user_test.dart
```diff
@@ -0,0 +1,121 @@
+/*
```
Jordan-Nelson, line 1: Do you want to import this into main_test.dart so that it runs with the rest of the test suite?

packages/amplify_auth_cognito/example/integration_test/utils/setup_utils.dart
```diff
@@ -17,10 +17,12 @@ import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
 import 'package:amplify_auth_cognito_example/amplifyconfiguration.dart';
 import 'package:amplify_flutter/amplify_flutter.dart';
 
-Future<void> configureAuth() async {
+Future<void> configureAuth({
+  List<AmplifyPluginInterface> additionalConfigs = const [],
```
Jordan-Nelson, line 21: *nit:* I think `additionalConfigs` is what we have in the authenticator tests, but `additionalPlugins` probably makes more sense.

haverchuck: Yeah you're absolutely correct.

Jordan-Nelson (approved):

HuiSF (approved):

Issue #1238, opened 2022-01-02
Navigation using new Authenticator UI component is not working


andrewcrayden:
**Describe the bug**
[As suggested in the documentation](https://docs.amplify.aws/lib/auth/getting-started/q/platform/flutter/#authentication-with-amplify), "the Authenticator component should be a descendant of your MaterialApp". The issue is that this Amplify UI Component does not seem to work with the `MaterialApp` navigation system, and there doesn't seem to be any guidance for navigation in the following documentation:
- [Authenticator Amplify UI documentation](https://docs.amplify.aws/ui/auth/authenticator/q/framework/flutter/)
- [Flutter Authentication library documentation](https://docs.amplify.aws/lib/auth/getting-started/q/platform/flutter/#authentication-with-amplify)
- [amplify_authenticator pub.dev Readme and Example pages](https://pub.dev/packages/amplify_authenticator)

The `HomePage` component below renders fine, but the `SecondPage` component which includes the new `SignOutButton` throws the following error:  
> The following assertion was thrown building SignOutButton(dirty): No InheritedAuthViewModel widget found.
Make sure your app is wrapped with an Authenticator widget. The relevant error-causing widget was: SignOutButton

Is this new Authenticator UI component meant to work when configuring the `routes` property of `MaterialApp`? What is the correct way to implement navigation and the new Authenticator UI component together?

**main.dart**
```
import 'package:amplify_api/amplify_api.dart';
import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
import 'package:amplify_authenticator/amplify_authenticator.dart';
import 'package:amplify_flutter/amplify.dart';
import 'package:flutter/material.dart';
import 'pages/home_page.dart';
import 'pages/second_page.dart';
import 'package:example/providers/user_provider.dart';
import 'package:provider/provider.dart';
import 'dart:developer' as developer;
import 'amplifyconfiguration.dart';

void main() {
  runApp(
      MultiProvider(providers: [
        ChangeNotifierProvider(create: (_) => UserProvider())
      ],
          child: MyApp()
      )
  );
}

class MyApp extends StatefulWidget {
  const MyApp({Key? key}) : super(key: key);

  @override
  State<MyApp> createState() => _MyAppState();
}

class _MyAppState extends State<MyApp> {

  @override
  void initState() {
    super.initState();
    _configureAmplify();
  }

  Future<void> _configureAmplify() async {

[...]

PR #1241, opened 2022-01-03
fix(authenticator): Improved navigation support

dnys1:
*Issue #, if available:*
Fixes https://github.com/aws-amplify/amplify-flutter/issues/1238

*Description of changes:*
- Re-organizes Authenticator components to allow `Navigator` to work correctly.

### DIFF
--- a/packages/amplify_authenticator/example/lib/main.dart
+++ b/packages/amplify_authenticator/example/lib/main.dart
@@ -118,7 +118,7 @@ class _MyAppState extends State<MyApp> {
         // The Authenticator component must wrap your Navigator component which
         // can be done using the `builder` method.
         builder: (context, child) {
-          return AuthenticatorBody(child: child!);
+          return AuthenticatorBody(child: child);
         },
 
         initialRoute: '/routeA',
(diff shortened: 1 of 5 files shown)


packages/amplify_authenticator/example/lib/main.dart
```diff
@@ -96,30 +96,37 @@ class _MyAppState extends State<MyApp> {
...
-        Locale('en'), // English
-        Locale('es'), // Spanish
-      ],
-
-      // The Authenticator component should be a descendant of your MaterialApp.
-      home: authenticator,
+        // The Authenticator component must wrap your Navigator component which
+        // can be done using the `builder` method.
+        builder: (context, child) {
+          return AuthenticatorBody(child: child);
```
haverchuck, line 121: If we have to expose and force the use of AuthenticatorBody like this, maybe we should change the name? Perhaps we could do `AuthenticatorChild` with a positional argument?

dnys1: Updated 👍

Jordan-Nelson (approved):

haverchuck (approved):

PR #1252, opened 2022-01-08
feat(authenticator): l10n Chinese support

quirkeer:
*Description of changes:*

Add l10n Chinese translation.


dnys1:
Thanks for your PR, @quirkeer! Would you be willing to run the `tool/generate_l10n.sh` script and verify that the app looks and feels correct when using Chinese? There may be more changes necessary on our end and I'd love your feedback if so!

packages/amplify_authenticator/lib/src/l10n/src/titles/titles_zh.arb
```diff
@@ -0,0 +1,12 @@
+{
+    "@@locale": "zh",
+    "@@context": "amplify_authenticator:titles",
+    "signin": "登陆你的账户",
```
HuiSF, line 4:
```suggestion
    "signin": "登录你的账户",
```

packages/amplify_authenticator/lib/src/l10n/src/countries/countries_zh.arb
```diff
@@ -0,0 +1,250 @@
+{
+  "@@locale": "zh",
+  "@@context": "amplify_authenticator:countries",
+  "selectDialCode": "选择您的国家区号",
```
HuiSF, line 4: Translation of possessive determiner is inconsistent, e.g. "您的" vs "你的".

HuiSF (commented):
Thanks for contributing!

HuiSF:
Hello @quirkeer 
Currently, developers can provide and use their own localization for authenticator  ([reference](https://ui.docs.amplify.aws/components/authenticator?platform=flutter#internationalization-i18n)). We will be investigating possibility of enabling internationalization in the Authenticator based on App running device locale. Hence, I'm going to close this PR for now.

Thank you very much again for contributing. Please feel free to follow up if you have any questions and suggestions.

dJani97:
An updated link to the docs: https://ui.docs.amplify.aws/flutter/connected-components/authenticator/customization

May I ask why such community PRs are rejected? The documented way seems to involve overriding 7+ resolver classes and hundreds of methods, each returning a single localization value. Is this really the only way to localize the library?

Having premade ARB files contributed by the community would be so much easier for everyone.

Amplify-UI already does this: https://github.com/aws-amplify/amplify-ui/tree/main/packages/ui/src/i18n/dictionaries/authenticator. Plus, creating your own localization is a breeze, just copy one of these dictionaries and change the values. Can we have something like that?

Issue #230, opened 2020-12-03
[DataStore] Add custom conflict resolution handling


Amplifiyer:

mehrdad-shokri:
2 years old, what's status of this feature?

HuiSF:
amplify-flutter [0.5.0](https://github.com/aws-amplify/amplify-flutter/releases/tag/v0.5.0) has released, including capability of configuring `conflictHandler` for DataStore.
Examples please refer to [this documentation](https://docs.amplify.aws/lib/datastore/conflict/q/platform/flutter/#custom-configuration-fields).

Please feel free to follow up if anything. Thanks!

PR #1254, opened 2022-01-10
feat(datastore): Custom Conflict Handler

fjnoyp:
*Issue #, if available:* https://github.com/aws-amplify/amplify-flutter/issues/230

*Description of changes:*

Add ConflictHandler Support.  The API is similar to native, you pass a custom conflict handler method within the Datastore configure method. 

Tested and working on iOS. 

> **The Android team is actively working to fix conflict handler.  We will need to wait for them before we can release this PR.** 
> However, beyond updating our Android dependency version, this will not require a change of code in this PR.  Android team is just fixing a bug on their side only. 

The ConflictData object passed in the custom ConflictHandler method contains a local and remote model.  Currently, the remote model will not contain nested model data even if that field is required.  Thus in that case, if you return a remote.copyWith model, the operation will fail as a required field is missing. 
As this is the current behavior in native and JS and is expected behavior in AppSync we will resolve that issue separately. 


packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -202,6 +199,59 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
         }
         dataStoreConfigurationBuilder.errorHandler(errorHandler)
 
+        var conflictHandler: DataStoreConflictHandler
+        conflictHandler = if ((request["hasConflictHandler"] as? Boolean? == true)) {
```
HuiSF: The double parentheses may be uncessary.

```suggestion
        val errorHandler = if (request["hasErrorHandler"] as? Boolean? == true) {
```

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -202,6 +199,59 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
...
+                        "local" to FlutterSerializedModel(conflictData.local as SerializedModel).toMap(),
+                        "remote" to FlutterSerializedModel(conflictData.remote as SerializedModel).toMap()
+                )
+
+                uiThreadHandler.post {
+                    channel.invokeMethod("conflictHandler", args, object : Result {
+                        override fun success(result: Any?) {
+                            val resultMap: Map<String, Any>? = result.safeCastToMap()
+                            when (resultMap?.get("resolutionStrategy") as String) {
+                                "APPLY_REMOTE" -> onDecision.accept(DataStoreConflictHandler.ConflictResolutionDecision.applyRemote())
```
HuiSF: Maybe we can reuse the existing enum `DataStoreConflictHandler.ResolutionStrategy` from amplify-android for consistency.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -672,4 +655,78 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
             }
         }
     }
+
+    private fun createErrorHandler(request: Map<String, Any>) : DataStoreErrorHandler{
+        var errorHandler : DataStoreErrorHandler
+        errorHandler = if( (request["hasErrorHandler"] as? Boolean? == true) ) {
```
HuiSF, line 661:
```suggestion
        val errorHandler = if (request["hasErrorHandler"] as? Boolean? == true) {
```

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -672,4 +655,78 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
...
+            DataStoreErrorHandler {
+                val args = hashMapOf(
```
HuiSF, line 663: Could use `mapOf` as the args is used as an immutable map.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -672,4 +655,78 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
...
+            DataStoreErrorHandler {
+                LOG.error(it.toString())
+            }
+        }
+        return errorHandler
+    }
+
+    private fun createConflictHandler(request: Map<String, Any>) : DataStoreConflictHandler{
+        var conflictHandler: DataStoreConflictHandler
+        conflictHandler = if ((request["hasConflictHandler"] as? Boolean? == true)) {
```
HuiSF, line 680:
```suggestion
        val conflictHandler = if (request["hasConflictHandler"] as? Boolean? == true) {
```

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -672,4 +655,78 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
...
+                        "local" to FlutterSerializedModel(conflictData.local as SerializedModel).toMap(),
+                        "remote" to FlutterSerializedModel(conflictData.remote as SerializedModel).toMap()
+                )
+
+                uiThreadHandler.post {
+                    channel.invokeMethod("conflictHandler", args, object : Result {
+                        override fun success(result: Any?) {
+                            val resultMap: Map<String, Any>? = result.safeCastToMap()
+                            when (resultMap?.get("resolutionStrategy") as String) {
+                                "APPLY_REMOTE" -> onDecision.accept(DataStoreConflictHandler.ConflictResolutionDecision.applyRemote())
```
HuiSF, line 696: Could use `DataStoreConflictHandler.ResolutionStrategy` to replace hardcoded string `"APPLY_REMOTE"` for consistency.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/conflict_handler/FlutterConflictResolutionDecision.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import com.amplifyframework.core.model.Model
+import com.amplifyframework.datastore.DataStoreConflictHandler;
+
+data class FlutterConflictResolutionDecision(val map: Map<String, Any>) {
+
+    fun convertToNativeConflictResolutionDecision() : DataStoreConflictHandler.ConflictResolutionDecision<Model>? {
+        return null
+    }
+}
```
HuiSF, line 11:
```suggestion
}

```

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
```diff
@@ -601,9 +589,77 @@ public class SwiftAmplifyDataStorePlugin: NSObject, FlutterPlugin {
...
+                        "modelName" : modelName,
+                        "local" : try local.toMap(flutterModelRegistration: self.modelSchemaRegistry, modelName: modelName),
+                        "remote" : try remote.toMap(flutterModelRegistration: self.modelSchemaRegistry, modelName: modelName)
+                    ]
+
+                    self.channel!.invokeMethod("conflictHandler", arguments: map){ result in
+                        do {
+                            let resultMap : [String: Any] = result as! [String: Any]
+                            switch(resultMap["resolutionStrategy"] as! String){
+                                case "APPLY_REMOTE": onDecision(.applyRemote)
```
HuiSF, line 636: Ditto can we use exiting enum if any?

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
```diff
@@ -601,9 +589,77 @@ public class SwiftAmplifyDataStorePlugin: NSObject, FlutterPlugin {
...
+        }
+        return errorHandler
+    }
+    
+    func createConflictHandler(args: [String: Any])  -> DataStoreConflictHandler {
+        var conflictHandler:DataStoreConflictHandler
+        if((args["hasConflictHandler"] as? Bool) == true) {
+            conflictHandler = { conflictData, onDecision in
+                do {
+                    let modelName = conflictData.remote["__typename"] as! String
```
HuiSF, line 622: Question: is the `conflictData.remote` from AppSync directly without native model conversion always where `__typename` is always available?

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -38,10 +38,7 @@ import com.amplifyframework.core.model.Model
 import com.amplifyframework.core.model.SerializedModel
 import com.amplifyframework.core.model.query.QueryOptions
 import com.amplifyframework.core.model.query.predicate.QueryPredicates
-import com.amplifyframework.datastore.AWSDataStorePlugin
-import com.amplifyframework.datastore.DataStoreConfiguration
-import com.amplifyframework.datastore.DataStoreErrorHandler
-import com.amplifyframework.datastore.DataStoreException
+import com.amplifyframework.datastore.*
```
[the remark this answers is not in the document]
fjnoyp: alright changed back

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -166,8 +163,8 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
 
         val dataStoreConfigurationBuilder = DataStoreConfiguration.builder()
 
-        var errorHandler : DataStoreErrorHandler;
-        errorHandler = if( (request["hasErrorHandler"] as? Boolean? == true) ) {
+        var errorHandler: DataStoreErrorHandler
+        errorHandler = if ((request["hasErrorHandler"] as? Boolean? == true)) {
```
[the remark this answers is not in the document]
fjnoyp: thanks, updated to your second comment.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -202,6 +199,59 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
         }
         dataStoreConfigurationBuilder.errorHandler(errorHandler)
 
+        var conflictHandler: DataStoreConflictHandler
+        conflictHandler = if ((request["hasConflictHandler"] as? Boolean? == true)) {
+            DataStoreConflictHandler { conflictData,
+                                       onDecision ->
+
+                val modelName = conflictData.local.modelName
+                val args = hashMapOf(
```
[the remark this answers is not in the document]
fjnoyp: done

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
fjnoyp: changed to: 

`conflictHandler = if (request["hasConflictHandler"] as? Boolean? == true) {`

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
fjnoyp: switched to using enum :

```
try {
                                val resolutionStrategy: ResolutionStrategy =
                                        ResolutionStrategy.valueOf(resultMap?.get("resolutionStrategy") as String)
                                when (resolutionStrategy) {
```

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
fjnoyp: changed to inline return

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
fjnoyp: updated

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
fjnoyp: changed to inline return.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
fjnoyp: Changed to use enum

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/conflict_handler/FlutterConflictResolutionDecision.kt
fjnoyp: this class is not used, deleted it

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
fjnoyp: Hmm I don't think this will be so straightforward to do.  

I don't know how to directly create an enum from a string without declaring the enum with : string.  (see [stackoverflow](https://stackoverflow.com/questions/30009788/in-swift-is-it-possible-to-convert-a-string-to-an-enum/30009873)) 

In the example implementation I used above, the code fails: 
```

import Foundation

public protocol Model: Codable {
    var modelName: String { get }
}

public enum DataStoreConflictHandlerResult {

    /// Discard the local changes in favor of the remote ones. Semantically the same as `DISCARD` on Amplify-JS
    case applyRemote

    /// Keep the local changes. (semantic shortcut to `retry(local)`).
    case retryLocal

    /// Return a new `Model` instance that should used instead of the local and remote changes.
    case retry(Model)
}

print( DataStoreConflictHandlerResult(rawValue: "retry") )
```

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
fjnoyp: If I understand you correctly, yes -> the __typename field should always be there in `conflictData.remote`.  Had to use that field due to an Amplify iOS quirk where fields labeled modelName were always returning `FlutterSerializedModel` which would break the code with "unknown model"

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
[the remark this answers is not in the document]
HuiSF: Thanks! But you might have missed to include this change in the 3rd commit :D

HuiSF (approved):

packages/amplify_datastore/lib/method_channel_datastore.dart
```diff
@@ -54,12 +55,23 @@ class AmplifyDataStoreMethodChannel extends AmplifyDataStore {
...
+        errorHandler!(_deserializeExceptionFromMap(arguments));
         break;
 
       case 'conflictHandler':
-        break;
+        Map<String, dynamic> arguments =
+            Map<String, dynamic>.from(call.arguments);
+
+        String modelName = arguments["modelName"];
+        var modelType = modelProvider!.getModelTypeByModelName(modelName);
```
dnys1, line 66: Be consistent - either `Type name = ...` or `var name = ...`, but preferably use `final name = ...` for values which don't change:

```suggestion
        final modelName = arguments["modelName"] as String;
        final modelType = modelProvider!.getModelTypeByModelName(modelName);
```

packages/amplify_datastore_plugin_interface/lib/src/types/conflict_handler/datastore_conflict_handler.dart
```diff
@@ -0,0 +1,52 @@
+import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart';
+import 'package:flutter/foundation.dart';
+
+typedef DataStoreConflictHandler = ConflictResolutionDecision Function(
+    ConflictData);
+
+class ConflictData {
```
dnys1, line 7:
```suggestion
@immutable
class ConflictData {
```

packages/amplify_datastore_plugin_interface/lib/src/types/conflict_handler/datastore_conflict_handler.dart
```diff
@@ -0,0 +1,52 @@
...
+  Model local;
+  Model remote;
+
+  ConflictData(this.local, this.remote);
```
dnys1, line 11:
```suggestion
  const ConflictData(this.local, this.remote);
```

packages/amplify_datastore_plugin_interface/lib/src/types/conflict_handler/datastore_conflict_handler.dart
```diff
@@ -0,0 +1,52 @@
...
+
+  ConflictData.fromJson(ModelType modelType, Map<String, dynamic> localJson,
+      Map<String, dynamic> remoteJson)
+      : local = modelType.fromJson(
+            new Map<String, dynamic>.from(localJson["serializedData"])),
+        remote = modelType.fromJson(
+            new Map<String, dynamic>.from(remoteJson["serializedData"]));
+}
+
+enum ResolutionStrategy { APPLY_REMOTE, RETRY_LOCAL, RETRY }
```
dnys1, line 21: I won't mandate it here since we haven't cleaned up older code which does this - but please use camelCase for enum types per Dart conventions

packages/amplify_datastore_plugin_interface/lib/src/types/conflict_handler/datastore_conflict_handler.dart
```diff
@@ -0,0 +1,52 @@
...
+  Model remote;
+
+  ConflictData(this.local, this.remote);
+
+  ConflictData.fromJson(ModelType modelType, Map<String, dynamic> localJson,
+      Map<String, dynamic> remoteJson)
+      : local = modelType.fromJson(
+            new Map<String, dynamic>.from(localJson["serializedData"])),
+        remote = modelType.fromJson(
+            new Map<String, dynamic>.from(remoteJson["serializedData"]));
```
dnys1, line 18: `new` was deprecated in Dart 2.0

```suggestion
      : local = modelType.fromJson(
            (localJson["serializedData"] as Map).cast<String, dynamic()),
        remote = modelType.fromJson(
            (remoteJson["serializedData"] as Map).cast<String, dynamic());
```

packages/amplify_datastore_plugin_interface/lib/src/types/conflict_handler/datastore_conflict_handler.dart
```diff
@@ -0,0 +1,52 @@
...
+}
+
+enum ResolutionStrategy { APPLY_REMOTE, RETRY_LOCAL, RETRY }
+
+class ConflictResolutionDecision {
+  ResolutionStrategy _resolutionStrategy;
+  Model? customModel;
+
+  ConflictResolutionDecision(this._resolutionStrategy, customModel);
```
dnys1, line 27:
```suggestion
@immutable
class ConflictResolutionDecision {
  final ResolutionStrategy _resolutionStrategy;
  final Model? customModel;

  const ConflictResolutionDecision(this._resolutionStrategy, customModel);
```

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
```diff
@@ -622,9 +610,83 @@ public class SwiftAmplifyDataStorePlugin: NSObject, FlutterPlugin {
...
+                    ]
+
+                    self.channel!.invokeMethod("conflictHandler", arguments: map){ result in
+                        do {
+                            let resultMap : [String: Any] = result as! [String: Any]
+                            switch(resultMap["resolutionStrategy"] as! String){
+                                case "APPLY_REMOTE": onDecision(.applyRemote)
+                                case "RETRY_LOCAL": onDecision(.retryLocal)
+                                case "RETRY":
+                                    let modelMap : [String: Any] = resultMap["customModel"] as! [String : Any]
```
dnys1, line 666: You force unwrap customModel here but it is nullable in Dart

packages/amplify_datastore_plugin_interface/lib/src/types/conflict_handler/datastore_conflict_handler.dart
```diff
@@ -0,0 +1,52 @@
...
+
+  ConflictResolutionDecision.applyRemote()
+      : _resolutionStrategy = ResolutionStrategy.APPLY_REMOTE;
+
+  ConflictResolutionDecision.retryLocal()
+      : _resolutionStrategy = ResolutionStrategy.RETRY_LOCAL;
+
+  ConflictResolutionDecision.retry(this.customModel)
```
dnys1, line 35:
```suggestion
  ConflictResolutionDecision.retry(Model this.customModel)
```

dnys1 (changes requested):
Generally looks fine, but please:
- Reformat all code (Cmd+Option+Shift+L in Android Studio)
- I won't approve code with force unwraps (`!`), especially iOS code since it will crash the app. Please re-work to use `guard` statements and throw exceptions on missing values

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -692,4 +677,77 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
...
+                            val resultMap: Map<String, Any>? = result.safeCastToMap()
+                            try {
+                                val resolutionStrategy: ResolutionStrategy =
+                                        ResolutionStrategy.valueOf(resultMap?.get("resolutionStrategy") as String)
+                                when (resolutionStrategy) {
+                                    ResolutionStrategy.APPLY_REMOTE -> onDecision.accept(DataStoreConflictHandler.ConflictResolutionDecision.applyRemote())
+                                    ResolutionStrategy.RETRY_LOCAL -> onDecision.accept(DataStoreConflictHandler.ConflictResolutionDecision.retryLocal())
+                                    ResolutionStrategy.RETRY -> {
+                                        val serializedModel = SerializedModel.builder()
+                                                .serializedData(resultMap["customModel"] as HashMap<String, Any>)
```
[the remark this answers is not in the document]
fjnoyp: Alright, what is the benefit of calling cast() instead of what we had before?

packages/amplify_datastore/lib/method_channel_datastore.dart
```diff
@@ -54,12 +55,23 @@ class AmplifyDataStoreMethodChannel extends AmplifyDataStore {
       case 'errorHandler':
         Map<String, dynamic> arguments =
             Map<String, dynamic>.from(call.arguments);
-        _errorHandler!(_deserializeExceptionFromMap(arguments));
+        errorHandler!(_deserializeExceptionFromMap(arguments));
```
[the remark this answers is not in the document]
fjnoyp: it's guaranteed to never be called if errorHandler / conflictHandler was null in the first place.  

We call `method_channel_datastore` - `configureDataStore` with fields : 

'hasErrorHandler': errorHandler != null,
        'hasConflictHandler': conflictHandler != null,

That are used to determine if the methodChannel will be called again in native side.  For example: 

    ```
private fun createConflictHandler(request: Map<String, Any>) : DataStoreConflictHandler {
        return if (request["hasConflictHandler"] as? Boolean? == true) {
```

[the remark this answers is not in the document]
Comment (fjnoyp) [reply packages/amplify_datastore/lib/method_channel_datastore.dart:58]: I think you have a good point though ... should just have a redundant check to avoid a force cast exception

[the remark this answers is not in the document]
Comment (fjnoyp) [reply packages/amplify_datastore/lib/method_channel_datastore.dart:58]: if (errorHandler == null) return;

File: packages/amplify_datastore/lib/method_channel_datastore.dart (hunk 13 above)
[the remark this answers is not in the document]
Comment (fjnoyp) [reply packages/amplify_datastore/lib/method_channel_datastore.dart:66]: Our code is a bit strange ... the constructor of AmplifyDataStore has a required modelProvider field but the underlying field is nullable 

I believe this is to allow a custom ModelProvider instance to be passed in the configure call.  Most specifically this line: 

`amplify_datastore` - `configureDataStore` : 
    ModelProviderInterface provider = modelProvider ?? this.modelProvider!;

File: packages/amplify_datastore/lib/method_channel_datastore.dart [hunk 23]
```diff
@@ -54,12 +55,23 @@ class AmplifyDataStoreMethodChannel extends AmplifyDataStore {
...
+
+        String modelName = arguments["modelName"];
+        var modelType = modelProvider!.getModelTypeByModelName(modelName);
+
+        ConflictData conflictData = ConflictData.fromJson(
+            modelType,
+            Map<String, dynamic>.from(arguments["local"]),
+            Map<String, dynamic>.from(arguments["remote"]));
 
+        ConflictResolutionDecision decision = conflictHandler!(conflictData);
```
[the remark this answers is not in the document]
fjnoyp: same reason for error handler, added a new check 

> 
>    if (conflictHandler == null)
>           throw StateError(
>               "Native calling non existen ConflictHandler in Dart");

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
fjnoyp: switch with guard 

```
guard let modelMap =  resultMap["customModel"] as? [String : Any]
                                        else {
                                            throw DataStoreError.decodingError("Flutter CustomModel map is invalid", "Check the values that are being passed from Dart.")
                                        }
```

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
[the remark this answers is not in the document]
dnys1: To consolidate unchecked casts:

Before 
Screen Shot 2022-03-22 at 5 44 26 PM

After 
Screen Shot 2022-03-22 at 5 45 43 PM

packages/amplify_datastore/lib/method_channel_datastore.dart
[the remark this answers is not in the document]
dnys1: Yeah, we just wanna reduce the mental burden and if there's a comment or better an exception that explains why it's never null then we're golden

dnys1 (approved):
Looks good, thanks!

Issue #1277, opened 2022-01-17
How to use Authenticator component to protect only some routes


andrewcrayden:
**Describe the bug**
The team here seemed very receptive to some [feedback I submitted about Authenticator navigation](https://github.com/aws-amplify/amplify-flutter/issues/1238), so I wanted to use Github again to submit further feedback and ask for some guidance.

It is not clear how to use the Authenticator component to protect only part of an app. The documentation suggests putting the entire `MaterialApp` inside of the Authenticator widget, and then calling `builder: Authenticator.builder()` within the `MaterialApp`.

Has the Authenticator component been designed to allow for protecting only some of the app routes? Is this possible?

Documentation that I referenced:
[Authenticator UI Component docs](https://docs.amplify.aws/ui/auth/authenticator/q/framework/flutter/) - Amplify Docs
[amplify_authenticator: ^0.1.0-rc.3 readme and example](https://pub.dev/packages/amplify_authenticator) - pub.dev

```
@override
Widget build(BuildContext context) {
  return Authenticator(
    child: MaterialApp(
      title: 'Authenticator Demo',
      builder: Authenticator.builder(),
      home: const Scaffold(
        body: Center(
          child: Text('You are logged in!'),
        ),
      ),
    ),
  );
}
```

dnys1:
Hi @andrewcrayden - this is not possible yet, although we are planning to release support for this soon.

dnys1:
Great question, and I wish I had a great answer for you haha. I believe the best way is to check the `pub.dev` listing and Github releases page regularly. No firm timelines but within the next few weeks for sure!

andrewcrayden:
@dnys1 checking the `pub.dev` is more than satisfactory. Really appreciate that I can drop feedback here on Github.. Love what your team is doing for Flutter in Amplify!

dnys1:
My pleasure! Thanks for all the feedback :-)

Jordan-Nelson:
Hey @andrewcrayden - This is available in `0.1.0.rc.4`. You can find the docs for it [here](https://docs.amplify.aws/ui/auth/authenticator/q/framework/flutter/#unauthenticated-routes). Let me know if you have any questions or issues while using it. I'll leave this issue open for now if you have any feedback to provide while using it.

Jordan-Nelson:
The docs for this are now located here: https://ui.docs.amplify.aws/components/authenticator?platform=flutter#unauthenticated-routes

I am going to close this issue out.

[...]

PR #1281, opened 2022-01-18
feat(amplify_authenticator): Custom UI, protect subset of routes

Jordan-Nelson:
#### Issue #, if available:
- https://github.com/aws-amplify/amplify-flutter/issues/1277

#### Description of changes:

This PR introduces two main changes.
1. Allow the customer to provide a builder method to create a custom UI for the authenticator. This change involves making a number of things public that are currently private. In making them public, many of the classes/interfaces were modified to remove portions that should not be public or to make them easier to use and understand.
2. Allow for only a subset of routes to be protected by the authenticator, opposed to the entire application

#### Breaking Changes
- As part of the effort to clean up APIs that were being made public, some APIs that were already public were changed as well. An. example of this is some of the names of the content strings - `"signin"` changed to `"signIn"` to match `AuthenticatorStep.signIn`
- Button sizes were changed. The authenticator defaults most button heights to 48px, which is larger than the default in material. It became obvious that something was off when a custom UI was built composed of some standard material widgets and some prebuilt widgets. We could change the heights back to 48px for AmplifyTheme if desired.

#### Examples / Public API

Below are a few examples of the new functionality.

Example with only some routes protected:

```dart
@override
  return Authenticator(
    child: MaterialApp(
      initialRoute: '/routeA',
      routes: {
        '/routeA': (BuildContext context) => const RouteA(),
        '/routeB': (BuildContext context) {
          return const AuthenticatedView(
            child: RouteB(),
          );
        },
      },
    ),
  );
}
```

Example with a custom UI

```dart
@override
  return Authenticator(
    // builder method that adds a logo to the sign in and sign up screens,
    // and removes the TabBar navigation
    builder: (context, state) {
      switch (state.currentStep) {
        case AuthenticatorStep.signIn:
          return Scaffold(
            body: Padding(
              padding: const EdgeInsets.all(16),
              child: Column(
                children: [
                  // app logo (flutter logo used for example)
                  const Center(child: FlutterLogo(size: 100)),

                  // prebuilt sign in form from amplify_authenticator package
                  SignInForm(),

                  Expanded(child: Container()),

                  // custom button to take the user to sign up
                  NavigateToSignUpButton(state: state),
                ],
              ),
            ),
          );
        case AuthenticatorStep.signUp:
          return Scaffold(
            body: Padding(
              padding: const EdgeInsets.all(16),
              child: Column(
                children: [
                  // app logo (flutter logo used for example)
                  const Center(child: FlutterLogo(size: 100)),

                  // prebuilt sign up form from amplify_authenticator package
                  SignUpForm(),

                  Expanded(child: Container()),

                  // custom button to take the user to sign in
                  NavigateToSignInButton(state: state),
                ],
              ),
            ),
          );
        default:
          // return the prebuilt authenticator for all other steps
          return null;
      }
    },
    child: MaterialApp(
      builder: Authenticator.builder(),
      initialRoute: '/routeA',
      routes: {
        '/routeA': (BuildContext context) => const RouteA(),
        '/routeB': (BuildContext context) => const RouteB(),
      },
    ),
  );
}
}
```


packages/amplify_authenticator/lib/src/models/authenticator_builder.dart
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
Jordan-Nelson, line 2: The diff here is misleading. This is a new file. The file on the left was renamed `authenticator_step.dart`.

packages/amplify_authenticator/lib/src/widgets/button.dart
```diff
@@ -27,14 +25,14 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 
 extension AuthenticatorButtonSizes on AuthenticatorButtonSize {
-  double get height {
+  double? get height {
```
Jordan-Nelson, line 28: Some cleanup is needed here. Most button heights were defaulting to 48px. This may be desirable for AmplifyTheme, but is much larger than the default in material. Returning null makes the buttons fallback to the material default. I left `AuthenticatorButtonSize.medium` at 40 because it is only used for social buttons, which probably do need a specific height.

packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -229,11 +241,14 @@ class Authenticator extends StatefulWidget {
     this.onException,
     this.exceptionBannerLocation = ExceptionBannerLocation.auto,
     this.preferPrivateSession = false,
+    this.initialStep = AuthenticatorStep.signIn,
+    AuthenticatorBuilder? builder,
```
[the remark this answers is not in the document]
dnys1: Something like this could also work:

```dart
@override
  return Authenticator(
    child: MaterialApp(
      builder: Authenticator.builder((context, state, defaultScreen) {
        switch (state) {
          case AuthenticatorStep.loading:
            return const CustomLoadingScreen();
        }
        return defaultScreen;
      }),
      home: const RouteA(),
    ),
  );
}
```

packages/amplify_authenticator/example/lib/authenticator_with_custom_layout.dart
```diff
@@ -0,0 +1,173 @@
...
+      //
+      // state is the current state of the authenticator. It contains the
+      // current auth step (sign in, sign out, etc.), the form data, and
+      // methods for updating this state.
+      //
+      // child is widget that would be displayed without the use of a custom
+      // builder. If currentStep is AuthenticatorStep.authenticated, this will
+      // be the material app. Otherwise this will be a prebuilt widget from
+      // amplify_authenticator to handle the current AuthenticatorStep
+      builder: (context, state, child) {
```
dnys1, line 37: I think `child` is a confusing name - I'd prefer `defaultChild` or `defaultScreen` or just not even pass it in at all and let the method return `null` since every implementation of this should just return it as the default option and will never wrap it.

packages/amplify_authenticator/example/lib/authenticator_with_custom_forms.dart
```diff
@@ -0,0 +1,346 @@
+import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
```
haverchuck, line 1: Let's add a license.

packages/amplify_authenticator/example/lib/authenticator_with_custom_layout.dart
```diff
@@ -0,0 +1,173 @@
+import 'package:amplify_authenticator/amplify_authenticator.dart';
```
haverchuck, line 1: License here also.

packages/amplify_authenticator/example/lib/main.dart
```diff
@@ -126,6 +134,57 @@ class _MyAppState extends State<MyApp> {
...
+  //           return const AuthenticatedView(
+  //             child: RouteB(),
+  //           );
+  //         },
+  //       },
+  //     ),
+  //   );
+  // }
+
+  // Providing a `builder` argument to Authenticator.builder, allows you to
```
haverchuck, line 166:
```suggestion
  // Providing a `builder` argument to Authenticator.builder allows you to
```

packages/amplify_authenticator/example/lib/main.dart
```diff
@@ -126,6 +134,57 @@ class _MyAppState extends State<MyApp> {
...
+  //     child: MaterialApp(
+  //       theme: ThemeData.light(),
+  //       darkTheme: ThemeData.dark(),
+  //       themeMode: ThemeMode.system,
+  //       debugShowCheckedModeBanner: false,
+  //       initialRoute: '/routeA',
+  //       routes: {
+  //         '/routeA': (BuildContext context) => const RouteA(),
+  //         '/routeB': (BuildContext context) {
+  //           return const AuthenticatedView(
```
haverchuck, line 157: 👍

dnys1 (commented):
Initial comments

packages/amplify_authenticator/example/lib/main.dart
```diff
@@ -126,6 +134,57 @@ class _MyAppState extends State<MyApp> {
...
+  //
+  // Uncomment the build method below (and comment out the one above) to
+  // see a simple example of this
+
+  // @override
+  // Widget build(BuildContext context) {
+  //   return const AuthenticatorWithCustomLayout();
+  // }
+
+  // Below is another example of a custom authenticator, with a more
```
haverchuck, line 181:
```suggestion
  // Below is another example of a custom authenticator, with more
```

packages/amplify_authenticator/lib/src/blocs/auth/auth_event.dart
```diff
@@ -23,14 +23,10 @@ class AuthLoad extends AuthEvent {
   const AuthLoad();
 }
 
-class GetCurrentUser extends AuthEvent {
-  const GetCurrentUser();
-}
-
 class AuthChangeScreen extends AuthEvent {
-  final AuthScreen screen;
+  final AuthenticatorStep step;
```
haverchuck, line 27: +1 for this change.

packages/amplify_authenticator/example/lib/authenticator_with_custom_layout.dart
```diff
@@ -0,0 +1,217 @@
...
+      //
+      // state is the current state of the authenticator. It contains the
+      // current auth step (sign in, sign out, etc.), the form data, and
+      // methods for updating this state.
+      //
+      // child is widget that would be displayed without the use of a custom
+      // builder. If currentStep is AuthenticatorStep.authenticated, this will
+      // be the material app. Otherwise this will be a prebuilt widget from
+      // amplify_authenticator to handle the current AuthenticatorStep
+      builder: (context, state, child) {
```
haverchuck, line 37: Could we get an example of the landingPage?

packages/amplify_authenticator/lib/src/models/authenticator_exception.dart
```diff
@@ -49,20 +49,20 @@ class AuthenticatorException implements Exception {
...
 
 /// {@template amplify_authenticator.exception_banner_location}
-/// The location on the screen to show error banners.
+/// The location on the step to show error banners.
 ///
 /// Defaults to `auto`.
 /// {@endtemplate}
 enum ExceptionBannerLocation {
-  /// Decide based off screen size which to show. Smaller devices including
+  /// Decide based off step size which to show. Smaller devices including
```
haverchuck, line 57: Find/replace error?

packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -229,11 +241,14 @@ class Authenticator extends StatefulWidget {
     this.onException,
     this.exceptionBannerLocation = ExceptionBannerLocation.auto,
     this.preferPrivateSession = false,
+    this.initialStep = AuthenticatorStep.signIn,
```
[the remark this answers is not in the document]
haverchuck: Yeah, it seems like maybe some steps should be disallowed, which gets us into a bit of complexity that is avoidable. Maybe we should remove this option until somone asks for it (although I still think the landing page is a good feature)?

packages/amplify_authenticator/lib/src/enums/authenticator_step.dart
```diff
@@ -0,0 +1,80 @@
...
+  authenticated,
+
+  /// The user is on the Landing Page step.
+  ///
+  /// This step can be used to create and initial landing page
+  /// that will be displayed to users prior to showing sign in / sign up.
+  ///
+  /// This step is only relevant for custom authenticator UIs.
+  /// The prebuilt authenticator uses the sign in view as the landing page.
+  landingPage,
```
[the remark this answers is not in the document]
haverchuck: Personally I think the landingPage is a good feature. @dnys1 Can you elaborate a bit on what you mean by 'handled by loading'?

Side note (doesn't need to be addressed now): I think it might be helpful to display the loading widget while API calls are completing. These are very fast right now in the default flow under good network conditions, but with slow network connections or when we introduce custom flows with lambda triggers these calls might take a bit longer.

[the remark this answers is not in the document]
dnys1: I worded that poorly. Maybe I don't fully understand the use case of it, but it seems like a half-baked feature. If it's to show a splash screen, that should be done for `loading` and if it's for onboarding then that should follow authentication anyway.

There are loading indicators already, though? Flashing a full loading screen for a couple seconds continuously seems like very poor UX.

packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -513,22 +536,27 @@ class _AuthenticatorState extends State<Authenticator> {
...
-              confirmResetPasswordForm: const ConfirmResetPasswordForm(),
-              signInForm: widget.signInForm,
-              signUpForm: widget.signUpForm,
-              confirmSignUpForm: ConfirmSignUpForm(),
-              confirmSignInMFAForm: ConfirmSignInMFAForm(),
-              verifyUserForm: VerifyUserForm(),
-              confirmVerifyUserForm: ConfirmVerifyUserForm(),
-              child: widget.child,
+          state: _viewModel,
+          child: InheritedAuthenticatorBuilder(
```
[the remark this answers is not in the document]
Jordan-Nelson: oof. I did not realize that. Will update.

packages/amplify_authenticator/lib/src/enums/authenticator_step.dart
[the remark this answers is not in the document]
Jordan-Nelson: The docs I added might be describing the intended use case poorly. I added an example [here](https://github.com/aws-amplify/amplify-flutter/pull/1281/files#r787951259). If that clears it up, I can improve the doc comments here.

packages/amplify_authenticator/example/lib/authenticator_with_custom_layout.dart
```diff
@@ -0,0 +1,217 @@
...
+        switch (state.currentStep) {
```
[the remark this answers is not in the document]
Jordan-Nelson: Yeah. The custom forms example sort of has one, although it isn't really showing the intended use of it. I will update that example.

packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -216,24 +228,84 @@ export 'src/widgets/form_field.dart'
...
+    this.confirmSignInNewPasswordForm,
     this.stringResolver = const AuthStringResolver(),
     required this.child,
     this.useAmplifyTheme = false,
     this.onException,
     this.exceptionBannerLocation = ExceptionBannerLocation.auto,
     this.preferPrivateSession = false,
+    this.initialStep = AuthenticatorStep.signIn,
+    this.authenticatorBuilder,
+    this.padding = const EdgeInsets.all(32),
```
Jordan-Nelson, line 295: This is the padding that is used for each screen. I exposed padding as a configurable property. For custom UIs that only customize some of the screens (Sign In, but not Reset Password) customers may use a different padding on their custom screens. There will then be inconsistencies between the prebuilt screens and the custom ones. This was the case in some of the examples I built.

I think this should default to `EdgeInsets.all(16)` or `EdgeInsets.symmetric(horizontal: 16, vertical: 32)` because I think 16 is a more common horizontal padding. However, 32 is what we currently use. Material doesn't have super clear guidance on this, but their docs on [responsive layouts under "breakpoints"](https://material.io/design/layout/responsive-layout-grid.html#breakpoints), it does indicate that extra small screens (phones) should have a margin of 16px.

I am leaving this change out for now, but @haverchuck @dnys1 if you both agree I can include it.

packages/amplify_authenticator/example/lib/authenticator_with_custom_layout.dart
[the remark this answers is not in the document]
Jordan-Nelson: I added an example. See `/authenticator_with_onboarding.dart`

packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -557,54 +726,103 @@ class _AuthenticatorBody extends StatelessWidget {
...
+  @override
+  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
+    super.debugFillProperties(properties);
+    properties.add(
+        ObjectFlagProperty<Widget Function(AuthState state, Widget child)>.has(
+            'builder', builder));
+  }
+}
+
+// The body of the authenticator
```
dnys1, line 757: Can we update this comment? Something like 

```dart
/// First child of the [MaterialApp] when using [Authenticator.builder]. 
/// 
/// All routes are wrapped with a Navigator which allows for separation between the
/// the user's navigation and the Authenticator's.
```

packages/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -557,54 +726,103 @@ class _AuthenticatorBody extends StatelessWidget {
...
+  }
+}
+
+/// A widget for wrapping portions of the application (typically routes)
+/// that are intended to be protected by authentication.
+class AuthenticatedView extends StatelessWidget {
+  const AuthenticatedView({
+    Key? key,
+    required this.child,
+  }) : super(key: key);
```
dnys1, line 802: So it shows on the ctor as well

```suggestion
/// {@template amplify_authenticator.authenticated_view}
/// A widget for wrapping portions of the application (typically routes)
/// that are intended to be protected by authentication.
/// {@endtemplate}
class AuthenticatedView extends StatelessWidget {
  /// {@macro amplify_authenticator.authenticated_view}
  const AuthenticatedView({
    Key? key,
    required this.child,
  }) : super(key: key);
```

dnys1 (approved):
Much easier to follow now. And thanks for adding comments throughout! If you're comfortable merging despite some flaky integ tests, I'm good with it

haverchuck (approved):

abasu0713:
Can't seem to use the simple builder function.. Getting errors:
`error: The property 'currentStep' can't be unconditionally accessed because the receiver can be 'null'. (unchecked_use_of_nullable_value at [data_dashboard] lib\main.dart:91)
error: Case expressions must be constant. (non_constant_case_expression at [data_dashboard] lib\main.dart:92)
error: Undefined name 'AuthenticatorStep'. (undefined_identifier at [data_dashboard] lib\main.dart:92)`
Updated to latest cognito and authenticator packages.. :(

abasu0713:
Sorry I am still a flutter newbie!

abasu0713:
Never mind me. I figured it out. Thanks though.

PR #1303, opened 2022-01-24
test(amplify_datastore): Cognito + Datastore Integ testing

haverchuck:
*Description of changes:*
Datastore/auth integ tests (non-exhaustive)


packages/amplify_datastore/example/integration_test/auth_test/private_blog_allow_read.dart
```diff
@@ -0,0 +1,220 @@
...
+    return random.nextInt(500).toString();
+  }
+
+  /// Creates a blog as [user1], saves it, and signs in [user2]
+  Future<void> _blogSetupAndTest(String blogName) async {
+    // Sign in with user1 and start the store
+    await user1.signInCognitoUser();
+
+    // Create blog as user1
+    await user1.createModels(Blog(name: blogName));
```
fjnoyp, line 47: Can this redundant code be refactored into a helper method that can be called from different auth_test s?

fjnoyp: In future, maybe we can investigate a concise way to represent unit/integration tests

Appears most tests share a lot of common setup/teardown and inner testing logic.  But long code file makes it harder to read/parse/maintain them all. 

It would be interesting if we could have our own inner symbols to represent maybe : 

```
blogSetupAndTest() async {
    runIntegrationTest( new List<IntegrationTestSteps>() {
SignIn(User1), CreateAndVerifyModel(User1, Blog(blogName)), Signout(User1), Signout(User2) 
}); 
}
```

Or maybe represent it as a txt file we read in and parse. 

Such a representation could make it easier to make sure we capture each permutation that might cause an error, and if well represented, allow for some rudimentary auto generation of integration tests.

fjnoyp (commented):
Looks good. 

1) Comment about refactoring unit/integration tests to be more readable. 

2) We should stop duplicating the auto generated models Blog/Comment/Post throughout our codebase

Issue #1296, opened 2022-01-22
amplify_authenticator: 0.1.0-rc.3 confirm phone number issue


egyangel:
**Describe the bug**
1- Wrong phone number prefix on the confirmation page
![image](https://user-images.githubusercontent.com/4232680/150643188-65b96dc8-273b-4496-8a94-e46c211d650b.png)

2- click on the confirm button dosn't do any thing, with or without the right confirmatin number. (No error message)

**To Reproduce**
1. Create new authentication based on phone_number
2. Signup for new account
3. enter the confirmation code recived via sms
4. click confirm button 

**Expected behavior**
Confirm button should either show an error message or confirm the registration number

**Screenshots**
amplify_flutter: ^0.3.1
  amplify_datastore: ^0.3.1
  amplify_core: ^0.3.1
  amplify_auth_cognito: ^0.3.1
  amplify_api: ^0.3.1
  amplify_authenticator: ^0.1.0-rc.3

**Platform**
Amplify Flutter current supports iOS and Android.  This issue is reproducible in (check all that apply):
[x] Android
[] iOS

<details>

  <summary>Dependencies (<code>
  amplify_flutter: ^0.3.1
  amplify_datastore: ^0.3.1
  amplify_core: ^0.3.1
  amplify_auth_cognito: ^0.3.1
  amplify_api: ^0.3.1
  amplify_authenticator: ^0.1.0-rc.3
  </code>)</summary>
  
  ```
  Paste the contents of your "pubspec.lock" file here
  ```

</details>

Jordan-Nelson:
Hello @egyangel - I was able to reproduce this. Thanks for bringing it to our attention. I will let you know when we have a fix for this.

Jordan-Nelson:
@egyangel - I just merged a fix for this. I will let you know when there is a new RC available.

Jordan-Nelson:
@egyangel this should be resolved now in `0.1.0.rc.4`. I am going to close this issue out, but please do not hesitate to open an another if have any other issues or feedback to provide. Thanks again for bringing this to our attention.

egyangel:
hello @Jordan-Nelson ,
I've just tested it and it works greate 

Many Thanks for your attention 👍

PR #1116, opened 2021-11-18
chore: refactor authenticator form fields

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Create `AuthenticatorFormFieldType` to replace all the form specific FormField enums
- Create 5 new classes that extend `AuthenticatorFormField` - `AuthenticatorTextFormField`, `AuthenticatorDateFormField`, `AuthenticatorUsernameFormField`, `AuthenticatorPhoneFormField`, `AuthenticatorRadioFormField`.
- Remove mixins in favor of the classes above
- Move logic from `SignInFormField`, `SignUpFormField`, `ConfirmSignInFormField`, etc. into `AuthenticatorFormField` and the new classes

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1312, opened 2022-01-26
fix(amplify_authenticator): fix country code bugs

Jordan-Nelson:
*Issue #, if available:* https://github.com/aws-amplify/amplify-flutter/issues/1296

*Description of changes:*
- Store country code state in AuthenticatorState (formally view model) to persist it between sign up and confirm sign up
- Update phone number in AuthenticatorState when the country code changes
- Add tests for non US numbers
- Clean up tests (remove legacy tests, rename a test file)


packages/amplify_authenticator/lib/src/mixins/authenticator_phone_field.dart
```diff
@@ -158,9 +157,7 @@ mixin AuthenticatorPhoneFieldMixin<FieldType,
     );
 
     if (selectedCountry != null) {
-      setState(() {
-        _selectedCountry = selectedCountry;
-      });
+      state.country = selectedCountry;
```
dnys1, line 160: Seems like `setState` would still be needed, no?

packages/amplify_authenticator/example/integration_test/utils/mock_data.dart
```diff
@@ -34,10 +34,54 @@ String generatePassword() =>
...
-String randomPhoneDigits() => '${random.nextInt(10)}${random.nextInt(10)}';
+String randomPhoneDigits(int count) {
+  String value = '';
+  for (var i = 0; i < count; i++) {
+    value += random.nextInt(10).toString();
+  }
+  return value;
+}
+
+class _PhoneNumber {
```
dnys1, line 67: Seems like this might be a useful class in Authenticator, instead of working with strings

dnys1 (approved):
Looks good, few minor Qs

haverchuck (approved):

packages/amplify_authenticator/example/integration_test/utils/mock_data.dart
Jordan-Nelson: Agreed. I am a little hesitant to do this right now though. I started to, but it got messy quickly. I think we need to do some (not all) of the refactoring work around AuthenticatorFormField I proposed in https://github.com/aws-amplify/amplify-flutter/pull/1116. This would make the form field code easier to understand. In particular, the current definition of AuthenticatorFormField I think can be simplified.

The current definition is: 

```dart
abstract class AuthenticatorFormField<FieldType, FieldValue,
        T extends AuthenticatorFormField<FieldType, FieldValue, T>>
    extends AuthenticatorComponent<T> {
}
```

IMO it is hard to follow what FieldType, FieldValue, and T are in this context. I am pretty sure we could simplify it to something like:

```dart
abstract class AuthenticatorFormField<FieldType>
    extends AuthenticatorComponent {
}
```

In the above definition, FieldType would be the data type for the data in the form field (String, PhoneNumber, etc.). In the previous refactor PR I put up, I simplified it further remove both `FieldType` and `FieldValue`. While that worked, it meant that every field was a String, which isn't what we want.

packages/amplify_authenticator/lib/src/mixins/authenticator_phone_field.dart
Jordan-Nelson: The country is now stored in AuthenticatorState and the setter for country calls `notifyListeners()`.

PR #1319, opened 2022-01-28
chore: updated skeleton for amplify_flutter & amplify_auth_cognito refactor

haverchuck:
*Description of changes:*
This introduces the **very basic** implementations for amplify_flutter and amplify_auth_cognito. 

Using the amplify_auth_cognito example app, you should be able to:

1. Run the application on Android, iOS, web and MacOS. 

2. Run through the addPlugin and configure flow. On Android and iOS this is actual functioning code. On web and MacOS we are calling the correct APIs but true configuration is not actually happening.

3. Call the 'signup' API. On Android and iOS this will fail because the request is malformed. On web and MacOS you will receive a dummy response from the Dart implementation (but nothing is happening on the backend).

**What is included in this PR**
 
1. Restructuring of our class hierarchies for auth. The amplify_auth_cognito package and it's accompanying interface package have been replaced with a 'amplify_auth_cognito_plugin' directory containing:
  **amplify_auth_cognito**: an app facing package which delegates API calls to the registered instance.
  **amplify_auth_cognito_dart**: a non-flutter dart package containing a mixin which implements the auth category interface, and an abstract class to which the mixin is applied. This enables us to use the mixin as an interface for the cognito interface (discussed below) and also to apply its implementation directly to web and desktop plugins (and eventually mobile).
  **amplify_auth_cognito_platform_interface**: a package which contains a platform interface for the Cognito plugin, and the method channel implementation for android and ios.
  **amplify_auth_cognito_android and _ios**: the federated packages for Android and iOS. Contains our method channel implementations from the current amplify_auth_cognito packages /android and /ios directories.
  **amplify_auth_cognito_web and _macos**: thin registrars using the amplify_auth_cognito_dart mixin.
  
2. Similar restructuring for amplify_flutter, sans the category interface.

**What is NOT in this PR**

1. Finalized code. This is a pass at updating our POC. it should not be considered complete in any way.
2. Well crafted implementation of amplify_flutter. I think this plugin may need to change once we have done more investigation of configuration, so I didn't invest much time into it.
3. Fully fleshed-out readmes, licenses and pubspec boilerplate and branding copy. I will updated these once we have consensus on this PR.
4. Tests.
5. Support for other categories. I have disabled non-auth categories within amplify-flutter for now. I can re-enable them once we have consensus on our architecture and can port all the category code to amplify_core.


dnys1:
Core no longer has a Flutter dependency but that also means all the iOS/Android code in the core package is unavailable to other plugins. That code should either be moved to amplify_flutter, or amplify_core should be federated and we add an amplify_common package for shared Flutter/Dart code.

haverchuck:
@dnys1 I like the idea of keeping amplify_core as a clean, non-plugin base package. It seems like moving this to amplify-flutter could work, assuming we avoid a circular dependency situation (which I think we would, since we would be importing this code directly in platform code and not via pubspec dependencies).

ragingsquirrel3 (approved):

Issue #1255, opened 2022-01-11
Flutter app fails to configure AWS Amplify (Null check operator used on a null value)


pflugs30:
**Describe the bug**
I ran into a problem while upgrading a production app from AWS Amplify Flutter 0.1.X to ^0.2.0.  The Flutter app fails to configure the Amplify packages without error.  I then created a brand new Flutter application, following the steps on [the Getting Started documentation](https://docs.amplify.aws/lib/project-setup/prereq/q/platform/flutter/) and [the Authentication page](https://docs.amplify.aws/lib/auth/getting-started/q/platform/flutter/).  I experienced the exact same error, despite building a project from scratch, including the AWS resources.

**To Reproduce**
1. Follow the steps on [the Getting Started documentation](https://docs.amplify.aws/lib/project-setup/prereq/q/platform/flutter/).
2. Add the auth plugin following [these steps](https://docs.amplify.aws/lib/auth/getting-started/q/platform/flutter/).
3. Add a test file module, such as this one:
```dart
import 'package:flutter_test/flutter_test.dart';
    import 'package:amplify_flutter/amplify.dart';
    import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
    import 'package:aws_amplify_flutter_app/amplifyconfiguration.dart';
    
    void main() {
      test('can configure Amplify', () async {
        AmplifyAuthCognito authPlugin = AmplifyAuthCognito();
        await Amplify.addPlugin(authPlugin);
    
        try {
          await Amplify.configure(amplifyconfig);
        } on AmplifyAlreadyConfiguredException {
          print("Tried to reconfigure Amplify; this can occur when your app restarts on Android.");
        }
      });
    }
```
4. Run the test using a command like: `flutter test test\my_test.dart`

**Actual Behavior**
When I run the test file above, I get the following output:
```
F:\dev\personal\aws_amplify_flutter_app>flutter test test\widget_test.dart
00:36 +0: can configure Amplify
Amplify plugin was not added
00:36 +0 -1: can configure Amplify [E]
  AmplifyException(message: Amplify plugin AmplifyAuthCognito was not added successfully., recoverySuggestion: We currently don't have a recovery suggestion for this exception., underlyingException: Null check operator used on a null value)
  package:amplify_flutter/amplify.dart 101:9  AmplifyClass.addPlugin

00:36 +0 -1: Some tests failed.
```

[...]

PR #1325, opened 2022-02-02
feat(amplify_flutter): allow customers to override AmplifyClass methods

Jordan-Nelson:
*Issue #, if available:* n/a

*Goal:*
- Allow Amplify Auth to be stubbed for use in the browser, so that a live Amplify Authenticator demo (with a stubbed backend) can be imbedded in the docs website.

*Issue description:*

`AmplifyClass` has a setter for the private `_instance` field. This allows a customer to pass in their own AmplifyClass implementation. However, the instance they pass in isn't used (except in one location, where a private method is invoked, and private methods can't be overridden from outside the lib anyway). This means that customers can technically create their own AmplifyClass implementation, but they can't override any of the methods, which defeats the whole point.

*Description of changes:*
- Move all implementations from AmplifyClass to MethodChannelAmplify.
  - _Note_: I think it is debatable if **all** of the methods belong in the method channel class. We could create a new class, which would  be the default implementation of AmplifyClass, and this class could call through to the method channel class when needed. The pattern of putting all of the implementations in the method channel class is consistent with the other plugins though. I think it is worth revisiting this possibly in the future, but for now we just move them all over.
- Replace all implementation with calls to the appropriate method on the instance

*Example:*

[Here is an example where AmplifyClass is stubbed](https://github.com/Jordan-Nelson/amplify_authenticator_example/blob/main/device_preview_example/lib/stubs/amplify_class_stub.dart) for the purpose of running a (mocked) demo of the Amplify Authenticator in the browser. The implementation here is similar to the implementation in the real class. The key difference is that the method channel is never invoked. This is important because the method channel will not be available in the browser.

A similar stub could be used to allow Amplify.configure() and Amplify.addPlugin() to be called from within a widget/unit tests. Right now those methods can only be called from within integ tests. We had [an issue](https://github.com/aws-amplify/amplify-flutter/issues/1255) opened in the past related to this.

This also would in theory allow a custom to write their own implementation for a platform we don't currently support, although I think that is an unlikely use case.

### DIFF
--- a/packages/amplify_flutter/lib/amplify_flutter.dart
+++ b/packages/amplify_flutter/lib/amplify_flutter.dart
@@ -37,6 +37,10 @@ export 'src/utils/json.dart';
 export 'src/utils/serializable.dart';
 
 /// Top level singleton Amplify object.
-final AmplifyClass Amplify = AmplifyClass.protected();
+AmplifyClass get Amplify => AmplifyClass.instance;
+
+set Amplify(AmplifyClass amplifyClass) {
+  AmplifyClass.instance = amplifyClass;
+}
 
 // ignore_for_file: non_constant_identifier_names
--- a/packages/amplify_flutter/lib/amplify_flutter.dart
+++ b/packages/amplify_flutter/lib/amplify_flutter.dart
@@ -19,6 +19,9 @@ import 'src/amplify_impl.dart';
 
 export 'package:amplify_core/amplify_core.dart';
 
+// amplify class
+export 'src/amplify_impl.dart';
+
 // Config
 export 'src/config/amplify_config.dart';
 export 'src/config/amplify_plugin_config.dart' hide UnknownPluginConfigFactory;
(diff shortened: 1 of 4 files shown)


packages/amplify_flutter/lib/amplify_flutter.dart
```diff
@@ -28,7 +31,6 @@ export 'src/config/api/api_config.dart';
 export 'src/config/auth/auth_config.dart';
 export 'src/config/config_map.dart';
 export 'src/config/storage/storage_config.dart';
-
 // Utilities
 export 'src/utils/equatable.dart';
 export 'src/utils/json.dart';
```
dnys1, line 36: I can't leave a comment on the line for some reason but this:

```dart
/// Top level singleton Amplify object.
final AmplifyClass Amplify = AmplifyClass.protected();
```

should become this:

```dart
/// Top level singleton Amplify object.
AmplifyClass get Amplify => AmplifyClass.instance;
```

dnys1 (approved):

packages/amplify_flutter/lib/amplify_flutter.dart
dnys1: Also maybe a setter as well?

Jordan-Nelson: Good call. Updated and added a setter.

haverchuck (approved):

PR #1333, opened 2022-02-03
fix(amplify_flutter): updates for latest flutter and dart versions

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*


example/android/app/src/main/AndroidManifest.xml
```diff
@@ -6,7 +6,7 @@
          additional functionality it is fine to subclass or reimplement
          FlutterApplication and put your custom class here. -->
     <application
-        android:name="io.flutter.app.FlutterApplication"
+        android:name="${applicationName}"
```
Jordan-Nelson, line 9: See https://github.com/flutter/flutter/wiki/Upgrading-pre-1.12-Android-projects for explanation of changes.

example/android/build.gradle
```diff
@@ -1,5 +1,5 @@
 buildscript {
-    ext.kotlin_version = '1.3.50'
+    ext.kotlin_version = '1.6.10'
```
Jordan-Nelson, line 2: `1.3.50` was causing issues with Flutter 2.10.0. Flutter 2.10.0 doesn't have release notes yet, but https://docs.flutter.dev/release/breaking-changes/kotlin-version shows that as of 2.9.0, `1.5.31` is required. `1.6.10` is the version that it defaults to when running `flutter create` with 2.10.0. I think bumping it to either would be fine.

This change is in a sample app. I also bumped this in the individual packages. My understanding is that it doesn't really matter what version we have in the package, because each flutter app is required to have this defined as well.

packages/amplify_authenticator/lib/src/theme/amplify_theme.dart
```diff
@@ -262,103 +262,94 @@ class AmplifyColors {
...
+  static const _h6 = TextStyle(
     fontFamily: fontFamily,
     package: fontPackage,
     fontWeight: FontWeight.bold,
     fontSize: 24,
     height: 32 / 24,
   );
 
-  static const bodyLarge = TextStyle(
+  static const _bodyLarge = TextStyle(
```
Jordan-Nelson, line 328: A property on TextTheme was renamed to bodyLarge, which made this conflict. I made all these private just on the off chance this happens again.

packages/amplify_authenticator/test/ui/tab_view_test.dart
```diff
@@ -143,6 +143,7 @@ void main() {
   group('AuthenticatorTabView', () {
     setUp(() {
       binding.window.clearPlatformBrightnessTestValue();
+      binding.window.physicalSizeTestValue = const Size(2400, 2400);
```
Jordan-Nelson, line 146: This had ti be adjusted for the Amplify theme tests because the authenticator is now bigger than the screenshot. I updated for all the others as well.

dnys1 (approved):
You rock, thanks!

example/android/build.gradle
haverchuck: When you say 'each flutter app is required to have this as well', you mean each app is required to have this set to at least `1.5.31` in order to run Flutter `2.10.0`?

Jordan-Nelson: I'll test this out to be sure, but here is my understanding - If someone created an app with a version of flutter older than 2.10.0 (lets say 2.0.0) and is still using that version, they may have `ext.kotlin_version` set to `1.3.50` in their build.gradle file. If they upgrade to the latest version of amplify_flutter (with this change) that version would override the version set in the build.gradle file of the amplify_flutter packages. Basically, the version set in our packages doesn't mean anything for the customer.

I'll run some tests to confirm my understanding.

haverchuck: Ok thanks - I tried testing this change out as well and the build failed, but I think it was an unrelated issue. Didn't have time to dig into it too much.

Jordan-Nelson: After testing, I think my understanding was incorrect. We should leave it as 1.3.50 in the packages. We can change it to 1.6.10 in the sample apps though. I will push an update shortly. See testing details below.

I used the following testing steps:
- update `kotlin_version` to the values being tested
- `fvm use <version>`
- `cd packages/<package>/example
- `fvm flutter clean`
- `fvm flutter pub global run melos clean && fvm flutter pub global run melos bootstrap`
- `fvm flutter run -d <android-device-id>`

I tested the following scenarios with kotlin_version 1.6.10 in the packages:
- Flutter 2.10.0, kotlin_version 1.6.10 in example apps
  - Expected scenario for users on latest flutter version
  - Build Passed
- Flutter 2.8.1, kotlin_version 1.3.50 in example apps
  - Expected scenario for users on flutter 2.8.1
  - **Build Failed**
- Flutter 2.8.1, kotlin_version 1.6.10 in example apps
  - Unlikely scenario, unless we want to force users to do this.
  - Build Passed

If we change kotlin_version to 1.6.10 in the packages, while customers could still use older flutter versions, it will force customers to upgrade to kotlin_version 1.6.10. Probably not something we want to do.

I then tested the following scenarios with kotlin_version 1.3.50 in the packages:
- Flutter 2.10.0, kotlin_version 1.6.10 in example apps
  - Expected scenario for users on latest flutter version
  - Build Passed
- Flutter 2.8.1, kotlin_version 1.3.50 in example apps
  - Expected scenario for users on flutter 2.8.1

Jordan-Nelson: Updated. The only changes to ext.kotlin_version are in example apps now.

.circleci/config.yml
```diff
@@ -37,7 +37,7 @@ commands:
       - run:
           name: Install and set up melos
           command: |
-            flutter pub global activate melos 0.4.5
+            flutter pub global activate melos 1.2.1
```
Jordan-Nelson, line 40:
```suggestion
            flutter pub global activate melos 0.4.5
```

haverchuck (approved):

PR #1336, opened 2022-02-04
feat(datastore): Add QueryPredicate to Save/Delete

fjnoyp:
*Issue #, if available:*

*Description of changes:*

Add QueryPredicate support for Save/Delete

### DIFF
--- a/packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
+++ b/packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
@@ -41,6 +41,7 @@ import com.amplifyframework.core.model.ModelSchema
 import com.amplifyframework.core.model.SerializedCustomType
 import com.amplifyframework.core.model.SerializedModel
 import com.amplifyframework.core.model.query.QueryOptions
+import com.amplifyframework.core.model.query.predicate.QueryPredicate
 import com.amplifyframework.core.model.query.predicate.QueryPredicates
 import com.amplifyframework.datastore.AWSDataStorePlugin
 import com.amplifyframework.datastore.DataStoreConfiguration
@@ -278,6 +279,7 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
     @VisibleForTesting
     fun onDelete(flutterResult: Result, request: Map<String, Any>) {
         val modelName: String
+        val queryPredicate: QueryPredicate
         val serializedModelData: Map<String, Any?>
         val schema: ModelSchema
 
@@ -286,6 +288,9 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
             schema = modelProvider.modelSchemas()[modelName]!!
             serializedModelData =
                 deserializeNestedModel(request["serializedModel"].safeCastToMap()!!, schema)
+
+            queryPredicate = QueryPredicateBuilder.fromSerializedMap(
+            request["queryPredicate"].safeCastToMap()) ?: QueryPredicates.all()
         } catch (e: Exception) {
             uiThreadHandler.post {
                 postExceptionToFlutterChannel(
@@ -305,6 +310,7 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
 
         plugin.delete(
             instance,
+            queryPredicate,
             {
                 LOG.info("Deleted item: " + it.item().toString())
                 uiThreadHandler.post { flutterResult.success(null) }
@@ -328,6 +334,7 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
     @VisibleForTesting
     fun onSave(flutterResult: Result, request: Map<String, Any>) {
         val modelName: String
+        val queryPredicate: QueryPredicate
         val serializedModelData: Map<String, Any?>
         val schema: ModelSchema
 
@@ -336,6 +343,9 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
             schema = modelProvider.modelSchemas()[modelName]!!
             serializedModelData =
                 deserializeNestedModel(request["serializedModel"].safeCastToMap()!!, schema)
+
+            queryPredicate = QueryPredicateBuilder.fromSerializedMap(
+                    request["queryPredicate"].safeCastToMap()) ?: QueryPredicates.all()
         } catch (e: Exception) {
             uiThreadHandler.post {
                 postExceptionToFlutterChannel(
@@ -353,11 +363,9 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
             .modelSchema(schema)
             .build()
 
-        val predicate = QueryPredicates.all()
-
         plugin.save(
             serializedModel,
-            predicate,
+            queryPredicate,
             {
                 LOG.info("Saved item: " + it.item().toString())
                 uiThreadHandler.post { flutterResult.success(null) }
--- a/packages/amplify_datastore/example/integration_test/delete_test.dart
+++ b/packages/amplify_datastore/example/integration_test/delete_test.dart
@@ -42,5 +42,17 @@ void main() {
       expect(blogs.length, 1);
       expect(blogs[0].name, originalBlogName);
     });
+
+    testWidgets('predicate should not prevent delete for matching model',
+        (WidgetTester tester) async {
+      const originalBlogName = 'matching blog';
+      Blog testBlog = Blog(name: originalBlogName);
+      await Amplify.DataStore.save(testBlog);
+
+      await Amplify.DataStore.delete(testBlog,
+          where: Blog.NAME.contains("matching"));
+      var blogs = await Amplify.DataStore.query(Blog.classType);
+      expect(blogs.length, 0);
+    });
   });
 }
--- a/packages/amplify_datastore/example/integration_test/delete_test.dart
+++ b/packages/amplify_datastore/example/integration_test/delete_test.dart
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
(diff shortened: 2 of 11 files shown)


Jordan-Nelson (commented):
I don't understand the use case for this API. I know iOS and Android have it, but I am wondering if their APIs are different in some way that makes this more useful.

Is there an example use case that this would satisfy?

Jordan-Nelson:
After looking at the docs for Android, I think the point is to be able to delete based on a predicate without having to re-query the data, which seems reasonable. For example, the rating of a Post object may have changed since it was last queried, and it should only be deleted if some condition is true. Today this would require re-querying that post to check the condition before calling delete.

I will note that the example in the Android docs is not great because it involves re-querying the data anyway. We should make sure to get a better example in the docs.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -286,6 +288,9 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
             schema = modelProvider.modelSchemas()[modelName]!!
             serializedModelData =
                 deserializeNestedModel(request["serializedModel"].safeCastToMap()!!, schema)
+
+            queryPredicate = QueryPredicateBuilder.fromSerializedMap(
+            request["queryPredicate"].safeCastToMap()) ?: QueryPredicates.all()
```
Jordan-Nelson, line 293: *Question:* This could also default to null, correct?

I think it is fine how it is. Just making sure I understand the behavior?

packages/amplify_datastore/example/integration_test/delete_test.dart
```diff
@@ -0,0 +1,46 @@
...
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  group('delete', () {
+    setUp(() async {
+      await configureDataStore();
+      // clear data before each test
+      await clearDataStore();
+    });
+
+    testWidgets('predicate should prevent delete to non matching model',
```
Jordan-Nelson, line 33: *nit:* Might be nice to add a similar test where the model doesn't match the predicate?

packages/amplify_datastore/example/integration_test/save_test.dart
```diff
@@ -62,5 +62,21 @@ void main() {
         expect(updatedBlogs[0].name, updatedBlogName);
       },
     );
+
+    testWidgets('predicate should prevent save to non matching model',
```
Jordan-Nelson, line 66: *nit:* Same as above. Might be nice to add a similar test where the model doesn't match the predicate.

packages/amplify_datastore/lib/method_channel_datastore.dart
```diff
@@ -151,11 +151,12 @@ class AmplifyDataStoreMethodChannel extends AmplifyDataStore {
   }
 
   @override
-  Future<void> delete<T extends Model>(T model) async {
+  Future<void> delete<T extends Model>(T model, {QueryPredicate? where}) async {
     try {
       await _setUpObserveIfNeeded();
       await _channel.invokeMethod('delete', <String, dynamic>{
         'modelName': model.getInstanceType().modelName(),
+        'queryPredicate': where?.serializeAsMap(),
```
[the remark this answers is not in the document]
Jordan-Nelson: +1

Jordan-Nelson (commented):
Looks pretty good! Mostly small/nit comments.

packages/amplify_datastore/lib/method_channel_datastore.dart
```diff
@@ -164,11 +165,12 @@ class AmplifyDataStoreMethodChannel extends AmplifyDataStore {
   }
 
   @override
-  Future<void> save<T extends Model>(T model) async {
+  Future<void> save<T extends Model>(T model, {QueryPredicate? where}) async {
     try {
       await _setUpObserveIfNeeded();
       var methodChannelSaveInput = <String, dynamic>{
         'modelName': model.getInstanceType().modelName(),
+        'queryPredicate': where?.serializeAsMap(),
```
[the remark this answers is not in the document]
fjnoyp: Sounds good, updated

packages/amplify_datastore/lib/method_channel_datastore.dart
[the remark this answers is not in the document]
fjnoyp: Sounds good, updated

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
```diff
@@ -259,6 +259,17 @@ public class SwiftAmplifyDataStorePlugin: NSObject, FlutterPlugin {
         }
     }
     
+    // Initial Save fails for QueryPredicate.all, so we must pass nil instead
```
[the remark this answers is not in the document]
fjnoyp: The root cause is initial save cannot be called with a query predicate.  In Amplify-iOS if you try to make an initial save with any queryPredicate it will fail. 

For Amplify-Android initial save will fail for any queryPredicate except for .all

packages/amplify_datastore/example/integration_test/delete_test.dart
fjnoyp: The test is for a non matching predicate.  I can add a test that a matching model to predicate succeeds.

packages/amplify_datastore/lib/method_channel_datastore.dart
```diff
@@ -151,26 +151,31 @@ class AmplifyDataStoreMethodChannel extends AmplifyDataStore {
   }
 
   @override
-  Future<void> delete<T extends Model>(T model) async {
+  Future<void> delete<T extends Model>(T model, {QueryPredicate? where}) async {
     try {
       await _setUpObserveIfNeeded();
-      await _channel.invokeMethod('delete', <String, dynamic>{
+      var methodChannelDeleteInput = <String, dynamic>{
```
HuiSF, line 157:
```suggestion
      var methodChannelDeleteInput = <String, dynamic> {
```

packages/amplify_datastore/lib/method_channel_datastore.dart
```diff
@@ -151,26 +151,31 @@ class AmplifyDataStoreMethodChannel extends AmplifyDataStore {
...
         'modelName': model.getInstanceType().modelName(),
+        'queryPredicate': where?.serializeAsMap(),
```
HuiSF, line 159:
```suggestion
        if (where != null) 'queryPredicate': where.serializeAsMap(),
```

packages/amplify_datastore/lib/method_channel_datastore.dart
```diff
@@ -151,26 +151,31 @@ class AmplifyDataStoreMethodChannel extends AmplifyDataStore {
...
         'serializedModel': model.toJson(),
-      });
+      };
+      methodChannelDeleteInput.removeWhere((_, v) => v == null);
```
HuiSF, line 162:
```suggestion
```

packages/amplify_datastore/lib/method_channel_datastore.dart
```diff
@@ -151,26 +151,31 @@ class AmplifyDataStoreMethodChannel extends AmplifyDataStore {
...
   }
 
   @override
-  Future<void> save<T extends Model>(T model) async {
+  Future<void> save<T extends Model>(T model, {QueryPredicate? where}) async {
     try {
       await _setUpObserveIfNeeded();
       var methodChannelSaveInput = <String, dynamic>{
         'modelName': model.getInstanceType().modelName(),
+        'queryPredicate': where?.serializeAsMap(),
```
HuiSF, line 175:
```suggestion
       if (where != null) 'queryPredicate': where.serializeAsMap(),
```

packages/amplify_datastore/lib/method_channel_datastore.dart
```diff
@@ -151,26 +151,31 @@ class AmplifyDataStoreMethodChannel extends AmplifyDataStore {
...
         'serializedModel': model.toJson(),
       };
+      methodChannelSaveInput.removeWhere((_, v) => v == null);
```
HuiSF, line 178:
```suggestion
```

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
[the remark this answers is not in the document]
HuiSF: Thanks!

If we want to merge this iOS specific trick in, could we add a comment to link to the related amplify-ios? So we can do a follow up to clean up when the underlying iOS issue gets fixed.

HuiSF (commented):
Thanks! A few minor suggestions :)

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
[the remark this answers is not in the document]
fjnoyp: Sure I can add a small comment so we remember to fix this.

HuiSF (approved):
LGTM✨

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
Jordan-Nelson: I'm curious if it *could* default to null and result in the same behavior, not if it *has to* be null. As long as it *can* default to null, you could make it the same on both platforms. That would be my preference, but this works as well.

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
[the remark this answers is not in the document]
Jordan-Nelson: Is there an issue on the ios repo? If so, it would be nice to include it in the comment. That way in 3 years from now when no one has context on this anymore, anyone can just open a link, see the issue was resolved (hopefully), and know that it is safe to update the code 🙂

Jordan-Nelson (approved):

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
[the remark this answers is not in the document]
fjnoyp: Created an issue and linked to it

PR #1339, opened 2022-02-05
fix(datastore): fix observe may receive duplicate events in Android

HuiSF:
*Issue #, if available:*

*Description of changes:*

- Add a flag to prevent multiple closure registration to amplify-android observe API
- Place `flutterResult` call to more appropriate spots
-----


dnys1 (commented):
- Does setUpObserve ever need to be called more than once? Should we just reject all subsequent calls?
- This code is not thread-safe. Do we know for sure observe will only ever be called from the main thread or should we add atomics to avoid a race condition?
- If there is an error besides "Failed to start DataStore" on the first setup call, DataStore will never be set up since all other calls will be ignored. Is that the correct behavior? Also, this approach of checking strings is very brittle.
- The PR changes what's sent back with the result (`nil` -> `true` in iOS, and `true`/`false` in Android), but there is no Dart code in the PR. Are these values handled appropriately already? Does a return of `false` mean something from an error?

HuiSF:
> * Does setUpObserve ever need to be called more than once? Should we just reject all subsequent calls?

`setUpObserve` is called in each DataStore API to ensure the event channel to be set up regardless the order of APIs that are being called by a developer. The actual set up logic on native side is executed only once, and subsequent calls will be **ignored** (by invoke `flutterResult.success()`)

> * This code is not thread-safe. Do we know for sure observe will only ever be called from the main thread or should we add atomics to avoid a race condition?

The method channel call happens on main thread, and the platform implementation didn't use `TaskQueue` should it should execute on main thread as well. Do you have any examples that we can invoke observe on a different thread?

> * If there is an error besides "Failed to start DataStore" on the first setup call, DataStore will never be set up since all other calls will be ignored. Is that the correct behavior? Also, this approach of checking strings is very brittle.

I think there is a misunderstanding -
`setUpObserve` is for creating a bridge forwarding native `observe` events to Flutter event channel. As long as the native `observe` call is executed (where the callback closures are registered), the set up is completed and time to return the method channel call.

"Failed to start DataStore" doesn't cancel the set up, I picked it out here, in order to find a good reason to do `flutterResult.success(false)`. Even when "Failed to start DataStore" happened, and later DataStore restarted, the observe channel should still work.

However, I think calling `flutterResult.success()` with a boolean is confusing... The dart implementation doesn't expect a specific value, as above described behavior. So we could get rid of `flutterResult.success(false)` and invoke `flutterResult.success()` when the callback closures are registered, and further we won't have the brittle string check on errors.

> * The PR changes what's sent back with the result (`nil` -> `true` in iOS, and `true`/`false` in Android), but there is no Dart code in the PR. Are these values handled appropriately already? Does a return of `false` mean something from an error?

As I described above the value doesn't matter, we could unify both platform implementation.

HuiSF:
I've pushed a new commit (https://github.com/aws-amplify/amplify-flutter/pull/1339/commits/fdf01f67c120891d9be9628a045f70c236b16759) to use `AtomicBoolean` for setting and checking `isSettingUpObserve` to address the concern "This code is not thread-safe".

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -395,17 +397,19 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
     }
 
     fun onSetUpObserve(flutterResult: Result) {
-        if (this::observeCancelable.isInitialized) {
+        if (this::observeCancelable.isInitialized || isSettingUpObserve.get()) {
```
dnys1, line 400: Use `getAndSet(true)` here

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -419,15 +423,19 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
                 }
             },
             { failure: DataStoreException ->
-                LOG.error("Received an error", failure)
-                dataStoreObserveEventStreamHandler.error(
-                    "DataStoreException",
-                    createSerializedError(failure)
-                )
+                if (failure.message?.contains("Failed to start DataStore", true) == true) {
+                    isSettingUpObserve.set(false);
```
dnys1, line 427: If the `else` clause is triggered, `isSettingUpObserve` is never reset

packages/amplify_datastore/android/src/test/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePluginTest.kt
```diff
@@ -646,7 +694,6 @@ class AmplifyDataStorePluginTest {
 
         flutterPlugin.onSetUpObserve(mockResult)
 
-        verify(mockResult, times(1)).success(true)
```
dnys1, line 649: Can we update these with expected result? For example, `verify(mockResult, times(0))` or something

dnys1 (approved):

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
dnys1: And remove the `set(true)` below this

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -68,6 +69,7 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
     private val dataStoreHubEventStreamHandler: DataStoreHubEventStreamHandler
     private val uiThreadHandler = Handler(Looper.getMainLooper())
     private val LOG = Amplify.Logging.forNamespace("amplify:flutter:datastore")
+    private var isSettingUpObserve = AtomicBoolean(false);
```
cshfang, line 72: Nit: I think `AtomicBoolean()` would initially be `false` anyway

packages/amplify_datastore/example/ios/unit_tests/DataStorePluginUnitTests.swift
```diff
@@ -230,7 +230,7 @@ class DataStorePluginUnitTests: XCTestCase {
         pluginUnderTest = SwiftAmplifyDataStorePlugin(bridge: dataStoreBridge, modelSchemaRegistry: modelSchemaRegistry, customTypeSchemasRegistry: customTypeSchemaRegistry, dataStoreObserveEventStreamHandler: streamHandler)
 
         pluginUnderTest.onSetUpObserve(flutterResult: { result in
-            XCTAssertNil(result)
+            XCTAssertEqual(result as! Bool, true)
```
cshfang, line 233: Nit: Can we just use `XCTAssertTrue` for this and the others below?

packages/amplify_datastore/example/ios/unit_tests/DataStorePluginUnitTests.swift
```diff
@@ -321,6 +321,26 @@ class DataStorePluginUnitTests: XCTestCase {
...
+
+        class MockStreamHandler: DataStoreObserveEventStreamHandler {}
+
+        let dataStoreBridge: MockDataStoreBridge = MockDataStoreBridge()
+        let streamHandler: MockStreamHandler = MockStreamHandler()
+
+        pluginUnderTest = SwiftAmplifyDataStorePlugin(bridge: dataStoreBridge, modelSchemaRegistry: modelSchemaRegistry, customTypeSchemasRegistry: customTypeSchemaRegistry, dataStoreObserveEventStreamHandler: streamHandler)
+
+        pluginUnderTest.onSetUpObserve(flutterResult: { result in
+            XCTAssertEqual(result as! Bool, false)
```
cshfang, line 340: Nit: Likewise but with `XCTAssertFalse`

cshfang (approved):
Really only had some minor comments. Lgtm otherwise

cshfang (approved):

PR #1351, opened 2022-02-10
chore(datastore): Enable cloud sync in integration tests

HuiSF:
*Issue #, if available:*

*Description of changes:*

- Forward native `subscritpionDataProcessed` (Android) and `syncReceived` (iOS) events to Dart DataStore hub event stream, these two events are named as `subscriptionDataProcessed` in dart implementation
- Updated test suites to run with cloud sync
- Split 23 testing models into smaller groups, and run `flutter test` separately to avoid long subscription and initial sync time which made tests run unstable in Android. 

NOTE: Android integration tests stability depends on PR #1339 ...
NOTE2: Cloud sync is enabled for model relationship tests and basic model operation within this PR. More tests with cloud sync will be add separately.


dnys1 (commented):
There is a lot of duplicated code in the tests - is there anyway to DRY this up a bit?

packages/amplify_datastore/example/integration_test/main_test.dart
```diff
@@ -16,10 +16,8 @@
 import 'package:integration_test/integration_test.dart';
 import 'package:flutter_test/flutter_test.dart';
 
-import 'save_test.dart' as save_tests;
 import 'query_test/query_test.dart' as query_test;
 import 'model_type_test.dart' as model_type_tests;
-import 'relationship_type_test.dart' as relationship_type_tests;
```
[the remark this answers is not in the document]
HuiSF: Hum meant to make it optional so can run in either mode, will modify the shell script a bit to make configurable.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterSerializedModel.kt
```diff
@@ -51,34 +51,38 @@ data class FlutterSerializedModel(val serializedModel: SerializedModel) {
         )
 
         return serializedData.mapValues {
-            val field = modelSchema.fields[it.key]!!
```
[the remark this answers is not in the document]
HuiSF: amplify-android sends nested models in `ModelWithMetadata` with meta fields (`_deleted`, `_version` and `_lastChangedAt`) merged into the model, and their values can be null, which may crash on `!!`.

build-support/integ_test_android.sh
```diff
@@ -0,0 +1,28 @@
...
+
+TARGET=integration_test/main_test.dart
+if [ ! -e $TARGET ]; then
+    echo "$TARGET file not found" >&2
+    exit
+fi
+
+flutter test \
+    --no-pub \
+    -d sdk \
```
ragingsquirrel3, line 16: question: Can the device argument be an optional argument passed to the script, with "sdk" as the default value? For some reason, in CI, "sdk" does not resolve and I've been passing "emulator-5554" instead, but I think it would help if this script accepted an argument for this.

packages/amplify_datastore/example/integration_test/utils/setup_utils.dart
```diff
@@ -34,6 +56,45 @@ Future<void> configureDataStore() async {
...
+    hubSubscription = Amplify.Hub.listen([HubChannel.DataStore], (event) {
+      if (event.eventName == 'ready') {
+        print(
+            '🎉🎉🎉DataStore is ready to start running test suites with API sync enabled.🎉🎉🎉');
+        hubSubscription.cancel();
+        _completer.complete();
+      }
+    });
+    startTimeout();
+    await Amplify.DataStore.start();
```
ragingsquirrel3, line 81: Could this be changed to `await Amplify.DataStore.start().timeout(Duration(minutes: 2));` and remove the previous line and timeout util?

build-support/integ_test_android.sh
```diff
@@ -0,0 +1,28 @@
...
+TEST_ENTRIES=`ls integration_test/separate_integration_tests/*.dart`
+for ENTRY in $TEST_ENTRIES
+do
+    echo "Run $ENTRY WITH API Sync"
+    flutter test \
+        --no-pub \
+        --dart-define ENABLE_CLOUD_SYNC=true \
+        -d sdk \
+        $ENTRY
+done
```
ragingsquirrel3, line 28: request: Could you please make this script such that it will have a non-zero exit status if any test command fails? I just tried forcing the first test suite to fail and and forcing the rest to pass, and even though output says "some tests failed" from failed test suite, melos thinks it passed, I think bc exit status of this script is still 0 if last suite it runs passed.

A good way to do this is to keep a list/array of failed tests, appending to it every time the `flutter test` command fails. Then, make the exit status the length of the list. no fails = status 0. You can see an example of that in test_all_plugins.sh

Same request for ios script.

packages/amplify_datastore/example/integration_test/separate_integration_tests/basic_model_operation_test.dart
```diff
@@ -0,0 +1,128 @@
...
+        var testBlog = (await Amplify.DataStore.query(Blog.classType))[0];
+        // update model
+        var updatedTestBlog = testBlog.copyWith(name: "updated test blog");
+
+        if (enableCloudSync) {
+          var eventGetter = getExpectedSubscriptionDataProcessedEvent(
+            eventMatcher: (event) =>
+                (event.element.model as Blog).id == updatedTestBlog.id &&
+                // updated model at this step should have version: 2
+                event.element.version == 2,
```
ragingsquirrel3, line 80: question: Is it possible/desirable to test that it got the new name in addition to version bump?

packages/amplify_datastore/example/integration_test/separate_integration_tests/basic_model_operation_test.dart
```diff
@@ -0,0 +1,128 @@
...
+      var blogs = await Amplify.DataStore.query(Blog.classType);
+      expect(blogs.length, 1);
+      expect(blogs.contains(testBlog), isTrue);
+    });
+
+    testWidgets(
+      'should update existing model ${enableCloudSync ? 'and sync to cloud' : ''}',
+      (WidgetTester tester) async {
+        // get previously saved model
+        var testBlog = (await Amplify.DataStore.query(Blog.classType))[0];
```
ragingsquirrel3, line 71: question: Could there be some kind of race condition/error getting just the first blog from DB/API if some other client running the test? For API integ tests, uses a utility to create instances for each test. Maybe that's too much here, just asking. Same question for other tests.

While race condition might seem unlikely, always consider the CI will run android/ios in parallel.

packages/amplify_datastore/example/integration_test/separate_integration_tests/basic_model_operation_test.dart
```diff
@@ -0,0 +1,128 @@
...
+
+          await Amplify.DataStore.delete(testBlog);
+
+          var event = await eventGetter;
+          expect(event.element.deleted, isTrue);
+        } else {
+          await Amplify.DataStore.delete(testBlog);
+        }
+
+        var blogs = await Amplify.DataStore.query(Blog.classType);
```
ragingsquirrel3, line 121: question: Would it be better to try to query for the one you deleted in case race condition makes the backend not empty?

packages/amplify_datastore/example/integration_test/separate_integration_tests/has_one_explicit_child_test.dart
```diff
@@ -0,0 +1,200 @@
...
+      await configureDataStore(
+          enableCloudSync: enableCloudSync,
+          modelProvider: ModelProvider.instance);
+
+      childEvent = Amplify.DataStore.observe(HasOneChild.classType).first;
+      parentEvent = Amplify.DataStore.observe(HasOneParent.classType).first;
+    });
+
+    testWidgets('precondition', (WidgetTester tester) async {
+      var queriedChildren =
```
ragingsquirrel3, line 60: Would it be possible to have:
a) utilities for testing commonly repeated things like "precondition" "save child" "save parent" etc... as much of the code seems copy/pasted with small variations
b) a high-level utility, like `testCommonRelationshipOperations({bool saveChild = true, bool saveParent = true, ....})` that each of these tests could call, then just provide false for the operations they don't need to test

Higher-level goal to save some electricity/repetition here and make tests easier to maintain.

packages/amplify_datastore/example/integration_test/separate_integration_tests/multi_relationship_test.dart
```diff
@@ -0,0 +1,443 @@
...
+      }
+      var queriedRegistrations =
+          await Amplify.DataStore.query(MultiRelatedRegistration.classType);
+      expect(queriedRegistrations, isNotEmpty);
+    });
+
+    testWidgets('query meetings', (WidgetTester tester) async {
+      var queriedMeetings =
+          await Amplify.DataStore.query(MultiRelatedMeeting.classType);
+      for (var meeting in queriedMeetings) {
```
ragingsquirrel3, line 194: nit https://api.flutter.dev/flutter/package-matcher_matcher/containsAll.html, same for similar tests

packages/amplify_datastore/example/integration_test/utils/setup_utils.dart
```diff
@@ -34,6 +56,45 @@ Future<void> configureDataStore() async {
...
+  final Completer _completer = Completer();
+  late StreamSubscription hubSubscription;
+
+  Future<void> startDataStore() async {
+    hubSubscription = Amplify.Hub.listen([HubChannel.DataStore], (event) {
+      if (event.eventName == 'ready') {
+        print(
+            '🎉🎉🎉DataStore is ready to start running test suites with API sync enabled.🎉🎉🎉');
+        hubSubscription.cancel();
+        _completer.complete();
```
ragingsquirrel3, line 77: Is this expected to be done before/after `Amplify.Datastore.start()`? Appears should be after? It's a little confusing to me, appears that we expect to start listening to hub, start datastore, then hub completes later.

packages/amplify_datastore/example/integration_test/utils/setup_utils.dart
```diff
@@ -34,6 +56,45 @@ Future<void> configureDataStore() async {
 ///
 /// see: https:///github.com/aws-amplify/amplify-android/issues/1464
 Future<void> clearDataStore() async {
-  await Future.delayed(Duration(milliseconds: 100));
+  await Future.delayed(Duration(seconds: 2));
```
ragingsquirrel3, line 59: nit: Does it make sense to make this delay a constant for the whole suite so it can be changed in one place? I believe the non-cloud tests do something similar IIRC.

packages/amplify_datastore/example/integration_test/utils/wait_for_expected_event_from_hub.dart
```diff
@@ -0,0 +1,55 @@
...
+  Future<T> start() {
+    hubSubscription = Amplify.Hub.listen([HubChannel.DataStore], (event) {
+      if (event.eventName == this.eventName) {
+        if (this.eventMatcher(event.payload)) {
+          hubSubscription.cancel();
+          _completer.complete(event.payload as T);
+        }
+      }
+    });
+    startTimeout();
```
ragingsquirrel3, line 28: nit: Same comment as in setup_utils.dart, but can we get rid of this utility and just call `.timeout` on the future in line 29?

packages/amplify_datastore/lib/amplify_datastore_stream_controller.dart
```diff
@@ -43,58 +43,73 @@ StreamController<DataStoreHubEvent> _controller =
 _onListen() {
   if (eventStream == null) {
     eventStream = channel.receiveBroadcastStream(1).listen((event) {
-      switch (event["eventName"]) {
-        case "ready":
+      switch (event['eventName']) {
```
ragingsquirrel3, line 46: optional nit: I know it's already like this and a little tangential to PR, but can 'eventName' be made a constant? Ideally, lots of serialization keys would be in a constant somewhere,b ut this one is used several times right here, so could just be `const _eventName = 'eventName'` at top of file.

packages/amplify_datastore_plugin_interface/lib/src/types/hub/HubEventElement.dart
```diff
@@ -19,9 +19,9 @@ part 'HubEventElementWithMetadata.dart';
 
 /// The model associated with a DataStore `outboxMutationEnqueued` or
 /// `outboxMutationProcessed` Hub event.
-class HubEventElement {
+class HubEventElement<M extends Model> {
```
ragingsquirrel3, line 22: nice!

ragingsquirrel3 (changes requested):
Nice job with this integ test change. It seems pretty stable for me running locally. There might be some challenges with this in CI, but not in scope for this PR IMO. A few high-level callouts:
1) Do we want the subscription events to be separate PR? If not, please reflect both changes in commit message when merged to it's there for anyone who is debugging/bisecting commits.
2) Like Dillon said, would be nice to DRY up the testing logic, left a suggestion for that, but feel free to chat offline more.
3) Test script should return non-zero status if anything fails.
4) Mostly nits after that.

HuiSF:
> 1. Do we want the subscription events to be separate PR? If not, please reflect both changes in commit message when merged to it's there for anyone who is debugging/bisecting commits.

Could to a merge instead of squash merge on this PR (I'll squash comment resolving commits before merge).

packages/amplify_datastore/example/integration_test/utils/setup_utils.dart
HuiSF: The timeout is determined by how long it takes to receive the `ready` event from DataStore hub.
Using `timeout` to alter `Amplify.DataStore.start()` returned future is inaccurate I think...

Thinking of this point, `Amplify.DataStore.start()` is not need to be `await`'ed I'll remove it.

packages/amplify_datastore/example/integration_test/utils/setup_utils.dart
HuiSF: Hub gets ready in general for all possible events from any category.
So we can setup the listener to hub even before starting DataStore to ensure the even can be always captured.

In this implementation, `_completer.complete();` is invoked after `Amplify.DataStore.start()` completed. As the `start` API only starts storage engine and sync engine, as long as the starting completes, it returns, and not ensure if the ready event is dispatched or not.

Hope this explanation helps understanding.

packages/amplify_datastore/example/integration_test/separate_integration_tests/basic_model_operation_test.dart
HuiSF: There is no need to validate the updates on the name fields, as if the `_version` bumps, means AppSync + DynamoDB accepted the update mutation. The name field will certainly be updated.

build-support/integ_test_android.sh
HuiSF: Updated the script to allow passing flag arguments.

Supported arguments see [here](https://github.com/aws-amplify/amplify-flutter/pull/1351/files#diff-c2b60e08277abc187525d4594d19b5d56fe925f1e710af6f08bdd99b2f0eab28R13-R21).

build-support/integ_test_android.sh
```diff
@@ -0,0 +1,28 @@
...
+flutter test \
+    --no-pub \
+    -d sdk \
+    $TARGET
+
+TEST_ENTRIES=`ls integration_test/separate_integration_tests/*.dart`
+for ENTRY in $TEST_ENTRIES
+do
+    echo "Run $ENTRY WITH API Sync"
+    flutter test \
```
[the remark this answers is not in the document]
HuiSF: I will look into this when integrating the integ tests runs into CI.

packages/amplify_datastore/example/integration_test/separate_integration_tests/has_one_explicit_child_test.dart
HuiSF: Added reusable test util function to simplify the scripts in each test suite.

packages/amplify_datastore/example/integration_test/main_test.dart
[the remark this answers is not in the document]
HuiSF: Updated the melos run and shell script, that can run integration tests with cloud sync disabled.

```
melos run test:integration:android -- --enable-cloud-sync false
```

packages/amplify_datastore/example/integration_test/utils/setup_utils.dart
ragingsquirrel3: thanks

packages/amplify_datastore/example/integration_test/utils/setup_utils.dart
```diff
@@ -34,6 +58,38 @@ Future<void> configureDataStore() async {
...
+
+/// Am async operator that starts DataStore API sync.
+/// It returns a Future which complets when [Amplify.Hub] receives the ready
+/// event on [HubChannel.DataStore], or completes with an timeout error if
+/// the ready event hasn't been emitted with in 2 minutes.
+class DataStoreStarter {
+  final Completer _completer = Completer();
+  late StreamSubscription hubSubscription;
+
+  Future<void> startDataStore() async {
```
ragingsquirrel3, line 73: nit: can `async` be removed here as this returns future without calling `await`?

ragingsquirrel3:
I think the datastore provision script needs a little fix to work with cloud sync, at https://github.com/aws-amplify/amplify-flutter/blob/main/packages/amplify_datastore/example/tool/add_api_request.json#L11 add:


```json
    "conflictResolution": {
      "defaultResolutionStrategy": {
        "type": "AUTOMERGE"
      }
    }
```

build-support/integ_test_ios.sh
```diff
@@ -7,25 +7,105 @@ if [ ! -d ios ]; then
...
         --no-pub \
+        -d $deviceId \
+        $TARGET; then
+        resultsList+=(0)
+    else
+        resultsList+=(1)
+    fi
+fi
+
+TEST_ENTRIES=`ls integration_test/separate_integration_tests/*.dart`
```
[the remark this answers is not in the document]
ragingsquirrel3: I think bc `set -euo pipefail` at top of script.

ragingsquirrel3 (approved):
A few small fixes needed and a nit, otherwise looks great! This improved so much, thanks for chipping away at a big, complicated problem and leaving us with this great script.

💯

build-support/integ_test_ios.sh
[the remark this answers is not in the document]
HuiSF: Nice find! I will add the path check guard before running the script to run the special integration test suites.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/DataStoreHubEventStreamHandler.kt
```diff
@@ -189,6 +190,23 @@ class DataStoreHubEventStreamHandler : EventChannel.StreamHandler {
...
+                        if (eventData.model is SerializedModel) {
+                            val message = FlutterSubscriptionDataProcessedEvent(
+                                hubEvent.name,
+                                eventData.model.modelName,
+                                eventData
+                            )
+                            sendEvent(message.toValueMap())
+                        } else {
+                            LOG.error("Element is not an instance of SerializedModel.")
+                        }
```
dnys1, line 205:
```suggestion
                        val eventData = hubEvent.data as ModelWithMetadata<*>
                        val model = eventData.model
                        if (model !is SerializedModel) {
                            LOG.error("Element is not an instance of SerializedModel.")
                            return
                        }
                        val message = FlutterSubscriptionDataProcessedEvent(
                            hubEvent.name,
                            eventData.syncMetadata,
                            model,
                        )
                        sendEvent(message.toValueMap())
```

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/hub/FlutterSubscriptionDataProcessedEvent.kt
```diff
@@ -0,0 +1,46 @@
...
+                    "id" to modelWithMetadata.syncMetadata.id,
+                    "_deleted" to modelWithMetadata.syncMetadata.isDeleted,
+                    "_version" to modelWithMetadata.syncMetadata.version,
+                    "_lastChangedAt" to modelWithMetadata.syncMetadata.lastChangedAt?.secondsSinceEpoch
+                ),
+                "model" to serializedModel
+            )
+        )
+    }
+}
```
dnys1, line 46:
```suggestion
class FlutterSubscriptionDataProcessedEvent(
    override val eventName: String,
    private val syncMetadata: ModelMetadata,
    private val model: SerializedModel
) : FlutterHubEvent {
    override fun toValueMap(): Map<String, Any> {
        return mapOf(
            "eventName" to eventName,
            "modelName" to model.modelName,
            "element" to mapOf(
                "syncMetadata" to mapOf(
                    "id" to syncMetadata.id,
                    "_deleted" to isDeleted,
                    "_version" to syncMetadata.version,
                    "_lastChangedAt" to syncMetadata.lastChangedAt?.secondsSinceEpoch
                ),
                "model" to FlutterSerializedModel(model).toMap()
            )
        )
    }
}
```

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterSerializedModel.kt
```diff
@@ -51,34 +51,38 @@ data class FlutterSerializedModel(val serializedModel: SerializedModel) {
...
+                    is Temporal.Date -> value.format()
+                    is Temporal.Time -> value.format()
+                    is Model -> FlutterSerializedModel(value as SerializedModel).toMap()
+                    is Temporal.Timestamp -> value.secondsSinceEpoch
+                    is SerializedCustomType -> FlutterSerializedCustomType(value).toMap()
+                    is List<*> -> {
+                        if (field.isCustomType) {
+                            // for a list like field if its type is CustomType
+                            // Then the item type must be CustomType
+                            (value as List<SerializedCustomType>).map { item ->
```
dnys1, line 69:
```suggestion
                            (value.cast<SerializedCustomType>()).map { item ->
```

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterSerializedModel.kt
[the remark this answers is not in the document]
dnys1: Can you add that as a comment to the code?

packages/amplify_datastore/ios/Classes/types/hub/FlutterHubElement.swift
```diff
@@ -46,6 +46,39 @@ public struct FlutterHubElement {
...
     }
+
+    init(
+        hubElement: MutationEvent,
+        modelSchemaRegistry: FlutterSchemaRegistry,
+        customTypeSchemaRegistry: FlutterSchemaRegistry,
+        modelName: String
+    ) throws {
+        do {
+            let decodedModel = try ModelRegistry.decode(modelName: modelName, from: hubElement.json) as! FlutterSerializedModel
```
dnys1, line 57: Please use `guard` - force casts will crash the app if it ever fails

packages/amplify_datastore/ios/Classes/types/hub/FlutterHubElement.swift
```diff
@@ -46,6 +46,39 @@ public struct FlutterHubElement {
...
+            model = try decodedModel.toMap(
+                modelSchemaRegistry: modelSchemaRegistry,
+                customTypeSchemaRegistry: customTypeSchemaRegistry,
+                modelName: modelName
+            )
+            self.version = hubElement.version
+            let serializedData = model["serializedData"] as! [String: Any]
```
dnys1, line 64:
```suggestion
            let serializedData = model["serializedData"] as? [String: Any] ?? [:]
```

packages/amplify_datastore/ios/Classes/types/hub/FlutterHubElement.swift
```diff
@@ -46,6 +46,39 @@ public struct FlutterHubElement {
...
+            if let value = serializedData["_deleted"] as? Bool {
+                self.deleted = value
+            } else if serializedData["_deleted"] == nil {
+                self.deleted = false
+            }
```
dnys1, line 69:
```suggestion
            self.deleted = serializedData["_deleted"] as? Bool
```

packages/amplify_datastore/ios/Classes/types/hub/FlutterSyncReceivedEvent.swift
```diff
@@ -0,0 +1,48 @@
...
+        } catch {
+            throw FlutterDataStoreError.acquireSchemaForHub
+        }
+    }
+
+    func toValueMap() -> Dictionary<String, Any> {
+        return [
+            "eventName": self.eventName,
+            "modelName": self.modelName,
+            "element": self.element.toValueMap() as Any
```
dnys1, line 45:
```suggestion
            "element": self.element.toValueMap()
```

packages/amplify_datastore/lib/amplify_datastore_stream_controller.dart
```diff
@@ -20,7 +20,7 @@ import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_inte
 import 'package:flutter/services.dart';
 
 EventChannel channel =
-    const EventChannel("com.amazonaws.amplify/datastore_hub_events");
+    const EventChannel('com.amazonaws.amplify/datastore_hub_events');
```
dnys1, line 23:
```dart
const channel = EventChannel('com.amazonaws.amplify/datastore_hub_events');
```

packages/amplify_datastore_plugin_interface/lib/src/types/hub/SubscriptionDataProcessedEvent.dart
```diff
@@ -0,0 +1,28 @@
...
+class SubscriptionDataProcessedEvent<M extends Model> extends HubEventPayload {
+  late HubEventElementWithMetadata<M> element;
+  late String modelName;
+
+  SubscriptionDataProcessedEvent(Map<dynamic, dynamic> serializedData,
+      ModelProviderInterface provider, String eventName) {
+    element = HubEventElementWithMetadata.fromMap(serializedData, provider);
+    modelName = serializedData["modelName"] as String;
+  }
+}
```
dnys1, line 28:
```suggestion
class SubscriptionDataProcessedEvent<M extends Model> extends HubEventPayload {
  final HubEventElementWithMetadata<M> element;
  final String modelName;

  SubscriptionDataProcessedEvent(this.element, this.modelName);

  factory SubscriptionDataProcessedEvent.fromMap(
    Map<dynamic, dynamic> serializedData,
    ModelProviderInterface provider,
    String eventName,
  ) {
    final element =
        HubEventElementWithMetadata<M>.fromMap(serializedData, provider);
    final modelName = serializedData["modelName"] as String;
    return SubscriptionDataProcessedEvent(element, modelName);
  }
}
```

dnys1 (approved):
Approved, contingent on changes mentioned

packages/amplify_datastore/ios/Classes/types/hub/FlutterHubElement.swift
HuiSF: Then it fails, the force cast exception will be captured and handled at this [spot](https://github.com/HuiSF/amplify-flutter/blob/chore/ds-integration-tests-cloud/packages/amplify_datastore/ios/Classes/DataStoreHubEventStreamHandler.swift#L190-L192).

packages/amplify_datastore/ios/Classes/types/hub/FlutterHubElement.swift
HuiSF: At this point, `model["serializedData"] as! [String: Any]` should always valid, as `try decodedModel.toMap()` should've completed without issues.

packages/amplify_datastore/ios/Classes/types/hub/FlutterHubElement.swift
HuiSF: Need to assign the default value `false` when `_delete` is `null`.
Will change to

```swift
let value = serializedData["_deleted"] as? Bool ?? false;
```

packages/amplify_datastore/ios/Classes/types/hub/FlutterHubElement.swift
dnys1: No, that's not true. It will crash the program - unwraps are not propagated as thrown errors like they are in Dart and other languages.

Screen Shot 2022-03-24 at 1 13 25 PM

packages/amplify_datastore/ios/Classes/types/hub/FlutterHubElement.swift
dnys1: You know that. I might even know that. But someone looking at this program for the first time doesn't know that, and the compiler doesn't know that. And those are the people that matter most.

packages/amplify_datastore/ios/Classes/types/hub/FlutterHubElement.swift
dnys1: Then make the property non-null as well on the type

packages/amplify_datastore/ios/Classes/types/hub/FlutterHubElement.swift
HuiSF: Good to know. But for this specific spot -

* `ModelRegistry.decode()` returns a `Model` otherwise it throws (would be captured)
* `FlutterSerializedModel` extends `Model`
* Within the DataStore iOS implementation, everything presented by `Model` is actually `FlutterSerializedModel`

Therefore this force unwrap should be safe here, Xcode hasn't prompt any warning as well.

packages/amplify_datastore/ios/Classes/types/hub/FlutterHubElement.swift
HuiSF: I can't, as `_deleted` is defined as nullable in a model schema.

packages/amplify_datastore/ios/Classes/types/hub/FlutterHubElement.swift
HuiSF: And this is a good point.

packages/amplify_datastore/ios/Classes/types/hub/FlutterHubElement.swift
HuiSF: Oh you mean the class property, yes I'll change it.

packages/amplify_datastore/ios/Classes/types/hub/FlutterHubElement.swift
dnys1: Like I said in my other comment, these are things you know but are not apparent by just reading the code. So please either write this in a comment above this line or just change it to use a `guard`/`throw` which, as you say, will have the same effect. The difference is that by changing it, it's ✨ **guaranteed** ✨ to never crash no matter what changes in our code or iOS.

You don't get a warning because you have statements that throw in the body. But your original statement was incorrect, this will crash the program if it ever fails. I'm trying to make a general point about force unwrapping, no longer specific to this one line.

dnys1: In this case, you can also create a method with this comment and reuse that throughout our DataStore code, because it sounds like we may be doing this a lot throughout and there may be something we know that the compiler doesn't.

HuiSF: Yea, the native impl. could use a good clean up. I'll fine some other time to do so... or waiting for the "final big" change. I'll take your suggestion here, thanks.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/hub/FlutterSubscriptionDataProcessedEvent.kt
HuiSF: Thanks, any advances for this proposed change on the constructor?
I kept the `eventName` and `modelName` as the first two parameter to make it similar with other event classes in the codebase.

dnys1: Yes, there are no casts, and no duplicate data is being passed anymore.

dnys1: It's okay to keep the other parameters for consistency, I was just trying to get things more concise.

packages/amplify_datastore_plugin_interface/lib/src/types/hub/SubscriptionDataProcessedEvent.dart
HuiSF: I like this version better, but I will keep the current implementation following other event class implementations for now.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterSerializedModel.kt
HuiSF: This breaks for some cases when the value uses the `List` implementation is a variant other than `ArrayList`. E.g. `SingletonList` created by Kotlin `listOf`.

I think the `cast` function should keep the generic `List`?

dnys1: The definition of `cast` is incorrect... I'm not sure why because I had written it differently at one point. It should be

```kotlin
inline fun <reified T> List<*>.cast(): List<T> =
    this as List<T>
```

HuiSF: I've updated the `cast` function with the latest commit, thanks.

PR #1354, opened 2022-02-10
chore: update flutter analysis to throw an error for depend_on_refere…

ragingsquirrel3:
*Description of changes:*

We currently have a lint rule for depend_on_referenced_packages, which basically gives info warning if your code imports something that is not in the pubspec for that plugin. When that dependency is transitive, it will compile, but will throw an error when you publish on pubspec.

To prevent that, this PR makes it an error that will cause analysis to fail.

### DIFF
--- a/packages/amplify_lints/lib/dart.yaml
+++ b/packages/amplify_lints/lib/dart.yaml
@@ -3,6 +3,8 @@ include: package:lints/recommended.yaml
 analyzer:
   strong-mode:
     implicit-dynamic: false
+  errors:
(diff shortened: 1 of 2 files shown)


HuiSF (approved):

dnys1 (approved):
Please add to `dart.yaml` as well - thanks!

ragingsquirrel3:
> Please add to `dart.yaml` as well - thanks!

done

Jordan-Nelson (approved):

PR #1359, opened 2022-02-14
fix(amplify_api): GraphQL model helpers, make items nullable in list queries to improve error handling

ragingsquirrel3:
Currently, GraphQL response decoders related to model-based helpers have an issue decoding certain error scenarios correctly. The scenario arises only in list queries when one of the returned objects has an error which prevents AppSync from returning the entry (for example, a required field has no value). When this happens, AppSync will have a null entry in the `items` array where that object would have been, along with an error message that references the index of the null value. This means that AppSync can actually return an array of nulls, or a mix of valid objects and nulls.

Currently, this causes type errors in the decoder as the `items` array in the paginated response is of type `List<T>` where `T extends Model`, even though the response data itself is nullable. This type definition makes it impossible to decode the AppSync response with null entries to a list of models. It's basically the same reason that the response data is nullable: The need to accurately reflect AppSync responses in a null-safe environment.

This PR makes 2 changes:
1) change the type definition of `items` from `List<T>` to `List<T?>`
2) some runtime code changed in decoding to account for null values here, plus tests

It's nearly impossible to write an integration test for this as the only way I have found to reproduce it is to have invalid entries in the database, which AppSync will prevent writing. In the real world, this could occur either from a schema change or from modifying the data outside of AppSync, which is how I found this. I manually tested in the example app by:
1) create some valid blog entries
2) in example app, use `ModelQueries.list()` to successfully retrieve a list of valid blogs.
3) in dynamo DB console, remove the `createdAt` and `updatedAt` fields (which are not nullable and automatically populated by AppSync)
4) try query from step 2 again. This time, there is a null entry in the `items` array and an error which references it. The other blog entries are still returned as before. Errors/nulls are visible and correctly decoded into the response with no runtime exceptions.

### DIFF
--- a/packages/amplify_api/example/integration_test/graphql_tests.dart
+++ b/packages/amplify_api/example/integration_test/graphql_tests.dart
@@ -202,7 +202,11 @@ void main() {
         var secondRes = await Amplify.API.query(request: secondReq!).response;
         var secondData = secondRes.data;
         expect(secondData?.items.length, limit);
-        expect(secondData?.items[0]?.id, isNot(firstData?.items[0]?.id));
+        final firstId = firstData?.items[0]?.id;
+        final secondId = secondData?.items[0]?.id;
+        expect(firstId?.isNotEmpty, isTrue);
+        expect(secondId?.isNotEmpty, isTrue);
+        expect(firstId, isNot(secondId));
       });
 
       testWidgets('should LIST blogs with Model helper with query predicate',
--- a/packages/amplify_api/example/integration_test/graphql_tests.dart
+++ b/packages/amplify_api/example/integration_test/graphql_tests.dart
@@ -202,7 +202,7 @@ void main() {
         var secondRes = await Amplify.API.query(request: secondReq!).response;
         var secondData = secondRes.data;
         expect(secondData?.items.length, limit);
-        expect(secondData?.items[0].id, isNot(firstData?.items[0].id));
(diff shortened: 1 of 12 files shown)


dnys1 (approved):

packages/amplify_api/example/integration_test/graphql_tests.dart
```diff
@@ -202,7 +202,7 @@ void main() {
         var secondRes = await Amplify.API.query(request: secondReq!).response;
         var secondData = secondRes.data;
         expect(secondData?.items.length, limit);
-        expect(secondData?.items[0].id, isNot(firstData?.items[0].id));
+        expect(secondData?.items[0]?.id, isNot(firstData?.items[0]?.id));
```
HuiSF, line 205:
```
expect(secondData?.items[0]?.id, isNot(firstData?.items[0]?.id));
```
can theoretically produce
```
expect(null, isNot(null));
```
? Should we add some sort of constraint to this `expect`?

HuiSF (approved):
A nit-picking comment.

packages/amplify_api/example/integration_test/graphql_tests.dart
ragingsquirrel3: Good callout. I will add extra check to ensure both IDs are non-empty, non-null strings, in addition to ensuring they are not the same.

HuiSF (approved):

PR #1366, opened 2022-02-16
feat(datastore): implement ModelIndex

HuiSF:
*Issue #, if available:*

*Description of changes:*

Currently, codegen generated Dart models don't contain model index information. This PR is adding relevant interface and implementation to support upcoming changes in codegen to provide index information in model schema.

- Implement ModelIndex and add it to ModelSchemaDefinition
- Update test models
- Update unit tests

Note: this PR doesn't include Flutter native implementation to convert serialize model indexes into native objects. This implementation will be added in a separate PR.


packages/amplify_datastore_plugin_interface/test/testData/Blog.dart
```diff
@@ -152,6 +155,10 @@ class Blog extends Model {
     modelSchemaDefinition.name = "Blog";
     modelSchemaDefinition.pluralName = "Blogs";
 
+    modelSchemaDefinition.indexes = [
```
dnys1, line 158: Is there a difference between an empty array and an array with the `id` field marked with `@primaryKey`? Wondering if we need to always include the primary key field or not.

packages/amplify_datastore_plugin_interface/test/testData/Person.dart
```diff
@@ -69,34 +61,16 @@ class Person extends Model {
     }
   }
 
-  TemporalDateTime? get createdAt {
```
dnys1, line 72: What happened to `createdAt`/`updatedAt`?

dnys1 (approved):

packages/amplify_datastore_plugin_interface/test/testData/Person.dart
HuiSF: This is a custom type, should not contain model meta fields. I think the original generated code was outdated...

packages/amplify_datastore_plugin_interface/test/testData/Blog.dart
HuiSF: Thanks for reviewing :)
This is simply following GraphQL transformer + codegen convention of how to describe an index, e.g. `@primaryKey`

`@model` with normal `id` as the default primary key
```
{ name: 'key', arguments: { fields: [ 'id' ] } }
```

`@model` with `modelID` annotated by `@primaryKey`
```
{ name: 'key', arguments: { fields: [ 'modelID' ] } }
```

`@model` with `productId` annotated by `@primaryKey` + sortKeyFields
```
{
  name: 'key',
  arguments: { fields: [ 'productId', 'productName', 'warehouseId', 'region' ] }
}
```

And `@index`

`@model` has `postId` annotated with `@index` + sortKeyFields
```
{
  name: 'byPost',
  arguments: { fields: [ 'postId', 'content' ] }
}
```

packages/amplify_core/lib/src/types/models/model_schema.dart
```diff
@@ -1,100 +1,109 @@
...
-  @override
-  int get hashCode {
-    return name.hashCode ^
-        pluralName.hashCode ^
-        authRules.hashCode ^
-        fields.hashCode;
-  }
-}
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
haverchuck, line 2: Not sure why this is registering as a change?

haverchuck (approved):
Left one question about a diffing issue, but otherwise LGTM

packages/amplify_core/lib/src/types/models/model_index.dart
```diff
@@ -0,0 +1,89 @@
...
+      ModelIndex.fromMap(jsonDecode(source));
+
+  /// Generate a json string that represents [ModelIndex]
+  String toJson() => jsonEncode(toMap());
+
+  @override
+  String toString() => 'ModelIndex(name: $name, fields: $fields)';
+
+  @override
+  bool operator ==(Object other) {
```
Jordan-Nelson, line 75: Should we start using the `AWSEquatable` mixin instead of implementing these everywhere? It is currently in amplify_flutter, so you would have to move it if you want to use it. I don't think you have to use it for this PR, but I want to discuss if we want to start using it moving forward.

@dnys1 I think you added AWSEquatable. Any reason we can't move that to amplify_core and start using it throughout all packages?

Jordan-Nelson (approved):
few nit comments but lgtm

packages/amplify_core/lib/src/types/models/model_index.dart
HuiSF: Thanks, wasn't aware of this mixin. Can we also use it with generated models? @dnys1

packages/amplify_core/lib/src/types/models/model_schema.dart
HuiSF: As this PR converts this file from CRLF to LF mode (The first commit of this PR).
If you view the second commit it shows the actual changes on this file.

PR #1372, opened 2022-02-17
chore(amplify_authenticator): clean up public API, add documentation

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Remove classes/functions/etc from the public API that should not be public
- Add documentation to public classes where it is missing
- Add a "Prebuilt Widgets" category to the docs. See https://pub.dev/packages/dartdoc#categories
- Move some files around in the sample app. pub seems to think there is no example. I think this may because of the current structure.
- Update the package description. pub wants descriptions to be 60+ char.


packages/amplify_authenticator/dartdoc_options.yaml
```diff
@@ -0,0 +1,5 @@
+dartdoc:
+  categories:
+    "Prebuilt Widgets":
```
Jordan-Nelson, line 3: This adds a page that lists all the classes tagged with the "Prebuilt Widget" category. The markdown file is used as the header. See: https://pub.dev/packages/dartdoc#categories

My plan is to link to this from the UI docs site, so that we don't have to list them all out on the docs UI site as well.

This is what it will look like:
Screen Shot 2022-02-17 at 4 08 08 PM

packages/amplify_authenticator/lib/src/mixins/authenticator_phone_field.dart
```diff
@@ -161,3 +161,12 @@ mixin AuthenticatorPhoneFieldMixin<FieldType,
     }
   }
 }
+
+extension _StringPrefix on String {
```
Jordan-Nelson, line 165: I moved this and made it private to remove it from the public API

packages/amplify_authenticator/lib/src/widgets/button.dart
```diff
@@ -244,111 +273,12 @@ class BackToSignInButton extends StatelessAuthenticatorComponent {
   }
 }
 
-class GoToSignUpButton extends StatelessAuthenticatorComponent {
```
Jordan-Nelson, line 247: These next two buttons are not used. I think they could be useful for the Custom UI using authenticatorBuilder, but they haven't been tested (to my knowledge) in a while. We can always add them back if we want.

packages/amplify_authenticator/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: amplify_authenticator
-description: A prebuilt sign in/sign up experience for Amplify Auth.
+description: A prebuilt Sign In and Sign Up experience for the Amplify Auth category
```
Jordan-Nelson, line 2: Pub likes these to be 60+ characters

packages/amplify_authenticator/dartdoc_options.yaml
dnys1: Sweet!!

dnys1 (approved):
Thanks, man!

Can we change template tags to start with `amplify_authenticator.` instead of `authenticator.`? They need to be globally unique and I'm wondering if there could ever be a conflict in the future.

packages/amplify_authenticator/docs/prebuilt-widgets.md
```diff
@@ -0,0 +1 @@
+The amplify_authenticator package provides prebuilt widgets for [building custom UIs](https://ui.docs.amplify.aws/components/authenticator?platform=flutter#full-ui-customization). All of the prebuilt widgets are integrated into the authenticator's state. Forms and form fields will automatically update the authenticator's state on change, and buttons will use data from the state to initiate auth events such as sign in and sign up.
```
haverchuck, line 1: ultra nit: we need to align on whether 'authenticator' is capitalized in documentation or not. Doesn't need to block this PR.

haverchuck (approved):

PR #1377, opened 2022-02-18
fix(amplify_authenticator): allow multiple custom attributes in custom forms

Jordan-Nelson:
*Issue #, if available:* n/a

*Issue description:*

If 2+ custom sign up attributes are used in SignUpForm.custom(), only the first will be displayed.

Example: "Bio" will be displayed in the form below. "Age" will not.

```dart
  fields: [
    SignUpFormField.username(),
    SignUpFormField.email(),
    SignUpFormField.password(),
    SignUpFormField.passwordConfirmation(),
    SignUpFormField.custom(
      title: 'Bio',
      attributeKey: const CognitoUserAttributeKey.custom('bio'),
    ),
    SignUpFormField.custom(
      title: 'Age',
      attributeKey: const CognitoUserAttributeKey.custom('age'),
    )
  ],
```

*Issue Cause*

The key used for de-duping runtime/custom attributes uses the `field`, which will be `SignUpField.custom` for all custom attributes.

```dart
final fields = LinkedHashSet<AuthenticatorFormField>(
  equals: (a, b) {
    return a.runtimeType == b.runtimeType && a.field == b.field;
  },
  hashCode: (field) => hashValues(field.runtimeType, field.field),
);
```

*Description of changes:*
- Move runtimeFields logic from AuthenticatorForm to SignInForm
- Do not render runtimeFields when SignUpForm.custom() is used 
- Simplify the de-duping logic. Since runtimeFields are not used with SignUpForm.custom(), the logic to de-dup custom/runtime fields together isn't required

*Breaking Changes*

With the current API, the username/password and other signup/verification attributes will be rendered automatically when using SignUpForm.custom(). For example, the following code would still render the username/password/email (assuming email was a verification mechanism)

```dart
  fields: [
    SignUpFormField.address(),
  ],
```

With the changes, the customer will have to be explicit about which fields should be rendered. For example, the following code would be required to render the username/password/email + the address.

```dart
  fields: [
    SignUpFormField.username(),
    SignUpFormField.email(),
    SignUpFormField.password(),
    SignUpFormField.passwordConfirmation(),
    SignUpFormField.address(),
  ],
```

This is consistent with the JS Authenticator. See: https://ui.docs.amplify.aws/components/authenticator#sign-up-fields

*Alternatives*

The issue could be solved by updating the key used to de-dup attributes. For example, we could use `field` + `customAttributeKey` as the key if the field is `SignUpField.custom`.

I took the approach to force an explicit list of attributes because it is consistent with the JS API, and it should be easier to maintain (bugs like this one will be less likely to surface in the future). I also personally like the explicit API better. I think an explicit API is easier to understand.


packages/amplify_authenticator/lib/src/widgets/button.dart
```diff
@@ -232,19 +232,25 @@ class ConfirmSignInMFAButton extends AuthenticatorElevatedButton {
 ///
 /// Uses [ButtonResolverKey.signout] for localization
 /// {@endtemplate}
-class SignOutButton extends AuthenticatorElevatedButton {
+class SignOutButton extends StatelessAuthenticatorComponent {
```
Jordan-Nelson, line 235: The changes here don't really have anything to do with this PR. I just noticed while running integ tests that the sign out button was defaulting to occupying the full screen. This is happening because it extends `AuthenticatorElevatedButton`, which is by default full width.

I had changed this to extend `AuthenticatorElevatedButton` when removing amplify theme. These changes revert it back to a normal  material `ElevatedButton`

packages/amplify_authenticator/lib/src/widgets/form.dart
```diff
@@ -262,35 +232,66 @@ class SignUpForm extends AuthenticatorForm {
...
+  List<AuthenticatorFormField> get allFields {
+    if (!widget.useRunTimeFields) {
+      return widget.fields;
+    }
+
+    // combine fields and sort on priority
+    final fields = [
+      ...widget.fields,
+      ...runtimeFields(context),
+    ]..sort((a, b) {
```
dnys1, line 287: This will sort the custom fields and `sort` is not stable. So, fields could theoretically get jumbled up every time you load the form.

packages/amplify_authenticator/lib/src/widgets/component.dart
```diff
@@ -50,7 +50,7 @@ abstract class StatelessAuthenticatorComponent extends StatelessWidget {
 /// {@template amplify_authenticator.authenticator_component}
 /// Base class for all stateful Authenticator components.
 /// {@endtemplate}
-abstract class AuthenticatorComponent<T extends AuthenticatorComponent<T>>
+abstract class AuthenticatorComponent<T extends StatefulWidget>
```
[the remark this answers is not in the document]
Jordan-Nelson: These changes are needed to make the changes to AuthenticatorFormState class. Without the the changes to the AuthenticatorFormState class, `widget` within _SignUpFormState is of type `AuthenticatorForm` instead of the more specific `SignUpForm`.

That is an issue because `useRunTimeFields` doesn't exist on `AuthenticatorForm`. It exists on `SignUpForm`

packages/amplify_authenticator/lib/src/widgets/form.dart
Jordan-Nelson: I updated it to use mergeSort from flutter/foundation, which is stable.

dnys1: Noice

dnys1: The other issue, though, is that the order of the custom fields may be different when including runtime fields since it's not sorted then. Is that expected?

packages/amplify_authenticator/lib/src/widgets/component.dart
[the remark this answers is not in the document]
dnys1: Ohh, I see

dnys1 (approved):
LGTM

packages/amplify_authenticator/lib/src/widgets/form.dart
Jordan-Nelson: There are two options for creating a SignUpForm. The default constructor `SignUpForm()` and the named constructor `SignUpForm.custom()`. `includeDefaultFields` is always true when the default constructor is used, and is always false when the named constructor is used. When the default constructor is used, the consumer has no control over what fields are displayed. The required username/password fields are combined with the fields from the config, and should be sorted. When the named constructor is used, the consumer has full control over what fields are displayed, and they should be displayed in the order they are passed in (they should not be sorted).

So I don't think it is an issue that fields are sorted when `includeDefaultFields` is true and are not sorted when `includeDefaultFields` is false. It is intentional.

It would potentially be an issue if `SignUpForm.custom()` accepted `includeDefaultFields` as an argument. But `includeDefaultFields` is only used internally. It could actually be made private. I will push an update to make it private.

dnys1: Ah, yeah I got a bit confused. Sounds good.

haverchuck (approved):

PR #1404, opened 2022-02-24
chore(amplify_api): clean up some code to limit analysis infos

ragingsquirrel3:
No user-facing change here, just some small changes in API category to reduce lint infos as we lost some pub points in last release for non-fatal static analysis violations.


packages/amplify_api/test/graphql_helpers_test.dart
```diff
@@ -23,7 +23,7 @@ import 'package:flutter_test/flutter_test.dart';
 
 void main() {
   group('with ModelProvider', () {
-    final AmplifyAPI api = AmplifyAPI(modelProvider: ModelProvider.instance);
+    AmplifyAPI(modelProvider: ModelProvider.instance);
```
dnys1, line 26: Prefer to delete

packages/amplify_api/test/graphql_helpers_test.dart
```diff
@@ -787,9 +787,9 @@ void main() {
 
   group('without ModelProvider', () {
     test('should handle no ModelProvider instance', () {
-      AmplifyAPI api = AmplifyAPI();
+      AmplifyAPI();
```
dnys1, line 790: Prefer to delete

packages/amplify_api/example/integration_test/graphql_tests.dart
```diff
@@ -44,7 +44,7 @@ void main() {
       final req = GraphQLRequest<String>(
           document: graphQLDocument, variables: <String, String>{'id': id});
       final response = await Amplify.API.mutate(request: req).response;
-      Map data = jsonDecode(response.data!);
+      jsonDecode(response.data!);
```
dnys1, line 47: Prefer to delete or wrap with try catch so its intention is clear.

dnys1 (approved):

Jordan-Nelson (approved):

packages/amplify_api/test/graphql_helpers_test.dart
ragingsquirrel3: You want to delete this line completely? maybe I'm missing something, but this is needed to register the model provider in the plugin singleton, just the plugin as a value not needed after that (at least to test model helpers).

packages/amplify_api/test/graphql_helpers_test.dart
ragingsquirrel3: needed to test failure logic of not using model provider. If deleted, uses previously registered one.

packages/amplify_api/example/integration_test/graphql_tests.dart
ragingsquirrel3: done

PR #1408, opened 2022-02-25
chore: add pub analysis to CI

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Add a new step to CI to run pub analysis. This uses the package `pana`, which is what calculates the score on pub. The step will fail if the package is more than N points below the max score, where N is a threshold that we set
- Enable step for all non interface packages, and set appropriate threshold
- Fixed an issue in amplify_authenticator to get it to the max score

We should aim to get to the max score for all packages if possible, but right now this PR sets the threshold at the lowest it can for each package without failing.

[Here is a link to a previous CI run](https://app.circleci.com/pipelines/github/aws-amplify/amplify-flutter/2892/workflows/b7c5d977-9df8-4e02-bb86-c777a576cd7c/jobs/12807) where the amplift_flutter package threshold was too low (set to 10, but 20 points were missing). This failed and the logs show the score for the package.

Note: 120 is considered a "max" score for packages that don't support web, even though adding web support would bump the score to 130. It is still considered a "max" by pana.


packages/amplify_authenticator/lib/src/services/amplify_auth_service.dart
```diff
@@ -18,7 +18,6 @@ import 'dart:async';
 import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
 import 'package:amplify_flutter/amplify_flutter.dart';
 import 'package:collection/collection.dart';
-import 'package:collection/src/iterable_extensions.dart';
```
Jordan-Nelson, line 21: Removing this was needed to get the authenticator to a max pub score

.circleci/config.yml
```diff
@@ -181,6 +201,21 @@ workflows:
     jobs:
       - format_flutter
       - analyze_flutter
+      - pub_analysis:
+          matrix:
+            parameters:
+              plugin:
+                [
+                  "amplify_authenticator",
+                  # TODO: enable other packages once pub score is addressed
```
Jordan-Nelson, line 210: `amplify_authenticator` is the only package with a max score right now.

Note: 120 is considered a max score for packages that don't support web, even though adding web support would bump the score to 130. It is still considered a "max" by pana,

ragingsquirrel3: It seems like the other packages lose points for, among other things, a dev dependency on unpublished package amplify_test. Any ideas for how to resolve that? Without fixing that, it seems we will either a) have to use only in authenticator b) change the exit-code-threshold to 10 (I think?) to basically ignore that loss of points.

Jordan-Nelson: I don't know off the top of my head. If we cannot figure out how to address that, we could set the threshold to 10, or set different threshold per package.

packages/amplify_authenticator/lib/src/services/amplify_auth_service.dart
haverchuck: I'm assuming it was superfluous and we had access to the extensions via `collection.dart`?

Jordan-Nelson: Correct

.circleci/config.yml
ragingsquirrel3: ok, makes sense. For this PR, I would either

a) merge it like this and figure it out when we add other plugins
b) do this now for any packages that can be published with threshold of 10 or 0.

ragingsquirrel3 (approved):

dnys1 (approved):
I like this initiative! I think that striving for 120 is a bit arbitrary, but I think it will help us align with lint rules so I'm about it.

Thanks for adding this!

.circleci/config.yml
Jordan-Nelson: I updated the PR to use a different threshold for each plugin. The one quirk with this is that the threshold is now in the CI job name. That means we will have to change the list of required jobs each time we adjust the threshold. I think thats fine though as it shouldn't be changed frequently.

.circleci/config.yml
```diff
@@ -181,6 +202,20 @@ workflows:
     jobs:
       - format_flutter
       - analyze_flutter
+      - pub_analysis:
+          matrix:
+            parameters:
+              plugin_threshold: [
+                  # should be in format <plugin_name>:<threshold>
+                  "amplify_analytics_pinpoint:10",
+                  "amplify_api:20",
```
Jordan-Nelson, line 211: I believe I can decrease this now that https://github.com/aws-amplify/amplify-flutter/pull/1404 has merged. I will update with main and look to decrease it.

PR #1417, opened 2022-02-28
fix(amplify_api): model helpers query predicates correctly translates…

ragingsquirrel3:
Issue #1419 

API model helper `ModelQueries.list()` accepts a query parameter argument `where` and translates it to a GraphQL filter according to the model's schema. Right now, it does not correctly handle when querying a list of models by a belongsTo ID. It uses the key `field` value of the query predicate, which is not correct in this case.

This PR has a small fix that looks for associations when getting that field name, and looks for the schema's `targetName`, which is the name of the field that AppSync uses from the compiled schema. This is not present for all query predicates, only when querying associations. This PR has the fix, plus unit test and integ test.


haverchuck (approved):

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
```diff
@@ -200,9 +200,16 @@ class GraphQLRequestFactory {
 
     // e.g. { 'name': { 'eq': 'foo }}
     if (queryPredicate is QueryPredicateOperation) {
-      // check for the IDs where fieldName set to e.g. "blog.id" and convert to "id"
-      final isId = queryPredicate.field == '${schema.name.toLowerCase()}.id';
-      final fieldName = isId ? idFieldName : queryPredicate.field;
+      final associatedTargetName =
+          schema.fields?[queryPredicate.field]?.association?.targetName;
+      String fieldName = queryPredicate.field;
+      if (queryPredicate.field == '${schema.name.toLowerCase()}.$idFieldName') {
```
dnys1, line 206: I just realized, this should be "camel-cased" schema name, not the lower-cased one: https://github.com/aws-amplify/amplify-codegen/blob/8baa1d91db83911c6b7fbc2d1c4601a525f37ba1/packages/appsync-modelgen-plugin/src/visitors/appsync-dart-visitor.ts#L752

dnys1 (approved):

packages/amplify_api/lib/src/graphql/graphql_request_factory.dart
ragingsquirrel3: good catch! updated to only lowercase first letter like the transformer

Issue #816, opened 2021-08-18
Autogenerated code for fromJson in all models doesn't match AppSync's query results when there are nested models in the query


epomeranz:
**Describe the bug**
Amplify automatically generated the code for all the models, including the fromJson functions. Additionally amplify also generates the basic resolvers for basic queries. Unfortunately, the amplify queries do not work with the fromJson functions when there are nested tables...
i.e. if I have this query:

```graphql
query MyQuery {
  listConversations {
    items {
      id
      type
      messages {
        items {
          message
          createdAt
        }
      }
      participants {
        items {
          usersID
        }
      }
    }
  }
}
```

Conversation has a list of messages and a list of participants... In the query result messages and participants are nested under the "items" key, and not under the root keys for "participants" and "messages". This is OK because you are including additional information under those root keys. But the automatically generated fromJson functions for all models expects that nested queries be on the root and not under an "items" key...

Here is the fromJson autogenerated code:
```dart
messages = json['messages'] is List
        ? (json['messages'] as List)
          .where((e) => e?['serializedData'] != null)
          .map((e) => Message.fromJson(new Map<String, dynamic>.from(e['serializedData'])))
          .toList()
        : null,
      _participants = json['participants'] is List
        ? (json['participants'] as List)
          .where((e) => e?['serializedData'] != null)
          .map((e) => Participant.fromJson(new Map<String, dynamic>.from(e['serializedData'])))
          .toList()
        : null,
```

This should be:
```dart
messages = json['messages'][items] is List
        ? (json['messages'][items] as List)
          .where((e) => e?['serializedData'] != null)
          .map((e) => Message.fromJson(new Map<String, dynamic>.from(e['serializedData'])))
          .toList()
        : null,
      _participants = json['participants'][items] is List
        ? (json['participants'][items] as List)
          .where((e) => e?['serializedData'] != null)
          .map((e) => Participant.fromJson(new Map<String, dynamic>.from(e['serializedData'])))
          .toList()
        : null,
```
**To Reproduce**
1. Create at leas 2 entities. One nested under the other in a (0 to many relationship)
2. Autogenerate the flutter code for the models
3. Run a query that reads the parent and the many children (0 to many relationship)
4. The fromJson code will ignore the nested models (children) because they are under an "items" key.

**Expected behavior**
The fromJson code should expect the same structure that the queries provide. They are both part of amplify so a read and a fromJson should work seamlessly and not have different structures.

**Screenshots**

**Platform**
[...]

PR #1420, opened 2022-03-01
chore(amplify_api): support decoding custom list request

ragingsquirrel3:
## Issues

* https://github.com/aws-amplify/amplify-flutter/issues/816
* https://github.com/aws-amplify/amplify-flutter/issues/1425

Model helpers support the use case of manually creating a request with some metadata so that it will get decoded into the correct type (i.e. instance of a model rather than string/JSON). This was tested/documented for requests of a single instance (get, update, create, delete) but not for custom list queries. They require a certain class `PaginatedModelTypeImpl` (implementation of abstract class `PaginatedModelType` that is not publicly exported to use properly).

To fix that, this PR refactors class definitions and decoding logic related to paginated results. It gets rid of the abstract/implementation classes and instead has simple class definitions in `PaginatedModelType` and `PaginatedResult` which are already public. The result is that caller can use `PaginatedModelType` as the `modelType` for request and they don't need the old private version. It has some tests, too.

Generally, this is a simplification, however, it required moving a small bit of logic for getting the request for next page of results to the decoder.


packages/amplify_api/example/integration_test/graphql_tests.dart
```diff
@@ -313,6 +313,29 @@ void main() {
...
+            items {
+              id
+              name
+              createdAt
+            }
+          }
+        }''';
+        final request = GraphQLRequest<PaginatedResult<Blog>>(
+            document: graphQLDocument,
+            modelType: const PaginatedModelTypeImpl(Blog.classType),
```
ragingsquirrel3, line 334: Callout: when we named this class we did not intend for it to be developer-facing as we were not considering the customer request use case very much. Now that we are exposing it like this, don't know if we like a different name?

Jordan-Nelson: I don't have enough context to suggest a name, but I would vote for renaming it.

Are we expecting there to be other classes that extend `PaginatedModelType`? Do we need `PaginatedModelType` and `PaginatedModelTypeImpl`?

dnys1: What about:

```dart
extension PaginatedModelTypeExt<T extends Model> on ModelType<T> {
  PaginatedModelType<T> get list => PaginatedModelTypeImpl(this);
}

// so it becomes

GraphQLRequest<PaginatedResult<Blog>>(
  document: graphQLDocument,
  modelType: Blog.classType.list,
  // ...
```

This way the class can stay private and it's more in line with the current approach.

ragingsquirrel3:
I think this fix will help with use case mentioned in https://github.com/aws-amplify/amplify-flutter/issues/816

packages/amplify_api/example/integration_test/graphql_tests.dart
ragingsquirrel3: In latest revision refactored to follow @Jordan-Nelson 's suggestion here.

packages/amplify_api_plugin_interface/lib/src/types/pagination/paginated_result.dart
```diff
@@ -17,21 +17,45 @@ import 'package:amplify_core/amplify_core.dart';
 
 import '../../types.dart';
 
-abstract class PaginatedResult<T extends Model> extends Model {
+class PaginatedResult<T extends Model> extends Model {
+  /// Model instances for this set of results. An entry might be null if
```
Jordan-Nelson, line 21: *nit:* https://dart.dev/guides/language/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph

```suggestion
  /// Model instances for this set of results.
  ///
  /// An entry might be null if
```

Jordan-Nelson (approved):

dnys1 (approved):

packages/amplify_api_plugin_interface/lib/src/types/pagination/paginated_result.dart
ragingsquirrel3: done, thanks

PR #1429, opened 2022-03-03
chore: update integ provision scripts to use AWS profile

ragingsquirrel3:
As written, provision scripts for integration test resources assume default profile in local AWS configuration. This PR lets the caller specify the AWS profile with an env variable `AWS_PROFILE`, similar to existing change in auth example script.

### DIFF
--- a/packages/amplify_api/example/tool/provision_integration_test_resources.sh
+++ b/packages/amplify_api/example/tool/provision_integration_test_resources.sh
@@ -2,7 +2,7 @@
 set -e
 IFS='|'
 
-[ "$AWS_PROFILE" ] && profileName="$AWS_PROFILE" || profileName="default";
+profileName=${AWS_PROFILE:-default}
 
 FLUTTERCONFIG="{\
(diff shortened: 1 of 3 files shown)


packages/amplify_api/example/tool/provision_integration_test_resources.sh
```diff
@@ -2,6 +2,8 @@
 set -e
 IFS='|'
 
+[ "$AWS_PROFILE" ] && profileName="$AWS_PROFILE" || profileName="default";
```
dnys1, line 5:
```suggestion
profileName=${AWS_PROFILE:-default}
```

dnys1 (approved):

Jordan-Nelson (approved):

packages/amplify_api/example/tool/provision_integration_test_resources.sh
ragingsquirrel3: good suggestion, updated here and other scripts

Issue #1426, opened 2022-03-03
Custom primary key support


dnys1:
### Description

This is to track support of custom primary keys in model gen and model mutations.

- [ ] Model gen
- [ ] Model mutations

#### Current state of custom primary key support

| CPK Support | Stable ver. | Dev Preview ver. |
| ----- | :----: | :----: |
| GraphQL API | ❌ | ✅ |
| DataStore | ✅ | ✅ |

### Categories

- [ ] Analytics
- [X] API
- [ ] Auth
- [X] DataStore
- [ ] Storage

### Platforms

- [ ] iOS
- [ ] Android

### Environment

```bash
N/A
```

### Dependencies

```bash
N/A
```

### Device

N/A

### OS

N/A

### CLI Version

N/A

tr-asimpson:
Use case for using Custom PK. 

We have a BLE device that has its own identifier which cannot be changed for security reasons. This along with some other information is sent to a Lambda function which then searches for the device "key" in DynamoDB and updates via AppSync.

I did not want to have to send the generated "ID" as this means id have to then store it on flash (Limited space) and add it to the network packet which uses 3g/4g increasing our overhead. 

I believe AppSync requires the primary key field for updates hence me needing a custom one.

This is also how the subscriptions for updates work by key. I am not sure if I could have done the same via ID, Correct me if I am wrong.

magicCashew:
We have been confronted with almost the exact same use case as @tr-asimpson .  We are working with BLE devices and have natural keys currently defined in DynamoDB.  We really want to use AppSync, but if we do, we'll have to apply costly workarounds, and will impact our customer experience.  Is it possible to share where this enhancement is on the roadmap or an ETA so that we can appropriately plan.

Thanks!

veeramneni:
+1

AmanMasipeddi:
+1

Simon-PumpAndUp:
Hi there

I can't get a unique record from dynamodb through AppSync using:
Amplify.API.query(request: ModelQueries.get(Model.classType, 123, "abc"))
As the parameter ID takes a single String.

Also, since I can't use two owner type auth rules (https://github.com/aws-amplify/amplify-ios/issues/1590, https://github.com/aws-amplify/amplify-cli/issues/10110)
I need a way to retrieve data from other users. Is there a way to automatically sync down specific records from a model, anyone?

hanna-becker:
[...]

PR #1430, opened 2022-03-04
Feat/custom pk model identifier

HuiSF:
*Issue #, if available:*

#1426 

*Description of changes:*

- Deprecate `Model.getId`
  - Will make changes to `observeQuery` in DataStore which uses `getId()` in a separate PR
  - Required change in the API category will be followed up seperately
- Add `modelIdentifier` getter to `Model`
- Update unit test models
- Reconcile DataStore test models to `amplify_test` package
- Regenerate test models with codegen on this feature branch https://github.com/aws-amplify/amplify-codegen/tree/custom-pk-flutter (pending PR https://github.com/aws-amplify/amplify-codegen/pull/394)


packages/amplify_api/lib/src/graphql/paginated_result_impl.dart
```diff
@@ -28,6 +45,11 @@ class PaginatedResultImpl<T extends Model> extends PaginatedResult<T> {
     return '';
   }
 
+  @override
+  ModelIdentifier get modelIdentifier {
```
dnys1, line 49: Why would this not be a list of model identifiers?

packages/amplify_datastore/example/lib/models/BelongsToChildExplicit.dart
```diff
@@ -186,3 +190,38 @@ class _BelongsToChildExplicitModelType
...
+  const BelongsToChildExplicitModelIdentifier({required this.id});
+
+  Map<String, dynamic> serializeAsMap() => (<String, dynamic>{'id': id});
+
+  List<Map<String, dynamic>> serializeAsList() => serializeAsMap()
+      .entries
+      .map((entry) => (<String, dynamic>{entry.key: entry.value}))
+      .toList();
+
+  String serializeAsString() => serializeAsMap().values.join('#');
```
dnys1, line 211: These need `@override` annotations.

packages/amplify_datastore/example/lib/models/SimpleCustomType.dart
```diff
@@ -31,10 +31,10 @@ class SimpleCustomType {
     try {
       return _foo!;
     } catch (e) {
-      throw DataStoreException(
-          DataStoreExceptionMessages
+      throw AmplifyCodeGenModelException(
```
dnys1, line 34: This is a breaking change we should note

dnys1 (commented):
I could be missing it - I don't see any models using a custom primary key

packages/amplify_api/lib/src/graphql/paginated_result_impl.dart
dnys1: Or why does it need it at all?

packages/amplify_datastore/example/lib/models/SimpleCustomType.dart
HuiSF: Hum I missed this part somehow, I don't this this is an expected change, let me double check, could be just the test model is outdated (generated with WIP codegen of other features)

ragingsquirrel3: I believe this is expected change in codegen that is specific to > 0.4.0, though possible we did not note this change properly. However, I think any breakage was already with 0.3.x - 0.4.x

ragingsquirrel3 (commented):
LGTM, spoke offline about an integ test failure and want to resolve that before approving (or if non-issue with implementation and just tests, can be followup PR IMO).

packages/amplify_test/lib/test_models/Blog.dart
```diff
@@ -148,14 +152,13 @@ class Blog extends Model {
   }
 
   Blog copyWith(
-      {String? id,
```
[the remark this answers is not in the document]
HuiSF: By requirement, PK fields should be immutable within a "copy" factory/custructor/builder.

HuiSF:
> LGTM, spoke offline about an integ test failure and want to resolve that before approving (or if non-issue with implementation and just tests, can be followup PR IMO).

The integ test failure should be caused by the issue that we patched in 0.4.1. I will update the PR branch.

HuiSF:
> approved assuming integ test issue resolved w rebase

Thanks for the follow up @ragingsquirrel3 ! yes the integration tests all passed after merging the main branch.

packages/amplify_api/lib/src/graphql/paginated_result_impl.dart
HuiSF: Forgot to respond to this question. This dummy getter is added because `PaginatedResultImpl` extends from `Model`, added this following the original impl of `getId` in this class. Been syncing up with @ragingsquirrel3 regarding this part, this may be refactored when finalize custom PK support in API category.

dnys1 (approved):

Issue #815, opened 2021-08-18
DataStore Multi Auth


dnys1:
**Describe the bug**

Feature request: Multiple authorization modes in DataStore, as supported in [iOS](https://docs.amplify.aws/lib/datastore/setup-auth-rules/q/platform/ios/#configure-multiple-authorization-types) and [Android](https://docs.amplify.aws/lib/datastore/setup-auth-rules/q/platform/android/#configure-multiple-authorization-types).

matthewapps:
Hi, I've been wrestling with an auth issue for Flutter and datastore. This is the thread #9290[https://github.com/aws-amplify/amplify-js/issues/9290](url). Please can you help?

leadsizzle:
Hi @dnys1 , kindly would like to know if there could be any information on the approximate date range for multi-auth datastore support to be released in the next 12 months or so? I'm following this feature as well. Very much thanks in advance and I also appreciate everyone's performance here, you guys are great.

dorontal:
It would also be nice to also have an easy way to reconfigure Amplify at run-time, so that we can more easily take advantage of multi-auth. This is already possible in Javascript: to make witch your app to IAM mode, we use
```javascript
                    Amplify.configure({
                        aws_appsync_authenticationType: 'AWS_IAM'
                    });
```
or for Cognito auth mode
```javascript
                    Amplify.configure({
                        aws_appsync_authenticationType: 'AMAZON_COGNITO_USER_POOLS'
                    });
```
But that's Javascript. It would be nice to have an easy way to switch auth modes while your program is running with similar code in Dart.

dnys1:
This feature is available in preview as `0.4.4-rc.1`. To make use of it, pass in `authModeStrategy: AuthModeStrategy.multiAuth` to the constructor of the DataStore plugin.

bschulz87:
While setting the authModeStrategy using 0.4.4-rc.1 is possible, having different authorizationType for the same endpoint in amplifyconfig api.plugins.awsAPIPlugin does not work.

```
const amplifyconfig = ''' {
    "UserAgent": "aws-amplify-cli/2.0",
    "Version": "1.0",
    "api": {
        "plugins": {
            "awsAPIPlugin": {
                "api_AWS_IAM": {
                    "endpointType": "GraphQL",
                    "endpoint": "https://id.appsync-api.eu-central-1.amazonaws.com/graphql",
                    "region": "eu-central-1",
[...]

PR #1478, opened 2022-03-25
feat(datastore): Multi-auth

dnys1:
*Issue #, if available:*
#815 

*Description of changes:*
- Adds multi-auth option to DataStore configuration

### DIFF
--- a/packages/amplify_datastore_plugin_interface/lib/amplify_datastore_plugin_interface.dart
+++ b/packages/amplify_datastore_plugin_interface/lib/amplify_datastore_plugin_interface.dart
@@ -64,7 +64,7 @@ abstract class DataStorePluginInterface extends AmplifyPluginInterface {
     this.syncInterval,
     this.syncMaxRecords,
     this.syncPageSize,
-    this.authModeStrategy = AuthModeStrategy.default$,
+    this.authModeStrategy = AuthModeStrategy.defaultStrategy,
   }) : super(token: token);
 
   /// Internal use constructor
@@ -94,7 +94,7 @@ abstract class DataStorePluginInterface extends AmplifyPluginInterface {
     int? syncInterval,
     int? syncMaxRecords,
     int? syncPageSize,
-    AuthModeStrategy authModeStrategy = AuthModeStrategy.default$,
+    AuthModeStrategy authModeStrategy = AuthModeStrategy.defaultStrategy,
   }) {
     throw UnimplementedError('configureDataStore() has not been implemented.');
   }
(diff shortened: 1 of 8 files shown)


packages/amplify_datastore_plugin_interface/lib/amplify_datastore_plugin_interface.dart
```diff
@@ -60,12 +64,13 @@ abstract class DataStorePluginInterface extends AmplifyPluginInterface {
...
     this.syncMaxRecords,
     this.syncPageSize,
+    this.authModeStrategy = AuthModeStrategy.default$,
   }) : super(token: token);
 
   /// Internal use constructor
   @protected
   DataStorePluginInterface.tokenOnly({required Object token})
-      : super(token: token);
+      : this(token: token, modelProvider: null);
```
HuiSF, line 73: What's the reason changing this line?

packages/amplify_datastore_plugin_interface/lib/src/types/models/auth_mode_strategy.dart
```diff
@@ -0,0 +1,40 @@
...
+  /// them fail.
+  multiAuth,
+}
+
+/// Helpers for [AuthModeStrategy].
+extension AuthModeStrategyX on AuthModeStrategy {
+  /// The raw value used for interfacing with native SDKs.
+  String get rawValue {
+    switch (this) {
+      case AuthModeStrategy.default$:
```
HuiSF, line 34: It looks like that the codebase mostly uses `describeEnum` as convention. But using `describeEnum` doesn't work with this escaped keyword...
Do we have option to name `default` as `defaultAuth`, or use uppercase like Android?
Or do developers need to use `AuthModeStrategy.default` at any occasion instead of passing `null`?

HuiSF (approved):
lgtm. 
With couple of questions.

packages/amplify_datastore_plugin_interface/lib/amplify_datastore_plugin_interface.dart
dnys1: It's identical to before, just saves me repeating myself again. Alternative is:

```dart
  DataStorePluginInterface.tokenOnly({required Object token})
      : authModeStrategy = AuthModeStrategy.default$,
        super(token: token);
```

packages/amplify_datastore_plugin_interface/lib/src/types/models/auth_mode_strategy.dart
dnys1: Yeah, since `default` is a reserved word you can't use it as a member name. `defaultAuth` is fine, but it seemed redundant to say auth again. 

Personally, I don't like using `null` in this context if there is a default value hidden behind it.

Jordan-Nelson: I might make this `AuthModeStrategy.defaultValue` or `AuthModeStrategy.defaultStrategy` even though it is redundant.

My brain read this as `AuthModeStrategy.$default` and I was wondering if this was generated.

Jordan-Nelson (approved):
I left a minor comment, but looks good.

PR #1485, opened 2022-03-29
chore: android codebase maintenance

HuiSF:
*Issue #, if available:*

*Description of changes:*

Enable ktlint for additional packages, to capture issues such as "Call requires API level 24 (current min is 21)". Detailed changes see commits list of this PR.

Notes
* Most code changes were done automatically by `./gradlew ktlintFormat`, Import wild cards complaints were fixed manually
* Updated assertion for unit test `test_operation_with_invalid_api_name` in `GraphQLApiUnitTests` as the original assertion requires a very specific jvm env, and fails easily when configuration is different
* Bump AGP plugin version all the way to 7.1.2 (not a huge change as there was not 5.x and 6.x in between) and relevant cleaning up
* Enable kitlint, which will capture any usage of APIs that are added in the version > API 21 (`minSdkVersion` in android plugins)
* Android integration tests passed on this PR branch
  image


packages/amplify/amplify_flutter_android/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -4,4 +4,4 @@ zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
 
 # (nydillon) Must be <=6.1.1 until this issue is resolved: https://github.com/gradle/gradle/issues/14132
```
dnys1, line 6: Double check this is not still an issue when running integration tests

dnys1 (approved):
Well done. Thank you for making this a priority ✨

packages/amplify/amplify_flutter_android/android/gradle/wrapper/gradle-wrapper.properties
HuiSF: Thanks for calling this out. I'm keeping a couple of issues in mind that we put in temporary solutions earlier due to not able to upgrade gradle. 

I will do some more testing on the linter, then address relevant TODOs.

packages/amplify_datastore/android/proguard-rules.pro
```diff
@@ -1 +0,0 @@
--keep class io.reactivex.rxjava3.** { *; }
```
dnys1, line 1: This was needed to prevent a crash: https://github.com/aws-amplify/amplify-flutter/pull/1064

packages/amplify_datastore/example/android/app/build.gradle
```diff
@@ -60,4 +61,9 @@ flutter {
 
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
```
dnys1, line 63: This was removed in other gradle files - needed here?

dnys1 (approved):

packages/amplify_datastore/example/android/app/build.gradle
HuiSF: Oh darn missed scanning on the example Apps 😅 thanks I will add a commit to cleaning this up

packages/amplify_datastore/android/proguard-rules.pro
HuiSF: Yea according to the comment link you added in the original fix: https://stackoverflow.com/a/64506880/12626712
This should no longer be an issue when using gradle >= v6.5, we upgraded to 7.2.

dnys1: Sweet, thanks!

PR #1499, opened 2022-04-04
chore: merge amplify-ios-patch branch

HuiSF:
*Issue #, if available:*

*Description of changes:*

`amplify-ios-patch` branch is behind `main` which didn't have the commit migrating `amplify_flutter` package to federated plugin. Hence this PR bumps the version of these new packages but they are not published as a part of version `0.4.3`. Everything will be synced again when release `0.4.4`.


example/ios/Podfile
```diff
@@ -37,5 +37,8 @@ end
 post_install do |installer|
   installer.pods_project.targets.each do |target|
     flutter_additional_ios_build_settings(target)
+    target.build_configurations.each do |config|
+       config.build_settings.delete 'IPHONEOS_DEPLOYMENT_TARGET'
```
Jordan-Nelson, line 41: Why is this added? Is some change in the new version of amplify-iOS requiring this?

HuiSF: Forgot to explain -

The realtime sync iOS dependency seems introduced usage of `Combine` which requires deployment target 11.0. In the newly added federated plugin example Apps, their deployment target was 8.0 be default. Hence it caused [build failure](https://app.circleci.com/pipelines/github/aws-amplify/amplify-flutter/3073/workflows/9cf764f9-c04c-4b0e-a383-f03ba9684c50/jobs/17875). I had to modify the `Podfile`s to make the build pass.

Also I'm not sure why, the example App for Auth pulls DataStore plugins from amplify-ios, which requires deployment target 13.0, while Auth plugin requires only 11.0. Due to this I have to [packages/auth/amplify_auth_cognito_ios/example/ios/Runner.xcodeproj/project.pbxproj](https://github.com/aws-amplify/amplify-flutter/pull/1499/files#diff-5f87f842a6393af1c46580e5aaf80efd7e651d8281b00202941b577cfa8b968b) for the Auth category.

Jordan-Nelson: > Also I'm not sure why, the example App for Auth pulls DataStore plugins from amplify-ios

Hm. That seems a little concerning that we don't know why it is pulling in deps from DS.

Jordan-Nelson: Do we have an idea if that is because of a change in Amplify-iOS?

Jordan-Nelson: If any of our plugins that were supporting 11, now only support 13+, we need to call that out (and probably make this breaking). Do we know if that is that case?

HuiSF: `Amplify` in amplify-ios is still targeting 11.0. I think the issue is on the unit tests host Apps and example Apps settings. I haven't got time to look into it.

Jordan-Nelson: Yeah, I see that `Amplify` in amplify-ios still has 11.0 listed. I am confused by this error message in the build failure though.

```
module 'Amplify' has a minimum deployment target of iOS 13.0
```

HuiSF: It's noticeable that the auth example App Podfile is targeting platform 13.0. 
The built frameworks may be messed up.

Jordan-Nelson (approved):

Issue #1504, opened 2022-04-05
FlutterJNI was detached from native C++


crawford-jake:
### Description

When I configure Amplify using a custom FunctionAuthProvider, the method channel fails to call getLatestAuthToken when the app state changes. This does not occur for api authorizations that do not use the FunctionAuthProvider class. Seemingly because the method channel call in FlutterAuthProvider.kt is avoided.

When the app reopens from a state change ```Tried to send a platform message to Flutter, but FlutterJNI was detached from native C++``` is thrown on all API requests. I have confirmed that the method channel call to retrieve the custom token fails. 

Our QA team has reported the same issue coming from their iOS devices as well, however it is easiest to reproduce in Android.

### Categories

- [ ] Analytics
- [ ] API (REST)
- [X] API (GraphQL)
- [ ] Auth
- [ ] Authenticator
- [ ] DataStore
- [ ] Storage

### Steps to Reproduce

# Steps to Reproduce
1. Create a new flutter project on the stable channel
2. Install amplify_flutter 0.4.3 and amplify_api 0.4.3
3. Configure Amplify with FunctionAuthProvider and setup working Amplify Query
4. Break Amplify

## 1. Create a new flutter project on the stable channel
**Run the following commands in the terminal**
 ```flutter channel stable```
 ```flutter create amplify_bug```
 ```cd amplify_bug```
 ```flutter pub cache clean```
 ```flutter pub get```
 
 **Setup iOS for Amplify**
  Modify the ios/Podfile and replace the second line with: ```platform :ios, '11.0'```
  ```cd ios && pod install```
  
 **Setup Android for Amplify**
  Modify the android/app/build.gradle to the minSdkVersion 21
```
defaultConfig {
    minSdkVersion 21
}
```

**Confirm the project runs on the iOS simulator and Android emulator**
```flutter pub get```
```flutter run``` 
 
 ## 2. Install amplify_flutter 0.4.3 and amplify_api 0.4.3
 Open pubspec.yaml and add the following dependencies
```
amplify_flutter: 0.4.3
amplify_api: 0.4.3
```
Then confirm that these packages run 
 ```flutter pub get```
```flutter run``` 

## 3. Configure Amplify with FunctionAuthProvider and setup working Amplify Query
**Create a new .dart file in the lib directory, and copy the following. (To see this render your will need to set this as the home widget in main.dart)**
 ```
 import 'package:amplify_api/amplify_api.dart';
[...]

PR #1509, opened 2022-04-06
fix(api): JNI issues

dnys1:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-flutter/issues/1504

*Description of changes:*
Handles closure of method and event channels when sending apps to the background and adds reattach logic for each


packages/amplify_api/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/AmplifyApiPlugin.kt
```diff
@@ -15,7 +15,6 @@
 
 package com.amazonaws.amplify.amplify_api
```
haverchuck, line 16: You don't need `import com.amplifyframework.api.aws.AuthorizationType`?

haverchuck (approved):

HuiSF (approved):

PR #1516, opened 2022-04-07
chore: changelog updates, api example fixes and gitignores

haverchuck:
*Description of changes:*
Updates changelogs to match release branch. Also fixes api example apps with Flutter deps and gitignore files.


dnys1 (approved):
We should formalize what goes where in the changelogs at some point. Not a huge deal, but I would personally rearrange some things:

- ios/android/platform interface packages only show changes made to their packages, the app-facing package shows changes to all
- Likewise, the amplify_flutter package (and only this package) should show all changes to all packages since this is imported by *every* Amplify project

Jordan-Nelson (approved):

haverchuck:
@dnys1 One hundred percent agree, and I think what you proposed could be automated.

Issue #842, opened 2021-08-27
DataStore query throws Exception with empty QueryPagination on Android


Jordan-Nelson:
**Describe the bug**
QueryPagination() has no required arguments. `page` is optional and `limit` defaults to 100. However, the query will fail on Android if no value page is supplied. `page` should either be required or default to 0. On iOS, the first page is returned if no page is supplied.

**To Reproduce**
Below is an integration test which will throw an exception on Android.

```dart
group('pagination', () {
  var models = List.generate(1000, (i) => Blog(name: 'blog $i'));
  setUp(() async {
    await configureDataStore();
    // clear data before each test
    await clearDataStore();
    for (var model in models) {
      await Amplify.DataStore.save(model);
    }
  });
  testWidgets('should return the models for the given page and limit',
      (WidgetTester tester) async {
    var blogs = await Amplify.DataStore.query(Blog.classType,
        pagination: QueryPagination());
  });
});
```

**Expected behavior**
`page` should either be required or default to 0. If `page` is not required, the query should not throw an exception without a supplied value.

**Screenshots**

**Platform**
Amplify Flutter current supports iOS and Android.  This issue is reproducible in (check all that apply):
[x] Android
[] iOS

cshfang:
Should be fixed by #1533

PR #1533, opened 2022-04-16
fix(core): Update QueryPagination page field to default to 0

cshfang:
*Issue #, if available:* #842

*Description of changes:* This PR updates the default constructor for `QueryPagination`s to supply a default value of `0` for the `page` field. The Kotlin version of the `QueryPaginationBuilder` checks to see if the serialized map passed to it contains the `page` key, but since the `serializeAsMap()` method *always* adds the page key, a `NullPointerException` can arise as a result.

Another fix considered was to update the `serializeAsMap()` method to only add the `page` key if page is defined. This would be a reasonable fix but only when coupled with the current `QueryPaginationBuilder` utilities. Opting for the default value would not require the "is the `page` key defined check.
* Also of note - Rather than removing these checks from the utilities, opting to keep them there as they are, in theory, directly usable outside of Flutter method channel context as well.

### DIFF
--- a/packages/amplify_core/lib/src/types/query/query_pagination.dart
+++ b/packages/amplify_core/lib/src/types/query/query_pagination.dart
@@ -16,10 +16,10 @@
 part of 'query_field.dart';
 
 class QueryPagination {
-  final int? page;
+  final int page;
   final int limit;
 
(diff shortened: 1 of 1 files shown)


dnys1 (approved):
Thanks!

PR #1545, opened 2022-04-22
chore(authenticator): Fix failing integration tests

dnys1:
*Issue #, if available:*

*Description of changes:*
- Updates integration tests to fix failures
- Adds one new test for force new password


packages/amplify_authenticator/lib/src/mixins/authenticator_username_field.dart
```diff
@@ -57,7 +57,6 @@ mixin AuthenticatorUsernameField<FieldType,
         };
       case UsernameType.phoneNumber:
         return (input) {
-          state.phoneNumber = input.username;
```
Jordan-Nelson, line 60: Why is this being removed? If the username type is phone number, shouldn't the phone number be updated?

packages/amplify_authenticator/lib/src/state/authenticator_state.dart
```diff
@@ -454,11 +454,8 @@ class AuthenticatorState extends ChangeNotifier {
   }
 
   /// Change to a new step in the authentication flow
-  void changeStep(AuthenticatorStep step, {bool resetAttributes = true}) {
+  void changeStep(AuthenticatorStep step) {
     _authBloc.add(AuthChangeScreen(step));
-
-    /// Clean [ViewModel] when user manually navigates widgets
-    if (resetAttributes) _resetAttributes();
```
Jordan-Nelson, line 461: Is this change related to an issue in one of the tests?

packages/amplify_authenticator/lib/src/mixins/authenticator_username_field.dart
dnys1: It screws things up to set a phone number attribute when the username is the same phone number. Not sure why, but the force password test will fail with this line.

packages/amplify_authenticator/lib/src/state/authenticator_state.dart
dnys1: This was just clean up. It wasn't used anywhere.

Jordan-Nelson: `resetAttributes` was defaulting to true. If that option is not used, shouldn't just the if statement be removed?

```dart
void changeStep(AuthenticatorStep step) {
  _authBloc.add(AuthChangeScreen(step));
}
```

packages/amplify_authenticator/lib/src/mixins/authenticator_username_field.dart
dnys1: https://github.com/aws-amplify/amplify-flutter/pull/1545/files#diff-9ae1ea7228cafb6c75962a25ba589ed38fcc6aa278c3ad8a43c6571ee0778f41R169

Jordan-Nelson: Okay. I think we should figure out why it is failing. Was there an error message from Cognito?

Jordan-Nelson: I don't think removing this is that big of a deal since IIRC this was mostly a UX improvement to make it so the phone number is already populated when switching sign in modes, but it would be nice to know why it is failing, and if setting the email when email is the username is also an issue.

dnys1: Yeah, Cognito throws an error regarding parameter validation or something. And yeah, are other integ tests for phone number username which validate this behavior is still correct.

Setting email for username=email probably does the same thing.

Jordan-Nelson: This introduces a somewhat edge case bug for apps configured to use email OR phone number. Here are the repro steps:
1. User selects the email option on the sign up screen
2. User types phone number A (555-555-5555)
3. User switches to phone number based sign up
4. User enters phone number B (123-123-4444) and attempts to sign up
5. Sign up fails because the "Phone Number" attribute does not match the username

We could resolve this by clearing email/phone when the username type is changed. I think we tried to provide a nice UX where the username value was automatically updated to equal the phone number when the username type was changed to phone number (and same for email), but clearing email/phone/username when the user switches might make things more straight forward.

Whatever we do for phone number, we should replicate for email

dnys1: I think clearing the attributes makes sense when the type changes. I will update the PR and do the same for email. Good catch!

dnys1: This should be fixed now 👍

Jordan-Nelson (approved):

PR #1552, opened 2022-04-26
chore: migrates auth types (core migration 2)

haverchuck:
*Description of changes:*
Moves auth types to core.

Depends on #1550 


packages/auth/amplify_auth_cognito/lib/src/types.dart
```diff
@@ -46,32 +46,32 @@ export 'CognitoAttribute/CognitoUpdateUserAttributesOptions.dart';
...
-export 'package:amplify_auth_plugin_interface/src/Exceptions/SignedOutException.dart';
-export 'package:amplify_auth_plugin_interface/src/Exceptions/SoftwareTokenMFANotFoundException.dart';
-export 'package:amplify_auth_plugin_interface/src/Exceptions/TooManyFailedAttemptsException.dart';
-export 'package:amplify_auth_plugin_interface/src/Exceptions/TooManyRequestsException.dart';
-export 'package:amplify_auth_plugin_interface/src/Exceptions/UnknownException.dart';
-export 'package:amplify_auth_plugin_interface/src/Exceptions/UserCancelledException.dart';
-export 'package:amplify_auth_plugin_interface/src/Exceptions/UsernameExistsException.dart';
-export 'package:amplify_auth_plugin_interface/src/Exceptions/UserNotConfirmedException.dart';
-export 'package:amplify_auth_plugin_interface/src/Exceptions/UserNotFoundException.dart';
+// exception_s
```
Jordan-Nelson, line 49:
```suggestion
// exceptions
```

Jordan-Nelson (approved):

packages/auth/amplify_auth_cognito/lib/amplify_auth_cognito.dart
```diff
@@ -16,28 +16,21 @@
...
-  static final Object _token = Object();
-
   /// Constructs a AmplifyAuthCognito plugin
-  AmplifyAuthCognito() : super(token: _token);
+  AmplifyAuthCognito();
 
   static AmplifyAuthCognito _instance = AmplifyAuthCognitoMethodChannel();
   static AuthStreamController streamWrapper = AuthStreamController();
 
-  static set instance(AuthPluginInterface instance) {
```
dnys1, line 37: So we don't need this necessarily, but we do need a way to change where calls go depending on the platform. The instance + Dart's plugin registration system is one way, or you can have a factory method like this:

```dart
factory AmplifyAuthCognito() {
  if (zIsWeb || Platform.isMacOS || Platform.isWindows || Platform.isLinux) {
    return AmplifyAuthCognitoDart();
  } else {
    return AmplifyAuthCognitoMethodChannel();
  }
}
```

and then you can make this class abstract and remove all methods.

dnys1 (approved):

packages/auth/amplify_auth_cognito/lib/amplify_auth_cognito.dart
haverchuck: Yup - will bring this in once we actually have the Dart implementation.

PR #1568, opened 2022-05-02
chore: enable dependabot

ragingsquirrel3:
This PR adds needed code changes to enable dependabot for dart projects. See https://github.blog/changelog/2022-04-05-pub-beta-support-for-dependabot-version-updates/ as well as linked thread where there are some questions about monorepos. Per that thread, I believe we need separate entries for all the pubspec locations, so I made a little script to generate that as I expect it to need updates when directory structure changes.


dnys1 (approved):
Looks good. Script could be cleaned up but not a major issue IMO. Can rewrite in Dart if you want and use `Directory.list(recursive: true)` to make it future proof.

haverchuck (approved):
This is any annoying nit, but could we put the 'updates' block in alphabetical order? 
Otherwise looks good to me.

ragingsquirrel3:
> This is any annoying nit, but could we put the 'updates' block in alphabetical order? Otherwise looks good to me.

I changed the generation script to sort the output alphabetically.

ragingsquirrel3:
> Looks good. Script could be cleaned up but not a major issue IMO. Can rewrite in Dart if you want and use `Directory.list(recursive: true)` to make it future proof.

Good suggestion, ended up converting to a dart script to do this plus some other things.

PR #1572, opened 2022-05-04
fix(aws_signature_v4): Various fixes

dnys1:
*Issue #, if available:*

*Description of changes:*
- Fixes a problem with S3 path normalization (S3 paths should be un-normalized)
- Fixes a problem with S3 path encoding. S3 path segments should be encoded once, every other service expects double-encoding. The confusion was that the test suite expects single-encoding like S3.
- Fixes a problem with S3 options (chunked/signPayload). Adds tests to verify permutations work as expected.
- Removes `Host` and `Content-Length` headers on Web after signing to prevent browser errors warning against setting these manually

### DIFF
--- a/packages/aws_signature_v4/lib/src/configuration/services/s3.dart
+++ b/packages/aws_signature_v4/lib/src/configuration/services/s3.dart
@@ -77,6 +77,7 @@ class S3ServiceConfiguration extends BaseServiceConfiguration {
         super(
           normalizePath: false,
           omitSessionToken: false,
+          doubleEncodePathSegments: false,
         );
 
   int _calculateContentLength(AWSBaseHttpRequest request, int decodedLength) {
--- a/packages/aws_signature_v4/lib/src/configuration/services/s3.dart
+++ b/packages/aws_signature_v4/lib/src/configuration/services/s3.dart
@@ -50,8 +50,12 @@ class S3ServiceConfiguration extends BaseServiceConfiguration {
   // 64 KB is the recommended chunk size.
   static const int _defaultChunkSize = 64 * 1024;
 
-  static const _chunkedPayloadSeedHash = 'STREAMING-AWS4-HMAC-SHA256-PAYLOAD';
-  static const _unsignedChunkedPayloadSeedHash = 'UNSIGNED-PAYLOAD';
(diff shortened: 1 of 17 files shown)


HuiSF (approved):

packages/aws_signature_v4/lib/src/configuration/services/s3.dart
```diff
@@ -73,14 +77,18 @@ class S3ServiceConfiguration extends BaseServiceConfiguration {
     this.chunked = false,
     int chunkSize = _defaultChunkSize,
     this.encoding = S3PayloadEncoding.none,
-  })  : chunkSize = max(chunkSize, _minChunkSize),
+  })  : assert(
+          signPayload || !chunked,
+          'S3 does not accept unsigned, chunked payloads',
```
HuiSF, line 82: 👍 nice to save a network round trip

nit: Would `!signedPayload && chunked` be clearer?

dnys1: That's the check but we need to assert the converse is true

packages/aws_signature_v4/lib/src/request/canonical_request/canonical_path.dart
```diff
@@ -0,0 +1,48 @@
...
+  /// Returns the normalized path with double-encoded path segments.
+  ///
+  /// Uses [url] to normalize the path.
+  static String canonicalize(
+    final String path, {
+    required ServiceConfiguration serviceConfiguration,
+  }) {
+    var canonicalPath = path;
+
+    // `normalize` removes leading and trailing slashes which should be
```
Jordan-Nelson, line 30: > leading and trailing slashes

Outside of the scope of this PR since this is just a move, but it looks like only trailing slashes is handled below?

Jordan-Nelson (approved):

packages/aws_signature_v4/lib/src/request/canonical_request/canonical_path.dart
dnys1: No worries - appreciate the callout. Both are handled, though, with `ensureStartsWith` and `ensureEndsWith`

PR #1573, opened 2022-05-05
chore(aws_common): Crypto-secure UUID

dnys1:
*Issue #, if available:*
- Pentest discovered that `package:uuid` is not crypto-secure by default.

*Description of changes:*
- Adds a top-level helper for retrieving crypto-secure UUIDs, to be used in Amplify Dart code

### DIFF
--- a/packages/aws_common/lib/src/util/uuid.dart
+++ b/packages/aws_common/lib/src/util/uuid.dart
@@ -12,26 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-/// Common types and utilities used across AWS and Amplify packages.
-library aws_common;
+import 'package:uuid/uuid.dart';
+import 'package:uuid/uuid_util.dart';
 
-// Collection
-export 'src/collection/case_insensitive.dart';
-
-// Config
-export 'src/config/aws_service.dart';
-
-// HTTP
-export 'src/http/aws_headers.dart';
-export 'src/http/aws_http_method.dart';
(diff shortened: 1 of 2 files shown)


ragingsquirrel3 (approved):

Jordan-Nelson (approved):

packages/aws_common/lib/src/util/uuid.dart
```diff
@@ -0,0 +1,25 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'package:uuid/uuid.dart';
+import 'package:uuid/uuid_util.dart';
+
+/// Generates a crypto-secure UUID (v4).
+String uuid() => const Uuid().v4(
```
Jordan-Nelson, line 19: Is the expectation that all current uses of Uuid().v4 are replaced with this? Is there a downside to using this in locations where a non crypto secure uuid would be sufficient?

dnys1: > Is the expectation that all current uses of Uuid().v4 are replaced with this?
Yes, at least in Smithy/AWS we'll have to. I think we should also replace it throughout Amplify.

> Is there a downside to using this in locations where a non crypto secure uuid would be sufficient?
Maybe a slight performance hit. The converse, though, is that there is a security risk if we use it incorrectly.

PR #1576, opened 2022-05-09
chore: core-migration-8 (cleanup-1)

haverchuck:
**Do not review yet.**

*Description of changes:*
Cleans up migration code.


packages/amplify_core/lib/src/types/storage/download_file/download_file_request.dart
```diff
@@ -31,7 +31,7 @@ class DownloadFileRequest {
   Map<String, Object?> serializeAsMap() => {
         if (options != null) 'options': options!.serializeAsMap(),
         'key': key,
-        'local': local,
+        'path': local.absolute.path,
```
haverchuck, line 34: [Native code ](https://github.com/aws-amplify/amplify-flutter/blob/aa8fd3b3bf3b3420a0e7ac60b1867e64550910d2/packages/storage/amplify_storage_s3_android/android/src/main/kotlin/com/amazonaws/amplify/amplify_storage_s3/types/FlutterDownloadFileRequest.kt#L28)seems to require `path` and not `local`, right?

packages/amplify_core/lib/src/types/plugin/amplify_datastore_plugin_interface.dart
```diff
@@ -61,6 +61,10 @@ abstract class DataStorePluginInterface extends AmplifyPluginInterface {
     this.authModeStrategy = AuthModeStrategy.defaultStrategy,
   });
 
+  /// Internal use constructor
+  @protected
+  DataStorePluginInterface.emptyConstructor() : this(modelProvider: null);
```
haverchuck, line 66: Not sure if this is a great name. Open to suggestions here.

packages/amplify_core/lib/src/types/storage/download_file/download_file_request.dart
HuiSF: Yes it looks like that the currently implementation is wrong, as it doesn't respect API consumer provided `File` spec... 😅

packages/amplify_core/lib/src/types/plugin/amplify_datastore_plugin_interface.dart
HuiSF: Curious what's the use case of this constructor?

haverchuck: Previously there was a constructor called tokenOnly, which did not accept a modelProvider. The token was the verification token from the `[PlatformformInterface](https://pub.dev/documentation/plugin_platform_interface/latest/plugin_platform_interface/PlatformInterface-class.html)` class. I wasn't sure if we still needed a named constructor, or if the default would be ok.

packages/amplify_core/test/config/unknown_test.dart
```diff
@@ -35,7 +36,7 @@ void main() {
       };
 
       final config = AmplifyConfig.fromJson(json);
-      const expected = AmplifyConfig(
+      var expected = AmplifyConfig(
```
Jordan-Nelson, line 39: *nit:* If this cannot be `const` I would make it `final`
```suggestion
      final expected = AmplifyConfig(
```

Jordan-Nelson (approved):

packages/amplify_core/lib/src/types/plugin/amplify_datastore_plugin_interface.dart
dnys1: I think we should ditch this in all but DataStore. Even in DataStore, the plugin architecture should be reworked so we don't need this (if I remember it doesn't really do anything anyway), but that can happen later.

packages/amplify_core/test/config/unknown_test.dart
```diff
@@ -13,7 +13,8 @@
 // permissions and limitations under the License.
 //
 
-import 'package:amplify_flutter/amplify_flutter.dart';
+import 'package:amplify_core/amplify_core.dart';
+import 'package:amplify_core/src/types/config/amplify_plugin_config.dart';
```
dnys1, line 17: Let's make sure this is being exported from `amplify_core`

PR #1585, opened 2022-05-12
chore(ci): Update melos configuration

dnys1:
*Issue #, if available:*

*Description of changes:*
This change takes advantage of a new feature in `pub` called `pubspec_overrides` which allows for local overrides to the `pubspec.yaml` while developing without affecting publishing. This works by not checking in the `pubspec_overrides.yaml` file. As of the latest melos version, it will automatically generate this file for you.

The benefit of this approach is that `dart`/`flutter pub get` will correctly work now! 🎉

### DIFF
--- a/packages/aws_common/pubspec_overrides.yaml
+++ b/packages/aws_common/pubspec_overrides.yaml
@@ -1,0 +1,4 @@
+# melos_managed_dependency_overrides: amplify_lints
+dependency_overrides:
+  amplify_lints:
+    path: ../amplify_lints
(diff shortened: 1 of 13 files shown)


packages/aws_common/pubspec_overrides.yaml
```diff
@@ -0,0 +1,4 @@
+# melos_managed_dependency_overrides: amplify_lints
+dependency_overrides:
```
dnys1, line 2: These are tracked until mono_repo support expands or the pkgs are integrated into CircleCI

dnys1:
This functionality doesn't seem quite ready for our repo & doesn't play nice with mono_repo yet. Will track the status of the feature and continue experimenting with it.

Issue #1567, opened 2022-05-02
Bug when using an enum to filter an API query to list objects - expects Int instead of Enum.


djsjr:
### Description

When an Enum is used in a GraphQL schema, and is one of the items of a particular model, we should be able to filter by a certain value of the enum when querying for a list of objects. However, this results in a runtime error. For some reason, it expects an Int instead (which of course results in an empty list. 

This is only true for the API plugin. The same query done with Datastore is successful.

### Categories

- [ ] Analytics
- [ ] API (REST)
- [X] API (GraphQL)
- [ ] Auth
- [ ] Authenticator
- [ ] DataStore
- [ ] Storage

### Steps to Reproduce

1. create example schema:
```
enum Size {
  SMALL
  MEDIUM
  LARGE
}

type Todo @model @auth(rules: [{allow: private, operations: [read]}]) {
  id: ID!
  size: Size!
}
```
2. make example query: 
```
     final request = ModelQueries.list(Todo.classType,
          where: Todo.SIZE.eq(Size.SMALL));
      final response = await Amplify.API.query(request: request).response;
```
3. receive runtime error:
```[VERBOSE-2:ui_dart_state.cc(209)] Unhandled Exception: Invalid argument: Instance of 'Size'```

### Platforms

- [X] iOS
- [X] Android

### Android Device/Emulator API Level

API 32+

### Environment

```bash
[✓] Flutter (Channel stable, 2.10.3, on macOS 12.0.1 21A559 darwin-arm, locale en-US)
[✓] Android toolchain - develop for Android devices (Android SDK version 32.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 13.3.1)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2021.1)
[✓] VS Code (version 1.66.0)
[✓] Connected device (3 available)
    ! Error: iPhone is not connected. Xcode will continue when iPhone is connected. (code -13)
    ! Error: iPhone is busy: Waiting to connect and unlock the device. Xcode will continue when iPhone is finished. (code -10)
[✓] HTTP Host Availability

• No issues found!
```

### Dependencies

```bash
Dart SDK 2.16.1
Flutter SDK 2.10.3
see_app 2.5.5+66

dependencies:
- adaptive_dialog 1.6.2+1 [animations collection flutter intersperse macos_ui meta]
[...]

PR #1595, opened 2022-05-16
fix(api): support enums in query predicates for model helpers

ragingsquirrel3:
Issue #1567 

Model helpers for GraphQL accepts query predicates for some requests. The predicates get serialized into JSON objects which are included as request variables at runtime. There is a nasty error in the library when someone passes an enum into the query predicates. This small PR just calls `describeEnum` to convert them to a string.

Note that there is still a CLI issue at play here https://github.com/aws-amplify/amplify-cli/issues/10370 which makes the compiled schema look for int values instead of strings. That has been identified as a bug and this PR fixes the amplify-flutter end of it and will work if either a) the bug in CLI is fixed b) the appsync compiled schema is manually modified to accept strings.

### DIFF
--- a/packages/api/amplify_api/test/query_predicate_graphql_filter_test.dart
+++ b/packages/api/amplify_api/test/query_predicate_graphql_filter_test.dart
@@ -2,6 +2,7 @@ import 'package:amplify_api/amplify_api.dart';
 import 'package:amplify_api/src/graphql/graphql_request_factory.dart';
 import 'package:amplify_core/amplify_core.dart';
 import 'package:amplify_test/test_models/ModelProvider.dart';
+import 'package:flutter/foundation.dart';
 import 'package:flutter_test/flutter_test.dart';
 
 enum Size { SMALL, MEDIUM, LARGE }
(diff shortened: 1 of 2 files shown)


packages/api/amplify_api/test/query_predicate_graphql_filter_test.dart
```diff
@@ -209,5 +211,17 @@ void main() {
       _testQueryPredicateTranslation(queryPredicate, expectedFilter,
           modelType: Post.classType);
     });
+
+    test('query with enum should serialize to string', () {
+      // Note: enums are not actually in the codegen model. Nonetheless, we
+      // expect this predicate to translate as follows.
+      final queryPredicate = Post.TITLE.eq(Size.MEDIUM);
+      final expectedFilter = {
+        'title': {'eq': 'MEDIUM'}
```
HuiSF, line 220: nit: use `describeEnum` as well to ensure the same behavior

HuiSF (approved):

packages/api/amplify_api/test/query_predicate_graphql_filter_test.dart
ragingsquirrel3: good call, will change

dnys1 (approved):

Issue #1565, opened 2022-05-02
DataStore sync query fails in Android when using sync expression with the model id field


HuiSF:
### Description

This issue is identified in the discussion of #1501 (but this is not the cause of the original issue described in #1501).

Take schema

```graphql
type RegularParentModel @model {
  id: ID!
  title: String
  content: String
  children: [RegularChildModel] @hasMany
}
```

Using sync expression with DataStore to prevent sync any data of this model
```dart
final datastore = AmplifyDataStore(
  modelProvider: ModelProvider.instance,
  syncExpressions: [
    DataStoreSyncExpression(
      RegularParentModel.classType,
      () => RegularParentModel.ID.eq(null),
    ),
  ],
);
```

**The base sync triggered sync query failed in Android.**
Sync query request.
`regularParentModel.id` should be wrong here.
```json
{
	"query": "query SyncRegularParentModels($filter: ModelRegularParentModelFilterInput, $limit: Int) {\n  syncRegularParentModels(filter: $filter, limit: $limit) {\n    items {\n      __typename\n      _deleted\n      _lastChangedAt\n      _version\n      children {\n        items {\n          __typename\n          _deleted\n          _lastChangedAt\n          _version\n          createdAt\n          id\n          title\n          updatedAt\n        }\n        nextToken\n        startedAt\n      }\n      content\n      createdAt\n      id\n      title\n      updatedAt\n    }\n    nextToken\n    startedAt\n  }\n}\n",
	"variables": {
		"filter": {
			"and": [{
				"regularParentModel.id": {
					"eq": null
				}
			}]
		},
		"limit": 1000
	}
}
```
Response
```json
{
	"data": null,
	"errors": [{
		"path": null,
		"locations": null,
		"message": "The variables input contains a field name 'regularParentModel.id' that is not defined for input object type 'ModelRegularParentModelFilterInput' "
	}]
}
```

In the meantime, the exactly same sync expression works in iOS platform.
Sync query request
```json
{
	"variables": {
		"limit": 1000,
		"filter": {
			"and": [{
				"id": {
					"eq": null
				}
			}]
		}
	},
[...]

PR #1600, opened 2022-05-18
fix(datastore): fix invalid model id field name implication

HuiSF:
*Issue #, if available:*

https://github.com/aws-amplify/amplify-flutter/issues/1565

*Description of changes:*

Add logic in Android implement removing `<modelName>` from `<modelName>.id` when creating native query predicate.

* The same logic exists in [iOS implementation](https://github.com/aws-amplify/amplify-flutter/blob/5c6d8eab63c1ea11e63d1ec8e82a975a4dffeb28/packages/amplify_datastore/ios/Classes/types/query/QueryPredicateBuilder.swift#L23-L29)
* The field name `<modelName>.id` for a model id file is redundant and unnecessary and breaks query predicate usage outside of SQLite in Android 
* This issue didn't break `DataStore.query` because Android generated SQL command happened to compatible with `<modelName>.id`

**NOTE:** This PR resolves the issue identified in Android, and ensures backwards compatibility for model generated wrongly.

PR to amplify-codegen to correct this behavior: https://github.com/aws-amplify/amplify-codegen/pull/429

----


packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/query/QueryPredicateBuilder.kt
```diff
@@ -144,5 +144,14 @@ class QueryPredicateBuilder {
         fun fromSerializedMap(serializedMap: Map<String, Any>?): QueryPredicate? {
             return fromSerializedMap(serializedMap, null)
         }
+
+        @JvmStatic
+        fun extractIdFieldName(fieldName: String): String {
```
dnys1, line 149: Is this just to be backwards compatible with old codegen? The codegen fix alone should work, right?

If it's to be backwards compatible? Should we have a test which verifies this old behavior?

dnys1 (approved):

Jordan-Nelson (approved):

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/query/QueryPredicateBuilder.kt
HuiSF: Yes. This is for backwards compatibility.
I'll add a test to assert the original behavior.

HuiSF:
Update:

* Removed extracting id field name logics from both iOS and Android implementation
* Implemented the backwards compatibility guard in flutter implementation

PR #1608, opened 2022-05-20
chore: add github action to lint PR titles

Jordan-Nelson:
*Issue #, if available:* N/A

*Description of changes:*
- Add a GitHub action to lint PR titles according to conventional commits

This was validated in a fork. [See PR 10-14](https://github.com/Jordan-Nelson/amplify-flutter/pulls) for several good PR titles that pass CI and one bad title that fails.

### DIFF
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -13,6 +13,7 @@ jobs:
     name: Validate PR title
     runs-on: ubuntu-latest
     steps:
-      - uses: amannn/action-semantic-pull-request@v4
+        # amannn/action-semantic-pull-request@4.5
+      - uses: amannn/action-semantic-pull-request@6a7ed2d5046cf8a40c60494c83c962343061874a
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-title.yml
```diff
@@ -0,0 +1,18 @@
...
+      - synchronize
+
+jobs:
+  main:
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v4
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
dnys1, line 18: This line is probably the kicker for AppSec review. Otherwise, I'd say we should just review the code. We should also probably use the SHA hash of the action version instead of `@version` which will help our security posture.

.github/workflows/pr-title.yml
```diff
@@ -0,0 +1,19 @@
...
+      - edited
+      - synchronize
+
+jobs:
+  main:
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    steps:
+        # amannn/action-semantic-pull-request@4.5
+      - uses: amannn/action-semantic-pull-request@6a7ed2d5046cf8a40c60494c83c962343061874a
```
ragingsquirrel3, line 17: Were we planning on listing specific scopes here?

PR #1613, opened 2022-05-21
chore: Update lints

dnys1:
- Updates `amplify_lints` to use `package:lints` and `package:flutter_lints` v2.0
- Consolidates rulesets into `library` and `app`
- Renames current ruleset as the "core" rulesets for libraries/apps
- Expands current rulesets to include more lints. Packages can be migrated incrementally to the new ruleset.
- Fixes resulting lint violations of migrating a couple packages to the new rulesets.

PR #1620, opened 2022-05-23
refactor!(aws): Inline enum extensions

dnys1:
commit-id:aa7ed814

---

**Stack**:
- #1625
- #1624
- #1623
- #1621
- #1620 ⬅
- #1613


packages/aws_signature_v4/example/pubspec_overrides.yaml
```diff
@@ -0,0 +1,8 @@
+# This file is checked in until it can be easily generated locally
+# and in CI pipelines, e.g. with melos.
+#
+# It may cause issues with publishing until https://github.com/dart-lang/pub/pull/3419
+# is released in stable.
+dependency_overrides:
+  aws_common:
+    path: ../../aws_common
```
Jordan-Nelson, line 8: Does melos not already use the local version?

dnys1: melos does, but mono_repo runs `dart pub upgrade` as its first step, which will break melos

Jordan-Nelson: Oh I see. It looks like melos is moving toward using this method of overriding dependencies locally and in the future will generate this?

Jordan-Nelson: Okay. It looks like melos can generate these. Do we want to switch over to that? It appears to be a simple update to melos.yaml. I'm not sure if we would want to then gitignore all the generate files or commit them.

```yaml
command:
  bootstrap:
    usePubspecOverrides: true # pubspec_overrides.yaml will be autogenerate during bootstrap
```

dnys1: Exactly, yep! I tried that and it doesn't work for our repo :/ Still need to investigate more.

https://github.com/aws-amplify/amplify-flutter/pull/1585

Jordan-Nelson (approved):

HuiSF (approved):

PR #1629, opened 2022-05-24
feat(secure_storage): add `amplify_secure_storage`

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Rename `amplify_secure_storage_android` to  `amplify_secure_storage`
- Add support for non android platforms via `amplify_secure_storage_dart`
- Add all platforms to example apps

Originally I was planning to create a new package, `amplify_secure_storage`, that depended on `amplify_secure_storage_android` and `amplify_secure_storage_dart`. There were a few reasons (noted below) why I decided against this, and instead expanded the _android package to support all platforms and renamed it. 
- Less packages. (2 vs 3 or 4 depending on wether or not a platform interface package was created)
- Simpler implementation. The typical pattern used to register platform implementations used by federated plugins creates a singleton instance (usually held on the platform interface). Secure storage should not be a singleton. It should allow for multiple instances with different configurations. Once I started deviating from the typical federated plugin pattern, it didn't seem like there was much of an advantage of following it at all.
- Eventually we will have an iOS/macOS impl that uses platform channels like Android does. This impl will be able to share some things from pigeon with the Android impl if they are in the same package.


dnys1 (approved):
I agree this is the better approach. Missing some license headers throughout but otherwise LGTM!

packages/secure_storage/amplify_secure_storage/example/android/app/build.gradle
```diff
@@ -43,7 +44,9 @@ android {
 
     defaultConfig {
         // TODO: Specify your own unique Application ID (https://developer.android.com/studio/build/application-id.html).
-        applicationId "com.amazonaws.amplify.amplify_secure_storage.amplify_secure_storage_android_example"
+        applicationId "com.example.amplify_secure_storage_example"
+        // You can update the following values to match your application needs.
```
Jordan-Nelson, line 48: *Note:* The small updates to the comments throughout the android example came from running `flutter create` again with a more recent version of flutter. I did not want to attempt to manually rename everything so I actually deleted the existing example and recreated it with the new name.

HuiSF (approved):
LGTM.
(could be fun using Amplify icon for the example Apps)

PR #1632, opened 2022-05-24
feat(api): REST API methods in dart, auth mode "NONE"

ragingsquirrel3:
*Issue #, if available:*

*Description of changes:*


packages/api/amplify_api/test/amplify_rest_api_methods_test.dart
```diff
@@ -1,285 +1,105 @@
...
-import 'dart:typed_data';
+import 'dart:io' as io;
 
 import 'package:amplify_api/amplify_api.dart';
-import 'package:amplify_flutter/amplify_flutter.dart';
-import 'package:flutter/services.dart';
+import 'package:amplify_core/amplify_core.dart';
+import 'package:async/async.dart';
+import 'package:aws_common/aws_common.dart';
 import 'package:flutter_test/flutter_test.dart';
```
ragingsquirrel3, line 21: callout: Assuming we want to remove flutter as a testing dep here. Will do once mono repo generates test config for this package.

packages/api/amplify_api/lib/src/method_channel_api.dart
```diff
@@ -1,379 +0,0 @@
...
-  final Map<APIAuthorizationType, APIAuthProvider> _authProviders = {};
-
-  @override
-  final ModelProviderInterface? modelProvider;
-
-  @override
-  Future<void> addPlugin() async {
-    try {
-      setupAuthProviders();
-      await _channel.invokeMethod<void>('addPlugin', {
```
ragingsquirrel3, line 103: callout: datastore integration tests fail, I think bc needs this. Will eventually need some way to call this when using with datastore on supported platforms until datastore implemented.

packages/api/amplify_api/lib/amplify_api.dart
```diff
@@ -1,51 +1,20 @@
...
 export 'package:amplify_core/src/types/api/api_types.dart';
 
-export './model_mutations.dart';
-export './model_queries.dart';
-export './model_subscriptions.dart';
-
-/// {@template amplify_api.amplify_api}
-/// The AWS implementation of the Amplify API category.
-/// {@endtemplate}
-abstract class AmplifyAPI extends APIPluginInterface {
```
dnys1, line 33: So, since we're not planning to ship this for a while, can we leave the method channel stuff in place for iOS/Android and use the Dart impl for just Web/Desktop?

Basically you could use this class and just replace the `UnsupportedError` with a return of the Dart plugin

packages/api/amplify_api/test/amplify_rest_api_methods_test.dart
```diff
@@ -1,285 +1,105 @@
...
-  TestWidgetsFlutterBinding.ensureInitialized();
-  Amplify = MethodChannelAmplify();
-
-  const MethodChannel apiChannel = MethodChannel('com.amazonaws.amplify/api');
   late AmplifyAPI api;
 
   setUpAll(() async {
+    // mock
+    final jsonString =
+        io.File('test/test_data/fake_api_config.json').readAsStringSync();
```
Jordan-Nelson, line 45: You probably want to avoid using File since this won't run on web

PR #1638, opened 2022-05-26
chore(secure_storage): update ffi, ffigen to latest versions

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- update ffi, ffigen to latest versions.
- split keychain ffigen into two libs for CoreFoundation and Security. Prior to ffigen v5 types could not automatically be mapped, so it made sense to treat these two libs as one. Now that types can be mapped IMO it is cleaner to split them up.
- remove commented code from libsecret ffigen config, regenerate, remove type casts.


packages/secure_storage/amplify_secure_storage_dart/lib/src/ffi/cupertino/core_foundation.bindings.g.dart
```diff
@@ -90,127 +67,18 @@ class Keychain {
...
-  late final _SecItemUpdatePtr = _lookup<
-      ffi.NativeFunction<
-          OSStatus Function(
-              CFDictionaryRef, CFDictionaryRef)>>('SecItemUpdate');
-  late final _SecItemUpdate = _SecItemUpdatePtr.asFunction<
-      int Function(CFDictionaryRef, CFDictionaryRef)>();
-
-  int SecItemDelete(
-    CFDictionaryRef query,
+  void CFRelease1(
```
Jordan-Nelson, line 70: *Note:* There are some duplicate methods being generated (`CFRelease` and `CFRelease1`) with ffigen 5.0.0. They aren't an issue as they are unused. I opened an issue in the ffigen repo to figure out why though. Assuming it gets resolved in a later version, we can update and regenerate.

packages/secure_storage/amplify_secure_storage_dart/ffigen.cupertino.coreFoundation.config.yaml
```diff
@@ -0,0 +1,50 @@
...
+#
+# NOTE: You may need to update the path to the `/Frameworks/` directory depending on
+# your MacOS version.
+
+output: 'lib/src/ffi/cupertino/core_foundation.bindings.g.dart'
+name: 'CoreFoundation'
+description: 'Bindings for the CoreFoundation Framework'
+headers:
+  entry-points:
+    - '/Library/Developer/CommandLineTools/SDKs/MacOSX11.1.sdk/System/Library/Frameworks/CoreFoundation.framework/Headers/CFDictionary.h'
```
dnys1, line 14: I'm guessing it doesn't matter which version we use since these libs are not expected to change. I don't have 11.1, though, for ex

```
$ ls /Library/Developer/CommandLineTools/SDKs/
MacOSX.sdk	MacOSX11.sdk	MacOSX12.3.sdk
MacOSX11.3.sdk	MacOSX12.1.sdk	MacOSX12.sdk
```

Maybe just point to `MacOSX.sdk`?

dnys1 (approved):

packages/secure_storage/amplify_secure_storage_dart/ffigen.cupertino.coreFoundation.config.yaml
Jordan-Nelson: Huh. So I don't have `MacOSX.sdk` in that directory anymore. I was having issues with Command Line Tools not being able to update a while ago so I uninstalled it and reinstalled it. 11.1 is all I have now.

I'm thinking `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/` would be a more reliable location to reference. I have `MacOSX.sdk` & `MacOSX12.3.sdk` (my current OS version). I have updated it to use that location instead.

HuiSF (approved):

packages/secure_storage/amplify_secure_storage_dart/ffigen.cupertino.coreFoundation.config.yaml
HuiSF: Do we need to follow Xcode beta closely to update the paths? 😅

dnys1: No, I think we found a path that isn't expected to change

PR #1640, opened 2022-05-26
chore!(api): migrate API category type definitions

ragingsquirrel3:
Change the API type definitions to use `CancelableOperation` and different response types for REST in the existing method channel implementation.

While starting the dart implemention in https://github.com/aws-amplify/amplify-flutter/pull/1632, got some feedback that we should implement that interface in method channels first, then start dart implementation to de-risk, so that's what's going on this PR.


packages/api/amplify_api/lib/src/method_channel_api.dart
```diff
@@ -248,21 +236,51 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
...
+  CancelableOperation<AWSStreamedHttpResponse> _restFunctionHelper({
+    required String methodName,
+    required String path,
+    HttpPayload? body,
+    Map<String, String>? headers,
+    Map<String, dynamic>? queryParameters,
+    String? apiName,
+  }) {
+    // Send Request cancelToken to Native
+    String cancelToken = UUID.getUUID();
```
dnys1, line 269: Use `uuid()` from `aws_common`

packages/api/amplify_api/lib/src/method_channel_api.dart
```diff
@@ -248,21 +236,51 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
...
+    required String methodName,
+    required String path,
+    required String cancelToken,
+    HttpPayload? body,
+    Map<String, String>? headers,
+    Map<String, dynamic>? queryParameters,
+    String? apiName,
+  }) async {
+    final bodyBytes =
+        body != null ? Uint8List.fromList(await body.single) : null;
```
dnys1, line 250: You'll need to fold the bytes since there could be multiple chunks (see AWSStreamedHttpResponse.bodyBytes). Imagine if `body` is `Stream.fromIterable([[1,2,3], [4,5,6]])`, this would result in `Uint8List.fromList([1,2,3])` which is incorrect.

packages/api/amplify_api/lib/src/method_channel_api.dart
```diff
@@ -248,21 +236,51 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
...
+    final restOptions = RestOptions(
+        path: path,
+        body: bodyBytes,
+        apiName: apiName,
+        queryParameters: queryParameters as Map<String, String>?,
```
dnys1, line 255: There's a reason it's <String, dynamic> and not <String, String> in the request constructors. Same is true for AWSHttpRequest and would be important to have here

packages/api/amplify_api/lib/src/method_channel_api.dart
```diff
@@ -284,55 +302,116 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
...
+        body: Stream.value(rawResponseBody?.toList() ?? []));
   }
 
   @override
-  RestOperation get({required RestOptions restOptions}) {
-    return _restFunctionHelper(methodName: 'get', restOptions: restOptions);
+  CancelableOperation<AWSStreamedHttpResponse> get(
+    String path, {
+    Map<String, String>? headers,
+    Map<String, dynamic>? queryParameters,
```
dnys1, line 319: Let's make these `Map<String, Object>` instead of dynamic to match `AWSHttpRequest`. There's no good reason to pass `null` as a value, for example.

ragingsquirrel3: I was using dynamic here a) because that's the type used in Uri and it will convert to string as expected b) there are some cases where you might want to pass null, like if your backend implementation allowed for a null value from the client as opposed to omitted/blank string or whatever. Would prefer not to disallow people from doing something they would otherwise be allowed to do.

Would it cause an error if a Uri literal with a null value in query params passed to `AWSHttpRequest`?

dnys1: Yes, I know but my point was they allow <String, dynamic> so that you can pass <String, String> or <String, List<String>>. You're casting to <String, String> which will throw an error if they pass a <String, List<String>> or even a <String, String?> for that matter.

> Would it cause an error if a Uri literal with a null value in query params passed to AWSHttpRequest?

It wouldn't compile since type system doesn't allow it. That's the problem with `dynamic`. If you want to support null values, then make it <String, Object?>. Anything but `dynamic`.

From the Uri constructor:

```dart
Map<String, dynamic /*String|Iterable<String>*/ >? queryParameters,
```

packages/api/amplify_api/lib/src/method_channel_api.dart
ragingsquirrel3: Sure, ideally this cast is not necessary. The problem is that the existing platform channel implementation uses <String, String> in `RestOptions` which is used to enforce a contract on the native side. In android library, <String, String> is used https://github.com/aws-amplify/amplify-android/blob/main/core/src/main/java/com/amplifyframework/api/rest/RestOptions.java#L36 so I don't see a great way to handle the fact that the new dart interface is more permissive than the platform channel implementation actually allows. Casting here was 1st pass solution. It could be put in a try/catch that gives the caller a better error message. Do you have any other suggestions?

dnys1: This would be better than throwing. But it's still incorrect for lists and null values.

```dart
Uri(queryParameters: queryParameters).queryParameters
```

ragingsquirrel3: discussed offline and changed

packages/api/amplify_api/lib/src/method_channel_api.dart
ragingsquirrel3: thanks, took a shot at fixing in next revision, lmk if needs more changes.

packages/api/amplify_api/lib/src/method_channel_api.dart
```diff
@@ -248,21 +237,64 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
...
+  Future<AWSStreamedHttpResponse> _restResponseHelper({
+    required String methodName,
+    required String path,
+    required String cancelToken,
+    HttpPayload? body,
+    Map<String, String>? headers,
+    Map<String, String>? queryParameters,
+    String? apiName,
+  }) async {
+    Uint8List? bodyBytes;
```
dnys1, line 250: We should only be doing this for iOS/Android, not for Dart

packages/api/amplify_api/pubspec.yaml
```diff
@@ -2,6 +2,7 @@ name: amplify_api
 description: The Amplify Flutter API category plugin, supporting GraphQL and REST operations.
 version: 0.5.0
 homepage: https://github.com/aws-amplify/amplify-flutter/tree/main/packages/amplify_api
+publish_to: none # until finalized
```
dnys1, line 5: We'll be publishing this for dev preview

packages/amplify_core/lib/src/types/api/rest/http_payload.dart
```diff
@@ -0,0 +1,53 @@
...
+import 'dart:convert';
+
+import 'package:async/async.dart';
+
+/// {@template amplify_core.http_payload}
+/// An HTTP request's payload.
+/// {@endtemplate}
+class HttpPayload extends StreamView<List<int>> {
+  /// {@macro amplify_core.http_payload}
+  factory HttpPayload([Object? body]) {
```
dnys1, line 25: Let's add one for form values like `package:http` provides

packages/api/amplify_api/lib/src/method_channel_api.dart
ragingsquirrel3: Yep, my understanding as well. This method basically works around the limitations of existing serialization/deserialization implementation which makes streamed request bodies difficult. In dart, can (and should) allow request body to stream without converting all at once.

packages/api/amplify_api/pubspec.yaml
ragingsquirrel3: Yes I understand, comment unclear, but I wanted to keep this here for safety at least until API review finished and was thinking we would remove it once API set (for preview stage), then would include in dev preview.

packages/api/amplify_api/lib/src/method_channel_api.dart
```diff
@@ -284,55 +316,116 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
...
   }
 
-  bool _shouldThrow(int statusCode) {
-    return statusCode < 200 || statusCode > 299;
-  }
-
-  RestResponse _formatRestResponse(Map<String, dynamic> res) {
+  AWSStreamedHttpResponse _formatRestResponse(Map<String, dynamic> res) {
     final statusCode = res['statusCode'] as int;
     final headers = res['headers'] as Map?;
```
ragingsquirrel3, line 321: callout: realized I forgot to pass headers to response here, will fix in next rev

packages/amplify_core/lib/src/types/api/rest/rest_exception.dart
```diff
@@ -19,16 +19,8 @@ import 'package:amplify_core/amplify_core.dart';
 /// An HTTP error encountered during a REST API call, i.e. for calls returning
 /// non-2xx status codes.
 /// {@endtemplate}
+@Deprecated('No longer thrown for non-200 responses. Will soon be removed')
 class RestException extends ApiException {
```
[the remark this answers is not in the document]
ragingsquirrel3: no, this is only place it's used https://github.com/aws-amplify/amplify-flutter/blob/main/packages/api/amplify_api/lib/src/method_channel_api.dart#L300

made abstract for next rev

packages/amplify_core/lib/src/types/api/rest/http_payload.dart
ragingsquirrel3: Took a shot at this in next rev, lmk if needs more changes. Also, I'm assuming the caller would want to supply encoding header themselves when they do this? In package:http there is a check of the encoding type from the headers related to this. However, bc this is outside the request itself, not easy to replicate.

packages/api/amplify_api/lib/src/method_channel_api.dart
ragingsquirrel3: done

packages/api/amplify_api/lib/src/method_channel_api.dart
```diff
@@ -284,55 +316,119 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
...
-      headers: headers?.cast<String, String>(),
-      statusCode: statusCode,
-    );
-    if (_shouldThrow(statusCode)) {
-      throw RestException(response);
-    }
-    return response;
+    // Make type-safe version of response headers.
+    final serializedHeaders = res['headers'] as Map?;
+    final headers = serializedHeaders?.cast<String, String>();
```
ragingsquirrel3, line 323: callout: Do we want to do this some way that's safer? At one point I was creating a new map from this with only the string key/value pairs so it was type safe at runtime. This is obviously a little simpler.

packages/amplify_core/lib/src/types/api/rest/rest_exception.dart
```diff
@@ -19,16 +19,8 @@ import 'package:amplify_core/amplify_core.dart';
 /// An HTTP error encountered during a REST API call, i.e. for calls returning
 /// non-2xx status codes.
 /// {@endtemplate}
-class RestException extends ApiException {
-  /// The HTTP response from the server.
-  final RestResponse response;
-
+@Deprecated('No longer thrown for non-200 responses. Will soon be removed')
```
Jordan-Nelson, line 22: If a customer was using this to do something on non-200 responses they will have to make code changes if they want their app to function the same way, correct?

I'm on the fence about this, but I think we should consider removing this rather than marking it as deprecated. Marking it as deprecated seems misleading to me. I have always thought of the deprecation annotation as an indication that something *will* break. Not that it has already been broken.

I'm on the fence because 1) the message could be helpful, and 2) (I think) some customers will be using `RestException` for other reasons than catching non-200 responses.

If we keep the deprecation annotation, I think the message needs to make it clear that the app may not function as intended until the code is removed. Something like, "BREAKING CHANGE: RestException is no longer thrown for non-200 responses. See the breaking changes section of the latest release notes for migration instructions."

Here is what the Dart docs say about the [Deprecated annotation](https://api.flutter.dev/flutter/dart-core/Deprecated-class.html) FWIW

> The intent of the @Deprecated annotation is to inform authors who are currently using the feature, that they will soon need to stop using that feature in their code, even if the feature is currently still working correctly.
> 
> Deprecation is an early warning that the deprecated feature is scheduled to be removed at a later time, a time possibly specified in [message](https://api.flutter.dev/flutter/dart-core/Deprecated/message.html). A deprecated feature should no longer be used, code using it will break at some point in the future. If existing code is using the feature, that code should be rewritten to no longer use the deprecated feature.

packages/amplify_core/lib/src/types/api/async/legacy_api_operation.dart
```diff
@@ -13,11 +13,12 @@
...
-class GraphQLOperation<T> {
-  final Function cancel;
-  final Future<GraphQLResponse<T>> response;
-
-  const GraphQLOperation({required this.response, required this.cancel});
+/// Eventually this should be deprecated and just use [CancelableOperation].
+/// Until then, this is used to make `.response` available like it was for older
+/// [GraphQLOperation] and [RestOperation] classes.
+extension LegacyApiOperation<T> on CancelableOperation<T> {
+  @Deprecated('Use .value instead.')
```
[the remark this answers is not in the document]
Jordan-Nelson: Based on the updates in the example app, it seems like customers will (probably) still need to make changes even after changing `.response` to `.value` since value/response is now a different type than value use to be, correct? If so, maybe the auto migration is less useful. Still something to consider. I'll let you determine if it is worth it or not.

Since the type is changes, I'm actually not sure how useful it is to have `.response` at all. This is similar to a comment a left farther down, but to me the deprecated annotation should be used when something *will( break, not when it has already been broken. Since (based on the examples) it looks like codes changes are required, I would consider this already broken.

I guess it could be helpful to have the message. The message probably needs more info though. Again, will leave it up to you to decide how much value this adds to the customer.

packages/api/amplify_api/example/lib/rest_api_view.dart
```diff
@@ -79,16 +75,16 @@ class _RestApiViewState extends State<RestApiView> {
...
+      );
 
       _lastRestOperation = restOperation;
-      RestResponse response = await restOperation.response;
+      final response = await restOperation.value;
 
       print('Get SUCCESS');
-      print(response);
+      print(response.statusCode);
+      print(await response.decodeBody());
```
Jordan-Nelson, line 87: Having code similar to the following all over the place seems tedious.

```dart
final response = await restOperation.value;
final body = await response.decodeBody();
```

I would assume most customers would just want to decoded body. Maybe we extend `CancelableOperation` to make the common things customers would do less tedious. I had included the following extension in a comment above discussing the deprecation of `.response`. Something similar is what I was thinking.

```dart
extension CancelableOperationX on CancelableOperation<AWSStreamedHttpResponse> {
  Future<String> get response async {
    final httpResponse = await value;
    return httpResponse.decodeBody();
  }
}
```

I would need to spend quite a bit more time playing around with these changes to determine what would be most useful, so if you don't think this adds value, I trust your judgement here more than mine.

packages/api/amplify_api/lib/src/method_channel_api.dart
```diff
@@ -284,55 +316,119 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
...
+        path: path,
+        body: body,
+        headers: headers,
+        queryParameters: queryParameters,
+        apiName: apiName);
   }
 
-  @override
+  /// Cancels a request with a given request ID.
+  @Deprecated('Use .cancel() on CancelableOperation instead.')
```
Jordan-Nelson, line 431: Maybe another candidate for auto migration? See other comments about dart fix.

Jordan-Nelson (commented):
Overall looks pretty good. I left a handful of comments that are related to deprecation. I'll just summarize my thoughts here.

I'm not sure the deprecated annotation is used correctly in this PR. I think of the annotation as an indication that something will break in the future. There are two places where it is being used in places that I would still consider it a break even with the annotation. I think this could be misleading/confusing. If we can (with relatively low effort) find a way to make the changes non-breaking, that would be great. If not, I think the message in the annotation needs to clearly indicate that this is a breaking change, or we should just remove the deprecated member all together to force customers to make changes where needed.

Related to deprecation, we can try to use `dart fix` to auto migrate code for customers. I have some comments that include links and more info about it. I think this is a no brainer is situations where the effort on our side is low.

packages/amplify_core/lib/src/types/api/rest/rest_exception.dart
ragingsquirrel3: Agree we are not using it exactly as intended according to that blurb. I took your suggestion of changing the deprecation message to be a little more grabby given that we are kinda hacking the deprecation message to communicate why it's not there. I think it would be tough to use it more traditionally here because the old response type is a field on the exception, which is what started this thought process. I think we could consider a deprecation on main/stable as this moves toward GA.

packages/api/amplify_api/lib/src/method_channel_api.dart
```diff
@@ -284,55 +316,119 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
...
+    Map<String, String>? headers,
+    Map<String, String>? queryParameters,
+    String? apiName,
+  }) {
+    return _restFunctionHelper(
+        methodName: 'get',
+        path: path,
+        headers: headers,
+        queryParameters: queryParameters,
+        apiName: apiName);
```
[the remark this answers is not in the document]
ragingsquirrel3: thanks, changed these here

packages/api/amplify_api/lib/src/method_channel_api.dart
ragingsquirrel3: Based on my understanding of auto-migration tools (still limited) I think this one is harder to auto-migrate because it's not a simple transform, but rather, would require people to change logic in how they're managing cancelations. Instead of managing IDs somewhere and passing them, would have to manage the operations.

packages/api/amplify_api/example/lib/rest_api_view.dart
ragingsquirrel3: Spoke offline about this but I think good thing to consider for API review as changes here still pending that.

packages/amplify_core/lib/src/types/api/async/legacy_api_operation.dart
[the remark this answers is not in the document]
ragingsquirrel3: spoke offline about this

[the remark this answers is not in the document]
ragingsquirrel3: While yes, the operation types are changing in breaking way for someone who refers to those types, the .response => .value proxy is there because I suspect there are people  who don't refer to the types, just to the .response field, so this makes it easier for that subset of people. Agree it is breaking already and not textbook use of deprecation annotation.

packages/amplify_core/lib/src/types/api/rest/http_payload.dart
```diff
@@ -0,0 +1,76 @@
...
+
+  /// A UTF-8 encoded HTTP body.
+  HttpPayload.string(String body, {Encoding encoding = utf8})
+      : super(LazyStream(() => Stream.value(encoding.encode(body))));
+
+  /// A byte buffer HTTP body.
+  HttpPayload.bytes(List<int> body) : super(Stream.value(body));
+
+  /// Form-encodes the body.
+  HttpPayload.fields(Map<String, String> body, {Encoding encoding = utf8})
```
dnys1, line 55: I'd prefer `form` or `formValues` unless there's some precedence for this name

packages/amplify_core/lib/src/types/api/rest/http_payload.dart
```diff
@@ -0,0 +1,76 @@
...
+  const HttpPayload.streaming(Stream<List<int>> body) : super(body);
+}
+
+/// Converts a [Map] from parameter names to values to a URL query string.
+///
+///     mapToQuery({"foo": "bar", "baz": "bang"});
+///     //=> "foo=bar&baz=bang"
+///
+/// Copied from similar util. https://github.com/dart-lang/http/blob/06649afbb5847dbb0293816ba8348766b116e419/pkgs/http/lib/src/utils.dart#L15
+String _mapToQuery(Map<String, String> map, {Encoding? encoding}) {
```
dnys1, line 69: `encoding` is always non-null here.

packages/amplify_core/lib/src/types/api/rest/http_payload.dart
```diff
@@ -0,0 +1,76 @@
...
+  var pairs = <List<String>>[];
```
dnys1, line 70: This method will create a new list and requires potentially several allocations to grow it. Can probably be done with a single .map.join expression.

packages/amplify_core/lib/src/types/api/rest/http_payload.dart
dnys1: Would try to figure out how to do it so that we set the correct content type header for them

packages/api/amplify_api/lib/src/method_channel_api.dart
dnys1: It's safe when used here since the native plugin only sends back string header values (it does the mapping in native code). Just not safe when creating request.

packages/api/amplify_api/example/lib/rest_api_view.dart
dnys1: Agree we need something more fluent. Jordan's suggestion is pretty nice and saves a deprecation/new API

packages/amplify_core/lib/src/types/api/rest/http_payload.dart
```diff
@@ -0,0 +1,76 @@
...
+  const HttpPayload.empty() : super(const Stream.empty());
+
+  /// A UTF-8 encoded HTTP body.
+  HttpPayload.string(String body, {Encoding encoding = utf8})
+      : super(LazyStream(() => Stream.value(encoding.encode(body))));
+
+  /// A byte buffer HTTP body.
+  HttpPayload.bytes(List<int> body) : super(Stream.value(body));
+
+  /// Form-encodes the body.
```
dnys1, line 54:
```suggestion
  /// A form-encoded body of `key=value` pairs.
```

packages/amplify_core/lib/src/types/api/rest/http_payload.dart
ragingsquirrel3: changed

packages/amplify_core/lib/src/types/api/rest/http_payload.dart
ragingsquirrel3: thanks, made required

packages/amplify_core/lib/src/types/api/rest/http_payload.dart
ragingsquirrel3: I'll go with `formValues` in next rev. Wasn't wed to the name, went with that bc called bodyFields in http https://pub.dev/documentation/http/latest/http/Request-class.html and I thought "fields" was the best way to mirror that, but... not the best name by itself.

packages/amplify_core/lib/src/types/api/rest/http_payload.dart
ragingsquirrel3: tried doing that in next rev

packages/api/amplify_api/example/lib/rest_api_view.dart
ragingsquirrel3: revised this based on last convo, open to further suggestions

packages/api/amplify_api/example/lib/rest_api_view.dart
```diff
@@ -39,18 +38,16 @@ class _RestApiViewState extends State<RestApiView> {
...
+        body: HttpPayload.string('{"name":"Mow the lawn"}'),
       );
 
       _lastRestOperation = restOperation;
-      RestResponse response = await restOperation.response;
+      final response = await restOperation.response;
 
       print('Put SUCCESS');
-      print(response);
+      print(response.decodeBody());
```
dnys1, line 50: await

packages/api/amplify_api/example/lib/rest_api_view.dart
```diff
@@ -97,15 +91,14 @@ class _RestApiViewState extends State<RestApiView> {
...
+        _apiPathController.text,
       );
-
       _lastRestOperation = restOperation;
-      RestResponse response = await restOperation.response;
+      final response = await restOperation.response;
 
       print('Delete SUCCESS');
-      print(response);
+      print(response.decodeBody());
```
dnys1, line 101: await

packages/api/amplify_api/example/lib/rest_api_view.dart
```diff
@@ -140,15 +132,16 @@ class _RestApiViewState extends State<RestApiView> {
...
+        body: HttpPayload.string('{"name":"Mow the lawn"}'),
       );
 
       _lastRestOperation = restOperation;
-      RestResponse response = await restOperation.response;
+      final response = await restOperation.response;
 
       print('Patch SUCCESS');
-      print(response);
+      print(response.decodeBody());
```
dnys1, line 144: await

packages/api/amplify_api/lib/src/method_channel_api.dart
```diff
@@ -284,55 +325,125 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
...
-    return response;
+    // Make type-safe version of response headers.
+    final serializedHeaders = res['headers'] as Map?;
+    final headers = serializedHeaders?.cast<String, String>();
+    final rawResponseBody = res['data'] as Uint8List?;
+
+    return AWSStreamedHttpResponse(
+        statusCode: statusCode,
+        headers: headers,
+        body: Stream.value(rawResponseBody?.toList() ?? []));
```
dnys1, line 338: toList is not needed. Uint8List is already a List<int>

packages/api/amplify_api/example/lib/rest_api_view.dart
ragingsquirrel3: I don't think async is needed here because .response is non-streamed response and .decodeBody thus returns string.

dnys1: Ah I didn't see the change to non-streamed response

packages/api/amplify_api/lib/src/method_channel_api.dart
ragingsquirrel3: ty, changed for next rev

packages/api/amplify_api/example/lib/rest_api_view.dart
```diff
@@ -39,18 +38,16 @@ class _RestApiViewState extends State<RestApiView> {
...
   void onPutPressed() async {
     try {
-      RestOperation restOperation = Amplify.API.put(
-        restOptions: RestOptions(
-          path: _apiPathController.text,
-          body: ascii.encode('{"name":"Mow the lawn"}'),
-        ),
+      final restOperation = Amplify.API.put(
+        _apiPathController.text,
+        body: HttpPayload.string('{"name":"Mow the lawn"}'),
```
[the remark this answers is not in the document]
ragingsquirrel3: great idea! love it, added for next rev

PR #1642, opened 2022-05-27
feat(auth): Native Android bridge

dnys1:
Adds the Android implementation for the two-way Auth bridge allowing native Android plugins like API/DataStore to call into Flutter for retrieving auth credentials. Further a bridge is established to allow the Dart plugin to call the native addPlugin method with its configuration.

commit-id:08874ee6

---

**Stack**:
- #1655
- #1654
- #1652
- #1648
- #1647
- #1645
- #1642 ⬅
- #1639

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #1639, opened 2022-05-26
feat(auth): Native iOS auth bridge

dnys1:
Adds the iOS implementation for the two-way Auth bridge allowing native iOS plugins like API/DataStore to call into Flutter for retrieving auth credentials. Further a bridge is established to allow the Dart plugin to call the native `addPlugin` method with its configuration.

PR #1625, opened 2022-05-23
feat: State machine

dnys1:
Adds the base state machine types and the skeleton for the Auth plugin's state machine

commit-id:a2731ef0

---

**Stack**:
- #1648
- #1647
- #1645
- #1642
- #1639
- #1625 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #1647, opened 2022-05-27
feat(auth): fetchAuthSession

dnys1:
**Stack**:
- #1645
- #1642
- #1639
- #1625

### DIFF
--- a/packages/auth/amplify_auth_cognito/lib/src/auth_plugin_impl.dart
+++ b/packages/auth/amplify_auth_cognito/lib/src/auth_plugin_impl.dart
@@ -72,6 +72,15 @@ class AmplifyAuthCognito extends AuthPluginInterface implements Closeable {
       );
     }
 
+    // Configure this plugin to act as a native iOS/Android plugin.
+    if (!zIsWeb && (Platform.isAndroid || Platform.isIOS)) {
+      final nativePlugin = _NativeAmplifyAuthCognito(this);
+      NativeAuthPlugin.setup(nativePlugin);
+
+      final nativeBridge = NativeAuthBridge();
+      await nativeBridge.configure();
+    }
+
     await _init();
     _stateMachine.dispatch(AuthEvent.configure(config));
 
@@ -86,15 +95,33 @@ class AmplifyAuthCognito extends AuthPluginInterface implements Closeable {
           throw (state as AuthFailure).exception;
       }
     }
+  }
 
-    // Configure this plugin to act as a native iOS/Android plugin.
-    if (!zIsWeb && (Platform.isAndroid || Platform.isIOS)) {
-      final nativePlugin = _NativeAmplifyAuthCognito(this);
-      NativeAuthPlugin.setup(nativePlugin);
+  @override
+  Future<AuthSession> fetchAuthSession({
+    required AuthSessionRequest request,
+  }) async {
+    final options = request.options as CognitoSessionOptions?;
+    _stateMachine.dispatch(FetchAuthSessionEvent.fetch(options));
 
-      final nativeBridge = NativeAuthBridge();
-      await nativeBridge.configure();
+    await for (final state
+        in _stateMachine.stream.whereType<FetchAuthSessionState>()) {
+      switch (state.type) {
+        case FetchAuthSessionStateType.idle:
+        case FetchAuthSessionStateType.fetching:
+        case FetchAuthSessionStateType.refreshing:
+          continue;
+        case FetchAuthSessionStateType.success:
+          state as FetchAuthSessionSuccess;
+          return state.session;
+        case FetchAuthSessionStateType.failure:
+          state as FetchAuthSessionFailure;
+          throw state.exception;
+      }
     }
+
+    // This should never happen.
+    throw const UnknownException('fetchAuthSession could not be completed');
   }
 
   /* -- TODO: Replace -- */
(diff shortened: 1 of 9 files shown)


packages/auth/amplify_auth_cognito/lib/src/auth_plugin_impl.dart
```diff
@@ -72,6 +72,15 @@ class AmplifyAuthCognito extends AuthPluginInterface implements Closeable {
...
     }
 
+    // Configure this plugin to act as a native iOS/Android plugin.
+    if (!zIsWeb && (Platform.isAndroid || Platform.isIOS)) {
+      final nativePlugin = _NativeAmplifyAuthCognito(this);
+      NativeAuthPlugin.setup(nativePlugin);
+
+      final nativeBridge = NativeAuthBridge();
+      await nativeBridge.configure();
+    }
```
fjnoyp, line 82: Maybe it would be helpful to add some comments explaining why NativeAuthPlugin only provides FetchAuthSession functionality as the naming suggests it would have handled all Auth API related method calls.

fjnoyp: Is there a design document on our use of pigeon?

packages/auth/amplify_auth_cognito/lib/src/state/cognito_state_machine.dart
```diff
@@ -25,6 +25,7 @@ import 'state.dart';
 final stateMachineBuilders = <StateMachineToken, StateMachineBuilder>{
   AuthStateMachine.type: AuthStateMachine.new,
   CredentialStoreStateMachine.type: CredentialStoreStateMachine.new,
+  FetchAuthSessionStateMachine.type: FetchAuthSessionStateMachine.new,
```
fjnoyp, line 28: Minor nit, perhaps it's too late, but naming seems a bit misleading. Shouldn't it be cognito_state_machine_manager to denote it's a state machine manager instead of a single state machine instance? I had to read through the code and see that it extends StateMachineManager before being certain.

packages/auth/amplify_auth_cognito/lib/src/state/machines/auth_state_machine.dart
```diff
@@ -37,7 +37,15 @@ class AuthStateMachine extends AuthStateMachineBase {
   /// The AWS credentials provider, using cached credentials.
   late final AWSCredentialsProvider credentialsProvider =
       InlineCredentialsProvider(() async {
-    throw UnimplementedError();
+    // Get the latest credentials from storage.
+    dispatch(const FetchAuthSessionEvent.fetch());
+    final fetchState =
+        await expect(FetchAuthSessionStateMachine.type).getLatestResult();
+    final fetchedCredentials = fetchState?.session.credentials;
```
fjnoyp, line 44: minor nit/idea: would it be better to encapsulate the FetchAuth functionality into a separate class taht would handle the dispatch/expect taking of credentials instead of having to directly reference the different FetchAuth components directly from within auth_state_machine?

fjnoyp (approved):
Looks good, just some minor questions/comments

packages/auth/amplify_auth_cognito/lib/src/auth_plugin_impl.dart
dnys1: We don't have a design doc for using pigeon. It's being used throughout, though, including in secure storage as well. What info can I provide? All it does is generate the boilerplate to talk to iOS/Android. The code would be the same if we were to write it by hand, so the design is technically the same.

I will add some notes for the interface (why only fetchAuthSession)

packages/auth/amplify_auth_cognito/lib/src/state/machines/auth_state_machine.dart
dnys1: To an external observer, it is a single state machine. The goal is to abstract away the fact that it manages multiple state machines because all communication should happen with this state machine.

packages/auth/amplify_auth_cognito/lib/src/state/cognito_state_machine.dart
dnys1: To an external observer, it is a single state machine. The goal is to abstract away the fact that it manages multiple state machines because all communication should happen with this state machine.

packages/auth/amplify_auth_cognito/lib/src/state/machines/auth_state_machine.dart
dnys1: Great idea! I've extracted this into `AuthPluginCredentialsProvider`. This will make it easier to test now as well :)

packages/auth/amplify_auth_cognito/lib/src/state/machines/fetch_auth_session_state_machine.dart
```diff
@@ -0,0 +1,300 @@
...
+  /// Gets the latest fetchAuthSession result.
+  Future<FetchAuthSessionSuccess?> getLatestResult() async {
+    final fetchState = await stream.startWith(currentState).firstWhere((state) {
+      return state is FetchAuthSessionSuccess ||
+          state is FetchAuthSessionFailure;
+    });
+    if (fetchState is FetchAuthSessionFailure) {
+      throw fetchState.exception;
+    }
+    return fetchState is FetchAuthSessionSuccess ? fetchState : null;
```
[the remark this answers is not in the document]
dnys1: Yep, good catch

haverchuck (approved):

PR #1648, opened 2022-05-27
feat(auth): Hosted UI

dnys1:
- Dart-only Hosted UI platform and state machine
- iOS/Android bridge for platform webviews

commit-id:7c629b62

---

**Stack**:
- #1677
- #1676
- #1675
- #1670
- #1655
- #1654
- #1652
- #1648 ⬅

### DIFF
--- a/packages/amplify/amplify_flutter/lib/src/hybrid_impl.dart
+++ b/packages/amplify/amplify_flutter/lib/src/hybrid_impl.dart
@@ -80,6 +80,8 @@ class AmplifyHybridImpl extends AmplifyClass {
               AmplifyExceptionMessages.missingRecoverySuggestion,
         );
       }
+    } on AmplifyAlreadyConfiguredException {
+      return;
     } on Exception catch (e) {
       safePrint('Amplify plugin was not added');
       throw AmplifyException(
--- a/packages/auth/amplify_auth_cognito/lib/src/flows/hosted_ui/hosted_ui_platform_html.dart
+++ b/packages/auth/amplify_auth_cognito/lib/src/flows/hosted_ui/hosted_ui_platform_html.dart
@@ -1,0 +1,100 @@
+// Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'dart:html';
+
+import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
+import 'package:amplify_auth_cognito/src/crypto/oauth.dart';
+import 'package:amplify_core/amplify_core.dart';
+import 'package:flutter_web_plugins/flutter_web_plugins.dart';
+import 'package:path/path.dart' show url;
+
+/// {@macro amplify_auth_cognito.hosted_ui_platform}
+class HostedUiPlatformImpl extends HostedUiPlatform {
+  /// {@macro amplify_auth_cognito.hosted_ui_platform}
+  HostedUiPlatformImpl(super.dependencyManager) : super.protected();
+
+  /// The base URL
+  String get baseUrl {
+    final basePath = getBaseElementHrefFromDom() ?? '/';
+    return url.join(window.location.origin, basePath);
+  }
+
+  Never _noSuitableRedirect({required bool signIn}) {
+    final inOut = signIn ? 'in' : 'out';
+    throw InvalidUserPoolConfigurationException(
+      message: 'No sign $inOut redirect URLs registered for base URL: $baseUrl',
+      recoverySuggestion:
+          'Add a sign $inOut redirect URL on that starts with "$baseUrl". See '
+          'the docs for more info: '
+          'https://docs.amplify.aws/lib/auth/signin_web_ui/q/platform/flutter/',
+    );
+  }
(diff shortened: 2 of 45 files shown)


packages/auth/amplify_auth_cognito/pigeons/native_auth_plugin.dart
```diff
@@ -45,8 +45,33 @@ abstract class NativeAuthPlugin {
...
 @HostApi()
 abstract class NativeAuthBridge {
+  /// Configure the native platform/plugin.
+  ///
+  /// On iOS/Android, this calls `Amplify.addPlugin` with the [NativeAuthPlugin]
+  /// implementation.
   @async
   void configure();
+
+  /// Sign out by presenting [url] and waiting for a response to a URL with
```
Jordan-Nelson, line 55:
```suggestion
  /// Sign in by presenting [url] and waiting for a response to a URL with
```

packages/auth/amplify_auth_cognito/pigeons/native_auth_plugin.dart
```diff
@@ -45,8 +45,33 @@ abstract class NativeAuthPlugin {
...
+  /// [callbackUrlScheme].
+  ///
+  /// If [preferPrivateSession] is `true`, do not persist session cookies.
+  @async
+  Map<String, String> signIn(
```
Jordan-Nelson, line 60: *nit:* `signInWithUrl` would make it clear that this has nothing to do with username/password auth.

packages/auth/amplify_auth_cognito_android/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AtomicResult.kt
```diff
@@ -0,0 +1,76 @@
...
+import io.flutter.Log
+import kotlinx.coroutines.MainScope
+import kotlinx.coroutines.launch
+import java.util.concurrent.atomic.AtomicBoolean
+
+/**
+ * Thread-safe [NativeAuthPluginBindings.Result] wrapper which prevents multiple replies and
+ * automatically posts results to the main thread.
+ */
+class AtomicResult<T>(
```
Jordan-Nelson, line 26: Should the one in core eventually be replaced with this one?

Jordan-Nelson (approved):

packages/amplify/amplify_flutter/lib/src/hybrid_impl.dart
```diff
@@ -80,6 +80,8 @@ class AmplifyHybridImpl extends AmplifyClass {
               AmplifyExceptionMessages.missingRecoverySuggestion,
         );
       }
+    } on AmplifyAlreadyConfiguredException {
+      return;
```
haverchuck, line 84: Maybe we want to set the isConfigured flag to true here?

haverchuck, line 84: Maybe we want to set the isConfigured flag to true here?

packages/auth/amplify_auth_cognito_android/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AtomicResult.kt
dnys1: This wraps the generated bindings' result which is different than a generic Flutter result. So they're incompatible and we'll need both, unfortunately, unless at some point we migrate all the plugins to a single pigeon definition in core which could be pretty cool.

packages/auth/amplify_auth_cognito/lib/src/flows/hosted_ui/hosted_ui_platform_html.dart
```diff
@@ -0,0 +1,97 @@
...
+  /// The base URL
+  String get baseUrl {
+    final basePath = getBaseElementHrefFromDom() ?? '/';
+    return url.join(window.location.origin, basePath);
+  }
+
+  @override
+  Uri get signInRedirectUri => config.signInRedirectUris.firstWhere(
+        (uri) => uri.toString().startsWith(baseUrl),
+        orElse: () => throw UnknownException(
```
haverchuck, line 37: Do we want to have some sort `InvalidConfiguration` exception?

packages/auth/amplify_auth_cognito/pubspec.yaml
```diff
@@ -12,6 +12,7 @@ dependencies:
   amplify_auth_cognito_ios: 0.5.0
   amplify_auth_plugin_interface: 0.5.0
   amplify_core: 0.5.0
+  amplify_flutter: 0.5.0
```
haverchuck, line 15: Can we avoid this? I think it's preferable to avoid circular dependencies.

haverchuck (commented):
Left some questions/comments, but mostly LGTM

packages/auth/amplify_auth_cognito/pubspec.yaml
dnys1: amplify_flutter doesn't depend on amplify_auth_cognito, not sure what circular dependency you mean

packages/amplify/amplify_flutter/lib/src/hybrid_impl.dart
dnys1: It's set by completing asyncConfig. If we hot restart these will throw but by ignoring them, asyncConfig will be set

packages/auth/amplify_auth_cognito/pubspec.yaml
haverchuck: I think I may have been thinking about the old interface... this is fine.

haverchuck (approved):

packages/auth/amplify_auth_cognito/lib/src/flows/hosted_ui/hosted_ui_platform_html.dart
dnys1: We have InvalidUserPoolConfiguration which I guess works. I'll update it.

PR #1649, opened 2022-05-28
chore!(api): run background thread when cancel operation

metajet98:
1. **Issue**:
- When cancel multiple operation at the same time, android main thread will be freeze
2. **Description of changes:**
- Improvement: Use async task to execute the cancel job in background thread


dnys1:
Hi @metajet98 - thanks for your PR! Can you let me know how you were able to trigger this freeze so I can write tests for it? So far, I haven't been able to reproduce it.

Thanks again!

dnys1:
Hi @metajet98 would you be able to test with https://github.com/aws-amplify/amplify-flutter/pull/1691 and see if that fixes your issue?

PR #1656, opened 2022-05-31
chore: add canary workflow

yeung-wah:
*Issue #, if available:*

*Description of changes:*
Added canary workflow for getting started workflow(https://docs.amplify.aws/start/q/integration/flutter/). The new circleci job essentially doing the same step as described in the tutorial. 

After merging:
We will set up a trigger to run once a day(https://circleci.com/docs/2.0/scheduled-pipelines/#project-settings)

Note: We are using inline orb for now, we will extract the implementation and put in an actual orb after we've confirmed it runs successfully for couple of days, and when orb's namespace is ready to use

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,181 @@
 version: 2.1
 
+orbs:
+  # Using inline orb for now
+  getting-started-smoke-test:
+    orbs:
+      macos: circleci/macos@2.2.0
+      android: circleci/android@2.0.3
+      flutter-orb: circleci/flutter@1.1.0
+
+    executors:
+      mac-executor:
+        macos:
+          xcode: 13.2.1
+        resource_class: large
+
+      android-executor:
+        machine:
+          image: android:202102-01
+        resource_class: large
+
+    commands:
+      run-with-retry:
+        description: Run command with retry
+        parameters:
+          label:
+            description: Display name
+            type: string
+          command:
+            description: Command to run
+            type: string
+          retry-count:
+            description: Number of retry
+            type: integer
+            default: 3
+          sleep:
+            description: Wait duration until next retry
+            type: integer
+            default: 5
+          no_output_timeout:
+            description: Elapsed time the command can run without output
+            type: string
+            default: 10m
+        steps:
+          - run:
+              name: << parameters.label >>
+              command: |
+                retry() {
+                  MAX_RETRY=<< parameters.retry-count >>
+                  n=0
+                  until [ $n -ge $MAX_RETRY ]
+                  do
+                      << parameters.command >> && break
+                      n=$[$n+1]
+                      sleep << parameters.sleep >>
+                  done
+                  if [ $n -ge $MAX_RETRY ]; then
+                    echo "failed: ${@}" >&2
+                    exit 1
+                  fi
+                }
+                retry
+              no_output_timeout: << parameters.no_output_timeout >>
+      install-flutter:
+        description: Install Flutter and set up paths.
+        parameters:
+          flutter_branch:
+            description: Flutter branch or version tag.
+            type: string
+            default: stable
+        steps:
+          - run:
+              name: Set up Flutter
+              command: |
+                echo 'export FLUTTER_HOME=${HOME}/sdks/flutter' >> $BASH_ENV
+                echo 'export FLUTTER_BRANCH=<< parameters.flutter_branch >>' >> $BASH_ENV
+                echo 'export FLUTTER_ROOT=${FLUTTER_HOME}' >> $BASH_ENV
+                echo 'export PATH=${PATH}:${FLUTTER_HOME}/bin:${FLUTTER_HOME}/bin/cache/dart-sdk/bin:${HOME}/.pub-cache/bin:${FLUTTER_HOME}/.pub-cache/bin' >> $BASH_ENV
+                source $BASH_ENV
+                git clone --branch ${FLUTTER_BRANCH} https://github.com/flutter/flutter.git ${FLUTTER_HOME}
+                (yes || true) | flutter doctor --android-licenses && flutter doctor
+                flutter precache
+      setup-amplify-flutter-project:
+        description: Setup Amplify project
+        steps:
+          - run-with-retry:
+              label: Setting up dependences
+              command: flutter pub add amplify_flutter && flutter pub add amplify_datastore && flutter pub add amplify_storage_s3 && flutter pub add amplify_analytics_pinpoint && flutter pub add amplify_auth_cognito && flutter pub add amplify_api
+              no_output_timeout: 5m
+          - run:
+              name: Adding integration_test package
+              command: 'sed -i -e "s/dev_dependencies:/dev_dependencies:\n  integration_test:\n    sdk: flutter/" ./pubspec.yaml && cat ./pubspec.yaml'
+          - run:
+              name: Update outdated dependences
+              command: flutter pub upgrade --major-versions
+          - run:
+              name: Adding amplifyconfig file
+              command: mv canaries/dummy_amplifyconfiguration.dart canaries/amplifyconfiguration.dart && cp canaries/amplifyconfiguration.dart amplified_todo/lib
+              working_directory: ~/flutter-canaries/
+          - run:
+              name: Adding test code
+              command: cp canaries/main.dart amplified_todo/lib && cp -r canaries/integration_test amplified_todo/integration_test && cp -r canaries/models amplified_todo/lib/models
+              working_directory: ~/flutter-canaries/
+
+
+    jobs:
+      flutter-android:
+        parameters:
+          flutter-version:
+            type: string
+        executor:
+          name: android/android-machine
+          resource-class: large
+          tag: 2021.10.1
+        working_directory: ~/flutter-canaries/amplified_todo
+        steps:
+          - checkout:
+              path: ~/flutter-canaries
+          - install-flutter:
+              flutter_branch: << parameters.flutter-version >>
+          - run:
+              name: Setting up project
+              command: cd ../ && flutter create amplified_todo
+          - run:
+              name: Update Android version
+              command: sed -i -e "s/minSdkVersion .*/minSdkVersion 21/" ./android/app/build.gradle && cat ./android/app/build.gradle
+          - setup-amplify-flutter-project
+          - flutter-orb/install_android_gradle:
+              app-dir: ./
+          - android/create-avd:
+              avd-name: flutter
+              install: true
+              system-image: system-images;android-29;default;x86
+          - android/start-emulator:
+              avd-name: flutter
+              post-emulator-launch-assemble-command: ls -lrt
+              restore-gradle-cache-find-args: ./android -name 'build.gradle'
+          - run-with-retry:
+              label: Run Flutter Build
+              command: flutter build apk --debug
+              no_output_timeout: 20m
+          - run-with-retry:
+              label: Run Flutter Tests
+              command: flutter test integration_test
+              no_output_timeout: 1h
+              retry-count: 5
+
+      flutter-ios:
+        parameters:
+          flutter-version:
+            type: string
+        executor: mac-executor
+        working_directory: ~/flutter-canaries/amplified_todo
+        steps:
+          - checkout:
+              path: ~/flutter-canaries
+          - macos/preboot-simulator:
+              device: iPhone 13
+              version: "15.2"
+          - install-flutter:
+              flutter_branch: << parameters.flutter-version >>
+          - run:
+              name: Setting up project
+              command: cd ../ && flutter create amplified_todo
+          - setup-amplify-flutter-project
+          - run:
+              name: Update ios version
+              command: sed -i -e "s/# platform .*/platform :ios, '13.0'/" ./ios/Podfile && cat ./ios/Podfile
+          - run-with-retry:
+              label: Run Flutter Build
+              command: flutter build ios --debug --simulator
+              no_output_timeout: 20m
+          - run-with-retry:
+              label: Run Flutter Tests
+              command: flutter test integration_test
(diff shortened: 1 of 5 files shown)


iartemiev (changes requested):

.circleci/config.yml
```diff
@@ -254,3 +326,7 @@ workflows:
                   "amplify_datastore",
                   "amplify_flutter",
                 ]
+  getting_started_tests:
```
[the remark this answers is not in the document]
yeung-wah: Added, and refactored a bit to be consistent with other platforms

iartemiev (approved):
LGTM 👍🏻  Let's get someone from the Flutter team to review as well

iartemiev:
@yeung-wah - looks like the flutter/android test is failing: https://app.circleci.com/pipelines/github/yeung-wah/amplify-flutter/19/workflows/294287f8-72dc-49d9-8f2d-28c7e6b8d326/jobs/326

yeung-wah:
> @yeung-wah - looks like the flutter/android test is failing: https://app.circleci.com/pipelines/github/yeung-wah/amplify-flutter/19/workflows/294287f8-72dc-49d9-8f2d-28c7e6b8d326/jobs/326

yea, I am looking into it. It looks like a random issue, originally I set the timeout to longer period to avoid this issue in circleci, but seems like flutter command is also timing out. I will see what options we have to increase timeout in flutter

iartemiev:
> yea, I am looking into it. It looks like a random issue, originally I set the timeout to longer period to avoid this issue in circleci, but seems like flutter command is also timing out. I will see what options we have to increase timeout in flutter

Sounds good, thanks! We want these tests to be as reliable as possible, so let's try to eliminate any test flakiness if we can

.circleci/config.yml
```diff
@@ -1,5 +1,168 @@
...
+                source $BASH_ENV
+                git clone --branch ${FLUTTER_BRANCH} https://github.com/flutter/flutter.git ${FLUTTER_HOME}
+                (yes || true) | flutter doctor --android-licenses && flutter doctor
+                flutter precache
+      setup-amplify-flutter-project:
+        description: Setup Amplify project
+        steps:
+          - run-with-retry:
+              label: Setting up dependences
+              command: flutter pub add amplify_flutter && flutter pub add amplify_datastore
```
Jordan-Nelson, line 88: Do we want to install these from pub? I am not sure what the goal here is. If we install them from pub, we would not catch issues in the library until after release.

.circleci/config.yml
```diff
@@ -1,5 +1,168 @@
...
+                }
+                retry
+              no_output_timeout: << parameters.no_output_timeout >>
+      install-flutter:
+        description: Install Flutter and set up paths.
+        parameters:
+          flutter_branch:
+            description: Flutter branch or version tag.
+            type: string
+            default: stable
```
Jordan-Nelson, line 70: Do we want to run against `beta` as well? I'm not sure what the goal of the tests are. If we want to catch compatibility issues with new versions of flutter we will probably want to use beta as well to catch those issues before they are in a stable flutter release.

canaries/main.dart
```diff
@@ -0,0 +1,137 @@
...
+import 'amplifyconfiguration.dart';
+
+void main() {
+  runApp(const MyApp());
+}
+
+class MyApp extends StatelessWidget {
+  const MyApp({Key? key}) : super(key: key);
+
+  // This widget is the root of your application.
```
Jordan-Nelson, line 12: Would it be more useful to copy over the code from the app in `/example`? Not sure what the goal is here. The only category dependency is datastore so it won't catch issues in any other category.

.circleci/config.yml
iartemiev: The goal is to ensure changes to Flutter / other 3rd party tooling aren't breaking compatibility with Amplify, so I think it's fine to install Amplify from pub here. I agree with your comment below re: also testing against `beta`

.circleci/config.yml
yeung-wah: good idea, let me test against beta as well

.circleci/config.yml
Jordan-Nelson: Sounds good 👍

.circleci/config.yml
```diff
@@ -1,5 +1,181 @@
...
+                echo 'export FLUTTER_ROOT=${FLUTTER_HOME}' >> $BASH_ENV
+                echo 'export PATH=${PATH}:${FLUTTER_HOME}/bin:${FLUTTER_HOME}/bin/cache/dart-sdk/bin:${HOME}/.pub-cache/bin:${FLUTTER_HOME}/.pub-cache/bin' >> $BASH_ENV
+                source $BASH_ENV
+                git clone --branch ${FLUTTER_BRANCH} https://github.com/flutter/flutter.git ${FLUTTER_HOME}
+                (yes || true) | flutter doctor --android-licenses && flutter doctor
+                flutter precache
+      setup-amplify-flutter-project:
+        description: Setup Amplify project
+        steps:
+          - run-with-retry:
```
Jordan-Nelson, line 86: *Question:* Are you using `run-with-retry` because you have found this to be flaky?

.circleci/config.yml
```diff
@@ -1,5 +1,181 @@
...
+                      sleep << parameters.sleep >>
+                  done
+                  if [ $n -ge $MAX_RETRY ]; then
+                    echo "failed: ${@}" >&2
+                    exit 1
+                  fi
+                }
+                retry
+              no_output_timeout: << parameters.no_output_timeout >>
+      install-flutter:
```
Jordan-Nelson, line 64: Are we able to re-use the existing `install_flutter` command in this file?

.circleci/config.yml
```diff
@@ -1,5 +1,181 @@
...
+      setup-amplify-flutter-project:
+        description: Setup Amplify project
+        steps:
+          - run-with-retry:
+              label: Setting up dependences
+              command: flutter pub add amplify_flutter && flutter pub add amplify_datastore && flutter pub add amplify_storage_s3 && flutter pub add amplify_analytics_pinpoint && flutter pub add amplify_auth_cognito
+              no_output_timeout: 20m
+          - run:
+              name: Adding integration_test package
+              command: 'sed -i -e "s/dev_dependencies:/dev_dependencies:\n  integration_test:\n    sdk: flutter/" ./pubspec.yaml && cat ./pubspec.yaml'
```
Jordan-Nelson, line 92: Can this and the next few steps be broken out into a .sh file? Might be easier to manage in the future.

.circleci/config.yml
```diff
@@ -1,5 +1,181 @@
...
+              working_directory: ~/flutter-canaries/
+
+
+    jobs:
+      flutter-android:
+        parameters:
+          flutter-version:
+            type: string
+        executor:
+          name: android/android-machine
```
Jordan-Nelson, line 112: Can we use the same executor we use for running Android unit tests?

.circleci/config.yml
```diff
@@ -1,5 +1,181 @@
...
+              label: Run Flutter Tests
+              command: flutter test integration_test
+              no_output_timeout: 1h
+              retry-count: 5
+
+      flutter-ios:
+        parameters:
+          flutter-version:
+            type: string
+        executor: mac-executor
```
Jordan-Nelson, line 152: Similar to above, seems like this is using a different executor than the iOS unit tests use. I think it will be easier to maintain these if we use the same executors. If there is a reason to use this one over the other one we are using, maybe we change the unit tests to use this one.

.circleci/config.yml
```diff
@@ -197,11 +374,17 @@ releasable_branches: &releasable_branches
       - main
 
 workflows:
+  # Tells CircleCI to skip this workflow when the pipeline is triggered by the scheduled source,
```
Jordan-Nelson, line 377: *nit:* should this comment be below `test_deploy`, instead of above it? Seems like it is relevant for `when`.

.circleci/config.yml
```diff
@@ -1,5 +1,181 @@
...
+          - flutter-orb/install_android_gradle:
+              app-dir: ./
+          - android/create-avd:
+              avd-name: flutter
+              install: true
+              system-image: system-images;android-29;default;x86
+          - android/start-emulator:
+              avd-name: flutter
+              post-emulator-launch-assemble-command: ls -lrt
+              restore-gradle-cache-find-args: ./android -name 'build.gradle'
```
Jordan-Nelson, line 137: I believe these steps wouldn't be needed if the executor that is used for the Android unit tests is used.

.circleci/config.yml
```diff
@@ -1,5 +1,181 @@
...
+          - run-with-retry:
+              label: Run Flutter Build
+              command: flutter build apk --debug
+              no_output_timeout: 20m
+          - run-with-retry:
+              label: Run Flutter Tests
+              command: flutter test integration_test
+              no_output_timeout: 1h
```
Jordan-Nelson, line 145: Seems like a long time for the tests being run. Were you seeing it take longer than 20 min (which is what iOS is set to). I would think 20 would be plenty.

Jordan-Nelson (commented):
Looks pretty good. I left a few more comments. Most of them are around aligning our current unit tests for iOS and Android and the canary jobs. I think they can re-used some commands and can use the same executors.

.circleci/config.yml
yeung-wah: usually takes 2-3 mins to finish, but the test command sometime hang for 12 mins, and timeout. Setting it to 1 hour to give some buffers to retry

.circleci/config.yml
yeung-wah: Yes, specially with the android test(https://support.circleci.com/hc/en-us/articles/360000028928-Testing-with-the-Android-emulator-on-CircleCI). It fails pretty frequently, and adding retry help to solve the flakiness problem

Jordan-Nelson (dismissed):
lgtm. Discussed offline that eventually the inline orb will be removed, which means my comments about using the same executor don't need to be addressed.

.circleci/config.yml
```diff
@@ -1,5 +1,181 @@
 version: 2.1
 
+orbs:
+  # Using inline orb for now
+  getting-started-smoke-test:
+    orbs:
+      macos: circleci/macos@2
```
dnys1, line 7: Should this have a tighter constraint than "2"?

.circleci/config.yml
```diff
@@ -1,5 +1,181 @@
...
+      android: circleci/android@2.0.3
+      flutter-orb: circleci/flutter@1.0.1
```
dnys1, line 9: Latest version is `1.1.0` from what I can see. Any reason to not use that?

.circleci/config.yml
```diff
@@ -1,5 +1,181 @@
...
+
+    executors:
+      mac-executor:
+        macos:
+          xcode: 13.3.1
```
dnys1, line 14: Can we make this `13.2` since that will be our minimum version on our next release?

.circleci/config.yml
dnys1: There's one in the orb too: https://circleci.com/developer/orbs/orb/circleci/flutter#commands-install_sdk

.circleci/config.yml
```diff
@@ -1,5 +1,181 @@
...
+                git clone --branch ${FLUTTER_BRANCH} https://github.com/flutter/flutter.git ${FLUTTER_HOME}
+                (yes || true) | flutter doctor --android-licenses && flutter doctor
+                flutter precache
+      setup-amplify-flutter-project:
+        description: Setup Amplify project
+        steps:
+          - run-with-retry:
+              label: Setting up dependences
+              command: flutter pub add amplify_flutter && flutter pub add amplify_datastore && flutter pub add amplify_storage_s3 && flutter pub add amplify_analytics_pinpoint && flutter pub add amplify_auth_cognito
+              no_output_timeout: 20m
```
dnys1, line 89: This is a really long window. I would think `5m` is more than plenty. The problem is `pub` will retry indefinitely in certain situations so this will just burn compute time.

dnys1 (dismissed):
Looks good. There's no caching of pub dependencies though, from what I can tell, so this will have to download the entire dependency tree each run which seems inefficient.

.circleci/config.yml
yeung-wah: Yes we can update to latest version

.circleci/config.yml
yeung-wah: oh I see, didn't know pub command will hang for some cases, will change it to 5m

canaries/main.dart
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  void _configureAmplify() async {
+    if(!Amplify.isConfigured) {
+      print("Configuring amplify");
+      AmplifyAuthCognito auth = AmplifyAuthCognito();
+      AmplifyStorageS3 storage = AmplifyStorageS3();
+      AmplifyAnalyticsPinpoint analytics = AmplifyAnalyticsPinpoint();
+      AmplifyDataStore datastorePlugin =
+      AmplifyDataStore(modelProvider: ModelProvider.instance);
```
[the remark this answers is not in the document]
yeung-wah: Will add API category

.circleci/config.yml
yeung-wah: yea, I found the custom install_flutter more easier to work with as we can get the latest stable version easier than the one in the orb

yeung-wah:
> Looks good. There's no caching of pub dependencies though, from what I can tell, so this will have to download the entire dependency tree each run which seems inefficient.

I think it is find as the job will schedule to run once per day, shouldn't be a big problem

.circleci/config.yml
yeung-wah: changed to use 2.2.0

dnys1 (approved):
LGTM. Thanks!!

PR #1683, opened 2022-06-03
fix(aws_signature_v4): Presign Min/Max Expirations

wurikiji:
*Issue #, if available:*

*Description of changes:*

According to the [doc]( https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html), minimum exp value is one second, and maximum exp value is 7 days. 

- change `>` to `>=` and `<` to `<=`
- modify examples to print seven days and one second link

### DIFF
--- a/packages/aws_signature_v4/example/bin/example.dart
+++ b/packages/aws_signature_v4/example/bin/example.dart
@@ -162,13 +162,22 @@ Future<void> main(List<String> args) async {
       AWSHeaders.host: host,
     },
   );
-  final Uri signedUrl = await signer.presign(
+
+  final Uri signedUrl1Sec = await signer.presign(
     urlRequest,
     credentialScope: scope,
     serviceConfiguration: serviceConfiguration,
-    expiresIn: const Duration(minutes: 10),
+    expiresIn: const Duration(seconds: 1),
   );
-  stdout.writeln('Download URL: $signedUrl');
+  stdout.writeln('Download URL for 1 second: $signedUrl1Sec');
+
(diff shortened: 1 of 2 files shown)


dnys1:
Nice catch! Do you think we should add a test instead of (or in addition to) the example app change? Happy to knock that out if so.

wurikiji:
I think it's enough to have a comment line with a documentation link just like now. 
Actually, I thought the example did not need to be changed. 
If you think the other changes or the addition of a test is required, I will take it.

dnys1:
Sounds good to me! I agree, I think the example change might be a tad confusing. Although I would love to know where a 1-second presigned URL could be used in reality - there must be a good reason they allow it!

wurikiji:
I can't even imagine a good reason for less than 10 seconds link in reality 😃
For one second link, I tried several times to get content from that link by my hand, but I never succeded to use it without the help of code 😅
Do you think I should rollback `example.dart`?

dnys1:
> I can't even imagine a good reason for less than 10 seconds link in reality 😃
> For one second link, I tried several times to get content from that link by my hand, but I never succeded to use it without the help of code 😅

Haha nice!

> Do you think I should �rollback example.dart?

Yeah, let's do that. I'll go ahead and merge afterwards! :)

wurikiji:
Thanks for the quick comments!

dnys1 (approved):
Thanks for the PR!

haverchuck (approved):

dnys1:
@wurikiji This has been released in v0.1.1 of the library!

Issue #334, opened 2021-01-28
Support switching authentication flows at runtime


jcowley:
**Describe the bug**
It's not possible to change authenticationFlowType after configuring Amplify (which only supports a single authentidatioonFlowType at one time). This means that it is not possible to use more than one authenticationFlowType in an app.

The use case for this would be switching to CUSTOM_AUTH (from USER_SRP_AUTH or USER_PASSWORD_AUTH) in order to trigger a custom authentication flow for a user to enable a custom challenge. 

The ability to switch authentication flows is supported in amplify-js: 
https://docs.amplify.aws/lib/auth/switch-auth/q/platform/js

**To Reproduce**
1. Call with your default configuration, e.g. `Amplify.configure(amplifyconfig)`
2. Add the plugin `Amplify.addPlugin(AmplifyAuthCognito())`;
3. Attempt to reconfigure Amplify with a new configuration, e.g. `Amplify.configure(customconfig)`

**Expected behavior**
Amplify is reconfigured using the new configuration.

**Actional behavior**
Error thrown: `StateError (Bad state: Amplify has already been configured and re-configuration is not supported.)`

**Platform**
Amplify Flutter current supports iOS and Android.  This issue is reproducible in (check all that apply):
[x] Android
[x] iOS

Amplifiyer:
@jcowley Thanks for raising this issue. I'm marking this as a feature request as by architecture, Amplify for iOS and Android does not support reconfiguration.

dnys1:
Closing in favor of https://github.com/aws-amplify/amplify-flutter/issues/410. Updates on support for the custom auth flow will be posted there.

jcowley:
@dnys1 This is not at all the same as #410. They are completely different.

This ticket refers to the ability to *switch* authentication flows after amplify is configured (i.e. at runtime) like you can in the js client.

dnys1:
My apologies, @jcowley. I've re-opened the ticket. I believe this is a general request for re-configuring Amplify at runtime, is that correct?

twink1e:
This feature will be really helpful for supporting both CUSTOM_AUTH and USER_SRP_AUTH at run time, and JS SDK already supports it. Really appreciate it if can be included in a new release 🙏

abdallahshaban557:
We have released this feature! The documentation is available here https://docs.amplify.aws/lib/auth/signin/q/platform/flutter/#switching-authentication-flow-at-runtime

PR #1687, opened 2022-06-04
feat(auth): runtime authFlow attribute

haverchuck:
*Issue*
#334 

*Description of changes:*
Adds runtime authFlow type. Requires bumping amplify-android to 1.34.0 and amplify-ios to 1.25.0.


dnys1 (approved):
LGTM

packages/auth/amplify_auth_cognito/lib/src/CognitoSignIn/CognitoSignInOptions.dart
```diff
@@ -17,10 +17,13 @@ import 'package:amplify_core/amplify_core.dart';
 
 class CognitoSignInOptions extends SignInOptions {
   Map<String, String>? clientMetadata;
-  CognitoSignInOptions({this.clientMetadata}) : super();
+  final AuthenticationFlowType? authFlowType;
```
dnys1, line 20: How about just `authFlow`

dnys1, line 20: How about just `authFlow`. The `type` suffix originates in the SDK where they add `Type` to the end of all types for some reason

dnys1, line 20: How about just `authFlow`. The `type` suffix originates in the SDK where they add `Type` to the end of all types for some reason

haverchuck: Used `AuthenticationFlowType` because it matches what is in the config, which is what get's utilized as the default. Even though it's a little more verbose, I think having that consistency is helpful.

haverchuck: The API parameter `authFlowType` is what is used by the native libraries, so I decided to stay consistent with them. In terms of the enum `AuthenticationFlowType`, I used the existing one we are getting from the config. I realize it's a little odd to have a mismatch between the type name and the API parameter, but I didn't want to deviate from the native APIs or change/duplicate an existing enum.

HuiSF (approved):

dnys1 (approved):

PR #1700, opened 2022-06-09
feat(secure_storage): Remote worker

dnys1:
- Adds a remote worker bee for secure storage operations and makes it the default in Amplify libraries.
- Updates the Android impl to use background task queues

### DIFF
--- a/packages/secure_storage/amplify_secure_storage_dart/lib/src/amplify_secure_storage_dart.dart
+++ b/packages/secure_storage/amplify_secure_storage_dart/lib/src/amplify_secure_storage_dart.dart
@@ -60,8 +60,11 @@ class AmplifySecureStorageWorker extends AmplifySecureStorageInterface {
   late final SecureStorageWorker _worker;
   final _workerMemo = AsyncMemoizer<void>();
 
-  @override
-  Future<void> init() => _workerMemo.runOnce(() async {
+  /// {@template amplify_secure_storage_dart.secure_storage_interface.init}
+  /// Initializes the secure storage and performs any work which should be
+  /// performed once before any secure storage operations.
+  /// {@endtemplate}
+  Future<void> _init() => _workerMemo.runOnce(() async {
         _worker = SecureStorageWorker.create();
         // TODO(dnys1): Log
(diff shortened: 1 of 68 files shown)


haverchuck (approved):

packages/secure_storage/amplify_secure_storage_dart/lib/src/amplify_secure_storage_dart.dart
```diff
@@ -38,6 +43,58 @@ class AmplifySecureStorageDart extends AmplifySecureStorageInterface
...
+  /// {@macro amplify_secure_storage_dart.amplify_secure_storage_worker}
+  AmplifySecureStorageWorker({
+    required super.config,
+  });
+
+  late final SecureStorageWorker _worker;
+  final _workerMemo = AsyncMemoizer<void>();
+
+  @override
+  Future<void> init() => _workerMemo.runOnce(() async {
```
Jordan-Nelson, line 64: I am seeing the following exception when running the Auth example app and the Secure Storage example app on Mac.

```
LateError (LateInitializationError: Field '_worker@619435580' has not been initialized.)
```

For the Secure Storage example app that might be resolved by calling `storage.init()`. It looks like that was already added to the auth example though, I am still seeing that. I am not sure if it is not waiting for it.

It might make sense to confirm init() has completed before calling read/write/delete. `_workerMemo` is already an `AsyncMemoizer`, so I think you could just add `await init()` before each method. I am not super familiar with `AsyncMemoizer` so maybe it isn't quire that simple.

Jordan-Nelson (commented):
Thanks for taking care of this. Looks pretty good. I am seeing an issue on Mac that I left a comment for.

packages/secure_storage/amplify_secure_storage_dart/lib/src/amplify_secure_storage_dart.dart
dnys1: Ah okay, yeah that's probably easiest. Will do!

Jordan-Nelson (approved):
lgtm!

PR #1704, opened 2022-06-09
fix(authenticator): allow second signin attempt on web

haverchuck:
*Issue #, if available:*
The authenticator on web was not allowing a second sign in attempt after a failure, nor was it allowing the 'Forgot Password' button.

*Description of changes:*
This PR yields the UnauthenticatedStep.signIn state after a sign in failure.

### DIFF
--- a/packages/auth/amplify_auth_cognito/example/integration_test/sign_in_sign_out_test.dart
+++ b/packages/auth/amplify_auth_cognito/example/integration_test/sign_in_sign_out_test.dart
@@ -83,6 +83,19 @@ void main() {
       );
     });
 
+    testWidgets('should be able to sign in after a failed attempt',
+        (WidgetTester tester) async {
+      final incorrectUsername = generateUsername();
+      expect(
+        Amplify.Auth.signIn(username: incorrectUsername, password: password),
+        throwsA(isA<UserNotFoundException>()),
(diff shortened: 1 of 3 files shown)


packages/auth/amplify_auth_cognito/example/integration_test/sign_in_sign_out_test.dart
```diff
@@ -83,6 +83,19 @@ void main() {
       );
     });
 
+    testWidgets('should be able to sign in after a failed attempt',
```
dnys1, line 86: Let's remove/update this since it doesn't actually test the issue

dnys1 (approved):

packages/auth/amplify_auth_cognito/example/integration_test/sign_in_sign_out_test.dart
haverchuck: ha, yeah good call - wrote this before actually resolving the bug.

haverchuck: @dnys1 I added a corresponding test in the authenticator.

dnys1 (approved):

PR #1709, opened 2022-06-10
test: Uses non-amplify graphql request for adminCreateUser

haverchuck:
*Description of changes:*
Uses non-amplify graphql request for adminCreateUser


packages/amplify_authenticator/example/integration_test/sign_in_with_email_test.dart
```diff
@@ -79,49 +74,50 @@ void main() {
       await signInPage.expectUserNotFound();
     });
 
+    // TODO(dnnoyes): enable subscriptions with graphql package
```
dnys1, line 77: Can we just disable these on non-mobile platforms? You can use the `skip` flag on the group/test.

packages/amplify_test/lib/src/integration_test_utils/auth_cognito/integration_test_auth_utils.dart
```diff
@@ -71,77 +82,98 @@ Future<DeleteUserResponse> deleteUser(String username) async {
 /// The [verifyAttributes] flag will verify the email and phone, and should be used
 /// if tests need to bypass the verification step.
 /// The [attributes] list passes additional attributes.
-Future<AdminCreateUserResponse> adminCreateUser(
+Future<void> adminCreateUser(
```
dnys1, line 85: The response has the error in it, though, so how would we know if it failed or not?

packages/amplify_authenticator/example/integration_test/sign_in_with_email_test.dart
haverchuck: @dnys1 We can - but I believe we would conditionally need to add the AmplifyApi plugin for those platforms as well. If you and Jordan thinks that's worth it, I can make the change.

packages/amplify_test/lib/src/integration_test_utils/auth_cognito/integration_test_auth_utils.dart
haverchuck: We throw the exception if present in the response.

dnys1: Okay, so you changed the lambda as well? Last I looked at the lambda, it sent back a successful response no matter what pretty much, it just included the errors in a field.

packages/amplify_authenticator/example/integration_test/sign_in_with_email_test.dart
Jordan-Nelson: Is it just a matter of conditionally setting `additionalConfigs`?
```dart
additionalConfigs: !kIsWeb && (Platform.isIOS || Platform.isAndroid) ? [AmplifyAPI()] : null,
```

Jordan-Nelson (approved):

packages/amplify_test/lib/src/integration_test_utils/auth_cognito/integration_test_auth_utils.dart
haverchuck: @dnys1 Updated these in my personal lambda instance- I've included the changes in this PR now as well. Once the PR is approved, I'll update the instances we've already deployed.

dnys1 (approved):

PR #1723, opened 2022-06-12
fix(auth): Preserve Hosted UI options

dnys1:
Preserves the hosted UI options selected at sign in for use in sign out by saving the options in secure storage.

---

**Stack**:
- #1722
- #1721
- #1720
- #1719

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #1725, opened 2022-06-12
chore(auth): Migrate some types to AWS E/S/D

dnys1:
Migrates some Auth types to the AWS Equatable, Serializable, Debuggable stack.

---

**Stack**:
- #1724
- #1723
- #1722
- #1721
- #1720
- #1719


dnys1:
I just realized this includes migration of the Dart example apps as well which is why it's so big. A mishap in rebasing..

packages/amplify_core/lib/src/types/auth/attribute/cognito_user_attribute_key.dart
```diff
@@ -66,7 +66,6 @@ class CognitoUserAttributeKey extends UserAttributeKey
   List<Object?> get props => [
         // Cognito will lowercase these in API calls
         key.toLowerCase(),
-        readOnly,
```
haverchuck, line 69: Why are we removing this?

haverchuck (approved):
Left one question, otherwise LGTM

packages/amplify_core/lib/src/types/auth/attribute/cognito_user_attribute_key.dart
dnys1: You'd never have two attributes with the same key but different readOnly markers so it doesn't provide any additional info beyond just the key. Further, AWSDebuggable uses these properties to render `toString` and it doesn't add any value there.

Jordan-Nelson (approved):

PR #1737, opened 2022-06-14
chore(repo): Initial CLI tool work

dnys1:
Adds scaffolding for a CLI tool for managing our repo. Current commands:
- `list`: Lists all packages in the repo
- `generate-sdk`: Generates the AWS SDK for a package
- `deps check`: Checks package dependencies against a global config map of validated versions


Jordan-Nelson (approved):
lgtm.

Might want to add a short note about it in contributing.md. I don't think it needs a ton of info. Just a note that this tool is used to manage certain workflows, and it can be activated by running `dart pub global activate --source path packages/aft`

deps.yaml
```diff
@@ -0,0 +1,9 @@
+# Global dependency versions for third-party dependencies of
+# Amplify Flutter projects. These represent the values which
+# have been vetted by manual review and should be used consistently 
+# across all packages.
+#
+# Application of these values can verified/updated with the `aft` 
+# tool.
+dependencies:
```
ragingsquirrel3, line 8: love this 👏

ragingsquirrel3 (approved):
This is cool! I think this will give a us a nice place to do even fancier stuff with pubspec/yaml processing.

PR #1740, opened 2022-06-15
chore: Clean up

dnys1:
Clean up fragments of old PRs not caught by CI/CD runs because of issues with the integ tests


Jordan-Nelson (approved):

packages/auth/amplify_auth_cognito_dart/lib/src/state/event/fetch_auth_session_event.dart
```diff
@@ -114,9 +114,8 @@ class FetchAuthSessionRefresh extends FetchAuthSessionEvent {
 
   @override
   String? checkPrecondition(FetchAuthSessionState currentState) {
-    if (currentState.type == FetchAuthSessionStateType.refreshing ||
-        currentState.type == FetchAuthSessionStateType.fetching) {
-      return 'Credentials are already being fetched...';
+    if (currentState.type == FetchAuthSessionStateType.refreshing) {
```
haverchuck, line 117: Have you tested this?

haverchuck (commented):
Left one question, otherwise LGTM

packages/auth/amplify_auth_cognito_dart/lib/src/state/event/fetch_auth_session_event.dart
haverchuck: Ok - I recall needing this with some of the session logic, but I know other things have changed in that code path since then.

dnys1: I had added the second part of the check by mistake. Refreshes are always initiated while the current state is `fetching`. This still debounces concurrent refresh attempts.

PR #1746, opened 2022-06-16
feat(logging)

fjnoyp:


packages/logger/lib/src/log_level.dart
```diff
@@ -0,0 +1,87 @@
...
+    return this.index - other.index;
+  }
+
+  bool operator >(LogLevel value) => index > value.index;
+  bool operator >=(LogLevel value) => index >= value.index;
+  bool operator <(LogLevel value) => index < value.index;
+  bool operator <=(LogLevel value) => index <= value.index;
+
+  /// Creates a formatted string based on the log level.
+  String toFormattedString() {
```
fjnoyp, line 51: Jordan - I missed this in the first review. I think the default formatting can be simplified a bit. We want to use color and formatting to draw attention to things that need it. If we use it everywhere, the formatting loses meaning. Personally I would leave anything other than warn/error as unformatted. Warn/error obviously should have some attention drawn to it. Red/yellow makes sense, but personally I think the background colors are a bit much.

fjnoyp: @dnys1 what are your thoughts on Jordan's comments? I understand you've created a lot of Flutter applications, so you probably have the most experience dealing with logs from multiple libraries and your own code ...

To mitigate confusion, the less important log messages have a grey or blue background. The reason behind this was to differentiate default logs from Amplify from other logs that might be coming from other libraries / packages / native code that we or the customer might encounter when running the program.

fjnoyp:
Dillon - So this makes a lot of assumptions about how people will want to use the logger. Not everyone is going to like colors nor do people necessarily want their console swarmed with any messages from Amplify. The logger should dynamically react to logLevel changes - each logger should have it's own but there should also be a way to control the log level of "groups" of them (maybe all of them). The Dart team's logging package solves these problems pretty well - I recommend having a look and trying to emulate some of their behavior: https://pub.dev/packages/logging

fjnoyp:
Hey @dnys1 , good point - I wonder how much customizability we should expose for our default logger print behavior.

I agree with Jordan that it would be nice to have a LogLevel control for all modules but beyond that I am less certain.

In the initial iteration, the logging behavior was very customizable, but we ended up opting for exposing the 'Stream' instead so that external users could define any behavior they wanted.

So if users wanted, they could write custom code listening to the stream and dynamically react to logLevel changes.

So my takeaway:

Customizability is already provided via the AmplifyLogger.logStream

We should allow users to set the LogLevel for logs to be displayed in the default logger behavior as Jordan added via the AmplifyLogger logLevel variable.

fjnoyp:
Dillon - So consider the use case where I want to not use the default logger and listen to log stream.

I listen to the logStream instance and set logLevel to none. Is that the DX we want?

What if I want to silence Auth logs only, or only listen to API logs. Is that easy/doable with this setup?

What we want to find here is the architecture that allows for extension to accommodate these use cases and more. Right now, I don't think these use cases are particularly friendly or easy to implement in this set up.

I would check out the logging package from the Dart team and see how they approached things and maybe leverage the Amplify plugin system to create a pluggable architecture.

IMO, each category should have it's own logger which it can use internally and which can be listened to externally. Setting the log level on one of these affects the logs for only that category, while setting the master log level affects all loggers. I should be able to override the default logger and control the log level such that when I listen to the log stream I get the logs matching the log level I've set.

fjnoyp:
Dillon - Some more ideas to check out: https://github.com/dnys1/amplify-dart/pull/11

Jordan-Nelson:
Can you change the base branch to `next`?

Jordan-Nelson:
I think this is close, but I'm not sure the common use cases are as easy to achieve as they should be. Customers can always listen to the stream and do whatever they want, but I think we can improve the DX for both the default behavior and the stream.

Here are some use cases that I think should be very straight forward to achieve:
1. No logging at all
2. Logs printed to the console (with minimal formatting)
3. Logs handled via stream only (not printed to the console)
4. Logs printed to the console, above a certain level
5. Logs handled via stream only, above a certain level
6. Logs printed to the console, for a certain category
7. Logs handled via stream only, for a certain category
8. Logs printed to the console, for a certain category & above a certain level
9. Logs handled via stream only, for a certain category & above a certain level

Use cases 1, 2, & 4 seem easy to achieve.

Use cases 3 & 5 are achievable, but probably not quite the DX we want since the default behavior of logging to console has to explicitly be disabled otherwise they will be printed to the console and handled by the stream. I think there are several options to improve the DX here.

Use cases 6 through 9 are achievable by listening to the stream, filtering by the namespace, and either printing to the console or doing whatever else is desired. These use cases are a little more advanced than the others, so I think it's fine if customers have to listen to the stream to get this behavior, but I think some of Dillon's suggestions would improve the DX here. Using the `Category` enum instead of a namespace String seems like a clear improvement. Adding the logger instances to each category, opposed to keeping a map of them in one class, also seems like an improvement.

fjnoyp:
Hi @Jordan-Nelson  thanks for your comments.  I just pushed some changes to the API: 

Logs handled via stream only (not printed to the console) -> 

`AmplifyLogger.useDefaultLogHandler = false;
`

Logs handled via stream only, above a certain level ->

```
AmplifyLogger.useDefaultLogHandler = false;
AmplifyLogger.globalLogLevel = LogLevel.certainLevel;
AmplifyLogger.logStream.listen(logEvent){
// custom logging code 
}
```


Logs printed to the console, for a certain category -> 

```
AmplifyLogger.getCategoryInstance(Category.Analytics).logLevel = LogLevel.info; 
AmplifyLogger.getCategoryInstance(Category.Authentication).logLevel = LogLevel.none; 
```


Logs handled via stream only, for a certain category & above a certain level ->

```
// is the loglevel limit global or by category? 
// in the stream isn't it easy to filter by LogLevel? 
```


I have separated out the print log logic into a separate method that is called from logStream.listen.  It is currently private but perhaps we can make it public so that users can call that method within their custom logHandler (in case of complex filtering).

fjnoyp:
Hi @dnys1 I've updated the API. 

To disable the default logging behavior you just need to call:
`AmplifyLogger.useDefaultLogHandler = false;`

If you want to continue using the default log handler and selectively silence logs for a category you would do: 

```
AmplifyLogger.getCategoryInstance(Category.Analytics).logLevel = LogLevel.info; 
AmplifyLogger.getCategoryInstance(Category.Authentication).logLevel = LogLevel.none; 
```

If you had a custom log handler I think exposing the stream of all LogEvents would be the desirable behavior as they could add their own simple filtering logic. 

In summary, I've exposed the AmplifyLogger for each category via `getCategoryInstance` thereby allowing more fine grained control. 

The `AmplifyLogLevel.globalLogLevel` field will control all categories' log filter behavior.

packages/logger/lib/src/amplify_logger.dart
```diff
@@ -0,0 +1,144 @@
...
+  static final Map<String, AmplifyLogger> _logsMap = {};
+
+  /// A stream of logs from all logger instances.
+  static final Stream<LogEvent> logStream = _logStreamController.stream;
+
+  /// Default logger instance without a namespace.
+  static get defaultLogger => getInstance();
+
+  LogLevel logLevel = LogLevel.info;
+  static LogLevel globalLogLevel = LogLevel.info;
```
Jordan-Nelson, line 40: I am a little confused about how these two levels are used, and I can imagine customer may be as well.

`globalLogLevel` seems to only be used in the default handler, whereas `logLevel` is used in log method, meaning it gets applied to the custom stream handling and the default handling. Is that what we want?

I think whatever log level filter applies to the default handling should probably be applied to the custom stream handling.

Also, right now if a custom was using the default handler and wanted `debug` logs to print for auth, they would have to set the global level and the auth log level. I think that  having to set two levels will be confusing and cumbersome. I am not suggesting you get rid of the global level, just suggesting the API could be improved.

I am not sure what the ideal API would be here. I would have to put some thought into it. Maybe the `globalLogLevel` is really just a setter that sets all the category levels?

dnys1:
This is getting better. And, we're getting to a place where the benefit of our own logging implementation is starting to disappear. I've been using `package:logging` in `worker_bee` and `aws_common` packages because it's extremely simple and intuitive and provides the expected behavior for log levels (global vs. non-global) by establishing a hierarchy of loggers. The added benefit of using this package is that any project that also uses it immediately has access to all our logs and can filter based on the namespace, etc.

Here's a simple implementation which wraps `package:logging` and provides opt-in behavior for things like ANSI color printing.

`amplify_logger.dart`
```dart
import 'dart:async';

import 'package:ansicolor/ansicolor.dart';
import 'package:logging/logging.dart';

enum Category { Analytics, API, Auth, Storage, DataStore }

class AmplifyLogger {
  AmplifyLogger._(this._logger);

  static final Map<String, StreamSubscription<LogRecord>> _subscriptions = {};
  final Logger _logger;

  factory AmplifyLogger() => AmplifyLogger._(Logger('Amplify'));

  factory AmplifyLogger.category(Category category) {
    return AmplifyLogger._(Logger('Amplify.${category.name}'));
  }

  void registerPlugin(AmplifyLoggerPlugin plugin) {
    final currentSubscription = _subscriptions.remove(_logger.fullName);
    if (currentSubscription != null) {
      unawaited(currentSubscription.cancel());
    }
    _subscriptions[_logger.fullName] = _logger.onRecord
        .where((record) => record.loggerName == _logger.fullName)
        .listen(plugin.handleLogRecord);
  }

  /// Log a message with a log level of [Level.FINER].
  void verbose(String message) {
    _logger.finer(message);
  }

  /// Log a message with a log level of [Level.FINE].
  void debug(String message) {
    _logger.fine(message);
  }

  /// Log a message with a log level of [Level.INFO].
  void info(String message) {
    _logger.info(message);
  }

  /// Log a message with a log level of [Level.WARNING].
  void warn(String message) {
    _logger.warning(message);
  }

  /// Log a message with a log level of [Level.SEVERE].
  void error(String message, [Object? error, StackTrace? stackTrace]) {
    _logger.severe(message, error, stackTrace);
  }
}

abstract class AmplifyLoggerPlugin {
  void handleLogRecord(LogRecord record);
}

class AnsiPrettyPrinter implements AmplifyLoggerPlugin {
  @override
  void handleLogRecord(LogRecord record) {
    final buffer = StringBuffer();

    // Log Level
    buffer.write(record.level.formattedString);

    // Log Namespace
    buffer.write(' ');

    final namespace =
        record.loggerName == 'Amplify' ? '' : record.loggerName.split('.')[1];
    if (namespace.isNotEmpty) {
      buffer.write(_formatLogNamespace(namespace));
    }

    // Log Message
    buffer.write(' ');
    buffer.write(record.message);

    safePrint(buffer.toString());
  }

  String _formatLogNamespace(String? namespace) {
    return (AnsiPen()
      ..white(bold: true)
      ..gray(level: .2, bg: true))(' $namespace ');
  }
}

extension on Level {
  String get formattedString {
    if (this <= Level.FINER) {
      return (AnsiPen()
            ..white(bold: true)
            ..gray(level: .8, bg: true))(' V ')
          .toString();
    } else if (this <= Level.FINE) {
      return (AnsiPen()
        ..white(bold: true)
        ..gray(level: .6, bg: true))(' D ');
    } else if (this <= Level.INFO) {
      return (AnsiPen()
        ..white(bold: true)
        ..blue(bg: true))(' I ');
    } else if (this <= Level.WARNING) {
      return (AnsiPen()
        ..white(bold: true)
        ..yellow(bg: true))(' W ');
    } else {
      return (AnsiPen()
        ..white(bold: true)
        ..red(bg: true))(' E ');
    }
  }
}
```

`example.dart`
```dart
import 'package:amplify_logger/amplify_logger.dart';
import 'package:logging/logging.dart';

void main(List<String> arguments) {
  Logger.root.level = Level.ALL;
  AmplifyLogger()
    ..registerPlugin(AnsiPrettyPrinter())
    ..verbose('Verbose')
    ..debug('Debug')
    ..info('Info')
    ..warn('Warning')
    ..error('Error', Exception('An unknown error'));

  AmplifyLogger.category(Category.Auth)
    ..registerPlugin(AnsiPrettyPrinter())
    ..verbose('Verbose')
    ..debug('Debug')
    ..info('Info')
    ..warn('Warning')
    ..error('Error', Exception('An unknown error'));
}
```

Screen Shot 2022-06-22 at 4 12 26 PM

packages/logger/lib/src/amplify_logger.dart
```diff
@@ -0,0 +1,144 @@
...
+  static bool useDefaultLogHandler = true;
+
+  /// The namespace of the logger instance.
+  String get namespace {
+    return _namespace;
+  }
+
+  String _namespace;
+
+  AmplifyLogger({String namespace = 'default'}) : _namespace = namespace;
```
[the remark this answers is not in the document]
fjnoyp: What if users wanted to create their own AmplifyLogger?

packages/logger/lib/src/amplify_logger.dart
```diff
@@ -0,0 +1,144 @@
...
+    return _logsMap[namespace] ??= AmplifyLogger(namespace: namespace);
+  }
+
+  static AmplifyLogger getCategoryInstance({required Category category}) {
+    return getInstance(namespace: category.toString());
+  }
+
+  /// Initialize the logger by listening to the log stream.
+  static void initDefaultLogHandler() {
+    ansiColorDisabled = false;
```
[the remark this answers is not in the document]
fjnoyp: It will be included and exported from amplify_core.  In `amplify_class.dart` we should call `AmplifyLogger.initDefaultLogHandler();`

packages/logger/lib/src/amplify_logger.dart
```diff
@@ -0,0 +1,144 @@
...
+  ///
+  /// If no instance exists with the given namespace, a new
+  /// instance will be created.
+  static AmplifyLogger getInstance({String namespace = 'default'}) {
+    return _logsMap[namespace] ??= AmplifyLogger(namespace: namespace);
+  }
+
+  static AmplifyLogger getCategoryInstance({required Category category}) {
+    return getInstance(namespace: category.toString());
+  }
```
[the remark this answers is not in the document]
fjnoyp: What if users's (external developers) wanted to use the logger outside of an existing Amplify Category?

packages/logger/lib/src/amplify_logger.dart
```diff
@@ -0,0 +1,144 @@
...
+  AmplifyLogger({String namespace = 'default'}) : _namespace = namespace;
+
+  /// Log a message with an optional log level.
+  ///
+  /// The log level defaults to [LogLevel.info].
+  void log({
+    required String message,
+    LogLevel level = LogLevel.info,
+  }) {
+    if (level < logLevel) return;
```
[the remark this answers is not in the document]
fjnoyp: The global log level is handled within `initDefaultLogHandler` 

```
 logStream
        .where((event) => event.level >= globalLogLevel)
        .listen((logEvent) {
      if (useDefaultLogHandler) _printLog(logEvent);
    });
```

packages/logger/lib/src/amplify_logger.dart
fjnoyp: @dnys1 has a good solution for this below by using AmplifyLogger as a wrapper on top of the `package:logging`. 

`package:logging` has built in support for hierarchal logs. 

Setting the logLevel of an individual 'package:logging' logger will set the logLevel of all its children loggers, unless those loggers define overrides. 

The main question now is how we would want to expose this behavior. 

In the solution Dillon recommends below, there is no access to the underlying `package_logging' -> `logger`.  But I think if we were expose it we would have a good API for setting global vs local log levels. 

Ex: 
// Assume AmplifyLogger has a getter for `package_logging->logger` 

hierarchicalLoggingEnabled = true; // must be true else all loggers share same log level 

var mainLog = AmplifyLogger(); // main logger 

var authLog = AmplifyLogger.category(Category.Auth) // all category loggers are automatically children loggers of main logger 

mainLog.logger.level = Level.INFO; // sets log level of main and all category loggers

authLog.logger.level = Level.WARNING; // sets log level of just auth and overrides mainLog.logger's log level

fjnoyp: The above proposal would require importing `amplify_logger` and also the `logging` package from pub.dev 

Instead of directly accessing the `logging` package's logger, we should hide it behind our own API. 

so: 

class AmplifyLogger{

void setLogLevel(LogLevel level){
_logger.level = level
}

Furthermore, the basic AmplifyLogger() implicitly returns the main logger.  Perhaps we can make that more clear by renaming the constructor as AmplifyLogger.main()

What are your thoughts @dnys1 @Jordan-Nelson

fjnoyp:
I really like this solution @dnys1 that you provide above.  

Thanks for taking the time to write it up. 

In my above comments I wrote out how we could handle setting individual logger log levels.  Curious to hear your thoughts for that approach.  I'll keep our discussion there.

packages/logger/lib/src/amplify_logger.dart
fjnoyp: Lastly, I think we should have a `factory` method for AmplifyLogger that doesn't take `Category` in case the user wants to use our package to define their own logging behavior: 

```
factory AmplifyLogger.category(String namespace) {
    return AmplifyLogger._(Logger('Amplify.${namespace}'));
}
```

dnys1:
Glad you like it! 

For external loggers, I'd prefer either 

```dart
AmplifyLogger.fromLogger(this._logger);
```

or change the default factory to be

```dart
AmplifyLogger([String namespace = 'Amplify']) => AmplifyLogger._(Logger(namespace));
```

And yeah, we'll want to expose the log level interface and maybe others too - this was just a rough draft. Log level I would expose like this, probably with some better naming.

```dart
// Our own enum which can map between the package:logging levels and our own.
enum LogLevel { 
 verbose, debug, info, warn, error;

  Level get level => ...
}

extension on Level {
  LogLevel get logLevel => ...
}

class AmplifyLogger {
  final Logger _logger;

  set logLevel(LogLevel logLevel) {
    _logger.logLevel = logLevel.level;
  }

  LogLevel get logLevel => _logger.logLevel.level;
}
```

I would not recommend exposing the wrapped logger. By selectively exposing functionality of `Logger`, we maintain control over the interface and do not bind ourselves to the logging package.

dnys1:
To your note about the imports, I think this functionality should be exported by one of our packages, probably `amplify_core`, and I'll likely want a similar `AWSLogger` exported from `aws_common`. Customers should not have to import anything besides one of these two packages.

packages/logger/lib/src/amplify_logger.dart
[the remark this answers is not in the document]
Jordan-Nelson: I left this on another comment:

> Our main focus right now should be logging within our library since that is an immediate need. I would actually suggest we don't expose this initially. We should think more about the requirements for a logger that customers would use (sending logs to cloud watch, etc.) before we expose this, as those requirements might change what we expose.

packages/logger/lib/src/amplify_logger.dart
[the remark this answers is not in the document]
Jordan-Nelson: Is `ansiColorDisabled` used?

packages/logger/lib/src/amplify_logger.dart
fjnoyp: Yeah we should not expose the dart logging package.   Instead we can just expose our own Amplify logging package and have methods for controlling log level. 

Second point sounds good.  Dillon has a good suggestion for that below whereby our factory method can just take a string: 

AmplifyLogger([String namespace = 'Amplify']) => AmplifyLogger._(Logger(namespace));

I'll update the main PR accordingly

packages/logger/lib/src/amplify_logger.dart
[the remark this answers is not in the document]
fjnoyp: okay sounds good, see updated code

packages/logger/lib/src/amplify_logger.dart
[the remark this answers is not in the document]
fjnoyp: see updated code, switched to factory

packages/amplify/amplify_flutter/pubspec.yaml
```diff
@@ -8,6 +8,7 @@ environment:
   flutter: ">=3.0.0"
 
 dependencies:
+  amplify_logger: ^0.1.0
```
dnys1, line 11: I don't think this needs its own package right now. I would move this to `amplify_core`.

packages/logger/lib/src/amplify_logger.dart
```diff
@@ -0,0 +1,76 @@
...
+  final Logger _logger;
+
+  factory AmplifyLogger([String namespace = 'Amplify']) =>
+      AmplifyLogger._(Logger(namespace));
+
+  factory AmplifyLogger.category(Category category) {
+    return AmplifyLogger._(Logger('Amplify.${category.name}'));
+  }
+
+  void registerPlugin(AmplifyLoggerPlugin plugin) {
```
dnys1, line 39: I've been thinking about this and I think we can come up with a better experience than just automatically de-registering the previous plugin. Maybe something like:

```dart
  final Map<AmplifyLoggerPlugin, StreamSubscription<LogRecord>> _subscriptions =
      {};

  void registerPlugin(AmplifyLoggerPlugin plugin) {
    unregisterPlugin(plugin);
    _subscriptions[plugin] = _logger.onRecord.listen(plugin.handleLogRecord);
  }

  void unregisterPlugin(AmplifyLoggerPlugin plugin) {
    final currentSubscription = _subscriptions.remove(plugin);
    if (currentSubscription != null) {
      unawaited(currentSubscription.cancel());
    }
  }

  void unregisterAllPlugins() {
    for (final plugin in List.of(_subscriptions.keys)) {
      unregisterPlugin(plugin);
    }
  }
```

packages/logger/lib/src/amplify_logger.dart
```diff
@@ -0,0 +1,76 @@
...
+
+  static final Map<String, StreamSubscription<LogRecord>> _subscriptions = {};
+  final Logger _logger;
+
+  factory AmplifyLogger([String namespace = 'Amplify']) =>
+      AmplifyLogger._(Logger(namespace));
+
+  factory AmplifyLogger.category(Category category) {
+    return AmplifyLogger._(Logger('Amplify.${category.name}'));
+  }
```
dnys1, line 37: Let's change the default namespace to `AWS.Amplify`, so that the parent logger `AWSLogger` (which will have namespace `AWS`) can control Amplify + SDK operations.

```dart
  factory AmplifyLogger([String namespace = 'AWS.Amplify']) =>
      AmplifyLogger._(Logger(namespace));

  factory AmplifyLogger.category(Category category) {
    return AmplifyLogger._(Logger('AWS.Amplify.${category.name}'));
  }
```

packages/logger/lib/src/log_event.dart
```diff
@@ -0,0 +1,25 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'log_level.dart';
+
+/// Representation of log information emitted when calling a log method
+/// on an AmplifyLoggerInterface
+class LogEvent {
```
dnys1, line 19: Still needed? Should we transform `LogRecord`s?

packages/logger/lib/src/level_extension.dart
```diff
@@ -0,0 +1,60 @@
...
+      return LogLevel.error;
+    } else if (this.value >= Level.WARNING.value) {
+      return LogLevel.warn;
+    } else if (this.value >= Level.INFO.value) {
+      return LogLevel.info;
+    } else if (this.value >= Level.ALL.value) {
+      return LogLevel.verbose;
+    }
+    return LogLevel.verbose;
+  }
```
dnys1, line 59: LogLevel.none is not correctly reflected here

packages/logger/lib/src/amplify_logger.dart
```diff
@@ -0,0 +1,76 @@
...
+import 'package:logging/logging.dart';
+
+import '../amplify_logger.dart';
+
+enum Category { Analytics, API, Auth, Storage, DataStore }
+
+class AmplifyLogger {
+  AmplifyLogger._(this._logger);
+
+  set logLevel(LogLevel logLevel) {
```
dnys1, line 25: Let's add a getter as well

dnys1:
Let's write some tests as well to verify the leveling and plugin operations

packages/logger/lib/src/amplify_logger.dart
fjnoyp: Good call, updated the code.

packages/amplify_core/lib/src/logger/amplify_logger.dart
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  void unregisterAllPlugins() {
+    for (final plugin in List.of(_subscriptions.keys)) {
+      unregisterPlugin(plugin);
+    }
+  }
+
+  set logLevel(LogLevel logLevel) {
+    hierarchicalLoggingEnabled = true;
```
dnys1, line 54: Probably should move this to the constructor

```dart
AmplifyLogger._(this._logger) {
  hierarchicalLoggingEnabled = true;
}
```

packages/amplify_core/lib/src/logger/ansi_pretty_printer.dart
```diff
@@ -0,0 +1,84 @@
...
+    final buffer = StringBuffer();
+
+    // Log Level
+    buffer.write(_printLevel(record.level));
+
+    // Log Namespace
+    buffer.write(' ');
+
+    final namespace =
+        record.loggerName == 'Amplify' ? '' : record.loggerName.split('.')[1];
```
dnys1, line 37: This needs to be updated as well since we changed it to `AWS.Amplify` - maybe make this a constant on AmplifyLogger?

packages/amplify/amplify_flutter/pubspec.yaml
dnys1, line 11: Remove

packages/amplify_core/lib/src/logger/log_level.dart
```diff
@@ -0,0 +1,66 @@
...
+    return index - other.index;
+  }
+
+  bool operator >(LogLevel value) => index > value.index;
+  bool operator >=(LogLevel value) => index >= value.index;
+  bool operator <(LogLevel value) => index < value.index;
+  bool operator <=(LogLevel value) => index <= value.index;
+
+  Level get level {
+    switch (this) {
```
dnys1, line 52: This is missing `LogLevel.debug`

dnys1 (commented):
Please get CI checks passing and all analysis issues cleaned up and then I think we can merge this!

packages/amplify_core/lib/src/logger/ansi_pretty_printer.dart
```diff
@@ -0,0 +1,85 @@
...
+
+    // Log Level
+    buffer.write(_printLevel(record.level));
+
+    // Log Namespace
+    buffer.write(' ');
+
+    final namespace = record.loggerName == AmplifyLogger.loggerNamespace
+        ? ''
+        : record.loggerName.split('.')[1];
```
dnys1, line 38: This needs to be updated too now

packages/amplify_core/lib/src/logger/log_level.dart
```diff
@@ -0,0 +1,66 @@
...
+      case LogLevel.verbose:
+        return Level.FINEST;
+      case LogLevel.debug:
+        return Level.CONFIG;
```
dnys1, line 55: I think CONFIG falls better into our `info` - I would start debug at FINE.

packages/amplify_core/pubspec.yaml
```diff
@@ -7,12 +7,14 @@ environment:
...
 
 dependencies:
+  ansicolor: ^2.0.1
   async: ^2.8.2
   aws_common: ^0.1.0
   aws_signature_v4: ^0.1.0
   collection: ^1.15.0
   intl: ^0.17.0
   json_annotation: ^4.4.0
+  logging: ^1.0.2
```
dnys1, line 17:
```suggestion
  logging: ^1.0.0
```

dnys1 (approved):

packages/amplify_core/lib/src/logger/amplify_logger.dart
```diff
@@ -0,0 +1,91 @@
...
+    }
+  }
+
+  set logLevel(LogLevel logLevel) {
+    _logger.level = logLevel.level;
+  }
+
+  LogLevel get logLevel => _logger.level.logLevel;
+
+  /// Log a message with a log level of [Level.FINER].
```
Jordan-Nelson, line 63: Should these comments be updated to reflect the `LogLevel` instead of the `Level`? Seems like the use of `logger` and `Level` are internal implementation details. I don't know that we want to expose them.

packages/amplify_core/test/logger/logger_test.dart
```diff
@@ -0,0 +1,224 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
Jordan-Nelson, line 2: *nit*: 2022. Also the header is a different format than the others in this PR. We have been moving to the other format.

packages/amplify_core/lib/src/logger/amplify_logger.dart
```diff
@@ -0,0 +1,91 @@
...
+      AmplifyLogger._(Logger(namespace));
+
+  factory AmplifyLogger.category(Category category) =>
+      AmplifyLogger._(Logger('$loggerNamespace.${category.name}'));
+
+  AmplifyLogger._(this._logger) {
+    hierarchicalLoggingEnabled = true;
+  }
+
+  void registerPlugin(AmplifyLoggerPlugin plugin) {
```
Jordan-Nelson, line 39: Should these methods be static?

packages/amplify_core/lib/src/amplify_class.dart
```diff
@@ -103,6 +103,8 @@ abstract class AmplifyClass {
       );
     }
 
+    AmplifyLogger().registerPlugin(AnsiPrettyPrinter());
```
Jordan-Nelson, line 106: I wanted to see what this looks like by default, so I added the following log statements

```dart
final logger = AmplifyLogger()
      ..registerPlugin(AnsiPrettyPrinter())
      ..logLevel = LogLevel.verbose;

    logger.verbose('verbose message');
    logger.debug('debug message');
    logger.info('info message');
    logger.warn('warn message');
    logger.error('error message');
```

And this is what I see (In VS Code and in a zsh and bash shell). Im I doing something wrong?

Screen Shot 2022-07-12 at 10 42 50 AM

packages/amplify_core/lib/src/logger/amplify_logger.dart
dnys1: No, each logger maintains it's own map of plugins. Due to the hierarchical nature, plugins applied to parent loggers receive all child events so doing `AmplifyLogger().registerPlugin` will receive all Amplify events whereas `AmplifyLogger.category(Category.auth).registerPlugin` registers for only Auth events.

packages/amplify_core/pubspec.yaml
fjnoyp: Why would we use an older version of logging?

packages/amplify_core/lib/src/amplify_class.dart
fjnoyp: Thanks for catching that Jordan.  That's a bummer it won't work there.  I created a SimplePrinter and switched it to be the new default.

packages/amplify_core/lib/src/logger/ansi_pretty_printer.dart
```diff
@@ -0,0 +1,85 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'package:amplify_core/amplify_core.dart';
+import 'package:ansicolor/ansicolor.dart';
+import 'package:logging/logging.dart';
+
+class AnsiPrettyPrinter implements AmplifyLoggerPlugin {
```
Jordan-Nelson, line 19: I would remove this if we have questions about where it does and doesn't work.

packages/amplify_core/lib/src/logger/simple_printer.dart
```diff
@@ -0,0 +1,44 @@
...
+
+class SimplePrinter implements AmplifyLoggerPlugin {
+  SimplePrinter();
+
+  @override
+  void handleLogRecord(LogRecord record) {
+    final buffer = StringBuffer();
+
+    // Log Level
+    buffer.write(record.level.toString());
```
Jordan-Nelson, line 26: Do we want a separator after the level and/or namespace? Probably wasn't needed with the pretty print. With the simple print, it might be nice to have a separator. It all runs together as is.

Current
```
INFO namespace message
INFO namespace message
WARN namespace message
```

Some other options
```
INFO: namespace | message
INFO: namespace | message
WARN: namespace | message
```

```
INFO | namespace | message
INFO | namespace | message
WARN | namespace | message
```

packages/amplify_core/pubspec.yaml
dnys1: This widens the acceptable range from "all versions above 1.0.2 but below 2.0.0" to "all versions between 1.0.0 and 2.0.0". There's not much benefit to us but it allows for people to use our library who have `logging: 1.0.0` in their pubspec without adding a dependency override and we don't particularly care if we resolve to 1.0.0 since there's nothing in 1.0.2 we need.

packages/amplify_core/lib/src/logger/quick_test.dart
```diff
@@ -0,0 +1,13 @@
+import 'package:amplify_core/src/logger/simple_printer.dart';
```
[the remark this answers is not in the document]
dnys1: Yeah, let's remove this.

packages/amplify_core/lib/src/logger/log_level.dart
dnys1: What I mean is:

```
LogLevel.error -> Level.SEVERE, Level.SHOUT
LogLevel.warn -> Level.WARN
LogLevel.info -> Level.CONFIG, Level.INFO
LogLevel.debug -> Level.FINE, Level.FINER
LogLevel.verbose -> Level.FINEST
```

packages/amplify_core/lib/src/logger/ansi_pretty_printer.dart
dnys1: Agreed, we can always add it back later

packages/amplify_core/lib/amplify_core.dart
```diff
@@ -36,11 +36,14 @@ export 'src/config/api/api_config.dart';
 export 'src/config/auth/auth_config.dart';
 export 'src/config/config_map.dart';
 export 'src/config/storage/storage_config.dart';
-
```
dnys1, line 39: Still seeing weird spacing in this file - can we return it to how it was?

packages/amplify_core/lib/src/logger/amplify_logger.dart
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  /// Log a message with level [LogLevel.error]
+  void error(String message, [Object? error, StackTrace? stackTrace]) {
+    _logger.severe(message, error, stackTrace);
+  }
+}
+
+abstract class AmplifyLoggerPlugin {
+  void handleLogRecord(LogRecord record);
```
Jordan-Nelson, line 90: LogRecord from `logger` uses `Level` not `LogLevel`. As a result, `AmplifyLogger.error('message')` prints `SEVERE message` not `ERROR message` which seems confusing. It also means that `Level` is sort of part of the public API.

Do we need our own `LogLevel` and log method names? Can we just use loggers naming convention, and then export `Level` and `LogRecord` from logger?

Side note - I'm wondering if we have a testing gap if this wan't caught. Maybe SimplePrinter needs some tests?

dnys1: Kind of dicey - that's starting to test implementation which makes unit tests really brittle.

The question is whether we want to align with other platforms around LogLevel or whether we're willing to diverge and use `Level` instead. If the former, then we can just print the `LogLevel`, otherwise, we can export `Level`.

packages/amplify_core/lib/src/amplify_class.dart
Jordan-Nelson: Hm. Idk what I did differently, but it seems to be working in VS code debugger now.

I would reduce the default formatting though. It is a lot IMO. Whatever is written in the yellow background is not legible. Definitely would not pass the contrast ratio on an accessibility test 🙂 .

Screen Shot 2022-07-12 at 6 03 45 PM

Jordan-Nelson: We also need to make sure this looks reasonable in all themes if we support it. It looks somewhat odd in light mode.

Screen Shot 2022-07-12 at 6 11 18 PM

packages/amplify_core/lib/src/logger/amplify_logger.dart
Jordan-Nelson: I see. I didn't know those levels were what other platforms use. Let's just make sure we aren't including both the API. We might need our own `LogRecord` because right now if you want to create your own plugin that does something with the `LogLevel`, you have to know how `Level` maps to `LogLevel`.

fjnoyp: Good call Jordan.  I think it would be best to keep our implementation standard with the other Amplify platforms so I'll implement our own LogRecord class that hides the Level and only provides LogLevel to end users.

packages/amplify_core/lib/src/logger/simple_printer.dart
fjnoyp: Sounds good, added some very basic formatting: 

image

packages/amplify_core/lib/src/logger/log_level.dart
fjnoyp: Okay updated to be smallest value equivalent in Level: 

  Level get level {
    switch (this) {
      case LogLevel.verbose:
        return Level.FINEST;
      case LogLevel.debug:
        return Level.FINE;
      case LogLevel.info:
        return Level.CONFIG;
      case LogLevel.warn:
        return Level.WARNING;
      case LogLevel.error:
        return Level.SEVERE;
      case LogLevel.none:
        return Level.OFF;
    }
  }

packages/amplify_core/lib/src/logger/amplify_logger.dart
Jordan-Nelson: Got it. @fjnoyp - API docs would be good here, but could be in a follow up PR.

Jordan-Nelson (approved):

dnys1 (approved):

PR #1749, opened 2022-06-16
chore(auth): Make IndexedDB default on web

dnys1:
Hosted UI requires an asynchronous exchange so some state must be persisted in between these sessions. The inMemory option cannot accommodate this currently.


Jordan-Nelson (approved):
Is there anyway that the in-mem option could handle that? Do we need to have some way to force some keys to be written to disk even for the in-mem option?

ragingsquirrel3 (approved):

PR #1754, opened 2022-06-16
chore: enable integration tests in gh actions

ragingsquirrel3:
This PR has the central config files/scripts to enable integration tests in github actions. It only actually runs them for API with the plan to enable the other categories once API passing.

### DIFF
--- a/.github/composite_actions/fetch_backends/action.yaml
+++ b/.github/composite_actions/fetch_backends/action.yaml
@@ -1,0 +1,47 @@
+name: "Fetch backends"
+description: "Downloads Amplify configurations after getting temporary AWS credentials"
+inputs:
+  aws-region:
+    required: true
+  role-to-assume:
+    required: true
+  # scope for melos, e.g. "amplify_api_example"
+  scope:
+    required: true
+  # Amplify app IDs for specific categories
+  api-app-id:
+    required: true
+  auth-app-id:
+    required: true
+  datastore-app-id:
+    required: true
+  storage-app-id:
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@05b148adc31e091bafbaf404f745055d4d3bc9d2 # 1.6.1
+      with:
+        role-to-assume: ${{ inputs.role-to-assume }}
+        aws-region: ${{ inputs.aws-region }}
+        role-duration-seconds: 900
+
+    - name: Create AWS profile
+      run: ./build-support/create_integration_test_profile.sh
+      shell: bash
+
+    - name: Pull Amplify Configurations
+      run: |
+        API_APP_ID=${{ inputs.api-app-id }} \
+        AUTH_APP_ID=${{ inputs.auth-app-id }} \
+        DATASTORE_APP_ID=${{ inputs.datastore-app-id }} \
+        STORAGE_APP_ID=${{ inputs.storage-app-id }} \
+          melos exec --scope=${{ inputs.scope }} ./tool/pull_test_backend.sh
+      shell: bash
+
+    - name: Delete AWS profile
+      run: rm -rf ~/.aws
+      shell: bash
+
--- a/.github/workflows/amplify_integration_tests.yaml
+++ b/.github/workflows/amplify_integration_tests.yaml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        scope: ["amplify_api_example", "amplify_storage_s3_example", "amplify_auth_cognito_example"]
+        scope: ["amplify_api_example"] #, "amplify_storage_s3_example", "amplify_auth_cognito_example"]
     steps:
       - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
         with:
--- a/.github/workflows/amplify_integration_tests.yaml
+++ b/.github/workflows/amplify_integration_tests.yaml
@@ -53,9 +53,8 @@ jobs:
         uses: reactivecircus/android-emulator-runner@e790971012b979513b4e2fe70d4079bc0ca8a1ae # 2.24.0
         timeout-minutes: 25
         with:
-          api-level: 31
-          # retry once
-          script: melos exec --scope=${{ matrix.scope }} "flutter test --no-pub integration_test/main_test.dart -d emulator-5554" || melos exec --scope=${{ matrix.scope }} "flutter test --no-pub integration_test/main_test.dart -d emulator-5554"
+          api-level: 29
+          script: melos exec -c 1 --scope ${{ matrix.scope }} -- "deviceId=emulator-5554 retries=1 \$MELOS_ROOT_PATH/build-support/integ_test_android.sh"
 
   ios:
     runs-on: macos-latest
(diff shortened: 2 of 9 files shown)


.github/workflows/amplify_integration_tests.yaml
```diff
@@ -0,0 +1,108 @@
...
+    - cron:  '0 13 * * *'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  android:
+    runs-on: macos-latest
+    permissions:
```
dnys1, line 16: Would add a note here why these are needed

.github/workflows/amplify_integration_tests.yaml
```diff
@@ -0,0 +1,108 @@
...
+        run: |
+          melos exec --scope=${{ matrix.scope }} --file-exists="integration_test/main_test.dart" --fail-fast flutter build apk --verbose
+
+      - name: Run Android integration tests
+        uses: reactivecircus/android-emulator-runner@76c2bf6f95ed6458fd659a1bdb680a0f8df232dc # 2.23.0
+        timeout-minutes: 25
+        with:
+          api-level: 29
+          # retry once
+          script: melos exec --scope=${{ matrix.scope }} "flutter test --no-pub integration_test/main_test.dart -d emulator-5554" || melos exec --scope=${{ matrix.scope }} "flutter test --no-pub integration_test/main_test.dart -d emulator-5554"
```
dnys1, line 58: Can we update our integ test script so it can be used here?

.github/workflows/amplify_integration_tests.yaml
```diff
@@ -0,0 +1,108 @@
...
+
+      - name: Build example app with integration tests
+        run: |
+          melos exec --scope=${{ matrix.scope }} --file-exists="integration_test/main_test.dart" --fail-fast flutter build apk --verbose
+
+      - name: Run Android integration tests
+        uses: reactivecircus/android-emulator-runner@76c2bf6f95ed6458fd659a1bdb680a0f8df232dc # 2.23.0
+        timeout-minutes: 25
+        with:
+          api-level: 29
```
dnys1, line 56: Should be 31 to match compileSdk

.github/workflows/amplify_integration_tests.yaml
```diff
@@ -0,0 +1,108 @@
...
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # 2.4.0
+        with:
+          persist-credentials: false
+      
+      - name: Install dependencies
+        uses: ./.github/composite_actions/install_dependencies
+
+      - name: Boot iOS Simulator
+        run: |
+          xcrun simctl boot "iPhone 12"
```
dnys1, line 79: Would always use latest here, in this case iPhone 13

.github/workflows/amplify_integration_tests.yaml
```diff
@@ -0,0 +1,108 @@
...
+        run: |
+          melos exec --scope=${{ matrix.scope }} flutter build ios --simulator
+
+      - name: Run integration tests
+        # retry once
+        id: iosTestFirstAttempt
+        continue-on-error: true
+        timeout-minutes: 10
+        run: |
+          melos exec --scope=${{ matrix.scope }} "flutter test --no-pub integration_test/main_test.dart -d iPhone"
```
dnys1, line 102: Same comment as above - can we update integ script if needed so it can be used here?

.github/workflows/amplify_integration_tests.yaml
```diff
@@ -0,0 +1,108 @@
...
+
+      - name: run integration tests (2nd attempt)
+        if: steps.iosTestFirstAttempt.outcome == 'failure'
+        timeout-minutes: 30
```
dnys1, line 106: Why is timeout higher?

packages/api/amplify_api/example/tool/pull_test_backend.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -e
+
+APP_ID=$API_APP_ID ../../../../build-support/pull_backend_by_app_id.sh
```
dnys1, line 4: Is this needed? Do we really want to make this in every package?

.github/composite_actions/fetch_backends/action.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - name: Pull Amplify Configurations
+      run: |
+        API_APP_ID=${{ inputs.api-app-id }} \
+        AUTH_APP_ID=${{ inputs.auth-app-id }} \
+        DATASTORE_APP_ID=${{ inputs.datastore-app-id }} \
+        STORAGE_APP_ID=${{ inputs.storage-app-id }} \
+          melos exec --scope=${{ inputs.scope }} ./tool/pull_test_backend.sh
+      shell: bash
+
+    - name: Delete AWS profile
```
dnys1, line 44: What's the benefit of this? Is it cached by default?

dnys1 (commented):
Looking good!

.github/composite_actions/fetch_backends/action.yaml
ragingsquirrel3: paranoia. Not cached by default, just limits the amount of time this is persisted so it will not be available to later steps.

.github/workflows/amplify_integration_tests.yaml
ragingsquirrel3: Agree this is hacky thing to avoid that, but I will take a shot at doing that in next rev.

.github/workflows/amplify_integration_tests.yaml
ragingsquirrel3: Is this required? It seems there are no default images for 30 or later published so to get it working, have to change some settings which results in much slower build (adds 5-10min per build per my rough observations) https://github.com/ReactiveCircus/android-emulator-runner/issues/222

ragingsquirrel3: actually beginning to think 30+ actually too slow to use after trying it more

.github/workflows/amplify_integration_tests.yaml
ragingsquirrel3: changed retries

.github/workflows/amplify_integration_tests.yaml
ragingsquirrel3: done

.github/workflows/amplify_integration_tests.yaml
ragingsquirrel3: done

packages/api/amplify_api/example/tool/pull_test_backend.sh
dnys1: Hmm yeah, I see. I'm sure there's a better way, but we don't need to block on it.

dnys1 (approved):

.github/workflows/amplify_integration_tests.yaml
```diff
@@ -0,0 +1,108 @@
+name: Amplify Integration Tests
+on:
+  push:
+    branches: [main, next, stable, feat/*]
+  schedule:
+    # 6am pacific time daily, only runs on default branch
+    - cron:  '0 13 * * *'
```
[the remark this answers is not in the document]
ragingsquirrel3: replied offline

.github/workflows/amplify_integration_tests.yaml
ragingsquirrel3: I agree. In general, the android emulator in this environment provided a whole slew of challenges, this just being the latest. It also seems to block building on a non-mac env bc the linux instances don't run the emulator fast enough for flutter to not timeout. Long-term, would like to spike/investigate running on real devices in device farm to avoid the emulator requirement and gain other benefits.

PR #1761, opened 2022-06-17
fix(amplify_authenticator): fix form field label overlap, enable other platforms

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Fix an issue present on web where the label of the form field overlapped with the content.
- Enable the authenticator example app on all platforms


packages/amplify_authenticator/example/macos/Runner/Info.plist
```diff
@@ -0,0 +1,32 @@
...
+	<key>CFBundleVersion</key>
+	<string>$(FLUTTER_BUILD_NUMBER)</string>
+	<key>LSMinimumSystemVersion</key>
+	<string>$(MACOSX_DEPLOYMENT_TARGET)</string>
+	<key>NSHumanReadableCopyright</key>
+	<string>$(PRODUCT_COPYRIGHT)</string>
+	<key>NSMainNibFile</key>
+	<string>MainMenu</string>
+	<key>NSPrincipalClass</key>
+	<string>NSApplication</string>
```
haverchuck, line 30: I can't recall - does macOS need the info.plist update for Hosted UI like iOS? If so, it might be convenient to just add it here (individual devs can update the actual redirect value)

```
	<key>CFBundleURLTypes</key>
	<array>
		<dict>
			<key>CFBundleURLSchemes</key>
			<array>
				<string>myapp</string>
			</array>
		</dict>
	</array>
```

haverchuck (approved):
Left one comment, but otherwise LGTM

PR #1763, opened 2022-06-17
chore(authenticator): Exception handling

dnys1:
Parse `message` from all thrown exceptions in Authenticator. Don't show `e.toString()` since this is probably never what we want to do. Instead, print "an unknown error has occurred" and log the message to the console.

### DIFF
--- a/packages/amplify_authenticator/pubspec.yaml
+++ b/packages/amplify_authenticator/pubspec.yaml
@@ -19,6 +19,9 @@ dependencies:
   flutter_localizations:
     sdk: flutter
   intl: ^0.17.0
+  smithy:
+    hosted: https://pub.dillonnys.com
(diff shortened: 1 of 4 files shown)


Jordan-Nelson (approved):

haverchuck (approved):

packages/amplify_authenticator/pubspec.yaml
```diff
@@ -19,6 +19,9 @@ dependencies:
   flutter_localizations:
     sdk: flutter
   intl: ^0.17.0
+  smithy:
+    hosted: https://pub.dillonnys.com
```
haverchuck, line 23: Can we add a todo to change this later?

haverchuck (approved):
Looks good, left one comment.

PR #1764, opened 2022-06-17
fix(authenticator): Collect MFA types

dnys1:
Collect phone number during sign up if registered as an MFA type


Jordan-Nelson (approved):
This should go to main as well, correct?

dnys1:
Yep, good catch. Opened: https://github.com/aws-amplify/amplify-flutter/pull/1781

PR #1767, opened 2022-06-18
feat(auth): Username/password flow

dnys1:
Adds support for the `USER_PASSWORD_AUTH` flow

### DIFF
--- a/packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_in_state_machine.dart
+++ b/packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_in_state_machine.dart
@@ -28,7 +28,9 @@ import 'package:amplify_auth_cognito_dart/src/flows/srp/srp_password_verifier_wo
 import 'package:amplify_auth_cognito_dart/src/jwt/jwt.dart';
 import 'package:amplify_auth_cognito_dart/src/model/cognito_user.dart';
 import 'package:amplify_auth_cognito_dart/src/model/sign_in_parameters.dart';
-import 'package:amplify_auth_cognito_dart/src/sdk/cognito_identity_provider.dart';
+import 'package:amplify_auth_cognito_dart/src/sdk/cognito_identity_provider.dart'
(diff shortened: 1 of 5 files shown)


packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_in_state_machine.dart
```diff
@@ -182,47 +179,30 @@ class SignInStateMachine extends StateMachine<SignInEvent, SignInState> {
...
+      case AuthFlowType.userSrpAuth:
+        expectPassword();
+        return initiateSrpAuth(event);
+      case AuthFlowType.customAuth:
+        return initiateCustomAuth(event);
+      case AuthFlowType.userPasswordAuth:
+        return initiateUserPasswordAuth(event, expectPassword());
+      case AuthFlowType.refreshToken:
+      case AuthFlowType.refreshTokenAuth:
+      case AuthFlowType.adminNoSrpAuth:
```
haverchuck, line 200: Why do we have the admin flow types here? These require developer credentials and I don't think would ever be in a client-side app.

haverchuck (commented):
Left one question; otherwise LGTM

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_in_state_machine.dart
dnys1: It's part of the SDK's enum

dnys1: I think it's just the diff formatting. The next line after the break is `throw StateError('Unsupported auth flow')`: https://github.com/aws-amplify/amplify-flutter/pull/1767/files/6add824531df9a3d7fd1336bec88acb25f5e2182#diff-9d1c23e771518da5d10ac1f888bd38c5f865868332b126a16ab6d19fd3ef028fR205

Jordan-Nelson (approved):

PR #1780, opened 2022-06-21
feat: exposes secure-storage as plugin constructor parameter

haverchuck:
*Description of changes:*
Exposes secure-storage as an auth plugin constructor parameter.

### DIFF
--- a/packages/amplify/amplify_flutter/lib/amplify_flutter.dart
+++ b/packages/amplify/amplify_flutter/lib/amplify_flutter.dart
@@ -21,6 +21,9 @@ import 'src/amplify_impl.dart';
 
 export 'package:amplify_core/amplify_core.dart' hide Amplify;
 
+// ignore: depend_on_referenced_packages
+export 'package:amplify_secure_storage/amplify_secure_storage.dart';
+
 /// Top level singleton Amplify object.
 AmplifyClass get Amplify {
   return AmplifyClass.instance ??= AmplifyClassImpl();
--- a/packages/amplify_authenticator/example/lib/main.dart
+++ b/packages/amplify_authenticator/example/lib/main.dart
@@ -34,6 +34,15 @@ class _MyAppState extends State<MyApp> {
   void _configureAmplify() async {
     try {
       await Amplify.addPlugin(AmplifyAuthCognito());
+      // Uncomment this block, and comment out the one above, in order to persist credentials
+      // await Amplify.addPlugin(AmplifyAuthCognito(credentialStorage: AmplifySecureStorage(
+      //       config: AmplifySecureStorageConfig(
+      //         scope: 'authtest',
+      //         webOptions: WebSecureStorageOptions(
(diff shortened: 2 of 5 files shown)


packages/amplify/amplify_flutter/lib/amplify_flutter.dart
```diff
@@ -21,6 +21,9 @@ import 'src/amplify_impl.dart';
 
 export 'package:amplify_core/amplify_core.dart' hide Amplify;
 
+// ignore: depend_on_referenced_packages
```
dnys1, line 24: Since you added it to the pubspec, this is no longer needed. In general, `ignore` is almost never the right option. The lint prevents publishing errors.

packages/amplify_authenticator/example/lib/main.dart
```diff
@@ -34,6 +34,15 @@ class _MyAppState extends State<MyApp> {
   void _configureAmplify() async {
     try {
       await Amplify.addPlugin(AmplifyAuthCognito());
+      // Uncomment this block, and comment out the one above, in order to persist credentials
```
dnys1, line 37: I would use a different example since indexedDB will become the default option: https://github.com/aws-amplify/amplify-flutter/pull/1749

dnys1: I feel like this should be in the auth example, too, not the authenticator example.

dnys1 (approved):

packages/auth/amplify_auth_cognito/lib/src/auth_plugin_impl.dart
```diff
@@ -30,9 +30,11 @@ import 'package:flutter/services.dart';
 /// {@endtemplate}
 class AmplifyAuthCognito extends AmplifyAuthCognitoDart {
   /// {@macro amplify_auth_cognito.amplify_auth_cognito}
-  AmplifyAuthCognito()
+  AmplifyAuthCognito({
```
Jordan-Nelson, line 33: We should add a doc comment to the constructor to callout that `credentialStorage` defaults to `AmplifySecureStorage` with a scope of "auth" and that changing the credential storage, or the credential storage config, will likely result in end users having to re-authenticate.

Something along the lines of:
```dart
  /// To change the default behavior of credential storage,
  /// provide a [credentialStorage] value. If no value is provided,
  /// [AmplifySecureStorage] will be used with a `scope` of "auth".
  ///
  /// **Warning:** Changing the credential provider, or customizing 
  /// the config value for [AmplifySecureStorage], will likely result in 
  /// end users having to re-authenticate then next time the app is opened.
```

packages/amplify/amplify_flutter/lib/amplify_flutter.dart
haverchuck: I thought this was the case as well - but the linter is still complaining.

PR #1782, opened 2022-06-21
chore(version): Bump `aws_signature_v4`

dnys1:
Patch release for sigv4 pacakge

## Cherry Picks
- ede360097c6ccd53aa6d377a2d59a42fdb18515d

### DIFF
--- a/packages/aws_signature_v4/pubspec.yaml
+++ b/packages/aws_signature_v4/pubspec.yaml
@@ -1,6 +1,6 @@
 name: aws_signature_v4
 description: Dart implementation of the AWS Signature Version 4 algorithm, for communication with AWS services.
-version: 0.1.1
+version: 0.1.1+1
 homepage: https://docs.amplify.aws/lib/q/platform/flutter/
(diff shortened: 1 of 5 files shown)


packages/aws_signature_v4/pubspec.yaml
```diff
@@ -19,9 +19,9 @@ dependencies:
   meta: ^1.7.0
   path: ^1.8.0
 
-dependency_overrides:
-  aws_common:
-    path: ../aws_common
+# dependency_overrides:
```
haverchuck, line 22: Why are we leaving this commented out?

haverchuck (commented):
left one inline question, otherwise lgtm

packages/aws_signature_v4/pubspec.yaml
dnys1: Just so I don't have to type it out when I'm developing. And because I can't use `pubspec_overrides` in main yet.

haverchuck (approved):

ragingsquirrel3 (approved):

PR #1802, opened 2022-06-22
feat(core): AWS config file

dnys1:
Adds support for the AWS config and credentials files, as defined in the SEP. This is critical for desktop support where configuration profiles are the main way credentials are managed.


haverchuck (commented):
Can you please give a bit more detail on why this is critical for desktop support?

dnys1:
Have you ever tried to use the Amplify or AWS CLI without a config file? If you do, you're gonna have a bad time lol. You're either copy/pasting creds into `export` statements or, in the case of Amplify, SOL.

It was the first thing Vlad asked about when I showed him the demo. For Windows, esp., environment variables are not as ubiquitous as on UNIX.

haverchuck:
I understand why that is necessary for the CLIs, which require the developer's credentials to provision resources or call other operations which require developer credentials. I'm just trying to get clarity on why access to these credentials is necessary for our library.

dnys1:
Well, so the signer is the only escape hatch we offer developers right now. In lieu of a full AWS SDK, it's the least we can do to provide proper ergonomics around that experience.

haverchuck:
Ah ok, that makes sense.

Jordan-Nelson (approved):
lgtm. Did not test at all but test cases look pretty exhaustive.

pytomtoto:
> 您能否详细说明为什么这对桌面支持至关重要？

I have multiple aws accounts ， I need switch accounts one by one。

packages/aws_common/lib/src/config/aws_config_value.dart
```diff
@@ -16,31 +16,58 @@ import 'dart:async';
...
+  region._(
+    'AWS_REGION',
+    'us-east-1',
+  ),
+
+  /// The location (filepath) of the AWS configuration file.
+  ///
+  /// Defaults to `~/.aws/config`.
+  configFile._(
+    'AWS_CONFIG_FILE',
```
haverchuck, line 51: Sorry, I think we talked about this at one point, but can you remind me what this is for?

dnys1: No problem. It's a standard environment variable for setting the location of the config file: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html#envvars-list-AWS_CONFIG_FILE

haverchuck (approved):

PR #1808, opened 2022-06-23
chore(auth): Better Hosted UI exceptions

dnys1:
Improves Hosted UI exception messages on Android, e.g. when no browser is available, and when the exception originated from a Lambda.

### DIFF
--- a/packages/auth/amplify_auth_cognito/lib/src/flows/hosted_ui/hosted_ui_platform_flutter.dart
+++ b/packages/auth/amplify_auth_cognito/lib/src/flows/hosted_ui/hosted_ui_platform_flutter.dart
@@ -107,7 +107,7 @@ class HostedUiPlatformImpl extends io.HostedUiPlatformImpl {
           recoverySuggestion = "Ensure you've added the <queries> tag to your "
               'AndroidManifest.xml as outlined in the docs';
         }
-        throw UnknownException(
+        throw UrlLauncherException(
           message ?? 'An unknown error occurred',
           recoverySuggestion: recoverySuggestion,
         );
--- a/packages/auth/amplify_auth_cognito/lib/src/flows/hosted_ui/hosted_ui_platform_flutter.dart
+++ b/packages/auth/amplify_auth_cognito/lib/src/flows/hosted_ui/hosted_ui_platform_flutter.dart
@@ -20,6 +20,7 @@ import 'package:amplify_auth_cognito/src/native_auth_plugin.dart';
 import 'package:amplify_auth_cognito_dart/src/flows/hosted_ui/hosted_ui_platform_io.dart'
     as io;
 import 'package:amplify_core/amplify_core.dart';
+import 'package:flutter/services.dart';
 
 /// {@template amplify_auth_cognito.hosted_ui_platform_flutter}
 /// The hybrid Flutter implementation of [HostedUiPlatform] which uses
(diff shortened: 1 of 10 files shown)


packages/auth/amplify_auth_cognito/lib/src/flows/hosted_ui/hosted_ui_platform_flutter.dart
```diff
@@ -88,8 +89,29 @@ class HostedUiPlatformImpl extends io.HostedUiPlatformImpl {
...
+            'The user cancelled the sign-in flow',
+          );
+        }
+        // Generated Android message is `CLASS_NAME: message`
+        var message = e.message;
+        if (message != null && message.contains(': ')) {
+          message = message.split(': ')[1];
+        }
+        String? recoverySuggestion;
+        if (e.code == 'NOBROWSER') {
```
Jordan-Nelson, line 106: Do we want to throw a more specific exception here? Customers might want to catch this and display something specific to the customer (Ex: "You default browser doesn't support Sign in with <Google/Facebook/etc.>. We recommend having Chrome installed as your default browser.").

We might actually want to do that in the authenticator. With these changes, the authenticator will display "No browser available for launching URL intent", which is certainly better than the current experience but could still be improved. That isn't super clear to and end user how to resolve the issue.

I see this even if Chrome is installed, but an unsupported browser (DuckDuckGo) is the default browser. So I don't think this is a very obscure edge case. Some end users will almost certainly run into this and likely be confused.

packages/auth/amplify_auth_cognito_dart/lib/src/flows/hosted_ui/hosted_ui_platform_io.dart
```diff
@@ -50,11 +50,25 @@ class HostedUiPlatformImpl extends HostedUiPlatform {
   </head>
   <body>
     <center>
-      <h2>Success! You can now close this window.</h2>
+      <h2>$message</h2>
     </center>
   </body>
 </html>''';
 
+  String _htmlForParams(Map<String, String> parameters) {
```
Jordan-Nelson, line 58: Nice. Wasn't sure how easy this was going to be to handle.

Jordan-Nelson (approved):
I left one comment about throwing a more specific exception for no browsers, but look good. I tested this and it addresses a lot of the edge case issues I was seeing yesterday!

packages/auth/amplify_auth_cognito/lib/src/flows/hosted_ui/hosted_ui_platform_flutter.dart
Jordan-Nelson: This use to cause the app to crash in Amplify-Android. When they resolved that issue, I think they updated it to be a generic AuthException. However, the "underlying" exception contained a specific exception, which would give customers a way to handle it if they wanted. I think we just need some way for a customer handle it reliably if they want to.

dnys1: Yep, totally agree. I've created a `UrlLauncherException` type and updated the `UnknownException`s throughout the Hosted UI platform code to use this, instead, as appropriate.

Jordan-Nelson (approved):

PR #1809, opened 2022-06-23
chore: Fix some analysis issues

dnys1:
Fixes some miscellaneous analysis issues

### DIFF
--- a/packages/auth/amplify_auth_cognito/lib/src/auth_plugin_impl.dart
+++ b/packages/auth/amplify_auth_cognito/lib/src/auth_plugin_impl.dart
@@ -27,28 +27,28 @@ import 'package:amplify_core/amplify_core.dart';
 import 'package:amplify_secure_storage/amplify_secure_storage.dart';
 import 'package:flutter/services.dart';
 
-
 /// {@template amplify_auth_cognito.amplify_auth_cognito}
 /// The AWS Cognito implementation of the Amplify Auth category.
-/// To change the default behavior of credential storage,
-/// provide a [credentialStorage] value. If no value is provided,
-/// [AmplifySecureStorage] will be used with a `scope` of "auth".
-///
-/// **Warning:** Changing the credential provider, or customizing 
-/// the config value for [AmplifySecureStorage], will likely result in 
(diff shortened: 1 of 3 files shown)


packages/auth/amplify_auth_cognito/lib/src/auth_plugin_impl.dart
```diff
@@ -27,28 +27,24 @@ import 'package:amplify_core/amplify_core.dart';
...
 import 'package:flutter/services.dart';
 
-
 /// {@template amplify_auth_cognito.amplify_auth_cognito}
 /// The AWS Cognito implementation of the Amplify Auth category.
-/// To change the default behavior of credential storage,
-/// provide a [credentialStorage] value. If no value is provided,
-/// [AmplifySecureStorage] will be used with a `scope` of "auth".
-///
-/// **Warning:** Changing the credential provider, or customizing
```
dnys1, line 37: I removed this because it felt a bit insulting to the developer. We have no reason to believe this is the case.

Jordan-Nelson: Maybe this just needs to be re-worded. I do think a warning is needed.

If a customer is using AmplifySecureStorage, any config change (scope, database name, schema name) will result in end users having to re-authenticate on at least one platform. We won't handle migrations in those cases.

You are right that we don't really know what will happen if they are using a custom implementation. I think it is likely that if they switch from using our solution to a custom implementation, that end users will need to re-authenticate, but not necessarily. I would be fine with omitting any mention of a custom impl. Presumably anyone implementing their own solution has thought through those scenarios.

Jordan-Nelson: Yeah that works

haverchuck (approved):

packages/auth/amplify_auth_cognito/lib/src/auth_plugin_impl.dart
dnys1: Done 👍

Jordan-Nelson (approved):

PR #1821, opened 2022-06-27
chore(repo): Update `mono_repo`

dnys1:
Updates mono_repo which contains security improvements like dependency pinning and explicit permissions

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,4 +1,4 @@
-# Created with package:mono_repo v6.2.1
+# Created with package:mono_repo v6.3.0
 name: Dart Lint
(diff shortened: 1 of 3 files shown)


haverchuck (approved):

Jordan-Nelson (approved):
I was hoping to see https://github.com/google/mono_repo.dart/pull/382 was included in the latest release 😄 

lgtm

dnys1:
Ikr! I think they just want to make sure that any additions are gonna solve multiple problems so that the API doesn't spiral out of control. If not exactly that PR, we should get something similar soon as there were discussions in the melos repo about this as well from what I could see.

PR #1842, opened 2022-07-01
Chore(Authenticator): Larger screen responsive fix

Equartey:
*Description of changes:*
Adds constraints and card view on larger screen sizes to avoid edge to edge widgets.

Verified changes on: iPhone Pro, iPad Pro (9.7in & 12.9), MacOS, Pixel, and Chrome.

Screen Shot 2022-07-01 at 10 25 45 AM


Equartey:
Removed the title above the form: 

Screen Shot 2022-07-01 at 1 00 20 PM

Screen Shot 2022-07-01 at 1 08 53 PM

packages/amplify_authenticator/lib/src/screens/authenticator_screen.dart
```diff
@@ -66,11 +66,15 @@ class AuthenticatorScreen extends StatelessAuthenticatorComponent {
     final bool isDesktop =
         screenSize.width > AuthenticatorContainerConstants.landScapeView;
```
Jordan-Nelson, line 67: I think we should add a min height here as well. If a phone in rotated, it switches to the "desktop" view, which probably isn't what we want.

```suggestion
    final bool isDesktop =
        screenSize.width > AuthenticatorContainerConstants.landScapeView &&
            screenSize.height > AuthenticatorContainerConstants.mediumWidth;
```

I know you didn't name this - but `isDesktop` implies we only want to use this layout on desktop devices. We want to use it on large tablets too. I'm not sure what a good name would be here. Maybe `isLargeLayout` or something. Or you could invert the conditions and name it `isMobileLayout`.

Jordan-Nelson (commented):
Looks really good. I played around with this locally and on desktop, iPhone, iPad, and web. I left a couple comments.

packages/amplify_authenticator/lib/src/screens/authenticator_screen.dart
dnys1: You can also use `screenSize.longestSide` which seems to give a decent behavior for me

packages/amplify_authenticator/lib/src/screens/authenticator_screen.dart
```diff
@@ -97,9 +101,21 @@ class AuthenticatorScreen extends StatelessAuthenticatorComponent {
...
     }
 
-    return Container(
-      constraints: BoxConstraints(maxWidth: containerWidth),
-      child: SafeArea(child: child),
+    return SingleChildScrollView(
+      child: Padding(
+        padding: EdgeInsets.all(containerPadding),
+        child: Center(
+          child: ConstrainedBox(
```
dnys1, line 108: I'm not sure the original intention of `containerPadding` but it's rather inflexible and looks weird for some screen sizes (like 800x600, where I need to scroll because of the padding even though it should all fit). The following stack seems to work better IMO:

```suggestion
    return Center(
      child: SingleChildScrollView(
        child: ConstrainedBox(
```

packages/amplify_authenticator/lib/src/screens/authenticator_screen.dart
```diff
@@ -132,13 +148,6 @@ class _FormWrapperView extends StatelessAuthenticatorComponent {
       padding: padding,
       child: Column(
         children: <Widget>[
-          Text(
```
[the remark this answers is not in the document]
Equartey: Good catch, I don't think we want to remove the title from the other screens. I've updated the code to reflect this. The generate script touched some other files in a minor way, let me know if this created any issues.

packages/amplify_authenticator/lib/src/screens/authenticator_screen.dart
Equartey: Moved to isMobile and used `screenSize.shortSide` to good effect. Thanks for the suggestions.

packages/amplify_authenticator/lib/src/screens/authenticator_screen.dart
Equartey: This is something we actually played around with, but chose against as it creates some unwanted behavior. For example, the card widget no longer is sticky to the top of the screen on Mobile which causes inconsistencies as you switch tabs/screens. With the changes from above, smaller screens like 800x600 should now reflect the mobile design and be more usable.

Jordan-Nelson (approved):
Looks pretty good. I think we might want to make some improvements to smaller tablet / large phone size screens as Dillon had pointed out, but other than that I think this is good to go.

packages/amplify_authenticator/lib/src/l10n/generated/input_localizations.dart
```diff
@@ -257,7 +257,7 @@ abstract class AuthenticatorInputLocalizations {
   ///
   /// In en, this message translates to:
   /// **'at least {numCharacters, plural, =1{1{characterType} character} other{{numCharacters}{characterType} characters}}'**
-  String passwordRequirementsAtLeast(int numCharacters, String characterType);
+  String passwordRequirementsAtLeast(num numCharacters, String characterType);
```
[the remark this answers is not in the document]
Equartey: Yeah this is from the script. I agree its probably not needed, but if someone runs the script again it'll show up so I figure might as well keep it 🤷

packages/amplify_authenticator/lib/src/screens/authenticator_screen.dart
Equartey: Not sure the padding is an issue with the current break points. For reference, this is the current behavior below:

Equartey: https://user-images.githubusercontent.com/6456935/178826198-adaa2565-b5a6-4217-9567-0724bf1d99af.mov

Jordan-Nelson: I think Dillon was thinking about the Sign Up form since it can be much longer. For example, see the screenshots below where the Sign In form looks good, but the Sign Up form requires scrolling.

Screen Shot 2022-07-14 at 10 03 16 AM

Screen Shot 2022-07-14 at 10 03 21 AM

I tested my idea of adding a medium breakpoint at 820px where the padding is set to 32px, and I'm not sure it looks very good. The Sign In screen seems too close to the top of the screen.

Screen Shot 2022-07-14 at 10 24 43 AM

I think it is going to be almost impossible to make this look great on every screen size for every configuration and number of sign up fields. I think we should try to ensure that:
1. The Sign In screen is not cut off in common scenarios. The maximum length would be a config that uses all 4 social providers, and allows phone OR email as sign up methods. I wouldn't necessarily consider this common though. I would make sure it isn't cut off with at least 2 or 3 social providers though.
2. The standard sign in and sign up flow (~5 sign up fields) isn't cut off on most screen sizes.

Just playing around with it, adding a medium breakpoint with a padding of 64px seems like a happy medium. The sign up screen will still be cut off a bit if you are just over the medium breakpoint and have all four social providers, but I think this is okay.

You might want to consider making an enum or class for breakpoint if you add a medium one. It will be easier to understand which values are used at which sizes. Something like below.

```dart
/// breakpoint
enum Breakpoint {
  small(maxScreenSize: 0, verticalPadding: 0),
  medium(maxScreenSize: 672, verticalPadding: 64),
  large(maxScreenSize: 900, verticalPadding: 96);

  const Breakpoint({
    required this.maxScreenSize,
    required this.verticalPadding,
  });

  final int maxScreenSize;
  final double verticalPadding;

  static Breakpoint of(BuildContext context) {
    final size = MediaQuery.of(context).size;
    if (size.shortestSide < medium.maxScreenSize) {
      return small;
    } else if (size.shortestSide < large.maxScreenSize) {
      return large;
    } else {
      return large;
    }
  }
}

/// example use
final breakpoint = Breakpoint.of(context);
final isMobile = breakpoint == Breakpoint.small;
final double containerPadding = breakpoint.verticalPadding;
```

dnys1: Note: In the video, when I stretch it to a really long width, it renders the mobile view. Is that expected?

Jordan-Nelson: I think centering it with animation is fine as well. We can go with that. Can we make sure that it still looks reasonable for very long Sign Up forms (one with every form field). By reasonable, I mean that the top of the form doesn't start off the top of the page. Having the top of the form cut off would be much worse than having the bottom cut off IMO.

FWIW - I don't think hard coding the padding at breakpoints is an unusual pattern, at least on web. Netflix and Facebook both hard code the top padding/margin on their sign in screens until the mobile breakpoint (Netflix at 90, Facebook at 136). Google centers it, but leaves a minimum top padding/margin. Those were the first three sites I looked at. I think both are common patterns.

dnys1: @Equartey This is the code for that video, starting at line 81 of `authenticator_screen.dart`. Probably needs to be re-worked a bit given the other conversations in this PR.

```dart
Widget child;
switch (step) {
  case AuthenticatorStep.onboarding:
  case AuthenticatorStep.signIn:
  case AuthenticatorStep.signUp:
    child = AnimatedSize(
      alignment: Alignment.topCenter,
      child: AuthenticatorTabView(
        tabs: signInUpTabs,
        initialIndex: step == AuthenticatorStep.signUp ? 1 : 0,
      ),
      duration: const Duration(milliseconds: 200),
    );
    break;
  case AuthenticatorStep.confirmSignUp:
  case AuthenticatorStep.confirmSignInCustomAuth:
  case AuthenticatorStep.confirmSignInMfa:
  case AuthenticatorStep.confirmSignInNewPassword:
  case AuthenticatorStep.resetPassword:
  case AuthenticatorStep.confirmResetPassword:
  case AuthenticatorStep.verifyUser:
  case AuthenticatorStep.confirmVerifyUser:
    child = _FormWrapperView(step: step);
    break;
  case AuthenticatorStep.loading:
    throw StateError('Invalid step: $this');
}

if (isMobile) {
  child = SingleChildScrollView(
    child: Center(
      child: Padding(
        padding: EdgeInsets.all(containerPadding),
        child: ConstrainedBox(
          constraints: BoxConstraints(maxWidth: containerWidth),
          child: child,
        ),
      ),
    ),
  );
} else {
  child = Center(
    child: SingleChildScrollView(
      child: ConstrainedBox(
        constraints: BoxConstraints(maxWidth: containerWidth),
        child: Padding(
          padding: const EdgeInsets.symmetric(vertical: 16),
          child: Card(child: child),
        ),
      ),
    ),
  );
}

return SafeArea(child: child);
```

Jordan-Nelson: > Note: In the video, when I stretch it to a really long width, it renders the mobile view. Is that expected?

Originally the Card view was showing on phones in landscape mode. I had told @Equartey that it looked odd, and the full screen mode should probably be used even in landscape mode on phones. So I think what you are seeing is the result of having a small height and wide width, which is being considered a phone in landscape mode with the current logic. The logic probably needs to be tweaked here as I would agree that very wide screens shouldn't show full screen.

I think right now we are using 672 as the breakpoint. If the view has a height or a width smaller than 672, it is being considered a phone.

One thing to note is that 672 is a pretty big breakpoint for mobile devices. I think 480 is a more common mobile breakpoint width. We could switch this. 672 was what we had previously, but I am not sure where that came from. @Equartey I will trust your judgement on what that looks like on screen sizes between 480 and 672 to decide if we should switch that.

Either way, we can add logic to prevent very wide screens from being considered a phone, even if the height is small. Something like

```dart
// `width < 480` -> phone in portrait, `(width < 960 && height < 480)` -> phone in landscape
final isMobile = width < 480 || (width < 960 && height < 480)
```

Equartey: Thanks all for the comments. I have implemented the suggestions and tested them on the various platforms. It looks great, it now behaving in a manner that satisfy all the comments brought up here. 

* Small height & Large Width displays desktop view, but landscape mobile is still the mobile layout
* Layout is now more dynamic as padding isn't being relied on for the layout
* The mobile to tablet breakpoints provide a seamless transition from full width to card layout

Jordan-Nelson: I think we are very very close, but the experience on web is a little goofy with a long sign up screen because the scrollable view only wraps the card. See in the screenshot below where the scrollbar is. This means that scrolling outside of the card doesn't do anything. It feels a bit unexpected.

Screen Shot 2022-07-15 at 1 22 13 PM

It should be a pretty simple change. The code below should push that scroll bar to the outside without impacting anything else.
```dart
      child = Center(
        child: SingleChildScrollView(
          child: Row(
            mainAxisAlignment: MainAxisAlignment.center,
            children: [
              ConstrainedBox(
                constraints: BoxConstraints(maxWidth: containerWidth),
                child: Padding(
                  padding: EdgeInsets.symmetric(vertical: containerPadding),
                  child: Card(child: SafeArea(child: child)),
                ),
              ),
            ],
          ),
        ),
      );
```

Result:
Screen Shot 2022-07-15 at 1 38 01 PM

dnys1 (approved):
Looks good - thanks!!

PR #1850, opened 2022-07-05
[POC] feat(authenticator): add theming API

Jordan-Nelson:
**This is a proof of concept for an improved theming API for the Authenticator**

*Issue #, if available:*

*Description of changes:*
- Add AuthenticatorTheme API


dnys1 (commented):
Nice! I definitely think users would prefer something like this to the ever-expansive Flutter Theme API. Do you have a sense of how Material 3 will affect our API? Should it be applied to the user's authenticated content as well? Otherwise, they would have to apply it separately, I think, right?

Jordan-Nelson:
Closing. Will re-open a new PR if we decide to move forward with this.

PR #1851, opened 2022-07-05
feat!(core,auth): auth providers definition and CognitoIamAuthProvider registers in Auth

ragingsquirrel3:
This PR has the initial implementation in core of auth providers, related changes to plugin configuration signature and auth plugin registers the first auth provider. In another PR, I plan to integrate this with API plugin to support IAM auth mode but this PR is intentionally downscoped to just upstream logic to register the auth provider.

In summary, the auth providers are a way for other plugins to call the auth plugin indirectly. By registering an auth provider from auth plugin, other plugins essentially get a callback they can use to call auth methods as long as auth is registered at runtime and other plugins don't need an auth dependency or to call `Amplify.Auth...` directly. This avoids compile-time errors.

BREAKING CHANGE: configure method for plugins now requires an instance of auth provider repo. This change is only breaking for plugin authors, not plugin users.


packages/auth/amplify_auth_cognito_dart/lib/src/util/authorization_providers.dart
```diff
@@ -0,0 +1,104 @@
...
+      throw const AuthException(
+        'Unable to authorize request with IAM. No region or service provided.',
+      );
+    }
+    final authSession = await Amplify.Auth.fetchAuthSession(
+            options: const CognitoSessionOptions(getAWSCredentials: true))
+        as CognitoAuthSession;
+    final credentials = authSession.credentials;
+    if (credentials == null) {
+      throw const ApiException(
```
ragingsquirrel3, line 40: shouldn't be API error, but general callout to rethink errors from auth providers, whether needs special exception types, etc...

packages/auth/amplify_auth_cognito_dart/lib/src/util/authorization_providers.dart
```diff
@@ -0,0 +1,109 @@
+// Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
ragingsquirrel3: I think that would be a good rename. I was originally thinking this could have the cog user pool provider, too, but separating them prob better. I'm reconsidering some aspects of the design here that might change this, however.

packages/auth/amplify_auth_cognito_dart/lib/src/util/authorization_providers.dart
dnys1: I would say, in general, the exception type should not matter to the caller, meaning all exceptions should be handled equally by the caller. The specific type thrown should be native to the plugin in which it's defined, though, so an `AuthException` or `InvalidCredentialsException` from sigv4 package.

packages/amplify_core/lib/src/types/common/amplify_auth_provider.dart
```diff
@@ -0,0 +1,66 @@
...
+}
+
+abstract class AmplifyAuthProvider {
+  Future<http.BaseRequest> authorizeRequest(
+    http.BaseRequest request, {
+    covariant AuthProviderOptions? options,
+  });
+}
+
+abstract class AWSCredentialsAmplifyAuthProvider extends AmplifyAuthProvider {
```
dnys1, line 37: I would remove this and just use `AWSCredentialsProvider` as discussed

packages/amplify_core/lib/src/types/common/amplify_auth_provider.dart
```diff
@@ -0,0 +1,66 @@
...
+    final token = await getLatestAuthToken();
+    request.headers.putIfAbsent(AWSHeaders.authorization, () => token);
+    return request;
+  }
+}
+
+class AmplifyAuthProviderRepository {
+  final Map<String, AmplifyAuthProvider> _authProviders = {};
+
+  AmplifyAuthProvider? getAuthProvider(String authorizationType) {
```
dnys1, line 58: Let's make this typed. You'd have to cast it for it to be useful in most cases so let's just do that for them.

packages/auth/amplify_auth_cognito_dart/lib/src/util/aws_iam_auth_provider.dart
```diff
@@ -0,0 +1,118 @@
...
+import 'package:amplify_auth_cognito_dart/amplify_auth_cognito_dart.dart';
+import 'package:amplify_core/amplify_core.dart';
+import 'package:aws_signature_v4/aws_signature_v4.dart';
+import 'package:http/http.dart' as http;
+import 'package:meta/meta.dart';
+
+/// [AmplifyAuthProvider] implementation that signs a request using AWS credentials
+/// from `Amplify.Auth.fetchAuthSession()` or allows getting credentials directly.
+@internal
+class AWSIAMAuthProvider extends AWSCredentialsAmplifyAuthProvider {
```
dnys1, line 29: Dart convention is to capitalize acronyms of 2 letters, but pascal case everything else: https://dart.dev/guides/language/effective-dart/style#do-capitalize-acronyms-and-abbreviations-longer-than-two-letters-like-words

My convention is to follow this for everything but `AWS`, which I think looks better capitalized, but only when it's the first word in a name. So this should be `AWSIamAuthProvider`

packages/amplify_core/lib/src/types/common/amplify_auth_provider.dart
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+import 'package:amplify_core/amplify_core.dart';
+import 'package:http/http.dart' as http;
+
+abstract class AuthProviderOptions {
+  const AuthProviderOptions();
+}
+
+class IAMAuthProviderOptions extends AuthProviderOptions {
```
dnys1, line 23: Same comment around naming, this should be `IamAuthProviderOptions`

packages/amplify_core/lib/src/types/common/amplify_auth_provider.dart
```diff
@@ -0,0 +1,66 @@
...
+  String region;
+  AWSService service;
+
+  IAMAuthProviderOptions({required this.region, required this.service});
```
dnys1, line 27:
```suggestion
  final String region;
  final AWSService service;

  const IamAuthProviderOptions({required this.region, required this.service});
```

packages/amplify_core/lib/src/types/common/amplify_auth_provider.dart
```diff
@@ -0,0 +1,66 @@
...
+}
+
+abstract class AmplifyAuthProvider {
+  Future<http.BaseRequest> authorizeRequest(
+    http.BaseRequest request, {
```
dnys1, line 32: I would use `AWSBaseHttpRequest`. You're gonna run into issues with paths using `http.BaseRequest` since you can't specify a request which doesn't use `Uri`.

packages/amplify/amplify_flutter/lib/src/hybrid_impl.dart
```diff
@@ -24,6 +24,8 @@ import 'package:amplify_flutter/src/amplify_impl.dart';
 class AmplifyHybridImpl extends AmplifyClassImpl {
   /// {@macro amplify_flutter.amplify_hybrid_impl}
   AmplifyHybridImpl() : super.protected();
+  final AmplifyAuthProviderRepository _authProviderRepo =
```
dnys1, line 27: Honestly, may just want to make this `@protected` instead of private and passing it into `configure`. If we ever make it public, which I think we should, then all the `configure` methods need to be updated again which is potentially a breaking change.

packages/amplify/amplify_flutter/lib/src/hybrid_impl.dart
```diff
@@ -16,6 +16,7 @@ import 'dart:convert';
 
 import 'package:amplify_flutter/amplify_flutter.dart';
 import 'package:amplify_flutter/src/amplify_impl.dart';
+import 'package:flutter/material.dart';
```
HuiSF, line 19: Why do we need the material library here? It provides only the UI widgets correct?

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
```diff
@@ -165,10 +166,21 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface implements Closeable {
...
+  Future<void> configure({
+    AmplifyConfig? config,
+    required AmplifyAuthProviderRepository authProviderRepo,
+  }) async {
     if (config == null) {
       throw const AuthException('No Cognito plugin config detected');
     }
+
+    // Register auth providers to provide auth functionality to other plugins
+    // without calling `Amplify.Auth...` directly.
```
haverchuck, line 178:
```suggestion
    // without requiring other plugins to call `Amplify.Auth...` directly.
```

packages/amplify_core/lib/src/types/common/amplify_auth_provider.dart
```diff
@@ -0,0 +1,79 @@
...
+  final String region;
+  final AWSService service;
+
+  const IamAuthProviderOptions({required this.region, required this.service});
+}
+
+abstract class AmplifyAuthProvider {
+  Future<AWSBaseHttpRequest> authorizeRequest(
+    AWSBaseHttpRequest request, {
+    covariant AuthProviderOptions? options,
```
HuiSF, line 40: Good to know about `covariant`, it should help us to avoid type casting required on consumer side like this?

```dart
    final result = await Amplify.Auth.fetchAuthSession(
      options: CognitoSessionOptions(getAWSCredentials: true),
    );
    String identityId = (result as CognitoAuthSession).identityId!;
```

HuiSF (approved):

packages/amplify/amplify_flutter/lib/src/hybrid_impl.dart
ragingsquirrel3: good catch, removed. I think it was left from a previous implementation that used `describeEnum`.

packages/amplify_core/lib/src/types/common/amplify_auth_provider.dart
ragingsquirrel3: Maybe I'm missing this, but don't think it helps here and casting would still be needed as I think it can only be used for parameters, not return types.

Separately, I think we will be able to define an auth provider that can provide identity IDs so casting would not be needed outside auth plugin for this use case.

PR #1858, opened 2022-07-06
feat!(api): GraphQL API key auth mode

Equartey:
*Description of changes:*
Adds API key auth mode to GraphQL Requests, error handling, and relevant unit tests.

Additionally, response errors are now nullable (breaking change).


packages/api/amplify_api/test/graphql_error_test.dart
```diff
@@ -68,6 +68,6 @@ void main() {
         .response;
```
ragingsquirrel3, line 68: request: now that we are changing the expected logic for errors, could you please test that a happy case response has null errors? You could just add an assertion on one of the query tests or add another test.

packages/api/amplify_api/test/dart_graphql_test.dart
```diff
@@ -0,0 +1,147 @@
...
+
+      final operation = Amplify.API.mutate(request: req);
+      final res = await operation.value;
+
+      final expected = json.encode(_expectedMutateSuccessResponseBody['data']);
+
+      expect(res.data, equals(expected));
+    });
+
+    test('Errors are decoded in response.errors', () async {
```
ragingsquirrel3, line 126: request: please add a test that errors are null when the graphql response JSON from server has no errors.

packages/api/amplify_api/test/dart_graphql_test.dart
```diff
@@ -0,0 +1,147 @@
...
+void main() {
+  group('GraphQL Dart API', () {
+    late AmplifyAPI dartApi;
+
+    setUpAll(() async {
+      await Amplify.addPlugin(MockAmplifyAPI());
+      await Amplify.configure(amplifyconfig);
+    });
+
+    test('Query returns proper response.data', () async {
```
ragingsquirrel3, line 86: If you look at the tests for method channel implementation in amplify_api_query_test.dart I think there is more coverage that would be good to have parity with here. Some of that stuff might not make sense to test in dart implementation, like serialization/deserialization logic. However, I think we should at least test queries that use model helpers to ensure non-string requests are converted properly.

packages/api/amplify_api/test/dart_graphql_test.dart
```diff
@@ -0,0 +1,147 @@
...
+        } ''';
+      final req = GraphQLRequest(document: graphQLDocument, variables: {});
+
+      final operation = Amplify.API.query(request: req);
+      final res = await operation.value;
+
+      final expected = json.encode(_expectedQuerySuccessResponseBody['data']);
+
+      expect(res.data, equals(expected));
+    });
```
ragingsquirrel3, line 104: Would it be possible to add a test in query and mutation for operation cancelation? You can look in the REST dart tests for an example. Arguably not necessary bc it's essentially testing that this is a cancelable operation but IMO good to have a check on that API surface.

packages/api/amplify_api/test/dart_graphql_test.dart
```diff
@@ -0,0 +1,147 @@
...
+
+    test('Mutate returns proper response.data', () async {
```
ragingsquirrel3, line 106: same comment for queries applies to mutations. Could we check coverage in method channel implementation and ensure we have parity? Specifically with model helper mutations.

packages/api/amplify_api/lib/src/util.dart
```diff
@@ -30,3 +30,21 @@ Map<String, String>? addContentTypeToHeaders(
...
+
+/// Grabs errors from GraphQL Response. Is used in method channel and Dart first code.
+/// TODO: Move to Dart first code when method channel GraphQL implementation is removed.
+@internal
+List<GraphQLResponseError> deserializeGraphQLResponseErrors(
+  Map<String, dynamic> response,
+) {
+  final errors = response['errors'] as List?;
+  if (errors == null || errors.isEmpty) {
+    return const [];
```
ragingsquirrel3, line 42: should this return null and then the return type of this method would be `List<GraphQLResponseError>?`

packages/api/amplify_api/test/dart_graphql_test.dart
```diff
@@ -0,0 +1,147 @@
...
+      'locations': _errorLocations,
+      'path': _errorPath,
+      'extensions': _errorExtensions,
+    },
+  ]
+};
+
+class MockAmplifyAPI extends AmplifyAPIDart {
+  @override
+  http.Client getGraphQLClient({String? apiName}) =>
```
ragingsquirrel3, line 61: nice job with this mocking logic. I think this will change as more cases get covered, or get copied/extended but this is a nice start.

packages/api/amplify_api/test/dart_graphql_test.dart
```diff
@@ -0,0 +1,147 @@
...
+      MockClient((request) async {
+        if (request.body.contains('TestMutate')) {
+          return http.Response(
+              json.encode(_expectedMutateSuccessResponseBody), 400);
```
ragingsquirrel3, line 65: nit/question: Does AppSync actually give a 400 status when there are errors like this? If not, would be good to match their server in the mock in case this had an impact on deserializing the server response (though I would not expect it to here).

packages/api/amplify_api/test/dart_graphql_test.dart
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        return http.Response(
+            json.encode(_expectedQuerySuccessResponseBody), 200);
+      });
+}
+
+void main() {
+  group('GraphQL Dart API', () {
+    late AmplifyAPI dartApi;
```
ragingsquirrel3, line 79: nit/question: Is this needed? I think it may have been left in the dart REST test by mistake, in which case you could remove there and here.

packages/api/amplify_api/lib/src/amplify_api_config.dart
```diff
@@ -29,7 +29,8 @@ class EndpointConfig with AWSEquatable<EndpointConfig> {
 
   /// Gets the host with environment path prefix from Amplify config and combines
   /// with [path] and [queryParameters] to return a full [Uri].
-  Uri getUri(String path, Map<String, dynamic>? queryParameters) {
+  Uri getUri(String? path, Map<String, dynamic>? queryParameters) {
```
ragingsquirrel3, line 32: request: Would it be possible to add unit test coverage for this class/utility here? I know it was not created in this PR but would be good to test the logic here regarding how it gets URIs from config, especially when inputs are null. Arguably this code could be simplified but would rather see a test, first.

packages/api/amplify_api/lib/src/api_plugin_impl.dart
```diff
@@ -130,6 +152,30 @@ class AmplifyAPIDart extends AmplifyAPI {
     _authProviders[authProvider.type] = authProvider;
   }
 
+  // ====== GraphQL ======
+
+  @override
+  CancelableOperation<GraphQLResponse<T>> query<T>(
```
ragingsquirrel3, line 158: I have an offline comment/question about this.

ragingsquirrel3:
Would it be possible to change the beginning of the PR title to be "feat!(api): ...." to indicate that this has a breaking change? When you eventually merge, could you please write: "BREAKING CHANGE: graphql response errors now nullable" or something similar at the bottom of the commit message that gets used as the merge commit by gh during squash merge.

ragingsquirrel3 (changes requested):

packages/api/amplify_api/lib/src/graphql/graphql_send_request.dart
```diff
@@ -0,0 +1,46 @@
...
+import 'graphql_response_decoder.dart';
+
+@internal
+Future<GraphQLResponse<T>> GraphQLSendRequest<T>(
+    {required GraphQLRequest<T> request,
+    required http.Client client,
+    required Uri uri}) async {
+  final body = {'variables': request.variables, 'query': request.document};
+  final graphQLResponse = await client.post(uri, body: json.encode(body));
+
```
ragingsquirrel3, line 32: I think there is room for improvement here regarding error handling/robustness. What if response is not a 200? What if JSON does not have expected structure? Would be good to have a look at other platforms and get an idea how this should work. Also suggest adding tests for that logic.

packages/api/amplify_api/lib/src/graphql/graphql_send_request.dart
```diff
@@ -0,0 +1,46 @@
...
+    List<GraphQLResponseError> errors =
+        deserializeGraphQLResponseErrors(responseBody);
+
+    return GraphQLResponseDecoder.instance
+        .decode<T>(request: request, data: responseData, errors: errors);
+  }
+
+  // TEMP
+  throw Exception('Bad types');
```
ragingsquirrel3, line 45: Going along with above comment about error handling, don't think we will throw this exception once that's refactored.

packages/api/amplify_api/lib/src/graphql/graphql_send_request.dart
```diff
@@ -0,0 +1,46 @@
...
+import 'package:meta/meta.dart';
+
+import '../util.dart';
+import 'graphql_response_decoder.dart';
+
+@internal
+Future<GraphQLResponse<T>> GraphQLSendRequest<T>(
+    {required GraphQLRequest<T> request,
+    required http.Client client,
+    required Uri uri}) async {
```
ragingsquirrel3, line 29: nit: I think you can add a trailing comma here by putting a comma after `Uri uri` and then a newline after the comma. Eventually we need to remove all the info warnings from API category so it's easier to automatically enforce this.

packages/api/amplify_api/lib/src/graphql/graphql_send_request.dart
```diff
@@ -0,0 +1,46 @@
...
+
+import 'package:amplify_core/amplify_core.dart';
+import 'package:http/http.dart' as http;
+import 'package:meta/meta.dart';
+
+import '../util.dart';
+import 'graphql_response_decoder.dart';
+
+@internal
+Future<GraphQLResponse<T>> GraphQLSendRequest<T>(
```
ragingsquirrel3, line 26: nit: Should `graphQLSendRequest` not be capitalized here bc a function not a class?

packages/api/amplify_api/example/integration_test/graphql_tests.dart
```diff
@@ -44,7 +44,7 @@ void main() {
       final req = GraphQLRequest<String>(
           document: graphQLDocument, variables: <String, String>{'id': id});
       final response = await Amplify.API.mutate(request: req).response;
-      if (response.errors.isNotEmpty) {
+      if (response.errors != null) {
```
dnys1, line 47: The [spec](https://spec.graphql.org/June2018/#sec-Errors) does say that this should either be `null` or non-empty, but we should still probably check if it's empty since we don't want to report an empty array as having error'd and AppSync doesn't always conform perfectly to the spec if I remember correctly.

```suggestion
      if (response.errors != null && response.errors!.isNotEmpty) {
```

packages/api/amplify_api/lib/src/graphql/graphql_send_request.dart
```diff
@@ -0,0 +1,46 @@
...
+    required http.Client client,
+    required Uri uri}) async {
+  final body = {'variables': request.variables, 'query': request.document};
+  final graphQLResponse = await client.post(uri, body: json.encode(body));
+
+  final responseBody = json.decode(graphQLResponse.body);
+  if (responseBody is Map<String, dynamic>) {
+    final responseData = json.encode(responseBody['data']);
+
+    List<GraphQLResponseError> errors =
```
dnys1, line 37: Omitting types from local vars helps keep code consistent and concise and `final` is good documentation that the variable is not expected to change (vs. `var`)

```suggestion
    final errors =
```

packages/api/amplify_api/lib/src/graphql/graphql_send_request.dart
dnys1: The [spec](https://spec.graphql.org/June2018/#sec-Response-Format) implies that a 200 response guarantees a certain shape to the data. We can still verify that we get a Map back, but we must check if the status code is 200 first. Any other code means something went wrong.

packages/api/amplify_api/test/dart_graphql_test.dart
Equartey: Doesn't look like there are tests for mutations in `amplify_api_query_test.dart` currently.

packages/api/amplify_api/test/dart_graphql_test.dart
Equartey: Depending on the error, AppSync will return a 400, ie a bad GraphQL document will return a 400 with messaging. There is also no effect currently of the status code during decoding.

packages/api/amplify_api/example/integration_test/graphql_tests.dart
```diff
@@ -561,7 +561,7 @@ void main() {
         // With stream established, exec callback with stream events.
         final subscription = await _getEstablishedSubscriptionOperation<T>(
             subscriptionRequest, (event) {
-          if (event.errors.isNotEmpty) {
+          if (event.errors != null) {
```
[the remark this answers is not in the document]
Equartey: Good idea, added.

packages/api/amplify_api/lib/src/graphql/graphql_send_request.dart
```diff
@@ -0,0 +1,51 @@
...
+
+  final responseBody = json.decode(graphQLResponse.body);
+
+  if (responseBody is! Map<String, dynamic>) {
+    throw Exception('Response from GraphQL client incorrect type.');
+  }
+
+  // Handle Model responses that do not return a 'data' key
+  final responseData =
+      request.decodePath != null ? responseBody : responseBody['data'];
```
ragingsquirrel3, line 42: Is this ternary necessary here? I ran the integ tests and saw lots of errors with the model helpers. I was thinking you would always want `responseBody['data']` because model helpers won't change the server response structure, just changes how the client decodes the response and doesn't require a change here. I think it was working before, just was suggesting adding some tests for it.

packages/api/amplify_api/lib/src/graphql/graphql_send_request.dart
```diff
@@ -0,0 +1,51 @@
...
+
+import 'package:amplify_core/amplify_core.dart';
+import 'package:http/http.dart' as http;
+import 'package:meta/meta.dart';
+
+import '../util.dart';
+import 'graphql_response_decoder.dart';
+
+@internal
+Future<GraphQLResponse<T>> graphQLSendRequest<T>({
```
ragingsquirrel3, line 26: nit: could you please add a little doc comment for this function saying what it does. I think this file will have no lint warnings after that. maybe something like:

```
/// Converts the [GraphQLRequest] to an HTTP POST request and sends with ///[client]. 
```

packages/api/amplify_api/example/integration_test/graphql_tests.dart
```diff
@@ -561,7 +561,7 @@ void main() {
         // With stream established, exec callback with stream events.
         final subscription = await _getEstablishedSubscriptionOperation<T>(
             subscriptionRequest, (event) {
-          if (event.errors.isNotEmpty) {
+          if (event.hasErrors) {
             fail('subscription errors: ${event.errors}');
           }
           dataCompleter.complete(event);
```
ragingsquirrel3, line 567: request: Could we comment out or have an early return in the group below about subscriptions along with a TODO comment to re-enable them once subscriptions implemented. I would prefer to have failing integ test suites and I think all these should pass here (except subscriptions), which is awesome!

packages/api/amplify_api/lib/src/graphql/graphql_send_request.dart
```diff
@@ -0,0 +1,51 @@
...
+  required http.Client client,
+  required Uri uri,
+}) async {
+  final body = {'variables': request.variables, 'query': request.document};
+  final graphQLResponse = await client.post(uri, body: json.encode(body));
+
+  final responseBody = json.decode(graphQLResponse.body);
+
+  if (responseBody is! Map<String, dynamic>) {
+    throw Exception('Response from GraphQL client incorrect type.');
```
ragingsquirrel3, line 37: nit: Could this be `ApiException` and change the message to something like "unable to parse GraphQLResponse from server response which was not a JSON object." Just thinking it might be more helpful to folks than something about type errors.

ragingsquirrel3: Also, for the underlying error argument to `ApiException`, would it make sense to include the status code and the response body? If someone had an error here, they might have no way of seeing the server error unless we surface along with this error.

packages/api/amplify_api/test/dart_graphql_test.dart
ragingsquirrel3: Thanks for checking. I also saw that non-200 responses still included valid gql responses with the errors formatted correctly.

packages/api/amplify_api/lib/src/graphql/graphql_send_request.dart
ragingsquirrel3: It might also make sense to wrap the `json.decode(` in a try/catch as I would expect that to throw an error if the body wasn't JSON-able.

packages/api/amplify_api/lib/src/amplify_authorization_rest_client.dart
```diff
@@ -50,8 +52,12 @@ class AmplifyAuthorizationRestClient extends http.BaseClient
...
     if (!request.headers.containsKey(AWSHeaders.authorization) &&
         endpointConfig.authorizationType != APIAuthorizationType.none) {
-      // ignore: todo
-      // TODO: Use auth providers from core to transform the request.
+      // TODO(ragingsquirrel3): Use auth providers from core to transform the request.
+      final apiKey = endpointConfig.apiKey;
+      if (endpointConfig.authorizationType == APIAuthorizationType.apiKey &&
+          apiKey != null) {
+        request.headers.putIfAbsent(_xApiKey, () => apiKey);
+      }
```
dnys1, line 60: I would rearrange this a bit. If the authorization type is API key but there is no API key we should alert the user (throw an error)

packages/api/amplify_api/lib/src/graphql/graphql_send_request.dart
```diff
@@ -0,0 +1,51 @@
...
+import 'graphql_response_decoder.dart';
+
+@internal
+Future<GraphQLResponse<T>> graphQLSendRequest<T>({
+  required GraphQLRequest<T> request,
+  required http.Client client,
+  required Uri uri,
+}) async {
+  final body = {'variables': request.variables, 'query': request.document};
+  final graphQLResponse = await client.post(uri, body: json.encode(body));
```
dnys1, line 32:
```suggestion
  final graphQLResponse = await client.post(uri, body: json.encode(body));
  
  if (graphQLResponse.statusCode != 200) {
    throw ApiException(...);
  }
```

packages/api/amplify_api/lib/src/graphql/graphql_send_request.dart
dnys1: +1. I would also rename it to `sendGraphQLRequest` per Dart [naming conventions](https://dart.dev/guides/language/effective-dart/design#prefer-an-imperative-verb-phrase-for-a-function-or-method-whose-main-purpose-is-a-side-effect).

packages/api/amplify_api/lib/src/graphql/graphql_send_request.dart
Equartey: I'm not sure we want to short circuit here on non 200 codes. App sync returns valuable error message information despite a non successful status code (ie 400) and the process to decode those errors are what follows this.

packages/api/amplify_api/test/amplify_api_config_test.dart
```diff
@@ -0,0 +1,72 @@
...
+
+void main() {
+  late EndpointConfig endpointConfig;
+
+  group('GraphQL Config', () {
+    const endpointType = EndpointType.graphQL;
+    const endpoint =
+        'https=//abc123.appsync-api.us-east-1.amazonaws.com/graphql';
+    const region = 'us-east-1';
+    const authorizationType = APIAuthorizationType.apiKey;
```
ragingsquirrel3, line 17: nit: if mocking api key type might as well add a fake api key here in case there was some code path that assumed those things connected (though that's not the case here).

packages/api/amplify_api/test/amplify_api_config_test.dart
```diff
@@ -0,0 +1,72 @@
...
+      const config = AWSApiConfig(
+          endpointType: endpointType,
+          endpoint: endpoint,
+          region: region,
+          authorizationType: authorizationType);
+
+      endpointConfig = const EndpointConfig('GraphQL', config);
+    });
+
+    test('should return valid URI wtih null params', () async {
```
ragingsquirrel3, line 29: nit, spelling

packages/api/amplify_api/test/amplify_api_config_test.dart
```diff
@@ -0,0 +1,72 @@
...
+      final uri = endpointConfig.getUri(path: null, queryParameters: null);
+      final expected = Uri.parse('$endpoint/');
+
+      expect(uri, equals(expected));
+    });
+
+    test('should returns invalid URI wtih params', () async {
```
ragingsquirrel3, line 36: here, too

packages/api/amplify_api/test/amplify_api_config_test.dart
```diff
@@ -0,0 +1,72 @@
...
+      expect(uri, isNot(expected));
+    });
+  });
+
+  group('REST Config', () {
+    const endpointType = EndpointType.rest;
+    const endpoint =
+        'https=//abc123.appsync-api.us-east-1.amazonaws.com/graphql';
+    const region = 'us-east-1';
+    const authorizationType = APIAuthorizationType.apiKey;
```
ragingsquirrel3, line 50: nit: please do IAM w REST for this mock as that's the only thing the CLI will do for you at the moment and api key not possible w REST.

packages/api/amplify_api/test/amplify_api_config_test.dart
```diff
@@ -0,0 +1,72 @@
...
+          endpointType: endpointType,
+          endpoint: endpoint,
+          region: region,
+          authorizationType: authorizationType);
+
+      endpointConfig = const EndpointConfig('REST', config);
+    });
+
+    test('should return valid URI wtih params', () async {
+      final path = 'path/to/nowhere';
```
ragingsquirrel3, line 63: request: please add a test where the path has a leading slash like '/path' bc that's something we account for

packages/api/amplify_api/lib/src/graphql/graphql_send_request.dart
Equartey: Resolved offline. Ternary is not needed.

packages/api/amplify_api/test/amplify_api_config_test.dart
```diff
@@ -0,0 +1,84 @@
...
+
+import 'test_data/fake_amplify_configuration.dart';
+
+void main() {
+  late EndpointConfig endpointConfig;
+
+  group('GraphQL Config', () {
+    const endpointType = EndpointType.graphQL;
+    const endpoint =
+        'https=//abc123.appsync-api.us-east-1.amazonaws.com/graphql';
```
ragingsquirrel3, line 15: nit: could you please make the first part "https://" so it's more like a normal URI instead of with "="?

packages/api/amplify_api/test/amplify_api_config_test.dart
```diff
@@ -0,0 +1,84 @@
...
+      final params = {'foo': 'bar', 'bar': 'baz'};
+      final uri = endpointConfig.getUri(path: path, queryParameters: params);
+
+      final expected = Uri.parse('$endpoint/$path?foo=bar&bar=baz');
+
+      expect(uri, equals(expected));
+    });
+
+    test('should handle a leading slash', () async {
+      final path = 'path/to/nowhere/';
```
ragingsquirrel3, line 75: request: for leading slash, could this be '/path/to/nowwhere' ?

packages/api/amplify_api/test/amplify_api_config_test.dart
```diff
@@ -0,0 +1,84 @@
...
+    });
+
+    test('should return valid URI with null params', () async {
+      final uri = endpointConfig.getUri(path: null, queryParameters: null);
+      final expected = Uri.parse('$endpoint/');
+
+      expect(uri, equals(expected));
+    });
+
+    test('should returns invalid URI with params', () async {
```
ragingsquirrel3, line 38: question: What is this testing and why is it invalid? Seems like this logic is tested below (although for REST) so not sure this test is necessary. If you do keep it, nit on grammar of description "should return ...."

packages/api/amplify_api/lib/src/graphql/send_graphql_request.dart
```diff
@@ -0,0 +1,57 @@
...
+    if (responseBody is! Map<String, dynamic>) {
+      throw ApiException(
+          'unable to parse GraphQLResponse from server response which was not a JSON object.',
+          underlyingException: graphQLResponse.body);
+    }
+
+    final responseData = responseBody['data'];
+    // Preserve `null`. json.encode(null) returns "null" not `null`
+    final responseDataJson =
+        responseData != null ? json.encode(responseData) : null;
```
ragingsquirrel3, line 47: callout (no change needed in this PR): Looks like we are decoding the response from string to map, then re-encoding part of it to a string to pass to decoder, which then converts it back to JSON again. Looks like we are doing this bc we are basing it off the decoder which was built around the method channel implementation that returns a string. We should look into changing the decoder to accept a map so that we only have to decode it once.

ragingsquirrel3 (approved):

packages/api/amplify_api/test/amplify_api_config_test.dart
Equartey: Think you're right, this test isn't needed. It's not invalid per se, and it's not a unique test case.

packages/api/amplify_api/test/dart_graphql_test.dart
```diff
@@ -0,0 +1,216 @@
+import 'dart:convert';
```
dnys1, line 1: Missing license header on this file. Hoping to create an `aft` command soon but in the meantime you can use `license.sh` in the repo to automatically add them in a package. So from the API package root, you could run `../../../tool/license.sh`

packages/api/amplify_api/lib/src/util.dart
```diff
@@ -30,3 +30,21 @@ Map<String, String>? addContentTypeToHeaders(
   modifiedHeaders.putIfAbsent(AWSHeaders.contentType, () => contentType);
   return modifiedHeaders;
 }
+
+/// Grabs errors from GraphQL Response. Is used in method channel and Dart first code.
+/// TODO: Move to Dart first code when method channel GraphQL implementation is removed.
```
dnys1, line 35: TODOs shouldn't be doc comments generally, and lets add usernames for all TODOs (this is a lint in the new packages, but this one hasn't been migrated yet)

```suggestion
// TODO(Equartey): Move to Dart first code when method channel GraphQL implementation is removed.
```

packages/api/amplify_api/lib/src/graphql/send_graphql_request.dart
```diff
@@ -0,0 +1,57 @@
...
+/// Converts the [GraphQLRequest] to an HTTP POST request and sends with ///[client].
+@internal
+Future<GraphQLResponse<T>> sendGraphQLRequest<T>({
+  required GraphQLRequest<T> request,
+  required http.Client client,
+  required Uri uri,
+}) async {
+  try {
+    final body = {'variables': request.variables, 'query': request.document};
+    final graphQLResponse = await client.post(uri, body: json.encode(body));
```
dnys1, line 34: Need to check for `200` status code here and throw otherwise

packages/api/amplify_api/test/dart_graphql_test.dart
```diff
@@ -0,0 +1,216 @@
...
+          'id': 'TEST_ID',
+          'name': 'Test App Blog',
+          'createdAt': '2022-06-28T17:36:52.460Z'
+        }
+      ]
+    }
+  }
+};
+
+final _modelQueryId = UUID.getUUID();
```
dnys1, line 32: Use `uuid` from `aws_common`. We're gonna deprecate this soon.

packages/api/amplify_api/test/amplify_api_config_test.dart
```diff
@@ -0,0 +1,75 @@
...
+          endpoint: endpoint,
+          region: region,
+          authorizationType: authorizationType,
+          apiKey: apiKey);
+
+      endpointConfig = const EndpointConfig('GraphQL', config);
+    });
+
+    test('should return valid URI with null params', () async {
+      final uri = endpointConfig.getUri(path: null, queryParameters: null);
```
dnys1, line 32:
```suggestion
      final uri = endpointConfig.getUri();
```

dnys1 (approved):
Some minor comments. LGTM, though!

PR #1871, opened 2022-07-12
chore: More ESD work

dnys1:
Migrates more types to the E/S/D stack to improve the debugging experience around core types


packages/amplify_core/lib/src/config/auth/cognito/appsync.g.dart
```diff
@@ -21,7 +21,7 @@ Map<String, dynamic> _$CognitoAppSyncConfigToJson(
   final val = <String, dynamic>{
     'ApiUrl': instance.apiUrl,
     'Region': instance.region,
-    'AuthMode': _$APIAuthorizationTypeEnumMap[instance.authMode],
+    'AuthMode': _$APIAuthorizationTypeEnumMap[instance.authMode]!,
```
haverchuck, line 24: Are we sure that authMode is always present? I'm wondering if the libraries have some sort of default in case it is missing in the config, in which case some customers may be successfully using a handwritten config.

packages/amplify_core/lib/src/config/auth/cognito/password_protection_settings.g.dart
```diff
@@ -33,7 +33,7 @@ Map<String, dynamic> _$PasswordProtectionSettingsToJson(
       const _PasswordPolicyMinLengthConverter()
           .toJson(instance.passwordPolicyMinLength));
   val['passwordPolicyCharacters'] = instance.passwordPolicyCharacters
-      .map((e) => _$PasswordPolicyCharactersEnumMap[e])
+      .map((e) => _$PasswordPolicyCharactersEnumMap[e]!)
```
haverchuck, line 36: Same question re: authMode.

haverchuck (commented):
Looks good, although I'm not sure about the postfix `!` operators in the `.g` files. It seems like some of these could break customers using older configs? This might be ok depending on the benefit, but its something we should avoid if possible.

dnys1:
@haverchuck the null checks are only on the map access - it's not saying anything about the presence of the fields

Jordan-Nelson (approved):

packages/amplify_core/lib/src/types/auth/attribute/fetch_user_attributes_options.dart
```diff
@@ -13,13 +13,16 @@
...
 /// {@endtemplate}
-class FetchUserAttributesOptions {
+abstract class FetchUserAttributesOptions
+    with AWSSerializable<Map<String, Object?>> {
   /// {@macro amplify_core.fetch_user_attributes_options}
   const FetchUserAttributesOptions();
 
-  /// Serialize the object to a map
-  Map<String, Object?> serializeAsMap() => {};
+  @Deprecated('Use toJson instead')
```
haverchuck, line 26: I'm happy to see these go.

haverchuck (approved):

PR #1879, opened 2022-07-13
fix(core, auth): analysis issue in state machine events

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Update `StateMachineEvent` to accept a `StateType` and remove the use of `covariant` and dynamic types

### DIFF
--- a/packages/amplify_core/test/state_machine/my_state_machine.dart
+++ b/packages/amplify_core/test/state_machine/my_state_machine.dart
@@ -43,7 +43,9 @@ class MyEvent extends StateMachineEvent<MyType, MyType> {
   final MyType type;
 
   @override
-  MyPreconditionException? checkPrecondition(MyState currentState) {
+  MyPreconditionException? checkPrecondition(
+    StateMachineState<MyType> currentState,
+  ) {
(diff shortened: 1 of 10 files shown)


packages/amplify_core/test/state_machine/my_state_machine.dart
```diff
@@ -43,7 +43,9 @@ class MyEvent extends StateMachineEvent<MyType> {
   final MyType type;
 
   @override
-  MyPreconditionException? checkPrecondition(MyState currentState) {
+  MyPreconditionException? checkPrecondition(
+    StateMachineState<MyType> currentState,
```
dnys1, line 47: If `covariant` is needed for this then I was mistaken and we should add it back.

```suggestion
    MyState currentState,
```

dnys1 (approved):

packages/amplify_core/test/state_machine/my_state_machine.dart
Jordan-Nelson: It was. I reverted the commit that removed covariant.

haverchuck (approved):

PR #1888, opened 2022-07-14
chore: Use logger

dnys1:
Use the logger throughout core and auth


Jordan-Nelson (approved):

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_in_state_machine.dart
```diff
@@ -149,6 +156,7 @@ class SignInStateMachine extends StateMachine<SignInEvent, SignInState> {
           return worker;
         }
         worker = ConfirmDeviceWorker.create();
+        // TODO(dnys1): Log
```
fjnoyp, line 159: Should these be switched with logs or removed?  Would we just need to update `print.dart -> safePrint` here?

fjnoyp (commented):
Looks great, thanks for making those changes to logger.   I like your use of logging_ext.dart for better code separation.  There were some TODO comments I was wondering were intentional or not.

fjnoyp (approved):

fjnoyp (approved):

PR #1892, opened 2022-07-19
fix: DDC build issue

dnys1:
**ISSUE**: https://github.com/dart-lang/sdk/issues/49484

This fixes a pretty weird issue in DDC, similar to this one: https://github.com/dart-lang/sdk/issues/47645. I say similar because, despite my best efforts, I haven't been able to reproduce it outside our repo. The issue has to do with the ordering of statements in compiled DDC modules - for example, a constructor calling another constructor which hasn't been defined yet. In all my reproductions, the ordering was correct, so I'm not sure what is special with our set up.

As noted by a Dart team member [here](https://github.com/dart-lang/sdk/issues/47645#issuecomment-967718351), DDC modules are created with all interdependent types in a package. So, one way to reduce DDC compile errors due to module ordering is to reduce the number of interdependencies i.e. cyclic imports.

This "fix", then, is simply to remove bulk imports in the Auth types which were causing the issue, reducing the average size of the compiled DDC modules. Luckily, this fixes the issue for now.

### DIFF
--- a/packages/amplify_core/lib/src/types/auth/attribute/auth_next_update_attribute_step.dart
+++ b/packages/amplify_core/lib/src/types/auth/attribute/auth_next_update_attribute_step.dart
@@ -13,7 +13,9 @@
  * permissions and limitations under the License.
  */
 
-import 'package:amplify_core/amplify_core.dart';
+import 'package:amplify_core/src/types/auth/auth_code_delivery_details.dart';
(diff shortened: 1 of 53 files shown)


dnys1:
This was an incomplete resolution to the issue. I will open a new PR with a better workaround.

PR #1899, opened 2022-07-20
chore(auth): Better Lambda error handling

dnys1:
Adds better error handling around Lambda errors thrown from non-hosted UI methods. Un-deprecates the `LambdaException` class since it actually provides a lot of utility and makes it easy for users to catch when they're only concerned with whether an error originated in a lambda.

### DIFF
--- a/packages/auth/amplify_auth_cognito_dart/lib/src/sdk/sdk_exception.dart
+++ b/packages/auth/amplify_auth_cognito_dart/lib/src/sdk/sdk_exception.dart
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// ignore_for_file: deprecated_member_use_from_same_package
-
 /// Exception types bridged from generated SDKs to their legacy counterparts.
 library amplify_auth_cognito_dart.sdk.sdk_exception;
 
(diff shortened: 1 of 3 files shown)


ragingsquirrel3 (approved):

packages/auth/amplify_auth_cognito_dart/lib/src/sdk/sdk_exception.dart
```diff
@@ -83,49 +81,41 @@ export 'cognito_identity_provider.dart'
 /// service SDKs.
 @internal
 Exception transformSdkException(Exception e) {
+  if (e is! SmithyException) {
+    return e;
+  }
+  final message = e.message ?? e.toString();
+  // Some exceptions are returned as non-Lambda exceptions even though they
+  // orginated in user-defined lambdas.
+  if (LambdaException.isLambdaException(message) ||
```
haverchuck, line 90: I realize this was already part of the code (and so maybe we can fix later), but I don't know if we want to collapse these flavors of Lambda Exception into one type, since according to the [Cognito docs](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_InitiateAuth.html#API_InitiateAuth_Errors) they have different connotations (even if they aren't super clear).

packages/auth/amplify_auth_cognito_dart/lib/src/sdk/sdk_exception.dart
```diff
@@ -154,24 +144,46 @@ class InvalidParameterException extends core.AmplifyException
...
+  }
+
+  const LambdaException._(
     super.message, {
+    this.lambdaName,
     super.underlyingException,
   });
 
+  /// Whether [exception] originated in a user Lambda.
+  static bool isLambdaException(String exception) =>
```
haverchuck, line 173: I think you're correct that this is really the only way to infer that these were originating in Lambda, but it seems like kind of a fragile approach in that a change to a string sent back from Cognito in a non-Lambda exception using this patter would result in an incorrect inference. Would it be better to call these UnknownExceptions and be correct-but-vague, rather than call them LambdaExceptions and be more-precise-but-possibly-wrong?

haverchuck: Alrighty  - I can definitely see that avoiding the UserNotFoundException is kind of imperative.

haverchuck (approved):

packages/auth/amplify_auth_cognito_dart/lib/src/sdk/sdk_exception.dart
haverchuck: I like the idea of letting them drill down with an `is` check.

PR #1906, opened 2022-07-20
chore: 0.6.4-rc.1 release

HuiSF:
*Issue #, if available:*

*Description of changes:*

Bump version to `0.6.4-rc.1`.

### DIFF
--- a/packages/amplify_authenticator/pubspec.yaml
+++ b/packages/amplify_authenticator/pubspec.yaml
@@ -10,9 +10,9 @@ environment:
   flutter: ">=2.10.0"
 
 dependencies:
-  amplify_auth_cognito: ^0.5.0
-  amplify_core: ^0.5.0
-  amplify_flutter: ^0.5.0
+  amplify_auth_cognito: ^0.6.4-rc.1
+  amplify_core: ^0.6.4-rc.1
+  amplify_flutter: ^0.6.4-rc.1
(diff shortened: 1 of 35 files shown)


dnys1 (dismissed):

dnys1 (approved):

packages/amplify_authenticator/pubspec.yaml
```diff
@@ -10,9 +10,9 @@ environment:
   flutter: ">=2.10.0"
 
 dependencies:
-  amplify_auth_cognito: ^0.5.0
-  amplify_core: ^0.5.0
-  amplify_flutter: ^0.5.0
+  amplify_auth_cognito: ^0.6.4-rc.1
```
ragingsquirrel3, line 13: nit: because authenticator will not be released here, is this needed? Doesn't seem to do any harm.

ragingsquirrel3 (approved):

packages/amplify_authenticator/pubspec.yaml
dnys1: Good catch. Yeah, we shouldn't need to change the Authenticator constraints at all

ragingsquirrel3: I guess makes sense to have this so everything matches.

PR #1915, opened 2022-07-21
feat(api): .subscribe() for GraphQL

ragingsquirrel3:
This PR has subscriptions for GraphQL in dart. It implements the appsync real-time client described on https://docs.aws.amazon.com/appsync/latest/devguide/aws-appsync-real-time-data.html. 

It does not implement reconnect on error or backoff/retry.

It does not include increased integration tests for other auth modes.


packages/api/amplify_api/lib/src/decorators/authorize_websocket_message.dart
```diff
@@ -0,0 +1,68 @@
...
+  final body = inputMessage.payload?.toJson()['data'];
+  if (inputMessage is WebSocketConnectionInitMessage) {
+    inputMessage.authorizationHeaders =
+        await _generateAuthorizationHeaders(config, authRepo: authRepo);
+  } else if (body is String) {
+    inputMessage.payload?.authorizationHeaders =
+        await _generateAuthorizationHeaders(config,
+            authRepo: authRepo, body: body);
+  }
+  return Future.value(inputMessage);
```
dnys1, line 39:
```suggestion
  return inputMessage;
```

packages/api/amplify_api/lib/src/graphql/ws/websocket_connection.dart
```diff
@@ -0,0 +1,186 @@
...
+
+import 'package:amplify_api/src/decorators/authorize_websocket_message.dart';
+import 'package:amplify_core/amplify_core.dart';
+import 'package:async/async.dart';
+import 'package:web_socket_channel/web_socket_channel.dart';
+
+import 'websocket_message.dart';
+import 'websocket_message_stream_transformer.dart';
+
+/// {@template websocket_connection}
```
dnys1, line 12: Template names should be globally unique

```suggestion
/// {@template amplify_api.ws.websocket_connection}
```

packages/api/amplify_api/lib/src/graphql/ws/websocket_connection.dart
```diff
@@ -0,0 +1,186 @@
...
+  /// Closes the WebSocket connection.
+  @override
+  void close() {
+    _subscription.cancel();
+    _channel.sink.close();
+  }
+
+  /// Initializes the connection.
+  Future<void> init() {
+    return _initFuture ??= _init();
```
dnys1, line 72: Use `AsyncMemoizer` from `package:async`

packages/api/amplify_api/lib/src/graphql/ws/websocket_connection.dart
```diff
@@ -0,0 +1,186 @@
...
+  static const webSocketProtocols = ['graphql-ws'];
+
+  final AmplifyAuthProviderRepository authProviderRepo;
+
+  final AWSApiConfig _config;
+  late final WebSocketChannel _channel;
+  late final StreamSubscription<WebSocketMessage> _subscription;
+  late final RestartableTimer _timeoutTimer;
+
+  // Add connection error variable to throw in `init`.
```
dnys1, line 25: Let's handle this - should've been a TODO

packages/api/amplify_api/lib/src/graphql/ws/websocket_connection.dart
```diff
@@ -0,0 +1,186 @@
...
+        );
+  }
+
+  /// Cancels a subscription.
+  void _cancel(String subscriptionId) {
+    _send(WebSocketMessage(
+      id: subscriptionId,
+      messageType: MessageType.stop,
+    ));
+    // TODO(equartey): if this is the only susbscription, close the connection.
```
dnys1, line 113: Is this how other clients do it? I guess it could make sense to save on pings but idk

packages/api/amplify_api/lib/src/decorators/authorize_websocket_message.dart
dnys1, line 39: This function took me so long to grok, I would suggest reworking it so that the significance of each branch is more clear. Grabbing body on the first line and then only using it in the `else` branch makes no sense to me. It's also not clear what the significance of `body is String` is or why it's okay to return in some cases without adding authorization headers or why for one branch the authorization headers are on the message itself but then the other branch they're in the payload.

Creating an enum for the message types would be better than `is` checks, IMO, and would make this a lot more readable. Also linking to the docs would help give context to why authorization headers land in two different spots - or alternatively, put the authorization headers on the message itself for all types. 

I think finding a way to not modify input parameters directly would be good too, since 1) there's no reason to return the message if you're just modifying it in the function and 2) it would help avoid the confusion of where to put the authorization headers if this is a pure function.

packages/api/amplify_api/lib/src/decorators/authorize_websocket_message.dart
```diff
@@ -0,0 +1,68 @@
...
+}
+
+Future<Map<String, dynamic>> _generateAuthorizationHeaders(
+  AWSApiConfig config, {
+  required AmplifyAuthProviderRepository authRepo,
+  String body = '{}',
+}) async {
+  final endpointHost = Uri.parse(config.endpoint).host;
+  // Create canonical HTTP request to authorize.
+  final maybeConnect = body != '{}' ? '' : '/connect';
```
dnys1, line 49: Needs a better variable name and link to docs - what is the significance of `{}`?

packages/api/amplify_api/lib/src/decorators/authorize_websocket_message.dart
```diff
@@ -0,0 +1,68 @@
...
+  final canonicalHttpRequest =
+      http.Request('POST', Uri.parse('${config.endpoint}$maybeConnect'));
+  canonicalHttpRequest.headers.addAll({
+    AWSHeaders.accept: 'application/json, text/javascript',
```
dnys1, line 53: This value should be in a constant - it's weird to accept javascript from AppSync so we should note that it's just a requirement

packages/api/amplify_api/lib/src/decorators/authorize_websocket_message.dart
```diff
@@ -0,0 +1,68 @@
...
+    AWSHeaders.contentEncoding: 'amz-1.0',
```
dnys1, line 54: In fact all of these could just be in a const variable with a link to the AppSync doc that provides these values

packages/api/amplify_api/lib/src/decorators/authorize_websocket_message.dart
```diff
@@ -0,0 +1,68 @@
...
+  canonicalHttpRequest.body = body;
+
+  final authorizedHttpRequest = await authorizeHttpRequest(
+    canonicalHttpRequest,
+    endpointConfig: config,
+    authProviderRepo: authRepo,
+  );
+  return {
+    ...authorizedHttpRequest.headers,
+    AWSHeaders.host: endpointHost,
```
dnys1, line 66: Shouldn't set this on web

packages/api/amplify_api/lib/src/graphql/ws/websocket_message.dart
```diff
@@ -0,0 +1,202 @@
...
+class WebSocketMessage {
+  final String? id;
+  final MessageType messageType;
+  final WebSocketMessagePayload? payload;
+
+  WebSocketMessage({
+    String? id,
+    required this.messageType,
+    this.payload,
+  }) : id = id ?? UUID.getUUID();
```
dnys1, line 145:
```suggestion
  }) : id = id ?? uuid();
```

packages/api/amplify_api/lib/src/graphql/ws/websocket_message_stream_transformer.dart
```diff
@@ -0,0 +1,54 @@
...
+      switch (event.messageType) {
+        case MessageType.startAck:
+          onEstablished?.call();
+          break;
+        case MessageType.data:
+          final payload = event.payload as SubscriptionDataPayload;
+          final errors = deserializeGraphQLResponseErrors(payload.toJson());
+          yield GraphQLResponseDecoder.instance.decode<T>(
+              request: request,
+              data: json.encode(payload.data),
```
dnys1, line 42: Ew so this data will have gone through 4 encode/decode passes before finally being returned to the user?

packages/api/amplify_api/lib/src/graphql/ws/websocket_message_stream_transformer.dart
```diff
@@ -0,0 +1,54 @@
...
+import '../graphql_response_decoder.dart';
+
+class WebSocketMessageStreamTransformer
+    extends StreamTransformerBase<dynamic, WebSocketMessage> {
+  const WebSocketMessageStreamTransformer();
+
+  @override
+  Stream<WebSocketMessage> bind(Stream<dynamic> stream) {
+    return stream.cast<String>().map<Map>((str) {
+      return json.decode(str) as Map;
```
dnys1, line 17:
```suggestion
    return stream.cast<String>().map<Map<String, Object?>>((str) {
      return json.decode(str) as Map<String, Object?>;
```

packages/api/amplify_api/lib/src/graphql/ws/websocket_message.dart
```diff
@@ -0,0 +1,202 @@
...
+        if (payload != null) 'payload': payload?.toJson(),
+      };
+
+  @override
+  String toString() {
+    return prettyPrintJson(this);
+  }
+}
+
+class WebSocketConnectionInitMessage extends WebSocketMessage {
```
dnys1, line 183: Following the logic on these classes is really difficult since they get mutated after instantiation. I would suggest making them all `const`. This will force a better data flow and more functional programming.

packages/api/amplify_api/lib/src/decorators/authorize_websocket_message.dart
ragingsquirrel3: Why not? I know that's the case with http requests but I was noticing the handshake/connection fails on web (or any platform) if I remove this part (though don't have error message handy) and it's specified in appsync docs. On web is it possible to replace with some variant of this header here? like "x-amz-host"?

packages/api/amplify_api/lib/src/graphql/ws/websocket_message.dart
ragingsquirrel3: I think that's a good suggestion. I haven't made them const yet bc they have a uuid that gets generated if not passed by caller, but I changed the logic to avoid mutating them and generally refactored the auth logic.

ragingsquirrel3:
I'm still working on this, especially testing/logging, many comments not addressed. However, I did refactor the auth logic to use more pure functions and added lots of comments there. Let me know if still difficult to follow.

packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
```diff
@@ -0,0 +1,210 @@
...
+  // and subscription registration messages.
+  final AmplifyAuthProviderRepository _authProviderRepo;
+  final AWSApiConfig _config;
+
+  // Manages all incoming messages from server. Primarily handles messages related
+  // to the entire connection. E.g. connection_ack, connection_error, ka, error.
+  // Other events (for single subscriptions) rebroadcast to _rebroadcastController.
+  late final WebSocketChannel _channel;
+  late final StreamSubscription<WebSocketMessage> _subscription;
+  late final RestartableTimer _timeoutTimer;
```
Equartey, line 45: nit: Does this timer need to be cleaned up on close/cancel? What happens when connection ends, but time remains and then ends?

packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
```diff
@@ -0,0 +1,210 @@
...
+  Stream<GraphQLResponse<T>> subscribe<T>(
+    GraphQLRequest<T> request,
+    void Function()? onEstablished,
+  ) {
+    if (!_connectionReady.isCompleted) {
+      init();
+    }
+
+    final subscriptionId = uuid();
+    return _messageStream
```
Equartey, line 122: Request: Can you verify the onDone() function triggers on the client side? When I was playing around with the code it didn't seem to trigger when I thought it might. Could be my assumptions were wrong, but let's make sure it's wired up correctly. 

Tests around these callback functions might be appropriate too.

packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
```diff
@@ -0,0 +1,210 @@
...
+        timeoutDuration,
+      ),
+    );
+  }
+
+  /// Handles incoming data on the WebSocket.
+  ///
+  /// Here, handle connection-wide messages and pass subscription events to
+  /// `_rebroadcastController`.
+  void _onData(WebSocketMessage message) {
```
Equartey, line 168: Observation: There are other message types not being handled here. Do we need to handle any of those?

Equartey (commented):
Nice work!

Question: Does the example app need to be updated to reflect any api changes?

packages/api/amplify_api/lib/src/decorators/authorize_websocket_message.dart
dnys1: Oh yeah, I forgot the "headers" actually go in the query string

packages/api/amplify_api/lib/src/graphql/ws/websocket_message.dart
dnys1: Yeah, if not `const` then you can mark them `@immutable` which will help catch the same issues.

packages/api/amplify_api/lib/src/graphql/ws/websocket_connection.dart
```diff
@@ -0,0 +1,186 @@
...
+      TimeoutException(
+        'Connection timeout',
+        timeoutDuration,
+      ),
+    );
+  }
+
+  /// Handles incoming data on the WebSocket.
+  void _onData(WebSocketMessage message) {
+    // print('Received: message $message');
```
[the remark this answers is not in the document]
dnys1: Still seeing some print statements - I do think they should be logged at `verbose` level, though, in the same way the iOS client does, some of them at `debug` level.

packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
ragingsquirrel3: Good call! For this PR, I think I will call cancel on the timeout when closing. I think with reconnect, we may need to reconsider some of this, maybe making some of these not final and re-assigning. Same thing for futures used during initial connection.

packages/api/amplify_api/lib/src/graphql/ws/websocket_connection.dart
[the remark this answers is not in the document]
ragingsquirrel3: Still working on addressing this.

packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
ragingsquirrel3: Some of the other message types are handled in the transformer, the ones related to individual subscriptions. Are there other types missing?

packages/api/amplify_api/lib/src/graphql/ws/websocket_message_stream_transformer.dart
ragingsquirrel3: yeah, not ideal, still working on improving this.

ragingsquirrel3:
> Nice work!
> 
> Question: Does the example app need to be updated to reflect any api changes?

No API changes for subscriptions, though I did just make a change to example app with last rev so unsubscribe button works.

packages/api/amplify_api/lib/src/graphql/ws/websocket_connection.dart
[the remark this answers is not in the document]
ragingsquirrel3: Just changed this w last revision. I made some cancel sub calls log with info bc in current implementation we are calling print() so I wanted to do something similar. Not wed to that and open to further suggestions regarding logging (or anything else here).

packages/api/amplify_api/lib/src/graphql/ws/websocket_message_stream_transformer.dart
ragingsquirrel3: done, thanks!

packages/api/amplify_api/lib/src/graphql/ws/websocket_message_stream_transformer.dart
ragingsquirrel3: I would rather improve this in a separate PR as this is all based on this method taking a string (based on method channel implementation) and changing it would touch several places. I think my simply making this method take a map (could be with errors as well) it would reduce 2 of those conversions in many cases. Would TODO comment here suffice for that? Do you have other suggestions for simplifying it?

packages/api/amplify_api/test/dart_graphql_test.dart
```diff
@@ -12,10 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import 'dart:async';
```
dnys1, line 15: I guess I never noticed this - but why are the other method channel tests/infra hanging around still? Outside of the configure API, we have no use for them anymore. These tests should just replace the old subscribe tests.

packages/api/amplify_api/lib/src/graphql/ws/websocket_message_stream_transformer.dart
dnys1: Either a TODO or FIXME. Let's not lose track of these changes though. Why are we waiting to remove method channel code?

packages/api/amplify_api/example/lib/graphql_api_view.dart
```diff
@@ -45,13 +45,19 @@ class _GraphQLApiViewState extends State<GraphQLApiView> {
...
-        print('Subscription event data received: ${event.data}');
-      }
-    } on Exception catch (e) {
-      print('Error in subscription stream: $e');
-    }
+    final streamSubscription = operation.listen(
+      (event) {
+        final result = 'Subscription event data received: ${event.data}';
+        print(result);
+        setState(() {
```
Equartey, line 52: Nice, this was a much needed improved imo

packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
```diff
@@ -0,0 +1,240 @@
...
+    return ready;
+  }
+
+  /// Subscribes to the given GraphQL request. Returns the subscription object,
+  /// or throws an [Exception] if there's an error.
+  Stream<GraphQLResponse<T>> subscribe<T>(
+    GraphQLRequest<T> request,
+    void Function()? onEstablished,
+  ) {
+    final subscriptionId = uuid();
```
Equartey, line 143: nit: This might as well be changed to use request.id now. It'll be that way in reconnect PR

packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
Equartey: I was thinking of the `complete` type, but looks like you got that handled in that file 👍

Equartey (approved):
Good work on the revisions. I noticed some things that will be nice for reconnection :)

packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
ragingsquirrel3: yes good catch

packages/api/amplify_api/test/dart_graphql_test.dart
ragingsquirrel3: I would rather do the cleanup in a single PR. There was a task we mentioned in grooming for it.

packages/api/amplify_api/lib/src/graphql/ws/websocket_message_stream_transformer.dart
ragingsquirrel3: added TODO

packages/api/amplify_api/test/dart_graphql_test.dart
```diff
@@ -184,12 +223,34 @@ void main() {
...
       expect(res.errors, equals(null));
     });
+
+    test('subscribe() should decode model data', () async {
+      Completer<void> establishedCompleter = Completer();
+      Completer<Post> dataCompleter = Completer();
+      final subscriptionRequest = ModelSubscriptions.onCreate(Post.classType);
+      final subscription = Amplify.API.subscribe(
+        subscriptionRequest,
+        onEstablished: () => establishedCompleter.complete(),
```
dnys1, line 233: This pattern of completer/complete/expectLater can be simplified with the `expectAsync*` methods from `package:test` e.g. this does the same thing.

```dart
Amplify.API.subscribe(
  subscriptionRequest,
  onEstablished: expectAsync0(() {}),
);
```

It makes more sense when the callback takes parameters and you want to run more `expect`s on those parameters, but it works here to shorten the amount of code you need to write.

packages/api/amplify_api/test/dart_graphql_test.dart
```diff
@@ -184,12 +223,34 @@ void main() {
...
+      );
+
+      final streamSub = subscription.listen(
+        (event) => dataCompleter.complete(event.data),
```
dnys1, line 237: e.g.

```dart
subscription.listen(
  expectAsync1((event) {
    expect(event.data, isA<Post>());
  }),
);

File: packages/api/amplify_api/test/dart_graphql_test.dart [hunk 23]
```diff
@@ -218,7 +279,7 @@ void main() {
     });
 
     test('canceled mutation request should never resolve', () async {
-      final req = GraphQLRequest(document: '', variables: {});
+      final req = GraphQLRequest<String>(document: '', variables: {});
```
dnys1, line 282: These should still work - was this just to prevent implicit `dynamic`? If so, can we have some tests with explicit `dynamic` if we don't elsewhere, e.g. `GraphQLRequest<dynamic>(...)`

Or better yet, add a task to remove the need for this once and for all 😂

packages/api/amplify_api/test/util.dart
```diff
@@ -60,3 +67,111 @@ void validateSignedRequest(http.BaseRequest request) {
...
+      'file': null,
+      'files': null,
+      'updatedAt': '2022-08-15T18:22:15.164Z'
+    }
+  }
+};
+
+/// Extension of [WebSocketConnection] that stores messages internally instead
+/// of sending them.
+class MockWebSocketConnection extends WebSocketConnection {
```
dnys1, line 114: At some point we should be mocking/injecting the `WebSocketChannel` instead and not overriding methods since it becomes unclear what's being tested. Like here, I guess we're testing everything but `connect`/`send` - but those seem like good things to test too.

packages/api/amplify_api/lib/src/decorators/web_socket_auth_utils.dart
```diff
@@ -0,0 +1,120 @@
...
+  AWSApiConfig config, {
+  required AmplifyAuthProviderRepository authRepo,
+  required String body,
+}) async {
+  final endpointHost = Uri.parse(config.endpoint).host;
+  // Create canonical HTTP request to authorize but never send.
+  //
+  // The canonical request URL is a little different depending on if connection_init
+  // or start (subscription registration).
+  final maybeConnect = body != _emptyBody ? '' : '/connect';
```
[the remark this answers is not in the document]
ragingsquirrel3: This comment is a little misleading so at the very least, will change it. Not sure it makes sense to pass the message type bc when authorizing the connection URI for handshake. It's not exactly authorizing the websocket message itself for connection init (as comment suggests), just the URI for the connection with authorized query params. That being said, I agree with your criticism here that it's confusing bc implicitly using value of body string to switch on the 2 major use cases here. I may just add a boolean param here `isConnectionInit` to explicity state if for handshake connection URI or sub reg.

packages/api/amplify_api/lib/src/graphql/ws/web_socket_types.dart
```diff
@@ -0,0 +1,229 @@
...
+  }
+
+  @override
+  Map<String, dynamic> toJson() => <String, dynamic>{
+        'data': data,
+        'errors': errors,
+      };
+}
+
+@internal
```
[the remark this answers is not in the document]
ragingsquirrel3: nice, thanks!

packages/api/amplify_api/test/dart_graphql_test.dart
ragingsquirrel3: good tip, thanks, integrated this in some places but in other places it was easier to just have the completer/future.

packages/api/amplify_api/test/dart_graphql_test.dart
ragingsquirrel3: It's for implicit `dynamic` like you thought so I will just add a test for explicit dynamic query.

packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
```diff
@@ -0,0 +1,262 @@
...
+        final wsError = message.payload as WebSocketError;
+        _rebroadcastController.addError(wsError);
+        return;
+      default:
+        break;
+    }
+
+    // Re-broadcast other message types related to single subscriptions.
+
+    if (!_rebroadcastController.isClosed) _rebroadcastController.add(message);
```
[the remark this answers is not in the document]
ragingsquirrel3: made this hard fail

packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
```diff
@@ -0,0 +1,262 @@
...
+    send(WebSocketStopMessage(id: subscriptionId));
+    // TODO(equartey): if this is the only subscription, close the connection.
+  }
+
+  /// Serializes a message as JSON string and sends over WebSocket channel.
+  @visibleForTesting
+  void send(WebSocketMessage message) {
+    final msgJson = json.encode(message.toJson());
+    if (_channel == null) {
+      throw ApiException(
```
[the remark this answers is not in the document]
ragingsquirrel3: agreed, channel shouldn't be null here so just changed  to use `!` here. Channel is only null during connection init so that shouldn't be a factor. I think this kind of logic might get touched more when implementing reconnect but will leave that for Elijah.

packages/api/amplify_api/lib/src/graphql/ws/web_socket_types.dart
```diff
@@ -0,0 +1,229 @@
...
+
+// ignore_for_file: public_member_api_docs
+
+import 'dart:convert';
+
+import 'package:amplify_core/amplify_core.dart';
+import 'package:meta/meta.dart';
+
+@internal
+class MessageType {
```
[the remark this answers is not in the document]
ragingsquirrel3: just made this enum

packages/api/amplify_api/test/util.dart
ragingsquirrel3: makes sense, I put a task in backlog to improve this

packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
```diff
@@ -0,0 +1,258 @@
...
+
+  /// Serializes a message as JSON string and sends over WebSocket channel.
+  @visibleForTesting
+  void send(WebSocketMessage message) {
+    final msgJson = json.encode(message.toJson());
+    _channel!.sink.add(msgJson);
+  }
+
+  /// Times out the connection (usually if a keep alive has not been received in time).
+  void _timeout(Duration timeoutDuration) {
```
dnys1, line 205: Shouldn't this method do more, like close the WS connection? I guess Elijah's PR will address that

dnys1 (approved):
This looks good for now. Once the method channel stuff is gone, I would recommend enabling lints at that point and doing the clean up for that so that future PRs meet spec. I've withheld all my formatting/style comments because it's easier if the analyzer prompts you as you go.

packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
ragingsquirrel3: Yeah, I think there will be more cleanup here but will leave for Elijah.

PR #1917, opened 2022-07-22
feat(auth): Allow multiple Hosted UI sign-in attempts

dnys1:
Allows multiple calls to `signInWithWebUI` so that, for example, desktop UIs do not have to be aware of the state of the launched browser. Users can launch a Hosted UI browser only to close it and re-initiate a new sign-in.

### DIFF
--- a/packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
+++ b/packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
@@ -21,6 +21,8 @@ import 'package:amplify_authenticator/src/blocs/auth/auth_data.dart';
 import 'package:amplify_authenticator/src/services/amplify_auth_service.dart';
 import 'package:amplify_authenticator/src/state/auth_state.dart';
 import 'package:amplify_flutter/amplify_flutter.dart';
+import 'package:async/async.dart';
+import 'package:stream_transform/stream_transform.dart';
 
 part 'auth_event.dart';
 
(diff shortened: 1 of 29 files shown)


haverchuck (approved):

packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
```diff
@@ -119,6 +136,27 @@ class StateMachineBloc {
...
+  AuthState? _mapHubEvent(AuthHubEvent event) {
+    logger.debug('Handling hub event: ${event.type}');
+    switch (event.type) {
+      case AuthHubEventType.signedIn:
+        if (_currentState is! AuthenticatedState) {
+          return const AuthenticatedState();
+        }
+        break;
+      case AuthHubEventType.signedOut:
+      case AuthHubEventType.sessionExpired:
```
Jordan-Nelson, line 150: When a user's session expires they will be taken to the sign in screen. We will not provide any message to them explaining why. The assumption is that the customer will be responsible for listening to the hub events and providing some sort of info in the UI if desired... Will they be able to do that easily?

I think the UX is fine if this happens on app launch, but it is a poor UX if it happens while the user is in the middle of something. Should we attempt to show something in the UI in this case?

I think this could be a follow up. I suspect we already are not handling this case well. Can we create a task for it if we think this should be improved?

Jordan-Nelson (approved):

packages/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
dnys1: Done 👍 good callout

PR #1924, opened 2022-07-25
feat: Smithy Dart

dnys1:
Migrates Smithy Dart code to the Amplify Flutter repo


dnys1:
I pushed a big JSON file (`endpoints.json`) which tacked on another 20k lines to the PR. I tried to avoid tracking it but it's needed for the build_verify step, which I think makes it worth having.

haverchuck (approved):
It would be nice to add some documentation that provides a general overview of how these packages work together - not a blocker though.

packages/smithy/smithy/lib/src/http/http_operation.dart
```diff
@@ -0,0 +1,378 @@
...
+    };
+    final body = protocol.serialize(input, specifiedType: FullType(Input));
+    var host = uri.host;
+    if (!endpoint.isHostnameImmutable && request.hostPrefix != null) {
+      var prefix = request.hostPrefix!;
+      if (input is HasLabel) {
+        prefix = expandHostLabel(prefix, input.labelFor);
+      }
+      host = '$prefix$host';
+    }
```
Jordan-Nelson, line 166: There are a lot of variables in here that could be final with some refactoring. It makes it a little easier to read.

```suggestion
    final prefix = _createPrefix(
      endpoint: endpoint,
      request: request,
      input: input,
    );
    final host = '$prefix${uri.host}';
```

```dart
  String? _createPrefix({
    required Endpoint endpoint,
    required HttpRequest request,
    required Input input,
  }) {
    if (!endpoint.isHostnameImmutable && request.hostPrefix != null) {
      if (input is HasLabel) {
        return expandHostLabel(request.hostPrefix!, input.labelFor);
      }
      return request.hostPrefix;
    }
    return null;
  }
```

packages/smithy/smithy/lib/src/types/timestamp.dart
```diff
@@ -0,0 +1,111 @@
...
+// millisecond precision. If a string is provided, it MUST be a valid RFC 3339
+// string with no UTC offset and optional fractional precision
+// (for example, 1985-04-12T23:20:50.52Z).
+
+import 'package:http_parser/http_parser.dart';
+import 'package:intl/intl.dart';
+import 'package:smithy/ast.dart';
+
+/// {@template aws.smithy.timestamp}
+///
```
Jordan-Nelson, line 27: Add a comment or remove

packages/smithy/smithy/lib/src/types/union.dart
```diff
@@ -0,0 +1,113 @@
...
+import 'package:built_value/serializer.dart';
+import 'package:smithy/smithy.dart';
+
+/// An union representation in Smithy, i.e. a value which is one of a fixed set
+/// of types.
+///
+/// For example, the following Smithy model:
+///
+/// ```
+/// @enum("YES": {}, "NO": {})
```
Jordan-Nelson, line 25: I think the example needs to be updated

packages/smithy/goldens/analysis_options.yaml
```diff
@@ -0,0 +1 @@
+include: package:lints/recommended.yaml
```
Jordan-Nelson, line 1: Do we want to use amplify lints? Could be done later as well.

packages/smithy/smithy_aws/test/http/aws_retryer_test.dart
```diff
@@ -0,0 +1,379 @@
...
+              case Outcome.retryRequest:
+                expectation = completes;
+                break;
+              case Outcome.maxAttemptsExceeded:
+              case Outcome.retryQuotaExceeded:
+                expectation = throwsA(isA<_TransientSmithyException>());
+                break;
+            }
+
+            // TODO(dnys1): Try to get fake_async to work properly to avoid
```
Jordan-Nelson, line 156: Whats the unnecessary delay?

Jordan-Nelson (approved):
I left some comments, but I wouldn't consider any of them blocking. Nice work.

packages/smithy/smithy_aws/test/http/aws_retryer_test.dart
dnys1: There's a `Future.delayed` inside the retryer

packages/smithy/goldens/analysis_options.yaml
dnys1: Nah, generated code should meet `lints` but it's not worth trying to get it cleaner than that.

packages/smithy/smithy/lib/src/types/timestamp.dart
dnys1: Lol

PR #1929, opened 2022-07-26
chore(auth): Clean up `ConfirmDeviceWorker`

dnys1:
Small touch ups, including a fix to align with the Cognito [docs](https://aws.amazon.com/premiumsupport/knowledge-center/cognito-user-pool-remembered-devices/). The `password` param to `ConfirmDevice` should be a 40-byte random string, not the user's password.


haverchuck (approved):

Jordan-Nelson (approved):
Looks good. Js files may need to be regenerated before you merge?

dnys1:
Potentially, yeah although depending on what gets merged today, it may be moot. I may put up a PR to remove them from tracking if it doesn't cause too many conflicts.

PR #1930, opened 2022-07-26
chore(secure_storage): use in-memory when indexedDB fails to open

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Check that indexedDB can open without exceptions, add in-memory as fallback

### DIFF
--- a/packages/secure_storage/amplify_secure_storage_dart/lib/src/platforms/amplify_secure_storage_web.dart
+++ b/packages/secure_storage/amplify_secure_storage_dart/lib/src/platforms/amplify_secure_storage_web.dart
@@ -32,7 +32,7 @@ class AmplifySecureStorageWeb extends AmplifySecureStorageInterface {
     if (indexedDB == null) {
       logger.warn(
         'IndexedDB is not available. '
-        'Falling back on in-memory storage.',
+        'Falling back to in-memory storage.',
       );
       return const AmplifySecureStorageInMemory();
     }
@@ -42,11 +42,11 @@ class AmplifySecureStorageWeb extends AmplifySecureStorageInterface {
       final openRequest = indexedDB!.open('test', 1);
       await openRequest.future;
       return _IndexedDBStorage(config: config);
-    } on Object catch (_) {
+    } on Object {
       logger.warn(
         'Could not open IndexedDB database. '
         'It may not be supported by the current browser. '
-        'Falling back on in-memory storage.',
+        'Falling back to in-memory storage.',
       );
       return const AmplifySecureStorageInMemory();
     }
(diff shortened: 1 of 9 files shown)


packages/secure_storage/amplify_secure_storage_dart/lib/src/platforms/amplify_secure_storage_web.dart
```diff
@@ -14,16 +14,66 @@
...
+    try {
+      final openRequest = indexedDB!.open('test', 1);
+      await openRequest.future;
+      return _IndexedDBStorage(config: config);
+    } on Object catch (_) {
+      logger.warn(
+        'Could not open IndexedDB database. '
+        'It may not be supported by the current browser. '
+        'Falling back on in-memory storage.',
+      );
```
dnys1, line 50: If you think it's helpful to include, you could do this too :

```suggestion
    } on Object catch (e) {
      logger.warn(
        'Could not open IndexedDB database. '
        'It may not be supported by the current browser. '
        'Falling back on in-memory storage.',
        e,
      );
```

I generally prefer not catching anything if it's unused, though

```suggestion
    } on Object {
      logger.warn(
        'Could not open IndexedDB database. '
        'It may not be supported by the current browser. '
        'Falling back on in-memory storage.',
      );
```

packages/secure_storage/amplify_secure_storage_dart/lib/src/platforms/amplify_secure_storage_web.dart
```diff
@@ -14,16 +14,66 @@
...
 
+  /// The [SecureStorageInterface] instance to use.
+  late final Future<SecureStorageInterface> _instance = () async {
+    if (config.webOptions.persistenceOption == WebPersistenceOption.inMemory) {
+      return const AmplifySecureStorageInMemory();
+    }
+    if (indexedDB == null) {
+      logger.warn(
+        'IndexedDB is not available. '
+        'Falling back on in-memory storage.',
```
dnys1, line 35: Idk, just me?

```suggestion
        'Falling back to in-memory storage.',
```

dnys1 (approved):

packages/secure_storage/amplify_secure_storage_dart/lib/src/platforms/amplify_secure_storage_web.dart
```diff
@@ -14,16 +14,66 @@
...
 
+  /// The [SecureStorageInterface] instance to use.
+  late final Future<SecureStorageInterface> _instance = () async {
+    if (config.webOptions.persistenceOption == WebPersistenceOption.inMemory) {
+      return const AmplifySecureStorageInMemory();
+    }
+    if (indexedDB == null) {
+      logger.warn(
+        'IndexedDB is not available. '
+        'Falling back to in-memory storage.',
```
haverchuck, line 35: I'm assuming that most developers will understand the implication of this, but if we get GH issues to the contrary we should consider adding a statement making it explicit.

haverchuck (approved):

PR #1933, opened 2022-07-27
chore: authenticator and amplify_flutter test fixes

haverchuck:
*Description of changes:*
Updates authenticator integ tests and refactors amplify_flutter tests (including adding flutter driver support).


packages/amplify/amplify_flutter/example/integration_test/config_test.dart
```diff
@@ -154,141 +153,72 @@ void main() {
...
-        await expectLater(Amplify.configure(amplifyconfig), completes);
-        expect(Amplify.isConfigured, isTrue);
-
-        await expectLater(
-          Amplify.configure(amplifyconfig),
-          throwsAlreadyConfigured,
-        );
-      });
-
-      testWidgets('hot reload persists all state', (WidgetTester tester) async {
```
Jordan-Nelson, line 205: Is the idea behind removing the hot restart / hot reload tests that these will not be needed in once the native plugins are no longer used and they are not worth maintaining?

Jordan-Nelson (approved):

packages/amplify/amplify_flutter/example/integration_test/config_test.dart
haverchuck: Yeah, and will probably be replaced with Amplify.reset tests.

dnys1 (approved):

PR #1940, opened 2022-07-29
fix(auth): cognito integ test fixes

haverchuck:
*Description of changes:*
Fixes signin and custom auth integ tests.

The change to the `auth_plugin_impl.dart` file is to support retrieval of information from the cognito trigger, which currently is stored in the `additionalInfo` attribute and not `challengeParameters`. However, I believe that `challengeParameters` is used for other purposes in other flows, hence the carveout for custom auth. 

### DIFF
--- a/packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
+++ b/packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
@@ -56,8 +56,6 @@ import 'package:built_collection/built_collection.dart';
 import 'package:collection/collection.dart';
 import 'package:meta/meta.dart';
 
-import 'sdk/src/cognito_identity_provider/model/challenge_name_type.dart';
-
 /// {@template amplify_auth_cognito_dart.amplify_auth_cognito_dart}
 /// The AWS Cognito implementation of the Amplify Auth category.
 /// {@endtemplate}
--- a/packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
+++ b/packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
@@ -457,7 +457,6 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface
           continue;
         case SignInStateType.challenge:
           state as SignInChallenge;
-
           return SignInResult(
             isSignedIn: false,
             nextStep: AuthNextSignInStep(
@@ -471,7 +470,6 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface
               missingAttributes: state.requiredAttributes,
             ),
           );
-
         case SignInStateType.success:
           return SignInResult(
             isSignedIn: true,
--- a/packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
+++ b/packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
@@ -216,16 +216,6 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface
     );
   }
 
-  /// Retrieves the challenge parameters that are not part of standard Cognito InitiateAuth flow.
-  Map<String, String>? _getAdditionalInfo(
-    Map<String, String> challengeParameters,
-  ) {
-    final additionalInfo = {...challengeParameters}..removeWhere(
-        (key, value) => CognitoConstants.staticChallenges.contains(key),
-      );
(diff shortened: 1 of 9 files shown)


packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
```diff
@@ -445,17 +447,34 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface
...
-              signInStep: state.challengeName.signInStep.value,
-              codeDeliveryDetails: _getChallengeDeliveryDetails(
-                state.challengeParameters,
-              ),
-              challengeParameters: state.challengeParameters,
-              missingAttributes: state.requiredAttributes,
-            ),
-          );
+
+          switch (state.challengeName) {
```
dnys1, line 451: We should just get rid of `challengeParameters`. I added it because I didn't realize we had `additionalInfo`

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
```diff
@@ -56,6 +56,8 @@ import 'package:built_collection/built_collection.dart';
 import 'package:collection/collection.dart';
 import 'package:meta/meta.dart';
 
+import 'sdk/src/cognito_identity_provider/model/challenge_name_type.dart';
```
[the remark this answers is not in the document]
dnys1: Yeah let's do it - https://github.com/aws-amplify/amplify-flutter/pull/1943

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
haverchuck: Done

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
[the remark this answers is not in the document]
haverchuck: We actually don't need this import after I made another change... removed it and approved the lint rule change PR.

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
```diff
@@ -216,6 +216,16 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface
     );
   }
 
+  /// Retrieves the challenge parameters that are not part of standard Cognito InitiateAuth flow.
+  Map<String, String>? _getAdditionalInfo(
+    Map<String, String> challengeParameters,
+  ) {
+    final additionalInfo = {...challengeParameters}..removeWhere(
```
dnys1, line 223: Not following why this is needed over just returning the challenge params

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
dnys1: Let's actually remove it from the result object I mean, since I added them only recently

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
haverchuck: I believe that the data in `state.challengeParameters` (and ultimately from the sdk response) include parameters for a variety of auth challenges that Cognito may issue. In addition, it can include the `publicChallengeParameters` returned by customer-written lambda code. My expectation is that this member wouldn't include things like `'SALT` and other values used internally by our library when calling Cognito, but other information which may be present in the response (which basically boils down to just the lambda response, but I'm trying to confirm this).

Jordan-Nelson (approved):

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
dnys1: In general, yes, but the value of `challengeParameters` for the CUSTOM_CHALLENGE state only has the username plus any custom values sent from the backend, e.g.

```json
{
  "USERNAME": "test4",
  "trigger": "true"
}
```

Was there a case where you noticed more than this being returned?

haverchuck: Oh I see. I was just testing the size of the list for anything beyond the publicChallengeParameters configured by the customer in their lambda. If these are the only values being appended, we can probably just leave it for now.

dnys1 (approved):

PR #1941, opened 2022-07-29
fix(datastore): missing query field model name cause ambigous column …

HuiSF:
…SQL error

*Issue #, if available:*

*Description of changes:*

* Update QueryPredicateBuilder to supply `modelName` for native `QueryField`
* Update integration tests to cover related use case to verify the fix

### DIFF
--- a/packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/query/QueryPredicateBuilder.kt
+++ b/packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/query/QueryPredicateBuilder.kt
@@ -58,7 +58,13 @@ class QueryPredicateBuilder {
                     }
                 }
 
-                val queryField: QueryField = QueryField.field(field)
+                // Here we are using query root model name to create QueryField to let amplify-android
+                // generate correct SQL command.
+                // This is based on the current assumption: amplify-flutter doesn't support cross models nested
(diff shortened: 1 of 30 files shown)


poojamat (approved):

dnys1 (approved):

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/query/QueryPredicateBuilder.kt
```diff
@@ -58,7 +58,12 @@ class QueryPredicateBuilder {
                     }
                 }
 
-                val queryField: QueryField = QueryField.field(field)
+                // Here we are using query root model name to create QueryField to let amplify-android
+                // generate correct SQL command.
+                // This is based on the current assumption: amplify-flutter doesn't support cross models nested
+                // predicate e.g. query comments by post.id
```
Jordan-Nelson, line 64: I thought this was possible. `where: Comment.POST.eq(postID)`. Are you referring to something else?

HuiSF: this comment is for

```dart
await Amplify.DataStore.query(
  Comment.classType,
  where: Post.ID.eq('123),
)
```

(This is currently supported only in amplify-android)

Jordan-Nelson (approved):

Issue #1898, opened 2022-07-20
Authenticator banner text is misaligned on large screen sizes


dnys1:
### Description

Banner text should be aligned with the icons, but appears too close to the top currently.

<img width="912" alt="Screen Shot 2022-07-19 at 5 57 49 PM" src="https://user-images.githubusercontent.com/24740863/179872963-65e38584-5f4f-4dd3-81c7-cc44cdea5c90.png">

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [ ] Auth
- [X] Authenticator
- [ ] DataStore
- [ ] Storage

### Steps to Reproduce

Trigger an error on a large screen size

### Platforms

- [ ] iOS
- [ ] Android

### Environment

```bash
`next`
```

### Dependencies

```bash
N/A
```

### Device

macOS

### OS

12

### CLI Version

N/A

dnys1:
Also, should the banner be a different color/symbol for errors? Right now, it's the same for all messages.

Jordan-Nelson:
This have been addressed in our developer preview version.

Issue #1889, opened 2022-07-15
Authenticator throws unclear exception when Authenticator FormFields are used outside of an AuthenticatorForm


Jordan-Nelson:
### Description

A very long and unclear exception is thrown when Authenticator FormFields are used outside of an AuthenticatorForm.

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [ ] Auth
- [X] Authenticator
- [ ] DataStore
- [ ] Storage

### Steps to Reproduce

1. Run the Authenticator sample app with the `AuthenticatorWithCustomLayout` widget.
2. Replace the `AuthenticatorForm` in `/lib/customization/authenticator_with_custom_layout.dart` with a `Container`.
3. Navigate to Sign Up.
4. See that the error message is not helpful at all.

Expectation: We intended to account for this scenario and this is the error message that *should* be seen: https://github.com/aws-amplify/amplify-flutter/blob/e4ceb156cf9eb6d203db56648cbdb416e793822d/packages/amplify_authenticator/lib/src/widgets/form.dart#L122-L132

### Platforms

- [ ] iOS
- [ ] Android

### Environment

```bash
N/A
```

### Dependencies

```bash
N/A
```

### Device

N/A

### OS

N/A

### CLI Version

N/A

### Additional Context

`debugFillProperties` in `AuthenticatorFormField` is throwing (possibly recursively), which is making the error very  confusing and hiding the message we intended to show.

Jordan-Nelson:
This has been addressed in the developer preview version.

PR #1976, opened 2022-08-02
fix(authenticator): Banner styles & custom form errors

Equartey:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-flutter/issues/1898 https://github.com/aws-amplify/amplify-flutter/issues/1889

*Description of changes:*
This implements fixes for the notification banner and error handling when using custom forms. The banner colors are now tied to the type of message, text is aligned more consistently, and icons are more distinguishable. Missing `AuthenticatorForm` error now displays in the console. 


packages/amplify_authenticator/lib/src/widgets/authenticator_banner.dart
```diff
@@ -79,4 +77,17 @@ extension on StatusType {
         return Icons.error;
     }
   }
+
+  ColorSwatch<int> get color {
```
Equartey, line 81: Wondered if there is a way to tie these colors to the theme and allow customization by devs?

Jordan-Nelson: Yeah we need to make this so that customers can still set this via the theme. These colors seem like reasonable defaults, but customers need to be able to change them. You might be able to check to see if the theme has a color set and use this as a fallback.

The SnackBar foreground color is set to `theme.snackBarTheme.contentTextStyle?.color ?? theme.colorScheme.surface`. It is using `theme.colorScheme.surface` as a fallback. You could update it to to `theme.snackBarTheme.contentTextStyle?.color ?? type.color` so that it uses this value as a fallback, but still allows customers to theme it. That is for SnackBar, but the same could be applied to material banner.

We should (eventually) allow customers to provide a theme for each type (info, success, error) but that should be part of a larger theming API change. See https://github.com/aws-amplify/amplify-flutter/pull/1850/files#diff-fc4479f329ca30aeb3fc007095889bc66b798e3f77569534ca9bfcd1971bb216R83-R92

Whatever updates we make to material banner we should probably make to snack bar as well.

Do these colors look reasonable in light and dark mode (Does the text color invert? If so, is that what we want?)

packages/amplify_authenticator/lib/src/widgets/authenticator_banner.dart
```diff
@@ -79,4 +77,17 @@ extension on StatusType {
...
+    switch (this) {
+      case StatusType.info:
+        return Colors.blue;
+      case StatusType.success:
+        return Colors.green;
+      case StatusType.warning:
```
Jordan-Nelson, line 87: Is `StatusType.warning` used anywhere? if not, can you remove it?

Equartey: Good call. I don't think warning or success get used currently.

packages/amplify_authenticator/lib/src/widgets/authenticator_banner.dart
Equartey: Great insight. Here are some options I've uncovered based on what you shared 

1) Tie the colors to the snackbar theme as suggested.

2) Use the material banner specific theme settings, eg. `theme.bannerTheme.backgroundColor`.

3) Tie the different type colors individually to theme settings like `theme.errorColor` & `theme.hintColor`

Re: light/dark mode - the colors do invert based on the mode and they look readable/reasonable to me. Now if this is what we want is probably a better question for a designer haha.

Jordan-Nelson: Let's use the bannerTheme for MaterialBanner and the snackBarTheme for SnackBar 👍 . Also, I just realized I was mixing up background and foreground color. Obviously let's use the SnackBar / banner theme *background* to determine the background color, and user the theme's *foreground* color to determine the text/icon/etc. I mixed it up in the code snippet.

```dart
// SnackBar background
final backgroundColor = theme.snackBarTheme.backgroundColor ?? type.color;

// Banner background
final backgroundColor = theme.bannerTheme.backgroundColor ?? type.color;
```

Re: light/dark mode - Yeah, this is a bit of an opinion. Let's make sure it meets accessibility standards, and then try to use Material Design standards or defaults where possible to guide us.

Here is what Material has to say about [text on colored backgrounds](https://material.io/design/color/text-legibility.html#text-backgrounds).

> Black text is recommended for use on light backgrounds, and white text on dark backgrounds. If your app has both light and dark themes, make sure the text is available in a contrasting color against each theme.

You could probably interpret this multiple ways, but the way I interpret this is - Text color should be a function of background color. If your app has different background colors in light/dark mode, make sure your text responds to that. That is what flutter does by default. SnackBar has a different background color in light/dark mode, and the foreground color responds to that.

In our case, I think we are going to use red/blue/green as the background in both light and dark mode, so I don't think the foreground color needs to adapt. I would choose either white or black and use it as the (fallback) foreground color regardless of the theme.

So how do we know if we should use black or white? Material has a great interactive guide for this. You pick a Material theme (or define a custom one) and it will help you decide what color the text should be based on the color/size/weight. Here is the link for [Colors.red](https://material.io/resources/color/#!/?view.left=1&view.right=0&primary.color=F44336). The guide here says that both white and black text is fine if the text is "large" but will be illegible for "normal" text (tooltip shows the definition of large/normal text). In other terms, black text will always be legible on this background, but white will only be legible if the font is large/bold. So, I would choose black in this case. It happens to be the same for blue and green as well. If you chose indigo, it would be the opposite - white is always legible, and black is only legible if the text is large/bold.

And finally we have to consider if the user has set values for either foreground or background.

This is what I am thinking the logic should be. This is for SnackBar, but should be adaptable for banners.

```dart
  // foreground and background colors are based on material standards for text legibility
  // if they are not provided by the theme. See https://material.io/design/color/text-legibility.html#text-backgrounds

  // fallback on default color based on type.
  final backgroundColor = theme.snackBarTheme.backgroundColor ?? type.color;

  // If there is no background color provided by the theme,
  // use on Colors.black87 (High-emphasis) as fallback, otherwise 
  // set it to null to let flutter control the foreground.
  final hasCustomBackground = theme.snackBarTheme.backgroundColor != null;
  final foregroundColorFallBack = hasCustomBackground ? null : Colors.black87;

  // use contentTextStyle color if available, otherwise use fallback as defined above.
  final foregroundColor =
      theme.snackBarTheme.contentTextStyle?.color ?? foregroundColorFallBack;
```

packages/amplify_authenticator/lib/src/widgets/authenticator_banner.dart
```diff
@@ -79,4 +120,15 @@ extension on StatusType {
...
   }
+
+  ColorSwatch<int> get color {
+    switch (this) {
+      case StatusType.info:
+        return Colors.blue;
+      case StatusType.error:
+        return Colors.red;
+      default:
+        return Colors.blue;
```
dnys1, line 131: So, warning & success will be blue by default? Seems like yellow/green may be better choices for these, no?

packages/amplify_authenticator/lib/src/widgets/authenticator_banner.dart
```diff
@@ -21,46 +21,87 @@ import 'package:flutter/material.dart';
...
+  required Function() actionCallback,
 }) {
-  final margin = MediaQuery.of(context).viewPadding.top;
+  final theme = Theme.of(context);
+
+  // foreground and background colors are based on material standards for text legibility
+  // if they are not provided by the theme. See https://material.io/design/color/text-legibility.html#text-backgrounds
+
+  // fallback on default color based on type.
+  final backgroundColor = theme.bannerTheme.backgroundColor ?? type.color;
```
dnys1, line 33: My DRY sense are tingling when I see all this duplicated code. If you create a type to wrap backgroundColor/foregroundColor you could create a `ColorChoices Function(customForegroundColor?, customBackgroundColor?)` method to isolate this logic

dnys1 (commented):
Looking good!

packages/amplify_authenticator/lib/src/widgets/authenticator_banner.dart
Jordan-Nelson: @Equartey you can completely remove `StatusType.warning` & and `StatusType.success` from the codebase since they are unused. Then the default case can be removed.

packages/amplify_authenticator/lib/src/widgets/authenticator_banner.dart
```diff
@@ -79,4 +120,15 @@ extension on StatusType {
         return Icons.error;
     }
   }
+
+  ColorSwatch<int> get color {
+    switch (this) {
+      case StatusType.info:
+        return Colors.blue;
+      case StatusType.error:
+        return Colors.red;
```
Jordan-Nelson, line 129: Should have caught this before - `Colors.red` is probably not what we actually want. Errors are not typically displayed as true red. They are usually a darker shade.

Material's docs say the baseline for this (I think that means default) is `#B00020` (See "Surface, background, and error colors" [here](https://material.io/design/color/the-color-system.html#color-theme-creation)). Seems like flutter uses `#d32f2f` as a default for the error color. I can't find that documented, but that is the color used by default for theme.errorColor.

I don't have a strong preference of one over the other. Maybe we just use flutter's default.

Note: for either one of these new darker background colors, the text color should be white based on materials contrast tool.

packages/amplify_authenticator/lib/src/widgets/authenticator_banner.dart
```diff
@@ -21,46 +21,87 @@ import 'package:flutter/material.dart';
...
       children: [
         Icon(
           type.icon,
           color: foregroundColor,
         ),
         const SizedBox(width: 16),
-        Expanded(child: content),
+        Expanded(
+          child: Text(message.trimRight(),
+              textAlign: TextAlign.center,
```
Jordan-Nelson, line 102: The text is aligned off center here due to the icon and SizedBox on the left.

Here is the current state.
```
|-icon-|-SizedBox-|--------Content--------|
```


You could either remove the SizedBox to center it within the space not occupied by the icon, or you could center it within the full view (ignoring the icon).

However, I'm wondering if we actually want to left align this content for both SnackBar and MaterialBanner.

The Material Design specs have this content left aligned for both components. There isn't any guidance specifically NOT to center it, so I don't think we have to follow this if we have a good reason to deviate. I can see the argument that on large screens having the content in the center feels more natural, but I also think that having the content so far away from the icon is odd. My eye first jumps to the content, then to the icon, and then finally to the close button. With the icon and content left aligned my eye jumps right to the icon/content, and then to the close button.
- https://material.io/components/snackbars#specs
- https://material.io/components/banners#specs (see Mobile & Desktop sub headers)

packages/amplify_authenticator/lib/src/widgets/authenticator_banner.dart
Jordan-Nelson: Turns out `#d32f2f` is `Colors.red[700]` fyi. [Here](https://github.com/flutter/flutter/blob/dbadee003f344e0940308364f1080d82aa8b9b36/packages/flutter/lib/src/material/color_scheme.dart#L570) is where that default error color is set. I would use that.

packages/amplify_authenticator/lib/src/widgets/authenticator_banner.dart
```diff
@@ -79,4 +120,15 @@ extension on StatusType {
         return Icons.error;
     }
   }
+
+  ColorSwatch<int> get color {
+    switch (this) {
+      case StatusType.info:
+        return Colors.blue;
```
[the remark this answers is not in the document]
Equartey: Resolved offline.

packages/amplify_authenticator/lib/src/widgets/authenticator_banner.dart
Equartey: Resolved offline.

packages/amplify_authenticator/lib/src/widgets/authenticator_banner.dart
Equartey: Resolved offline.

packages/amplify_authenticator/lib/src/widgets/authenticator_banner.dart
```diff
@@ -21,60 +21,124 @@ import 'package:flutter/material.dart';
...
-        Icon(
-          type.icon,
-          color: foregroundColor,
+        Icon(type.icon, color: fallbackIconColor),
+        const SizedBox(
+          width: 16,
         ),
-        const SizedBox(width: 16),
-        Expanded(child: content),
+        Expanded(
```
Jordan-Nelson, line 90: *nit:* I think this expanded isn't needed now what this is not centered.

packages/amplify_authenticator/lib/src/widgets/authenticator_banner.dart
```diff
@@ -21,60 +21,124 @@ import 'package:flutter/material.dart';
...
+          child: Text(
+            message.trimRight(),
+            textAlign: TextAlign.left,
```
Jordan-Nelson, line 93: *nit:* this defaults to `TextAlign.start` which I think in this case has the same effect. You can probably just remove this.

packages/amplify_authenticator/lib/src/widgets/authenticator_banner.dart
Jordan-Nelson: This comment applies to the banner as well.

packages/amplify_authenticator/lib/src/widgets/authenticator_banner.dart
Jordan-Nelson: The comment applies the the banner as well.

Jordan-Nelson (approved):
Some nit comments, but looks good!

packages/amplify_authenticator/lib/src/widgets/authenticator_banner.dart
```diff
@@ -21,60 +21,124 @@ import 'package:flutter/material.dart';
...
-      bottom: 4.0,
+    backgroundColor: colorsChoices.background,
+    leading: Icon(type.icon, color: colorsChoices.foreground),
+    content: Center(
+      child: Row(
+        crossAxisAlignment: CrossAxisAlignment.center,
+        children: [
+          Expanded(
+            child: Text(
+              message.trimRight(),
```
dnys1, line 42: Doesn't need to be changed but I find this a little confusing over just `.trim()` or instead of trimming the text at the source. It doesn't really matter here, but just a general comment for the future.

packages/amplify_authenticator/lib/src/widgets/authenticator_banner.dart
dnys1: True if not centered, but it looks like it is being centered

dnys1 (approved):

packages/amplify_authenticator/lib/src/widgets/authenticator_banner.dart
Equartey: The Expanded is required here, it allows text overflow on multi-line messages.

dnys1: Good catch, you're right

PR #2004, opened 2022-08-04
feat(auth): migrate legacy credentials on Android

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Implement Android credential migration from native SDK


packages/auth/amplify_auth_cognito/lib/src/credentials/legacy_credential_provider_android.dart
```diff
@@ -0,0 +1,58 @@
...
+    CognitoOAuthConfig? hostedUiConfig,
+  }) async {
+    final bridge = _stateMachine.expect<NativeAuthBridge>();
+    final userPoolConfig = _stateMachine.expect<CognitoUserPoolConfig>();
+    final identityPoolConfig =
+        _stateMachine.expect<CognitoIdentityPoolConfig>();
+    final legacyCredentials = await bridge.getLegacyCredentials(
+      identityPoolConfig.poolId,
+      userPoolConfig.appClientId,
+    );
```
dnys1, line 45: Should these be required params? Seems weird to fetch em when they're being passed in

```suggestion
    final legacyCredentials = await bridge.getLegacyCredentials(
      identityPoolConfig!.poolId,
      userPoolConfig!.appClientId,
    );
```

packages/auth/amplify_auth_cognito/lib/src/credentials/legacy_ios_cognito_keys.dart
```diff
@@ -12,8 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/// Keys used by the legacy iOS SDK
+/// Reference: https://github.com/aws-amplify/aws-sdk-ios
```
dnys1, line 16: Is there a file we can point to?

packages/auth/amplify_auth_cognito_android/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
```diff
@@ -165,6 +167,57 @@ open class AuthCognito :
         return applicationContext!!.packageName
     }
 
+    /**
+     * Fetches the legacy credentials set by the Android SDK
+     */
+    override fun getLegacyCredentials(identityPoolId: String, appClientId: String, result: NativeAuthPluginBindings.Result<NativeAuthPluginBindings.LegacyCredentialStoreData>) {
+        val legacyUserPoolStore = getLegacyKeyValueStore("CognitoIdentityProviderCache")
```
dnys1, line 174: I liked the idea of having a link for reference to the iOS code. Can we do the same here just so we know where the values came from?

packages/auth/amplify_auth_cognito_android/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt
```diff
@@ -165,6 +167,57 @@ open class AuthCognito :
...
+            .setIdToken(idToken)
+            .build()
+        result.success(data)
+    }
+
+    /**
+     * Clears the legacy credentials set by the Android SDK
+     */
+    override fun clearLegacyCredentials(result: NativeAuthPluginBindings.Result<Void>) {
+        val legacyUserPoolStore = getLegacyKeyValueStore("CognitoIdentityProviderCache")
```
dnys1, line 202: These could probably be lazy vals on the class

packages/auth/amplify_auth_cognito/lib/src/credentials/legacy_credential_provider_impl.dart
```diff
@@ -0,0 +1,70 @@
...
+/// The implementation of [LegacyCredentialProvider] for migrating
+/// credentials from the legacy iOS and Android SDKs.
+/// {@endtemplate}
+class LegacyCredentialProviderImpl implements LegacyCredentialProvider {
+  /// {@macro amplify_auth_cognito.legacy_credential_provider_impl}
+  LegacyCredentialProviderImpl(CognitoAuthStateMachine stateMachine)
+      : _stateMachine = stateMachine;
+  final CognitoAuthStateMachine _stateMachine;
+
+  late final LegacyCredentialProvider? _instance = () {
```
dnys1, line 32: Since this only gets instantiated in `auth_plugin_impl` in a method where we've already checked if we're on iOS/Android, I think you can safely remove all the platform-related null checks since it should be impossible to get here otherwise.

dnys1 (approved):
Missing at least one license header.

Looks good, though!

packages/auth/amplify_auth_cognito/lib/src/credentials/legacy_credential_provider_android.dart
Jordan-Nelson: Yeah I don't know why I fetched them when they were already passed in. Maybe I just missed this when refactoring. Updated.

haverchuck (approved):

PR #2022, opened 2022-08-09
chore(aft): add self-update command

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Add a `self-update` command to update aft. Inspired by [yarn self-update](https://classic.yarnpkg.com/lang/en/docs/cli/self-update/) or `flutter upgrade`

This just runs `dart pub global activate -spath <root>/packages/aft`.

Motivation: I always forget what the command it to activate a package from a path and have to look it up each time I switch to a branch with new aft changes. This makes it a little easier.

### DIFF
--- a/packages/aft/lib/src/commands/self_update_command.dart
+++ b/packages/aft/lib/src/commands/self_update_command.dart
@@ -1,0 +1,41 @@
+// Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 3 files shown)


dnys1 (approved):
Cool!

packages/aft/lib/src/commands/self_update_command.dart
```diff
@@ -0,0 +1,41 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'package:aft/aft.dart';
+
+/// A command to update the aft package.
+class SelfUpdateCommand extends AmplifyCommand {
```
dnys1, line 18: Looking at this again - since the version committed in `next` will always be the "latest"/stable should we expose that as an option or as the default?

Issue #2039, opened 2022-08-16
`observe` and `observeQuery` do not support filtering by IDs of nested models


Jordan-Nelson:
### Description

`query()` supports filtering by the ID of a nested model. For example, the query below will return Comments that have a Post with an ID of `post-id-123`:

```dart
Amplify.DataStore.query(
  Comment.classType,
  where: Comment.POST.eq('post-id-123')
);
```

This same predicate does not work for `observe()` or `observeQuery()`. No data will be returned when this predicate is used.

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [ ] Auth
- [ ] Authenticator
- [X] DataStore
- [ ] Storage

### Steps to Reproduce

1. Set up and app with a typical Blog/Post/Comment schema.
2. Run the following query:
```dart
Amplify.DataStore.observeQuery(
  Comment.classType,
  where: Comment.POST.eq('post-id-123')
);
```
3. Save new models that meet the predicate, and notice new snapshots are not emitted.

### Platforms

- [ ] iOS
- [ ] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Environment

```bash
N/A
```

### Dependencies

```bash
N/A
```

### Device

N/A

### OS

N/A

### Deployment Method

Amplify CLI

### CLI Version

N/A

### Additional Context

Query predicates for `observe` and `observeQuery` are applied in-memory (in Dart), where as Query predicates for `query` are applied in SQL. The in-memory logic needs to be updated to account for nested querying.

### Amplify Config

N/A

Jordan-Nelson:
A fix for this was included in v0.6.9. 

Note: If you are using CPK, there is a change required in code generation to support this that will come in a later version of the CLI.

PR #2043, opened 2022-08-16
fix(datastore): support nested predicates for observe and observeQuery

Jordan-Nelson:
*Issue #, if available:* https://github.com/aws-amplify/amplify-flutter/issues/2039

*Description of changes:*
- Support nested predicates for observe and observeQuery by mapping the value being compared to the ID if certain conditions are met.

*Note:*

This allows query predicates like `Post.BLOG.eq('some-id')` to work with observe and observeQuery. It does not allow for predicates like `Post.BLOG.eq(BlogModelIdentifier(id: 'some-id'))`. The latter will require code gen changes. This is because, without reflection, there is no way to take the string name of the model ('blog') and translate that into the appropriate model identifier.

Adding a `Map<String, Object?> toMap()` method that serializes a model one level deep would resolve this. For example:

```dart
  Map<String, dynamic> toMap() => {
        'id': id,
        'title': _title,
        'blog': _blog,
        'comments': _comments,
      };
```


packages/amplify_core/lib/src/types/query/query_field_operators.dart
```diff
@@ -79,6 +79,22 @@ abstract class QueryFieldOperator<T> {
     // TODO sanitize other types appropriately
     return value;
   }
+
+  /// Gets the correct field to use as a comparison for nested models.
+  ///
+  /// TODO: Remove when codegen supports a `toMap()` method.
```
Jordan-Nelson, line 85: This would be replaced or updated if the codegen changes in the issue description were added.

packages/amplify_core/lib/src/types/query/query_field_operators.dart
```diff
@@ -79,6 +79,22 @@ abstract class QueryFieldOperator<T> {
...
+  Object? _getOtherField(Object? other, Object? value) {
+    // if `other` is type Map and `value` is type String, the
+    // only valid query predicate is for a nested model by ID.
+    //
+    // For example: `Comment.POST.eq('post-id-123')`
+    //
+    // Return the ID to use as the comparison field.
+    if (other is Map && value is String) {
+      return other['id'];
```
dnys1, line 94: Does this work with CPK?

Jordan-Nelson: No. It already does not work with CPK though. We need code changes to support this with CPK. We have no way to get the `ModelIdentifier` of the nested model without reflection in Dart, or a codegen change to fill the gap (i.e. a `toMap()` method). This is because the query predicate just has the string field name (`"blog"` for a predicate of `Post.BLOG.eq("blog-id")`).

Those codegen changes will be non-breaking, but consuming them will be breaking (customers will need to regenerate models) unfortunately. 

I think it is debatable if it is worth moving forward with this. The benefit would be to get this fixed for those not using CPK before we 1) get the codgen change in, and 2) do out next breaking change.

If we can do 1 & 2 somewhat quickly, we could just opt to fix this the right way from the start.

Jordan-Nelson: @HuiSF - Curious to get your thoughts on this. I don't think the codegen changes will be too complicated, but I haven't had a chance to look at them yet.

HuiSF: Support of creating predicate on nested model is a bit wonky :/ The current implementation is a bit hacky already to adapt to different patterns between amplify-ios and amplify-android. I think it makes sense to make codegen necessary codegen change to make this working.

Re: the breaking change. We've added a deprecation warning on `Model.getId()` method, that after upgrading amplify-flutter library 0.6.6, this warning gets shown when build App. Hopefully this helps making the migration smoother.

Jordan-Nelson: Okay. I think that is fair. I can look into the codegen changes. Will mark this PR as draft until then.

packages/amplify_core/lib/src/types/query/query_field_operators.dart
```diff
@@ -156,8 +181,7 @@ class LessOrEqualQueryOperator<T extends Comparable>
     if (other == null) {
       return false;
     }
-    dynamic serializedValue = serializeDynamicValue(value);
```
Jordan-Nelson, line 159: Note: The addition of `toMap()` means that `other` will no longer be serialized, so `value` no longer needs to be serialized.

packages/amplify_core/lib/src/types/query/query_field_operators.dart
```diff
@@ -99,6 +109,31 @@ class EqualQueryOperator<T> extends QueryFieldOperatorSingleValue<T> {
 
   @override
   bool evaluate(T? other) {
+    // if `other` is Model, the query predicate is on a
+    // nested model, such as `Post.BLOG.eq(myBlog.modelIdentifier))`,
+    // and the value should be compared against the model ID.
+    if (other is Model) {
+      // TODO: Update to `return value == other.modelIdentifier`
```
Jordan-Nelson, line 116:
```suggestion
      // TODO(Jordan-Nelson): Update to `return value == other.modelIdentifier`
```

packages/amplify_core/lib/src/types/query/query_field_operators.dart
```diff
@@ -126,8 +168,33 @@ class NotEqualQueryOperator<T> extends QueryFieldOperatorSingleValue<T> {
 
   @override
   bool evaluate(T? other) {
+    // if `other` is Model, the query predicate is on a
+    // nested model, such as `Post.BLOG.eq(myBlog.modelIdentifier))`,
+    // and the value should be compared against the model ID.
+    if (other is Model) {
+      // TODO: Update to `return value != other.modelIdentifier`
```
Jordan-Nelson, line 175:
```suggestion
      // TODO(Jordan-Nelson): Update to `return value != other.modelIdentifier`
```

dnys1 (approved):
LGTM!

packages/amplify_core/lib/src/types/query/query_field_operators.dart
```diff
@@ -40,8 +40,18 @@ abstract class QueryFieldOperator<T> {
...
+  /// Evaluates this QueryFieldOperator against [other],
+  /// where [other] is a field on the model.
   bool evaluate(T? other);
 
+  /// Similar to `evaluate`, except `other` should be the
+  /// *serialized* value of a field on the model.
+  ///
+  /// This should be used to support comparisons with models
+  /// that were generated prior to `toMap()` being added.
+  @Deprecated('To be removed in next major release')
```
HuiSF, line 52: Should mention the action item inside the `@Deprecated` instead of a comment here so developer will know what to do seeing the deprecation message while building App?

HuiSF (approved):
Thanks.
Left a comment.

packages/amplify_core/lib/src/types/query/query_field_operators.dart
Jordan-Nelson: Updated.

I don't expect customers to see this since this is used internally, although it is technically part of the public API. I could make it internal, but then it makes it a pain to use outside of amplify_core. Updating the deprecation message in case customers stumble upon this API seems reasonable 👍

dnys1 (approved):

HuiSF (approved):

PR #2044, opened 2022-08-17
feat(secure_storage): handle package uninstall/re-install on Linux and Windows

Jordan-Nelson:
*Issue #, if available:*

#### General / Utils
- Add `FileKeyValueStore` for file simple key/value storage in the file system
- Add `path_provider` in as dependency in `amplify_secure_storage` to get the application support directory
- Tests for App re-install and utils

#### Linux
- Add required `gio` bindings to get app ID on linux
- Add App ID as a namespace for all keys since Linux does not namespace secrets by app (unlike iOS, Android, MacOS)
- Add `accessGroup` to linux config as an optional override to App ID to allow for sharing of data across apps
- Add initialization logic in flutter to clear all data for a given scope upon first initialization of that scope, set flags in file storage to keep track of scopes using `FileKeyValueStore`

#### Windows
- Pass application support dir from flutter package to dart only package
- Add config option for customer to optionally override the storage location
- Add Win32 Data Protection API bindings 
- Use the Data Protection APIs to encrypt/decrypt data and `FileKeyValueStore` to store the data

### Motivation:
If a user uninstalls and re-installs an app installed via snap or the Windows app store, the expectation would be that data from their previous install would not persist. 

### Implementation:
- Linux: Data is cleared immediately after the app is launched for the first time (after a re-install). Flags are set in application file storage to track the apps first launch post re-install.
- Windows: Encrypted data is now stored in application file storage. This removes the need to clear the data as it is removed automatically on app uninstall.


packages/secure_storage/amplify_secure_storage_dart/lib/src/ffi/utils/linux_utils.dart
```diff
@@ -0,0 +1,56 @@
...
+      xdg.dataHome.path,
+      appId,
+    ),
+  );
+  if (directory.existsSync()) {
+    return directory.path;
+  }
+
+  await directory.create(recursive: true);
+  return directory.path;
```
dnys1, line 45: Generally only want to use sync APIs in like scripts or CLI tools where it's okay to block. Also DRY-d it up a bit.

```suggestion
  if (!await directory.exists()) {
    await directory.create(recursive: true);
  }
  return directory.path;
}
```

packages/secure_storage/amplify_secure_storage_dart/lib/src/utils/file_key_value_store.dart
```diff
@@ -0,0 +1,98 @@
...
+import 'package:meta/meta.dart';
+import 'package:path/path.dart' as path;
+
+/// {@template amplify_secure_storage_dart.file_key_value_store}
+/// A Key-Value Store that stores data in the provided file.
+/// {@endtemplate}
+@internal
+class FileKeyValueStore {
+  /// {@macro amplify_secure_storage_dart.file_key_value_store}
+  FileKeyValueStore({required this.directory, required this.fileName});
```
dnys1, line 29: I would do something like:

```dart
FileKeyValueStore({
  required this.directory, 
  required this.fileName,
  @visibleForTesting this.fs = const local_file.LocalFileSystem(),
});

@visibleForTesting
final pkg_file.FileSystem fs;
```

packages/secure_storage/amplify_secure_storage_dart/lib/src/utils/file_key_value_store.dart
```diff
@@ -0,0 +1,98 @@
...
+    required String key,
+  }) async {
+    final data = await readAll();
+    return data.containsKey(key);
+  }
+
+  /// Reads all the key-value pairs from storage.
+  Future<Map<String, Object>> readAll() async {
+    if (file.existsSync()) {
+      final stringMap = file.readAsStringSync();
```
dnys1, line 88:
```suggestion
    if (await file.exists()) {
      final stringMap = await file.readAsString();
```

packages/secure_storage/amplify_secure_storage_dart/lib/src/utils/file_key_value_store.dart
```diff
@@ -0,0 +1,98 @@
...
+
+  /// Overwrites the existing data.
+  Future<void> writeAll(
+    Map<String, Object> data,
+  ) async {
+    if (!file.existsSync()) {
+      file.createSync(recursive: true);
+    }
+    final stringMap = json.encode(data);
+    file.writeAsStringSync(stringMap);
```
dnys1, line 66:
```suggestion
    if (!await file.exists()) {
      await file.create(recursive: true);
    }
    final stringMap = json.encode(data);
    await file.writeAsString(stringMap);
```

packages/secure_storage/amplify_secure_storage_test/test/linux_test.dart
```diff
@@ -0,0 +1,57 @@
...
+      expect(value1, isNotNull);
+
+      // clear the scope from file storage, simulating an app uninstall on linux
+      final appId = await getApplicationId();
+      final appDirectory = await getApplicationSupportPath(appId);
+      final fileStore = FileKeyValueStore(
+        directory: appDirectory,
+        fileName: 'amplify_secure_storage_scopes.json',
+      );
+      fileStore.writeAll({});
```
dnys1, line 47: Should this use the in-mem fs?

packages/secure_storage/amplify_secure_storage_dart/lib/src/ffi/utils/linux_utils.dart
dnys1: Oh I forgot this will run in an isolate... still, probably just worth changing them to async anyway

packages/secure_storage/amplify_secure_storage_dart/lib/src/utils/file_key_value_store.dart
```diff
@@ -0,0 +1,98 @@
...
+
+  /// The directory of the file.
+  final String directory;
+
+  /// The name of the file.
+  ///
+  /// The file will be created if it does not yet exist.
+  final String fileName;
+
+  late final File file = fs.file(
```
dnys1, line 39: If `accessGroup` is set, this file could potentially be written to from multiple processes, right? In that case, I don't think this would work. Maybe accessGroup becomes the name of a folder and each appId gets an individual file. If `accessGroup` is a Linux group name, then the folder's permissions could be configured around that.

dnys1: Oh nvm this is only used for the one flag. The appID is still used for everything else - my b

Jordan-Nelson: I think you bring up a good point. We don't want to clear keys if an access group is set. With the current impl, if two apps use the same access group, the second app will clear all the keys set by the first app.

I will update this to skip initialization if there is an access group set. I think that is all we can do since App N will have no way to know if another app set those keys, or if they were from a previous install.

packages/secure_storage/amplify_secure_storage_dart/lib/src/utils/file_key_value_store.dart
Jordan-Nelson: Nice. Will update this and add some tests for this 👍

packages/secure_storage/amplify_secure_storage_test/test/linux_test.dart
Jordan-Nelson: It could be updated to use in-mem. `AmplifySecureStorageDart` would also have to be updated to use in-mem for these tests in that case.

This is more of an integration tests since it isn't mocking libsecret / secret service, so I think it is fine as is.

packages/secure_storage/amplify_secure_storage_dart/lib/src/ffi/utils/linux_utils.dart
Jordan-Nelson: Yeah I will update these to use the async APIs. I based this off of shared preferences 😅 . Not sure why they use the sync APIs since the APIs they expose are all async.

packages/secure_storage/amplify_secure_storage_dart/lib/src/utils/file_key_value_store.dart
Jordan-Nelson: I have updated this in the linux impl

packages/secure_storage/amplify_secure_storage_test/lib/amplify_secure_storage_test.dart
```diff
@@ -148,31 +155,26 @@ void runStandardTests(
     });
   });
 
-  group('write', () {
```
Jordan-Nelson, line 151: Note: These tests were duplicates.

haverchuck (approved):

packages/secure_storage/amplify_secure_storage/lib/src/amplify_secure_storage.vm.dart
```diff
@@ -27,25 +33,66 @@ class AmplifySecureStorage extends AmplifySecureStorageInterface {
...
+  ///
+  /// Checks for an initialization flag in file storage.
+  /// If the flag is not present storage will be cleared
+  /// and then the flag will be set.
+  ///
+  /// Intended to clear storage after an app uninstall & re-install.
+  Future<void> _initializeScope(String? accessGroup) async {
+    // if accessGroup is set, do not clear data on initialization
+    // since the data can be shared across applications.
+    if (accessGroup != null) return;
```
dnys1, line 88: Can we reformat this to not call `_initalizeScope` if accessGroup is set since the only reason this method has a parameter is to check if it's not null which is kind of confusing.

packages/secure_storage/amplify_secure_storage_dart/lib/src/platforms/amplify_secure_storage_windows.dart
```diff
@@ -11,119 +11,69 @@
...
-        CRED_FLAG_DEFAULT,
-        credential,
+  Future<String?> read({required String key}) async {
+    final rawData = await keyValueStore.readKey(key: key);
+    if (rawData == null) {
+      return null;
+    }
+    final Uint8List encrypted;
+    try {
+      encrypted = Uint8List.fromList(List<int>.from(rawData as List));
```
dnys1, line 63: I think this makes two copies of the list (List.from, then Uint8List.fromList). I would do `Uint8List.fromList((rawData as List).cast());` to avoid that. Is the only failure because of the cast? It's probably more reasonable to do an `is` check if that's the case.

packages/secure_storage/amplify_secure_storage_test/lib/amplify_secure_storage_test.dart
```diff
@@ -148,31 +155,26 @@ void runStandardTests(
...
-
-      // write new value to the existing key
-      await storage.write(key: key1, value: 'test_update');
-
-      // confirm new value was written
-      final value2 = await storage.read(key: key1);
-      expect(value2, 'test_update');
-    });
+  group('read/write/delete can handle key value pairs of varying length', () {
+    for (var entry in keyValuePairs.entries) {
```
dnys1, line 159:
```suggestion
    for (final entry in keyValuePairs.entries) {
```

Don't necessarily need to change since I don't care as much about the style of test code but I thought this was a lint

dnys1 (approved):
Minor comments. Looks good, though. Thanks!

packages/secure_storage/amplify_secure_storage_dart/lib/src/platforms/amplify_secure_storage_windows.dart
```diff
@@ -11,119 +11,69 @@
...
-  /// Reference: https://docs.microsoft.com/en-us/windows/win32/api/wincred/ns-wincred-credentiala
-  String get _username => config.windowsOptions.targetNamePrefix != null
-      ? config.windowsOptions.targetNamePrefix!
-      : config.defaultNamespace;
+  late final FileKeyValueStore keyValueStore = () {
+    final directory = config.windowsOptions.storagePath ?? applicationDirectory;
+    if (directory != null) {
+      return FileKeyValueStore(fileName: _fileName, path: directory);
+    }
+    return FileKeyValueStore(
```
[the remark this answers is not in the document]
Jordan-Nelson: Yeah, tests/examples. Could also be a reasonable fallback if/when we support non-flutter use cases for amplify.

packages/secure_storage/amplify_secure_storage_dart/lib/src/interfaces/amplify_secure_storage_interface.dart
```diff
@@ -24,11 +25,29 @@ abstract class AmplifySecureStorageInterface extends SecureStorageInterface
...
 
   /// Configuration options for Secure Storage.
   final AmplifySecureStorageConfig config;
 
+  /// Path to a directory where the application should
+  /// place application support files.
+  ///
+  /// Used to store encrypted data on Windows. If no value
+  /// is provided, data will be stored in memory on Windows.
+  final String? applicationDirectory;
```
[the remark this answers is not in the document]
Jordan-Nelson: Yeah. This was unnecessary. I had added `storagePath` to `WindowsSecureStorageOptions` already. I was passing both to the _dart package, with `storagePath` taking priority over `applicationDirectory`. This made sense when I was using `applicationDirectory` in the _dart package on linux as well to read/write scope names, but with that happening on the flutter side, passing two values for the same thing didn't make sense.

Updated 👍

packages/secure_storage/amplify_secure_storage_dart/lib/src/platforms/amplify_secure_storage_windows.dart
Jordan-Nelson: Updated to use .cast() over List.from.

I kept the try/catch block. `rawData` could not be a list, or it could be a List that contains data other than integers. Either of those would be problematic. I could do an `is` check to confirm it is a list, but I would still need to cast to get it to List<int> rather than List<dynamic>.

PR #2047, opened 2022-08-18
feat: custom signIn method without build in form;

andr3i4o:
*Added a way to custom signIn with a username and password without the build in form fields*
*The business case scenario: having a server that generates the Cognito user and returns the username and password to the application which then will use it to login*

### DIFF
--- a/packages/amplify_authenticator/lib/src/state/authenticator_state.dart
+++ b/packages/amplify_authenticator/lib/src/state/authenticator_state.dart
@@ -381,6 +381,21 @@ class AuthenticatorState extends ChangeNotifier {
     _setIsBusy(false);
   }
 
+  /// Sign in with [username], and [password] without a build in form
+  Future<void> customSignIn(String userName, String password) async {
+    if(userName.isEmpty || password.isEmpty) {
+      return;
+    }
+    _setIsBusy(true);
+    AuthSignInData signIn = AuthUsernamePasswordSignInData(
+      username: userName,
+      password: password,
+    );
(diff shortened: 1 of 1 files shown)


Jordan-Nelson:
@andr3i4o - Thanks for opening the PR!

Ideally we want to allow all successful calls to `Amplify.Auth.signIn()` to take the user to the authenticated state, as this is a more generic solution that could be used outside of `authenticatorBuilder`. I will look to see if this is something we can support in the short term, and if not - we can discuss an alternate solution such as a new API like you have included in your PR.

Jordan-Nelson:
Hello @andr3i4o - I took a look at this last week, and determined that we would be able to implement this in a way that allows for Amplify.Auth.signIn() calls to be handled. There was a somewhat nuanced edge case that had to be handled though. Because of this, I ended up opening PRs to address this (linked below).

I will update the original issue once these have been released. Let me know if you have any questions about the implementations:
- PR to the `next` branch (tracks the latest dev-preview release): https://github.com/aws-amplify/amplify-flutter/pull/2049
- PR to the `main` branch (tracks the latest stable release): https://github.com/aws-amplify/amplify-flutter/pull/2053

I think we can probably close this PR out. I really appreciate you taking the time to open a PR to address this though.

andr3i4o:
Hey @Jordan-Nelson, Sorry for the delayed answer. That's great news regarding the implementation you are preparing. I know my PR was more of a quick fix than a general solution, but thanks for the review and such quick reaction.
I will close this one.

Have a great day.

PR #2064, opened 2022-08-24
chore(api): Fix mocks

dnys1:
Removes `mockito` generated mocks and replaces them with simple `mocktail` mocks to prevent generation errors for types like enhanced enums.


HuiSF (approved):

packages/api/amplify_api/pubspec.yaml
```diff
@@ -29,7 +29,7 @@ dev_dependencies:
   build_runner: ^2.0.0
   flutter_test:
     sdk: flutter
-  mockito: ^5.0.0
+  mocktail: ^0.3.0
```
Equartey, line 32: I'm curious why you'd choose one over the other. Does mockito not support enhanced enums?

dnys1: Mockito is maintained by the Dart team - it's based on code generation and is pretty hairy to use in certain situations. Mocktail is maintained by VGV, is newer, and doesn't involve codegen which makes it play nicely with language changes. It's usually the better choice nowadays.

Equartey (approved):

PR #2067, opened 2022-08-26
chore(api): remove method channel impl and refactor decoder

ragingsquirrel3:
This PR does 3 things:
1) deletes unused method channel implementation/tests for API category both on flutter/native sides
2) fixes native `addPlugin` code via more pigeon generation to correctly setup API auth providers in android/ios so that OIDC/lambda auth modes work in Datastore (tested manually by separate modifications to Datastore example app/backend with Lambda auth mode).
3) refactors graphql decoder in Dart to take the map directly from JSON-decoded response. That way it reduces re-encoding/decoding from JSON to string back and forth a few times.


packages/api/amplify_api_ios/ios/Classes/SwiftAmplifyApiPlugin.swift
```diff
@@ -21,63 +21,25 @@ import amplify_flutter_ios
...
-    init(
-        bridge: ApiBridge = ApiBridge(),
-        graphQLSubscriptionsStreamHandler: GraphQLSubscriptionsStreamHandler = GraphQLSubscriptionsStreamHandler()
-    ) {
-        self.bridge = bridge
-        self.graphQLSubscriptionsStreamHandler = graphQLSubscriptionsStreamHandler
-    }
-
     public static func register(with registrar: FlutterPluginRegistrar) {
         methodChannel = FlutterMethodChannel(
```
ragingsquirrel3, line 27: callout: I wasn't totally sure if this part of method channel was still needed? I was thinking we still needed it for oidc/lambda communication over the channel related to getting token.

packages/api/amplify_api_android/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/AmplifyApi.kt
```diff
@@ -54,105 +39,19 @@ class AmplifyApi : FlutterPlugin, MethodCallHandler, NativeApiPluginBindings.Nat
...
-    ) {
-        graphqlSubscriptionStreamHandler = eventHandler
-        this.dispatcher = dispatcher
-    }
-
-    private val handler = Handler(Looper.getMainLooper())
 
     override fun onAttachedToEngine(@NonNull flutterPluginBinding: FlutterPlugin.FlutterPluginBinding) {
-        graphqlSubscriptionStreamHandler = graphqlSubscriptionStreamHandler ?: GraphQLSubscriptionStreamHandler()
         channel = MethodChannel(flutterPluginBinding.binaryMessenger, "com.amazonaws.amplify/api")
```
ragingsquirrel3, line 45: same question here as ios: Is this part of channel for auth token in oidc/lambda?

packages/api/amplify_api/lib/src/auth_token.dart
```diff
@@ -1,30 +0,0 @@
...
- * permissions and limitations under the License.
- */
-
-part of 'method_channel_api.dart';
-
-extension on APIAuthProvider {
-  /// Retrieves the latest auth token for [type].
-  ///
-  /// Any [Exception] is caught and treated as a `null` token.
-  Future<String?> get authToken async {
```
dnys1, line 22: This is still relevant since this is just the business logic to retrieve the token

dnys1 (approved):

packages/api/amplify_api/lib/src/graphql/graphql_response_decoder.dart
```diff
@@ -31,19 +31,23 @@ class GraphQLResponseDecoder {
...
-  GraphQLResponse<T> decode<T>(
-      {required GraphQLRequest request,
-      String? data,
-      List<GraphQLResponseError>? errors}) {
-    if (data == null) {
-      return GraphQLResponse(data: null, errors: errors);
-    }
+  GraphQLResponse<T> decode<T>({
+    required GraphQLRequest request,
+    required Map<String, dynamic>? response,
```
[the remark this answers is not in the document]
ragingsquirrel3: good call, will change

packages/api/amplify_api/lib/src/graphql/graphql_response_decoder.dart
```diff
@@ -31,19 +31,23 @@ class GraphQLResponseDecoder {
...
+  }) {
+    final errors = _deserializeGraphQLResponseErrors(response);
+    final data = response?['data'];
+
     // If no modelType fallback to default (likely String).
     final modelType = request.modelType;
     if (modelType == null) {
       if (T == String || T == dynamic) {
+        // Preserve `null`. json.encode(null) returns "null" not `null`
+        final encodedData = data != null ? json.encode(data) : null;
```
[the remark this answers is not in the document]
ragingsquirrel3: This is to handle case where response is like `{"errors": [...] }`. So response is non-null but data is null and should be present in response as such.

packages/api/amplify_api/lib/src/auth_token.dart
ragingsquirrel3: I ended up keeping this file deleted and just called `getLatestAuthToken` but let me know if I'm missing something and this is still needed.

packages/api/amplify_api/lib/src/api_plugin_impl.dart
```diff
@@ -240,7 +243,7 @@ class AmplifyAPIDart extends AmplifyAPI {
     return AWSStreamedHttpRequest.delete(
       uri,
       body: body,
-      headers: addContentTypeToHeaders(headers, body),
```
ragingsquirrel3, line 243: The new HTTP client PR moves this logic to AWS common so not needed here (happens without this code).

packages/api/amplify_api/lib/src/graphql/graphql_response_decoder.dart
```diff
@@ -31,19 +31,23 @@ class GraphQLResponseDecoder {
...
+  }) {
+    final errors = _deserializeGraphQLResponseErrors(response);
+    final data = response['data'];
+
     // If no modelType fallback to default (likely String).
     final modelType = request.modelType;
     if (modelType == null) {
       if (T == String || T == dynamic) {
+        // Preserve `null`. json.encode(null) returns "null" not `null`
+        final encodedData = data != null ? json.encode(data) : null;
```
dnys1, line 46: Wondering if it would be better to return a Map instead of a String by default since this is how most other libraries do it and a string is not really useful until you decode it anyway

dnys1 (approved):
LGTM

packages/api/amplify_api/lib/src/graphql/graphql_response_decoder.dart
ragingsquirrel3: Interesting idea. Agree string not really useful. The response data is always decoded to the `<T>` of the request so it doesn't seem like it would work when explicit `<String>` like is in lots of our docs. Maybe we could do that if `<dynamic>` or `<Map>`?

packages/api/amplify_api/lib/src/api_plugin_impl.dart
```diff
@@ -320,8 +323,36 @@ class AmplifyAPIDart extends AmplifyAPI {
...
+      throw PlatformException(
+        code: 'NO_PROVIDER',
+        message: 'No provider found for $authProvider',
+      );
+    }
+    return authProvider.getLatestAuthToken();
+  }
+
+  @override
+  String get runtimeTypeName => '_NativeAmplifyAApi';
```
HuiSF, line 357: nit: single "A"?

HuiSF (approved):

packages/api/amplify_api/lib/src/graphql/graphql_response_decoder.dart
dnys1: I'm thinking long-term we'll want to get rid of `String` and `dynamic` support. Not sure we're quite ready to do that - maybe as part of the modelgen changes

PR #2070, opened 2022-08-26
chore(api): increase integration test coverage

ragingsquirrel3:
This PR adds integration test coverage to dart API category implementation by adding:
* test for multiauth backend for GraphQL so that API key, IAM and Cognito User Pools auth modes all tested (and multiauth)
* tests for REST
* related utils
* provision script/example app changed by checking in amplify dir so that it uses `amplify init` with that directory instead of full headless script


packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
```diff
@@ -0,0 +1,83 @@
...
+      await signOutTestUser();
+    });
+
+    group('queries', () {
+      testWidgets('should query authorized model', (WidgetTester tester) async {
+        final req = authorizeRequestForApiKey(
+          ModelQueries.list<Blog>(Blog.classType),
+          apiKey,
+        );
+        final res = await Amplify.API.query(request: req).response;
```
dnys1, line 52: Does this throw on GraphQL error? If not, should we check that `res.errors` is empty?

packages/api/amplify_api/example/integration_test/graphql/iam_test.dart
```diff
@@ -0,0 +1,334 @@
...
+    tearDownAll(() async {
+      await deleteTestModels();
+      if (!useExistingTestUser) {
+        await deleteTestUser();
+      }
+    });
+
+    group('queries (authorized user)', () {
+      setUpAll(() async {
+        await signInTestUser();
```
dnys1, line 47: This will be called twice since `setUpAll`s stack when nesting groups - true for all setUp/tearDown methods

packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
```diff
@@ -0,0 +1,83 @@
...
+        final data = res.data;
+        expect(data?.items.length, greaterThanOrEqualTo(0));
```
dnys1, line 54: Wouldn't creating a blog and testing >=1 make more sense?

packages/api/amplify_api/example/integration_test/rest_test.dart
```diff
@@ -0,0 +1,107 @@
...
+      });
+
+      testWidgets('should send GET request', (WidgetTester tester) async {
+        final res = await Amplify.API.get(path).response;
+        _validateRestResponse(res);
+      });
+
+      testWidgets('should get an error for POST', (WidgetTester tester) async {
+        final res = await Amplify.API.head(path).response;
+        expect(res.statusCode, 403);
```
dnys1, line 61: Did we decide if we wanted to throw an error for these or not?

packages/api/amplify_api/example/integration_test/rest_test.dart
```diff
@@ -0,0 +1,107 @@
...
+      setUpAll(() async {
+        await signOutTestUser();
+      });
+
+      testWidgets('should send GET request', (WidgetTester tester) async {
+        final res = await Amplify.API.get(path).response;
+        _validateRestResponse(res);
+      });
+
+      testWidgets('should get an error for POST', (WidgetTester tester) async {
```
dnys1, line 59:
```suggestion
      testWidgets('should get an error for HEAD', (WidgetTester tester) async {
```

packages/api/amplify_api/example/integration_test/rest_test.dart
```diff
@@ -0,0 +1,107 @@
...
+        await deleteTestUser();
+      }
+    });
+
+    group('guest user access', () {
+      setUpAll(() async {
+        await signOutTestUser();
+      });
+
+      testWidgets('should send GET request', (WidgetTester tester) async {
```
dnys1, line 54: Is this unique to the backend? I can't see why a GET should succeed and a HEAD should fail for guest access. Shouldn't both fail?

packages/api/amplify_api/example/integration_test/rest_test.dart
```diff
@@ -0,0 +1,107 @@
...
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  HttpPayload _generateTestPayload() => HttpPayload.json({'name': 'mow lawn'});
+  void _validateRestResponse(AWSHttpResponse response) {
+    final body = response.decodeBody();
+    expect(response.statusCode, 200);
+    expect(body, expectedResponseText);
+  }
+
+  group('REST', () {
```
dnys1, line 34: Can we test custom headers?

packages/api/amplify_api/example/integration_test/graphql/iam_test.dart
```diff
@@ -0,0 +1,334 @@
...
+import 'package:amplify_flutter/amplify_flutter.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:integration_test/integration_test.dart';
+
+import '../util.dart';
+
+void main({bool useExistingTestUser = false}) {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  group('GraphQL IAM', () {
```
dnys1, line 28: Is there a way to test custom headers in any of these? Maybe with a lambda resolver? I'm not sure what the use cases are but at some point we'll want to make sure we have a test in place for those.

ragingsquirrel3: I think the easies way to test that would be to keep some test that uses the headers to do custom auth in cognito mode (like the test utils currently do) and keep that in place even when we replace that util to use `authorizationMode`. I will add a test to do that explicitly here.

packages/api/amplify_api/example/integration_test/rest_test.dart
ragingsquirrel3: In our team we did but I want to hold off making any API changes until those changes have finished formal API review so that change can be made in a single PR and save some churn.

packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
ragingsquirrel3: As long as we query and get a valid list result which we can check the length on I think it's testing the same thing whether or not the list is populated..

packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
ragingsquirrel3: does not throw so adding checks here and other tests

packages/api/amplify_api/example/integration_test/rest_test.dart
ragingsquirrel3: Good question, agree it's weird. It's part of this backend but I think more CLI thing that unique to this backend. In CLI it asks you if unauthenticated users can READ, WRITE, UPDATE, or DELETE. I said only yes for READ, authenticated users can do all, and this is the result. Not sure why they are mapping those CRUD to theseHTTP methods. Makes sense for everything except HEAD.

packages/api/amplify_api/example/integration_test/graphql/iam_test.dart
ragingsquirrel3: I might be missing something but I debugged this and only saw it called once so may sync offline.

packages/api/amplify_api/example/integration_test/rest_test.dart
ragingsquirrel3: added

packages/api/amplify_api/example/integration_test/graphql/iam_test.dart
ragingsquirrel3: added test like this

ragingsquirrel3:
Got this running in CI on my fork https://github.com/ragingsquirrel3/amplify-flutter/actions/runs/3062792536

packages/api/amplify_api/example/integration_test/util.dart
```diff
@@ -0,0 +1,255 @@
...
+}
+
+/// Throws if response `.hasErrors` (any GraphQL errors from server).
+void throwIfError(GraphQLResponse response) {
+  if (response.hasErrors || response.data == null) {
+    throw AmplifyException(
+      'GraphQL error while running request: ${response.errors.toString()}',
+    );
+  }
+}
```
Jordan-Nelson, line 206: A custom matcher makes the test code a little more readable. I haven't tested this, but I think this should work.


```diff
+  expect(response, hasNoGraphQLErrors);
-  throwIfError(response);
```

```suggestion
Matcher hasNoGraphQLErrors<T>(GraphQLResponse<T> response) {
  return predicate(
    (GraphQLResponse<T> response) =>
        !response.hasErrors && response.data != null,
    'Has no GraphQL Errors',
  );
}
```

packages/api/amplify_api/example/integration_test/util.dart
```diff
@@ -0,0 +1,255 @@
...
+
+// Wait for subscription established for given request.
```
Jordan-Nelson, line 208: Make these all doc comments so they can be seen in IDE tooling when writing/updating tests.

```suggestion
/// Wait for subscription established for given request.
```

packages/api/amplify_api/example/integration_test/util.dart
```diff
@@ -0,0 +1,255 @@
...
+    onError: (Object e) => fail('Error in subscription stream: $e'),
+  );
+
+  await establishedCompleter.future
+      .timeout(const Duration(seconds: _subscriptionTimeoutInterval));
+  return subscription;
+}
+
+// Establish subscription for request, do the mutationFunction, then wait
+// for the stream event, cancel the operation, return response from event.
```
Jordan-Nelson, line 232:
```suggestion
/// Establish subscription for request, do the mutationFunction, then wait
/// for the stream event, cancel the operation, return response from event.
```

packages/api/amplify_api/example/integration_test/graphql/iam_test.dart
```diff
@@ -0,0 +1,371 @@
...
+        expect(firstId, isNot(secondId));
+      });
+
+      testWidgets('should LIST blogs with Model helper with query predicate',
+          (WidgetTester tester) async {
+        String blogName = 'Integration Test Blog ${uuid()}';
+        Blog blog = await addBlog(blogName);
+
+        final req = ModelQueries.list<Blog>(Blog.classType,
+            where: Blog.NAME.eq(blogName) & Blog.ID.eq(blog.id));
```
Jordan-Nelson, line 148: *nit:* Doesn't have to be part of this PR, but we should get the lint rule that enforces trailing commas enabled in the branch/package.

This is so much easier to read imo

```dart
final req = ModelQueries.list<Blog>(
  Blog.classType,
  where: Blog.NAME.eq(blogName) & Blog.ID.eq(blog.id),
);
```

packages/api/amplify_api/example/integration_test/util.dart
```diff
@@ -0,0 +1,255 @@
...
+import 'package:amplify_api/amplify_api.dart';
+import 'package:amplify_api_example/amplifyconfiguration.dart';
+import 'package:amplify_api_example/models/ModelProvider.dart';
+import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
+import 'package:amplify_flutter/amplify_flutter.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+const _subscriptionTimeoutInterval = 5;
+
+String? testUsername;
```
Jordan-Nelson, line 27: I guess this is fine as long as you are only using one user per test. I might still put this in a class so that you don't run into some hard to track down global variable issue later.

```dart
class TestUser {
  TestUser({
    String? username,
    String? password,
  })  : _username = 'testUser${uuid()}',
        _password = uuid(secure: true);

  final String _username;
  final String _password;

  Future<void> signUp() async {
    await signOut();
    final testEmail = '$_username@amazon.com';
    final result = await Amplify.Auth.signUp(
      username: _username,
      password: _password,
      options: CognitoSignUpOptions(
        userAttributes: {CognitoUserAttributeKey.email: testEmail},
      ),
    );
    if (!result.isSignUpComplete) {
      throw const AmplifyException('Unable to sign up test user.');
    }
  }

  Future<void> signOut() async {
    final session = await Amplify.Auth.fetchAuthSession();
    if (!session.isSignedIn) return;
    await Amplify.Auth.signOut();
  }
}
```


```dart
final user = TestUser();
await user.signIn();
```

Jordan-Nelson (approved):
I left a handful of comments, but none of them are really blockers.

packages/api/amplify_api/example/integration_test/graphql/iam_test.dart
ragingsquirrel3: gotcha. My mistake, I was debugging the wrong test file when I claimed only called once. This was leftover from previous organization of file and removing now, ty.

packages/api/amplify_api/example/integration_test/graphql/iam_test.dart
ragingsquirrel3: agree, there is a task for this in asana as it's hard to enforce manually

packages/api/amplify_api/example/integration_test/util.dart
ragingsquirrel3: nice! I had to tweak this a little

packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
```diff
@@ -0,0 +1,84 @@
...
+import 'package:amplify_api/amplify_api.dart';
+import 'package:amplify_api_example/amplifyconfiguration.dart';
+import 'package:amplify_api_example/models/ModelProvider.dart';
+import 'package:amplify_flutter/amplify_flutter.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:integration_test/integration_test.dart';
+
+import '../util.dart';
+
+void main({bool useExistingTestUser = false}) {
```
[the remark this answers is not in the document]
ragingsquirrel3: That is the intention but it does not work for api key test as it needs to sign up test user so I will make that change. Other files should be individually executable.

packages/api/amplify_api/example/integration_test/util.dart
ragingsquirrel3: good suggestion, just added something w a few mods

dnys1 (approved):

packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
ragingsquirrel3: yes

packages/api/amplify_api/example/integration_test/graphql/iam_test.dart
```diff
@@ -0,0 +1,371 @@
...
+          document: graphQLDocument,
+          modelType: const PaginatedModelType(Blog.classType),
+          decodePath: listBlogs,
+        );
+        final res = await Amplify.API.query(request: request).response;
+        throwIfError(res);
+        expect(res.data?.items.first, isA<Blog>());
+      });
+
+      testWidgets('should allow custom headers', (WidgetTester tester) async {
```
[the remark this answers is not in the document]
ragingsquirrel3: good call, just modified to do that

Issue #416, opened 2021-03-01
Feature Request: Support the use of custom storage paths


exampaqs:
## Updated Description
Feature Request: The ability to specify a custom prefix for S3 storage, instead of using the public/protected/private prefixes. For reference, see feature requests and docs in other Amplify libraries.
- Amplify-iOS feature request: https://github.com/aws-amplify/amplify-ios/issues/916
- Amplify-Android feature request: https://github.com/aws-amplify/amplify-android/issues/910
- Amplify-JS docs: https://docs.amplify.aws/lib/storage/configureaccess/q/platform/js/#customization

## Original Description
Amplify.Storage.list() is showing empty list of items when used with existing buckets

haverchuck:
@exampaqs How have you configured the ACL on the existing bucket?

exampaqs:
First all public access to Bucket is blocked and the Bucket permission reads "Only authorized users of this account"

Second I have added a bucket permission like this
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "PublicReadGetObject",
            "Effect": "Allow",
            "Principal": "*",
            "Action": "s3:GetObject",
            "Resource": "arn:aws:s3:::{BUCKET-NAME}/*"
        }
    ]
}

and the Auth and UnAuth Roles of the Iam user each have a policy configured like:
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:ListBucket",
                "s3:GetBucketLocation"
            ],
            "Resource": "arn:aws:s3:::{BUCKET-NAME}"
        },
        {
            "Effect": "Allow",
            "Action": [
                "s3:GetObject"
            ],
            "Resource": "arn:aws:s3:::{BUCKET-NAME}/*"
        }
    ]
}

but few hours ago, i created a folder in the bucket and named it 'public'. then i could only access objects in this public folder but do not have access to other objects in the bucket.

smerchermek:
I have the same problem when trying to access an existing bucket, Amplify.Storage.list () returns an empty list.
The configuration is good, I can access the list of my objects in my bucket from the command line

smerchermek:
I have the same problem while accessing the objects in my bucket, I can only access objects inside "public" folder.
Amplify is well configured and I have access to my bucket using Command Line

Jordan-Nelson:
[...]

PR #2071, opened 2022-08-27
feat(storage) : Support custom prefix

fjnoyp:
Modify method channels to call native layer support for custom prefix.

*Issue #, if available:*
https://github.com/aws-amplify/amplify-flutter/issues/416

*Description of changes:*
Call custom prefix in native side. 

### DIFF
--- a/packages/storage/amplify_storage_s3/lib/amplify_storage_s3.dart
+++ b/packages/storage/amplify_storage_s3/lib/amplify_storage_s3.dart
@@ -18,6 +18,7 @@ library amplify_storage_plugin;
 import 'dart:io';
 
 import 'package:amplify_core/amplify_core.dart';
+import 'package:amplify_storage_s3/src/s3_prefix_resolver/amplify_storage_s3_prefix_resolver.dart';
 import 'package:aws_common/aws_common.dart';
 import 'package:meta/meta.dart';
 
@@ -30,11 +31,11 @@ export './src/types.dart';
 /// {@endtemplate}
 abstract class AmplifyStorageS3 extends StoragePluginInterface {
   /// {@macro amplify_storage_s3.amplify_storage_s3}
-  factory AmplifyStorageS3() {
+  factory AmplifyStorageS3({StorageS3PrefixResolver? prefixResolver}) {
     if (zIsWeb || Platform.isWindows || Platform.isMacOS || Platform.isLinux) {
       throw UnsupportedError('This platform is not supported yet');
     }
-    return AmplifyStorageS3MethodChannel();
+    return AmplifyStorageS3MethodChannel(prefixResolver: prefixResolver);
   }
 
   /// Protected constructor for subclasses.
--- a/packages/storage/amplify_storage_s3/lib/method_channel_storage_s3.dart
+++ b/packages/storage/amplify_storage_s3/lib/method_channel_storage_s3.dart
@@ -32,10 +32,30 @@ var _transferProgressionCallbackMap =
 
 /// An implementation of [AmplifyPlatform] that uses method channels.
 class AmplifyStorageS3MethodChannel extends AmplifyStorageS3 {
-  AmplifyStorageS3MethodChannel() : super.protected();
+  StorageS3PrefixResolver? prefixResolver;
+
+  AmplifyStorageS3MethodChannel({this.prefixResolver}) : super.protected();
+
+  Future<dynamic> _methodCallHandler(MethodCall call) async {
+    switch (call.method) {
+      case 'awsS3PluginPrefixResolver':
+        if (prefixResolver == null) {
+          throw StateError("Native calling nonexistent PrefixResolver in Dart");
+        }
+
+        Map<String, dynamic> arguments =
+            Map<String, dynamic>.from(call.arguments);
+
+        return _handlePrefix(arguments, prefixResolver!);
+
+      default:
+        throw UnimplementedError('${call.method} has not been implemented.');
+    }
+  }
 
   @override
   Future<void> addPlugin() async {
+    _channel.setMethodCallHandler(_methodCallHandler);
     try {
       _transferProgressEventChannel.receiveBroadcastStream(0).listen((event) {
         var eventData = (event as Map).cast<String, dynamic>();
@@ -51,7 +71,8 @@ class AmplifyStorageS3MethodChannel extends AmplifyStorageS3 {
         }
       });
 
-      return await _channel.invokeMethod('addPlugin');
+      return await _channel.invokeMethod('configureStorage',
+          <String, dynamic>{'hasPrefixResolver': prefixResolver != null});
     } on PlatformException catch (e) {
       if (e.code == "AmplifyAlreadyConfiguredException") {
         throw AmplifyAlreadyConfiguredException(
@@ -241,5 +262,45 @@ class AmplifyStorageS3MethodChannel extends AmplifyStorageS3 {
 
   StorageException _convertToStorageException(PlatformException e) {
     return StorageException.fromMap(Map<String, String>.from(e.details));
+  }
+
+  Future<dynamic> _handlePrefix(Map<String, dynamic> arguments,
+      StorageS3PrefixResolver prefixResolver) async {
+    if (!arguments.containsKey('accessLevel')) {
+      throw StorageException(AmplifyExceptionMessages.missingExceptionMessage,
+          recoverySuggestion:
+              AmplifyExceptionMessages.missingRecoverySuggestion,
+          underlyingException: arguments.toString());
+    }
+
+    final accessLevelString = arguments['accessLevel'];
+    StorageAccessLevel accessLevel;
+    switch (accessLevelString) {
+      case 'guest':
+        accessLevel = StorageAccessLevel.guest;
+        break;
+      case 'protected':
+        accessLevel = StorageAccessLevel.protected;
+        break;
+      case 'private':
+        accessLevel = StorageAccessLevel.private;
+        break;
+      default:
+        throw StateError('Native sent invalid accessLevelString');
+    }
+    String? targetIdentity = arguments['targetIdentity'];
+
+    try {
+      String prefix = await prefixResolver.resolvePrefix(
+          storageAccessLevel: accessLevel, identityId: targetIdentity);
+      return {'isSuccess': true, 'prefix': prefix};
+      // Note: Amplify Native error callbacks expect StorageExceptions and not generic Exceptions
+    } on Exception catch (e) {
+      return {
+        'isSuccess': false,
+        'errorMessage': e.toString(),
+        'errorRecoverySuggestion': 'Custom PrefixResolver threw an exception'
+      };
+    }
   }
 }
--- a/packages/storage/amplify_storage_s3/test/amplify_storage_s3_prefix_test.dart
+++ b/packages/storage/amplify_storage_s3/test/amplify_storage_s3_prefix_test.dart
@@ -1,0 +1,132 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import 'dart:async';
+
+import 'package:amplify_core/amplify_core.dart';
+import 'package:amplify_storage_s3/amplify_storage_s3.dart';
+import 'package:amplify_storage_s3/method_channel_storage_s3.dart';
+import 'package:amplify_test/amplify_test.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+// Utilized from: https://github.com/flutter/flutter/issues/63465 #CyrilHu
+// For mocking Native -> Dart
+extension MockMethodChannel on MethodChannel {
+  Future<void> invokeMockMethod(
+      String method, dynamic arguments, Function(dynamic)? callback) async {
+    const codec = StandardMethodCodec();
+    final data = codec.encodeMethodCall(MethodCall(method, arguments));
+
+    return ambiguate(ServicesBinding.instance)
+        ?.defaultBinaryMessenger
+        .handlePlatformMessage(
+      name,
+      data,
+      (ByteData? data) {
+        if (callback != null) callback(codec.decodeEnvelope(data!));
+      },
+    );
+  }
+}
+
+void main() {
+  const testPath = 'chosenPath';
+  const testAccessLevelString = 'guest';
+  const testAccessLevel = StorageAccessLevel.guest;
+  const testTargetIdentity = 'test-identity-id';
+  const testErrorMessage = 'some error message';
+  final testException = Exception(testErrorMessage);
+
+  const MethodChannel storageChannel =
+      MethodChannel('com.amazonaws.amplify/storage_s3');
+
+  TestWidgetsFlutterBinding.ensureInitialized();
+
+  late StorageAccessLevel receivedAccessLevel;
+  String? receivedIdentityId;
+  bool throwErrorInPrefixResolver = false;
+
+  String prefixResolver(
+      {required StorageAccessLevel storageAccessLevel, String? identityId}) {
+    receivedAccessLevel = storageAccessLevel;
+    receivedIdentityId = identityId;
+
+    if (throwErrorInPrefixResolver) throw testException;
+
+    return testPath;
+  }
+
+  setUp(() {
+    storageChannel.setMockMethodCallHandler((MethodCall methodCall) async {});
+    AmplifyStorageS3MethodChannel storage = AmplifyStorageS3MethodChannel(
+        prefixResolver: PrefixResolverTest(prefixResolver));
+    return storage.addPlugin();
+  });
+
+  test(
+      'PrefixResolver information from MethodChannel is properly serialized and called',
+      () async {
+    dynamic dataSentToNative;
+    await storageChannel.invokeMockMethod(
+        'awsS3PluginPrefixResolver',
+        {
+          'accessLevel': testAccessLevelString,
+          'targetIdentity': testTargetIdentity
+        },
+        (e) => dataSentToNative = e);
+    expect(receivedAccessLevel, testAccessLevel);
+    expect(receivedIdentityId, testTargetIdentity);
+
+    dynamic map = Map<String, dynamic>.from(dataSentToNative);
+    expect(map['isSuccess'], true);
+    expect(map['prefix'], testPath);
+  });
+
+  test('Exception in PrefixResolver is properly serialized to native',
+      () async {
+    dynamic dataSentToNative;
+    throwErrorInPrefixResolver = true;
+    await storageChannel.invokeMockMethod(
+        'awsS3PluginPrefixResolver',
+        {
+          'accessLevel': testAccessLevelString,
+          'targetIdentity': testTargetIdentity
+        },
+        (e) => dataSentToNative = e);
+
+    dynamic map = Map<String, dynamic>.from(dataSentToNative);
+    expect(map['isSuccess'], false);
+    expect(map['errorMessage'], testException.toString());
+    expect(map['errorRecoverySuggestion'],
+        'Custom PrefixResolver threw an exception');
+  });
+}
+
+class PrefixResolverTest implements StorageS3PrefixResolver {
+  String Function(
+      {required StorageAccessLevel storageAccessLevel,
+      String? identityId}) callback;
+
+  PrefixResolverTest(this.callback);
+
+  Future<String> resolvePrefix({
+    required StorageAccessLevel storageAccessLevel,
+    String? identityId,
+  }) async {
+    return callback(
+        storageAccessLevel: storageAccessLevel, identityId: identityId);
+  }
+}
--- a/packages/storage/amplify_storage_s3_android/android/src/main/kotlin/com/amazonaws/amplify/amplify_storage_s3/StorageS3.kt
+++ b/packages/storage/amplify_storage_s3_android/android/src/main/kotlin/com/amazonaws/amplify/amplify_storage_s3/StorageS3.kt
@@ -21,9 +21,11 @@ import android.util.Log
 import androidx.annotation.NonNull
 import com.amazonaws.amplify.amplify_core.AtomicResult
 import com.amazonaws.amplify.amplify_core.exception.ExceptionUtil.Companion.handleAddPluginException
+import com.amazonaws.amplify.amplify_storage_s3.types.FlutterPrefixResolver
 import com.amazonaws.amplify.amplify_storage_s3.types.TransferProgressStreamHandler
 import com.amplifyframework.core.Amplify
 import com.amplifyframework.storage.s3.AWSS3StoragePlugin
+import com.amplifyframework.storage.s3.configuration.AWSS3StoragePluginConfiguration
 import io.flutter.embedding.engine.plugins.FlutterPlugin
 import io.flutter.embedding.engine.plugins.activity.ActivityAware
 import io.flutter.embedding.engine.plugins.activity.ActivityPluginBinding
@@ -44,14 +46,13 @@ class StorageS3 : FlutterPlugin, ActivityAware, MethodCallHandler {
     private val transferProgressStreamHandler: TransferProgressStreamHandler = TransferProgressStreamHandler()
 
     override fun onAttachedToEngine(@NonNull flutterPluginBinding: FlutterPlugin.FlutterPluginBinding) {
-        channel =
-            MethodChannel(flutterPluginBinding.binaryMessenger, "com.amazonaws.amplify/storage_s3")
+        channel = MethodChannel(flutterPluginBinding.binaryMessenger, "com.amazonaws.amplify/storage_s3")
         channel.setMethodCallHandler(this)
         context = flutterPluginBinding.applicationContext
 
         transferProgressEventChannel = EventChannel(
-            flutterPluginBinding.binaryMessenger,
-            "com.amazonaws.amplify/storage_transfer_progress_events"
+                flutterPluginBinding.binaryMessenger,
+                "com.amazonaws.amplify/storage_transfer_progress_events"
         )
         transferProgressEventChannel.setStreamHandler(transferProgressStreamHandler)
     }
@@ -59,9 +60,22 @@ class StorageS3 : FlutterPlugin, ActivityAware, MethodCallHandler {
     override fun onMethodCall(@NonNull call: MethodCall, @NonNull _result: Result) {
         val result = AtomicResult(_result, call.method)
 
-        if (call.method == "addPlugin") {
+        val arguments = call.arguments as Map<String, *>
+
+        if (call.method == "configureStorage") {
             try {
-                Amplify.addPlugin(AWSS3StoragePlugin())
+                val hasPrefixResolver = arguments["hasPrefixResolver"] as? Boolean? == true
+                Amplify.addPlugin(
+                        AWSS3StoragePlugin(
+                                AWSS3StoragePluginConfiguration {
+                                    awsS3PluginPrefixResolver =
+                                            if (hasPrefixResolver)
+                                                FlutterPrefixResolver(methodChannel = channel)
+                                            else null
+                                }
+                        )
+                )
+
                 Log.i("AmplifyFlutter", "Added StorageS3 plugin")
                 result.success(null)
             } catch (e: Exception) {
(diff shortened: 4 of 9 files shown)


packages/storage/amplify_storage_s3/lib/method_channel_storage_s3.dart
```diff
@@ -32,10 +33,30 @@ var _transferProgressionCallbackMap =
...
+    switch (call.method) {
+      case 'awsS3PluginPrefixResolver':
+        if (prefixResolver == null) {
+          throw StateError("Native calling nonexistent PrefixResolver in Dart");
+        }
+
+        Map<String, dynamic> arguments =
+            (call.arguments as Map).cast<String, dynamic>();
+
+        return PrefixHandler.handlePrefix(arguments, prefixResolver!);
```
HuiSF, line 50: nit: maybe we can throw exception on null check of `prefixResolver` instead of force unwrap the nullable property

packages/storage/amplify_storage_s3/lib/src/S3PrefixResolver/prefix_handler.dart
```diff
@@ -0,0 +1,59 @@
...
+import '../types.dart';
+
+class PrefixHandler {
+  static Future<dynamic> handlePrefix(Map<String, dynamic> arguments,
+      AmplifyStorageS3PrefixResolver prefixResolver) async {
+    final accessLevelString = arguments['accessLevel'];
+    StorageAccessLevel accessLevel;
+    switch (accessLevelString) {
+      case 'public':
+        // Why are our storageAccessLevel enums different?
```
HuiSF, line 27: The default prefix for `StorageAccessLevel.guest` is always `"public"` I think, do you see any inconsistency in the native libraries?

fjnoyp: The StorageAccessLevel enum is not the same between platforms which requires the extra case in the switch statement as a result.  But yes they map to the same S3 concepts. 

iOS: 
```
public enum StorageAccessLevel: String {

    /// Objects can be read or written by any user without authentication
    case guest

    /// Objects can be viewed by any user without authentication, but only written by the owner
    case protected

    /// Objects can only be read and written by the owner
    case `private`
}
```

Android: 
```
public enum StorageAccessLevel {

    /**
     * Storage items are accessible by all users of your app.
     */
    PUBLIC,

    /**
     * Storage items are readable by all users, but writable only by the creating user.
     */
    PROTECTED,

    /**
     * Storage items are accessible for the individual user who performs the write.
     */
    PRIVATE
}
```

packages/storage/amplify_storage_s3/lib/method_channel_storage_s3.dart
fjnoyp: Sounds good, updated: 
```

        Map<String, dynamic>? arguments =
        call.arguments is Map<String, dynamic> ? call.arguments : null;
        
        if(arguments == null) throw StateError("Native send invalid argument to PrefixResolver in Dart");

```

packages/storage/amplify_storage_s3/lib/amplify_storage_s3.dart
```diff
@@ -18,6 +18,7 @@ library amplify_storage_plugin;
 import 'dart:io';
 
 import 'package:amplify_core/amplify_core.dart';
+import 'package:amplify_storage_s3/src/S3PrefixResolver/AWSS3StoragePluginPrefixResolver.dart';
```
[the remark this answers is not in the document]
fjnoyp: Updated

packages/storage/amplify_storage_s3/lib/src/S3PrefixResolver/AmplifyStorageS3PrefixResolver.dart
```diff
@@ -0,0 +1,26 @@
+/*
```
dnys1, line 1: Folders and files should be named using snake_case per Dart conventions: https://dart.dev/guides/language/effective-dart/style#do-name-libraries-and-source-files-using-lowercase_with_underscores

`s3_prefix_resolver/amplify_storage_s3_prefix_resolver.dart`

packages/storage/amplify_storage_s3/lib/src/S3PrefixResolver/prefix_handler.dart
```diff
@@ -0,0 +1,59 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import 'package:amplify_core/amplify_core.dart';
+
+import '../types.dart';
+
+class PrefixHandler {
+  static Future<dynamic> handlePrefix(Map<String, dynamic> arguments,
```
dnys1, line 21: Classes with just static methods is generally considered an anti-pattern. There's even a lint for it: https://dart.dev/tools/linter-rules#avoid_classes_with_only_static_members. Is there a reason this shouldn't be defined on the method channel?

packages/storage/amplify_storage_s3/lib/src/S3PrefixResolver/prefix_handler.dart
```diff
@@ -0,0 +1,59 @@
...
+        accessLevel = StorageAccessLevel.guest;
+        break;
+      case 'protected':
+        accessLevel = StorageAccessLevel.protected;
+        break;
+      case 'private':
+        accessLevel = StorageAccessLevel.private;
+        break;
+      default:
+        throw StateError("Native sent invalid accessLevelString");
```
dnys1, line 40: Use single quotes for strings: https://dart.dev/tools/linter-rules#prefer_single_quotes

packages/storage/amplify_storage_s3/lib/src/S3PrefixResolver/prefix_handler.dart
```diff
@@ -0,0 +1,59 @@
...
+    }
+    String? targetIdentity = arguments['targetIdentity'];
+
+    try {
+      String prefix = await prefixResolver.resolvePrefix(
+          storageAccessLevel: accessLevel, identityId: targetIdentity);
+      return {'isSuccess': true, 'prefix': prefix};
+      // Native catches StorageExceptions, but why would customer code throw an Amplify exception?
+    } on AmplifyException catch (e) {
```
dnys1, line 49: What about other exception types? If the user is implementing `resolvePrefix`, they may not (probably won't) be throwing AmplifyExceptions

packages/storage/amplify_storage_s3/lib/method_channel_storage_s3.dart
```diff
@@ -32,10 +33,32 @@ var _transferProgressionCallbackMap =
...
+  AmplifyStorageS3MethodChannel({this.prefixResolver}) : super.protected();
+
+  Future<dynamic> _methodCallHandler(MethodCall call) async {
+    switch (call.method) {
+      case 'awsS3PluginPrefixResolver':
+        if (prefixResolver == null) {
+          throw StateError("Native calling nonexistent PrefixResolver in Dart");
+        }
+
+        Map<String, dynamic>? arguments =
```
dnys1, line 47: This file needs to be formatted

packages/storage/amplify_storage_s3/lib/method_channel_storage_s3.dart
```diff
@@ -32,10 +33,32 @@ var _transferProgressionCallbackMap =
...
+        call.arguments is Map<String, dynamic> ? call.arguments : null;
```
dnys1, line 48:
```suggestion
        call.arguments as Map<String, dynamic>?;
```

packages/storage/amplify_storage_s3/test/amplify_storage_s3_prefix_test.dart
```diff
@@ -0,0 +1,88 @@
...
+  });
+
+  test(
+      'PrefixResolver information from MethodChannel is properly serialized and called',
+      () async {
+    await storageChannel.invokeMockMethod("awsS3PluginPrefixResolver",
+        {'accessLevel': 'guest', 'targetIdentity': 'test-identity-id'});
+    expect(receivedAccessLevel, StorageAccessLevel.guest);
+    expect(receivedIdentityId, 'test-identity-id');
+  });
```
dnys1, line 71: Can we add some error tests, e.g. if `resolvePrefix` throws?

packages/storage/amplify_storage_s3/lib/src/S3PrefixResolver/AmplifyStorageS3PrefixResolver.dart
fjnoyp: shoot my bad thanks for catching that.

packages/storage/amplify_storage_s3/lib/src/S3PrefixResolver/prefix_handler.dart
fjnoyp: The intention was to avoid 'polluting' the method channel class and to therefore separate out the prefix handler logic to another class. 

It's really not necessary since we're going to deprecate this code soon anyway so I can merge this back in.

packages/storage/amplify_storage_s3/lib/src/S3PrefixResolver/prefix_handler.dart
fjnoyp: Updated to: 

```
on Exception catch (e) {
      return {
        'isSuccess': false,
        'errorMessage': e.toString(),
        'errorRecoverySuggestion': ''
      };
```

packages/storage/amplify_storage_s3_ios/ios/Classes/SwiftStorageS3.swift
```diff
@@ -21,62 +21,71 @@ import AWSMobileClient
...
-                let configError = error as! ConfigurationError
+    
+    public func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
+        let result = AtomicResult(result, call.method)
+        let arguments = call.arguments as! Dictionary<String,AnyObject>
+        
+        
+        if(call.method == "configureStorage"){
+            do {
+                let hasPrefixResolver = arguments["hasPrefixResolver"] as! Bool
```
HuiSF, line 43: nit: avoid using force unwrap, can use `guard else` to assign `false` as default value?

packages/storage/amplify_storage_s3_ios/ios/Classes/SwiftStorageS3.swift
```diff
@@ -21,62 +21,71 @@ import AWSMobileClient
...
                 
-                var errorCode = "StorageException"
-                if case .amplifyAlreadyConfigured = configError {
-                   errorCode = "AmplifyAlreadyConfiguredException"
+                if(hasPrefixResolver){
+                    try Amplify.add(plugin: AWSS3StoragePlugin(configuration: .prefixResolver(FlutterPrefixResolver(flutterChannel: self.channel!))))
+                }
+                else{
```
HuiSF, line 48: nit:
```suggestion
                } else {
```

packages/storage/amplify_storage_s3_ios/ios/Classes/SwiftStorageS3.swift
```diff
@@ -21,62 +21,71 @@ import AWSMobileClient
...
+                    try Amplify.add(plugin: AWSS3StoragePlugin())
+                }
+                result(true)
+            } catch let error{
```
HuiSF, line 52: nit:
```suggestion
            } catch let error {
```

packages/storage/amplify_storage_s3_ios/ios/Classes/SwiftStorageS3.swift
```diff
@@ -21,62 +21,71 @@ import AWSMobileClient
...
+                    ErrorUtil.postErrorToFlutterChannel(
+                        result: result,
+                        errorCode: errorCode,
+                        details: [
+                            "message" : configError.errorDescription,
+                            "recoverySuggestion" : configError.recoverySuggestion,
+                            "underlyingError": configError.underlyingError != nil ? configError.underlyingError!.localizedDescription : ""
+                        ]
+                    )
+                } else{
```
HuiSF, line 80: nit:
```suggestion
                } else {
```

packages/storage/amplify_storage_s3_ios/ios/Classes/SwiftStorageS3.swift
```diff
@@ -21,62 +21,71 @@ import AWSMobileClient
...
+                    
+                    ErrorUtil.postErrorToFlutterChannel(
+                        result: result,
+                        errorCode: "StorageException",
+                        details: [
+                            "message" : storageError.errorDescription,
+                            "recoverySuggestion" : storageError.recoverySuggestion,
+                            "underlyingError": storageError.underlyingError != nil ? storageError.underlyingError!.localizedDescription : ""
+                        ])
+                } else if(error is ConfigurationError) {
```
HuiSF, line 64: nit: 
```suggestion
                } else if (error is ConfigurationError) {
```

packages/storage/amplify_storage_s3_ios/ios/Classes/SwiftStorageS3.swift
```diff
@@ -21,62 +21,71 @@ import AWSMobileClient
...
-                    ])
-            } else if(error is ConfigurationError) {
-                let configError = error as! ConfigurationError
+    
+    public func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
+        let result = AtomicResult(result, call.method)
+        let arguments = call.arguments as! Dictionary<String,AnyObject>
+        
+        
+        if(call.method == "configureStorage"){
```
HuiSF, line 41: nit: could merge this branch to the `switch` block below. And extract the logic here as a function to be invoked within the `switch` block.

HuiSF, line 41: nit: 
```suggestion
        if(call.method == "configureStorage") {
```

packages/storage/amplify_storage_s3_ios/ios/Classes/FlutterPrefixResolver.swift
```diff
@@ -0,0 +1,72 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import Foundation
+import Amplify
+import Flutter
+import AmplifyPlugins
+import AWSMobileClient
```
HuiSF, line 20: Is this import needed?

packages/storage/amplify_storage_s3_ios/ios/Classes/FlutterPrefixResolver.swift
```diff
@@ -0,0 +1,72 @@
...
+    
+    func resolvePrefix(for accessLevel: StorageAccessLevel,
+                       targetIdentityId: String?,
+                       completion: @escaping (Result<String, StorageError>) -> Void) {
+        
+        
+        var accessLevelString : String
+        
+        switch accessLevel{
+        case StorageAccessLevel.guest: accessLevelString = "public"
```
HuiSF, line 40: I think we should use `StorageAccessLevel.guest. serviceAccessPrefix` to get the prefix string instead of hardcoding it here.

packages/storage/amplify_storage_s3_ios/ios/Classes/FlutterPrefixResolver.swift
```diff
@@ -0,0 +1,72 @@
...
+                do {
+                    let resultMap: [String: Any] = result as! [String: Any]
+                    
+                    let isSuccess = resultMap["isSuccess"] as! Bool
+                    
+                    if(isSuccess){
+                        let prefix = resultMap["prefix"] as! String
+                        completion(.success(prefix))
+                    }
+                    else{
```
HuiSF, line 61: nit: 
```suggestion
                    } else {
```

packages/storage/amplify_storage_s3_ios/ios/Classes/FlutterPrefixResolver.swift
```diff
@@ -0,0 +1,72 @@
...
+        ]
+        
+        DispatchQueue.main.async {
+            self.channel.invokeMethod("awsS3PluginPrefixResolver", arguments: args) { result in
+                do {
+                    let resultMap: [String: Any] = result as! [String: Any]
+                    
+                    let isSuccess = resultMap["isSuccess"] as! Bool
+                    
+                    if(isSuccess){
```
HuiSF, line 57: nit:
```suggestion
                    if (isSuccess) {
```

packages/storage/amplify_storage_s3_ios/ios/Classes/FlutterPrefixResolver.swift
```diff
@@ -0,0 +1,72 @@
...
+            "accessLevel" : accessLevelString,
+            "targetIdentity" : targetIdentityId
+        ]
+        
+        DispatchQueue.main.async {
+            self.channel.invokeMethod("awsS3PluginPrefixResolver", arguments: args) { result in
+                do {
+                    let resultMap: [String: Any] = result as! [String: Any]
+                    
+                    let isSuccess = resultMap["isSuccess"] as! Bool
```
HuiSF, line 55: Should add null check and throw exception here instead of force unwrapping?

packages/storage/amplify_storage_s3_android/android/src/main/kotlin/com/amazonaws/amplify/amplify_storage_s3/types/FlutterPrefixResolver.kt
```diff
@@ -0,0 +1,92 @@
...
+                        override fun notImplemented() {
+                            onSuccess.accept("")
+                            LOG.error("No custom S3 Storage Prefix Resolution Provided. Applying default prefixes.")
+                        }
+                    }
+            )
+        }
+    }
+
+}
```
HuiSF, line 92: nit:🆕 line

packages/storage/amplify_storage_s3_android/android/src/main/kotlin/com/amazonaws/amplify/amplify_storage_s3/types/FlutterPrefixResolver.kt
```diff
@@ -0,0 +1,92 @@
...
+            accessLevel: StorageAccessLevel,
+            targetIdentity: String?,
+            onSuccess: Consumer<String>,
+            onError: Consumer<StorageException>
+    ) {
+
+        val accessLevelString = when (accessLevel) {
+            StorageAccessLevel.PUBLIC -> "public"
+            StorageAccessLevel.PROTECTED -> "protected"
+            StorageAccessLevel.PRIVATE -> "private"
```
HuiSF, line 45: Should use `StorageAccessLevel` provided by amplify-android instead of hardcoding the string here.
[Example](https://github.com/aws-amplify/amplify-android/blob/main/aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/utils/S3Keys.java#L77)
```
StorageAccessLevel.PUBLIC.name().toLowerCase(Locale.US)
```

packages/storage/amplify_storage_s3_ios/ios/Classes/FlutterPrefixResolver.swift
```diff
@@ -0,0 +1,72 @@
...
+                    
+                    if(isSuccess){
+                        let prefix = resultMap["prefix"] as! String
+                        completion(.success(prefix))
+                    }
+                    else{
+                        let errorMessage = resultMap["errorMessage"] as! String
+                        let recoverySuggestion = resultMap["errorRecoverySuggestion"] as! String
+                        
+                        completion(.failure( StorageError.unknown(errorMessage)))
```
HuiSF, line 65: nit:
```suggestion
                        completion(.failure(StorageError.unknown(errorMessage)))
```

packages/storage/amplify_storage_s3_android/android/src/main/kotlin/com/amazonaws/amplify/amplify_storage_s3/types/FlutterPrefixResolver.kt
```diff
@@ -0,0 +1,92 @@
...
+                                    onSuccess.accept(prefix)
+                                } else {
+                                    val errorMessage = resultMap["errorMessage"] as String
+                                    var recoverySuggestion = resultMap["errorRecoverySuggestion"] as String?
+                                    if (recoverySuggestion.isNullOrEmpty()) recoverySuggestion = "No recovery suggestion provided";
+
+                                    onError.accept(StorageException(errorMessage, recoverySuggestion))
+                                }
+                            } catch (e: Exception) {
+                                onSuccess.accept("")
```
HuiSF, line 73: Should use `onError` here to actually ripple back the error instead of an error log?

packages/storage/amplify_storage_s3_android/android/src/main/kotlin/com/amazonaws/amplify/amplify_storage_s3/StorageS3.kt
```diff
@@ -45,23 +47,39 @@ class StorageS3 : FlutterPlugin, ActivityAware, MethodCallHandler {
 
     override fun onAttachedToEngine(@NonNull flutterPluginBinding: FlutterPlugin.FlutterPluginBinding) {
         channel =
-            MethodChannel(flutterPluginBinding.binaryMessenger, "com.amazonaws.amplify/storage_s3")
+                MethodChannel(flutterPluginBinding.binaryMessenger, "com.amazonaws.amplify/storage_s3")
```
HuiSF, line 50: nit: extra whitespace

packages/storage/amplify_storage_s3_android/android/src/main/kotlin/com/amazonaws/amplify/amplify_storage_s3/StorageS3.kt
```diff
@@ -45,23 +47,39 @@ class StorageS3 : FlutterPlugin, ActivityAware, MethodCallHandler {
...
         val result = AtomicResult(_result, call.method)
 
-        if (call.method == "addPlugin") {
+        val arguments = call.arguments as Map<String, *>
+
+        if (call.method == "configureStorage") {
             try {
-                Amplify.addPlugin(AWSS3StoragePlugin())
+
+                var hasPrefixResolver = arguments["hasPrefixResolver"] as? Boolean? == true
```
HuiSF, line 69: nit:
```suggestion
                val hasPrefixResolver = arguments["hasPrefixResolver"] as? Boolean? == true
```

packages/storage/amplify_storage_s3_android/android/src/main/kotlin/com/amazonaws/amplify/amplify_storage_s3/StorageS3.kt
```diff
@@ -45,23 +47,39 @@ class StorageS3 : FlutterPlugin, ActivityAware, MethodCallHandler {
...
+
+                if (hasPrefixResolver) {
+                    Amplify.addPlugin(
+                            AWSS3StoragePlugin(
+                                    AWSS3StoragePluginConfiguration {
+                                        awsS3PluginPrefixResolver = FlutterPrefixResolver(methodChannel = channel);
+                                    }
+                            )
+                    )
+                } else {
```
HuiSF, line 79: nit: might not need the `if else` branch here as `awsS3PluginPrefixResolver` of `AWSS3StoragePluginConfiguration` is nullable.

packages/storage/amplify_storage_s3/lib/src/S3PrefixResolver/prefix_handler.dart
HuiSF: iOS enum is `guest`, but it looks like that [the prefix value it returns should always be `"public"`](https://github.com/aws-amplify/amplify-ios/blob/c6d870a837c7765439fc06a31ba2aa1472913f05/AmplifyPlugins/Storage/AWSS3StoragePlugin/Support/Utils/StorageRequestUtils%2BGetter.swift#L64).

packages/storage/amplify_storage_s3/lib/src/S3PrefixResolver/prefix_handler.dart
HuiSF: I think we can use the [`StorageException`](https://github.com/aws-amplify/amplify-flutter/blob/main/packages/amplify_core/lib/src/types/storage/exceptions/storage_exception.dart) here.

HuiSF (commented):
Thanks for adding this feature! 👍
Added some comments, mostly nit-picking, but we can improve on few spots.

packages/storage/amplify_storage_s3/test/amplify_storage_s3_prefix_test.dart
fjnoyp: Added please see updated code

packages/storage/amplify_storage_s3/lib/src/S3PrefixResolver/prefix_handler.dart
fjnoyp: Switched to catch generic Exception, see updated code

fjnoyp: @HuiSF would customer code throw an amplify StorageException though?

packages/storage/amplify_storage_s3_ios/ios/Classes/SwiftStorageS3.swift
```diff
@@ -1,17 +1,17 @@
...
-* A copy of the License is located at
-*
-*  http://aws.amazon.com/apache2.0
-*
-* or in the "license" file accompanying this file. This file is distributed
-* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
-* express or implied. See the License for the specific language governing
-* permissions and limitations under the License.
-*/
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
fjnoyp: Sorry what is the problem here?  The license header appears to be the same as other swift files

packages/storage/amplify_storage_s3_ios/ios/Classes/FlutterPrefixResolver.swift
HuiSF: `StorageAccessLevel.guest.serviceAccessPrefix` [returns `"public"`](https://github.com/aws-amplify/amplify-ios/blob/c6d870a837c7765439fc06a31ba2aa1472913f05/AmplifyPlugins/Storage/AWSS3StoragePlugin/Support/Utils/StorageRequestUtils%2BGetter.swift#L64).

packages/storage/amplify_storage_s3/lib/src/S3PrefixResolver/prefix_handler.dart
HuiSF: The generic exception type makes sense 👍

packages/storage/amplify_storage_s3_ios/ios/Classes/SwiftStorageS3.swift
```diff
@@ -21,62 +21,69 @@ import AWSMobileClient
...
-                    ])
-            } else if(error is ConfigurationError) {
-                let configError = error as! ConfigurationError
+        switch call.method {
+        case "configureStorage":
+            do {
+                guard let hasPrefixResolver = arguments["hasPrefixResolver"] as? Bool else{
+                    _ = false
+                    return
+                }
```
HuiSF, line 46: Sorry my original comment may be false. but this early return seems incorrect here... as it's not invoking `result(true)`.
How about

```suggestion
                let hasPrefixResolver = arguments["hasPrefixResolver"] as? Bool ?? false
```

packages/storage/amplify_storage_s3_ios/ios/Classes/SwiftStorageS3.swift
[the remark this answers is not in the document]
fjnoyp: This issue is weird because in xcode I cannot actually delete a space (it doesn't even exist there)

packages/storage/amplify_storage_s3_ios/ios/Classes/SwiftStorageS3.swift
fjnoyp: Thanks Hui, code has been updated.

packages/storage/amplify_storage_s3/lib/src/S3PrefixResolver/amplify_storage_s3_prefix_resolver.dart
```diff
@@ -0,0 +1,26 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import 'dart:async';
+
+import 'package:amplify_core/amplify_core.dart';
+
+abstract class AmplifyStorageS3PrefixResolver {
```
dnys1, line 20: Hui has updated the naming of this: https://github.com/aws-amplify/amplify-flutter/pull/2059/files#diff-778909e5ff89aa667611d1d080237284550304c40bb1d00667c9b42f37126dcc

packages/storage/amplify_storage_s3/lib/amplify_storage_s3.dart
```diff
@@ -18,6 +18,7 @@ library amplify_storage_plugin;
 import 'dart:io';
 
 import 'package:amplify_core/amplify_core.dart';
+import './src/S3PrefixResolver/amplify_storage_s3_prefix_resolver.dart';
```
dnys1, line 21: Organize

dnys1 (dismissed):

Jordan-Nelson:
Note: We will likely want to cherry pick this to next so that customers using auth dev preview can use this. It will probably cause conflicts with the storage dart first branch. CC @HuiSF

packages/storage/amplify_storage_s3/lib/amplify_storage_s3.dart
Jordan-Nelson: Also naming: `S3PrefixResolver` -> `s3_prefix_resolver`, and use package imports. We are enforcing this on next with [always_use_package_imports](https://dart-lang.github.io/linter/lints/always_use_package_imports.html). Good idea to use it everywhere.

fjnoyp: alright I can make that change.  Are you sure @Jordan-Nelson as this would be inconsistent with the existing naming existing for main: 

s3_prefix_resolver
S3DownloadFile
S3GetUrl
S3List
S3UploadFile

fjnoyp:
Sounds good @Jordan-Nelson @HuiSF the PR for next is created: https://github.com/aws-amplify/amplify-flutter/pull/2117

It's in draft mode.  Once this is merged I'll move it out.

dnys1 (approved):
It can be in a follow-up, but I'm thinking this deserves a couple integration tests as well

HuiSF (approved):
Thanks for implementing this feature.
Expecting integration tests in separate PRs :)

dnys1 (approved):

PR #2106, opened 2022-09-09
 feat(api): OIDC and lambda (function) auth modes

ragingsquirrel3:
This PR implements OIDC/lambda (aka "function") auth modes for API category in dart. It's pretty simple, just instantiates an `AmplifyAuthProvider` with the `APIAuthProvider` that the user would pass when they instantiate the plugin and forwards the call to `getLatestAuthToken` from that. Also has some tests.

This PR does not modify the OIDC/Lambda config that is passed to native (platform channels) if used with Datastore.


packages/amplify_core/lib/src/types/common/amplify_auth_provider.dart
```diff
@@ -15,11 +15,18 @@
...
 import 'package:aws_signature_v4/aws_signature_v4.dart';
+import 'package:meta/meta.dart';
 
 /// An identifier to use as a key in an [AmplifyAuthProviderRepository] so that
 /// a retrieved auth provider can be typed more accurately.
-class AmplifyAuthProviderToken<T extends AmplifyAuthProvider> extends Token<T> {
-  const AmplifyAuthProviderToken();
+///
+/// Does not extend [Token] because [Token] is equal to another [Token] of same
+/// type which has unintended consequences.
```
ragingsquirrel3, line 24: I was finding that user could not register multiple auth providers that implement same type as configured in enum token, could not register 2 of type `TokenAmplifyAuthProvider` bc they would be considered equal by equality fn in `Token` and thus get overwritten in the repo map when setting.

Another way to go was to have additional no-op abstract classes in core for function/oidc which extend the token class. I tried that first pass and got annoying bc they are essentially same shape. If anyone has other ideas let me know.

packages/api/amplify_api/lib/src/api_plugin_impl.dart
```diff
@@ -77,17 +78,31 @@ class AmplifyAPIDart extends AmplifyAPI {
...
   void _registerApiPluginAuthProviders() {
     _apiConfig.endpoints.forEach((key, value) {
       // Check the presence of apiKey (not auth type) because other modes might
       // have a key if not the primary auth mode.
       if (value.apiKey != null) {
         _authProviderRepo.registerAuthProvider(
-            value.authorizationType.authProviderToken,
-            AppSyncApiKeyAuthProvider());
+          value.authorizationType.authProviderToken,
+          AppSyncApiKeyAuthProvider(),
```
dnys1, line 94: On an unrelated note - why not include the API key in the auth provider, like `AppSyncApiKeyAuthProvider(value.apiKey)`? Does this mean you have to pass the API token each time you use the API key auth provider? That doesn't seem right..

dnys1 (approved):

Equartey (approved):
Nice work!

packages/api/amplify_api/lib/src/api_plugin_impl.dart
ragingsquirrel3: Good question. yes, the API key itself is an option passed by consumer and provider just formats request. I initially tried to implement the way you describe but it got a little complicated with multiple backends, in which case `apiName` would have to be the provided as option instead and API key provider would have to store all of them and use that apiName parameter to switch on it. It's for sure possible, I just didn't see the need for that complexity given caller will have access to the config w the key anyway.

PR #2111, opened 2022-09-12
 fix(analytics): clarifies lack of android support

haverchuck:
*Description of changes:*
Clarifies category/platform support on developer preview.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,11 +53,16 @@ We are iterating and looking for feedback and collaboration, so please [**let us
   <img src="https://img.shields.io/pub/v/amplify_api.svg">
 </a>
 
-## Platform Support (Developer Preview)
+## Category / Platform Support (Developer Preview)
 
-| Category       | Android               | iOS                        | Web                        | Windows                    | MacoOS                     | Linux                      |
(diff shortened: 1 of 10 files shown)


Jordan-Nelson (approved):

README.md
```diff
@@ -53,13 +53,23 @@ We are iterating and looking for feedback and collaboration, so please [**let us
   <img src="https://img.shields.io/pub/v/amplify_api.svg">
 </a>
 
+## Category / Platform Support (Developer Preview)
```
dnys1, line 56: Can we add one for stable as well? Even though it's a bit redundant, it might be good to highlight for comparison with this.

dnys1 (approved):

README.md
haverchuck: @dnys1 Do you mean an equivalent table in the `stable` branch, or two tables in `next`?

dnys1: I think we need this table and a matching one for `stable` in the top-level `README` on `main` branch and in the package READMEs on both branches.

dnys1 (approved):

PR #2115, opened 2022-09-12
chore: adds platform/category table

haverchuck:
*Description of changes:*
Adds platform/category support table to READMEs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,8 @@ We are iterating and looking for feedback and collaboration, so please [**let us
 
 ## Category / Platform Support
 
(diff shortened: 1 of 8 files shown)


dnys1:
I think this is probably only worth adding if it's completed by the dev-preview table. Also, not sure it needs to be on stable packages' READMEs - maybe just the top-level README and the amplify_flutter package.

dnys1 (approved):
LGTM. Thanks!

ragingsquirrel3 (approved):
nice ASCII art

Issue #2105, opened 2022-09-09
Downloaded file is empty in iOS


jtn-d:
### Description

`downloadFile` seems to work fine (there is no errors) but when `readAsString` called content is empty. 

Same code works in Android. Using `getApplicationDocumentsDirectory` for directory doesn't make a difference.

```
final dir = await getTemporaryDirectory();
final key = 'foo/bar.txt';
final file = File('${dir.path}/$key');
await file.create(recursive: true);

final result = await Amplify.Storage.downloadFile(
  key: key,
  local: file,
);

print(await result.file.length()); // 0
print(await result.file.readAsString()); // ""
```

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [ ] Auth
- [ ] Authenticator
- [ ] DataStore
- [X] Storage

### Platforms

- [X] iOS
- [ ] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Environment

```bash
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.3.0, on macOS 12.3.1 21E258 darwin-arm, locale en-FI)
Checking Android licenses is taking an unexpectedly long time...[✓] Android toolchain - develop for Android devices (Android SDK version 31.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 13.4.1)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2021.1)
[✓] Android Studio (version 2021.2)
[✓] Android Studio (version 2020.3)
[✗] Cannot determine if IntelliJ is installed
    ✗ Directory listing failed
[✓] VS Code (version 1.71.0)
[✓] Connected device (4 available)
[✓] HTTP Host Availability

! Doctor found issues in 1 category.
```

### Dependencies

```bash
dependencies:
- amplify_analytics_pinpoint 0.6.6 [amplify_analytics_pinpoint_android amplify_analytics_pinpoint_ios amplify_core aws_common flutter meta]
- amplify_api 0.6.6 [amplify_api_android amplify_api_ios amplify_core amplify_flutter aws_common collection flutter meta plugin_platform_interface]
- amplify_auth_cognito 0.6.6 [amplify_auth_cognito_android amplify_auth_cognito_ios amplify_core aws_common collection flutter meta plugin_platform_interface]
[...]

PR #2116, opened 2022-09-13
fix(storage): Download to existing file

dnys1:
Fixes https://github.com/aws-amplify/amplify-flutter/issues/2105.

Fixes an issue where downloading to an existing file will not work on iOS. For whatever reason, it will report progress and success, but will ultimately not overwrite the contents of the file.


packages/storage/amplify_storage_s3/lib/method_channel_storage_s3.dart
```diff
@@ -152,7 +152,12 @@ class AmplifyStorageS3MethodChannel extends AmplifyStorageS3 {
       if (onProgress != null) {
         _transferProgressionCallbackMap[request.uuid] = onProgress;
       }
-
+      // Delete the file if it already exists since iOS will only write to
+      // a non-existent file; it will not overwrite the contents of an existing
+      // one.
+      if (Platform.isIOS && await request.local.exists()) {
+        await request.local.delete();
+      }
```
dnys1, line 160: This is the actual fix. The rest is updating tests and Xcode files.

Jordan-Nelson: If the download fails this will result in the original file being deleted without the new file being downloaded. Is that expected? Should this instead write to a temp directory and then copy it over?

dnys1: Since the intention is to overwrite whatever is passed in, then I think this is probably fine, although to be safe I can update it 👍

HuiSF:
Thanks for fixing this issue!

amplify-android implements the logic if file exists ([1.x](https://github.com/aws-amplify/aws-sdk-android/blob/main/aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java#L412), [2.x](https://github.com/aws-amplify/amplify-android/blob/dev-preview/aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/TransferManager.kt#L162)), it deletes the original file before write the new file. IMO, we should also file an issue to amplify-ios to implement the same logic instead of fixing in amplify-flutter (we may merge this PR as a temporary fix?). This behavior should be consistent cross Amplify platform libraries.

Jordan-Nelson (approved):

HuiSF (approved):

PR #2118, opened 2022-09-13
feat(secure_storage): remove items on re-install for iOS & MacOS

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Set flag in NSUserDefaults when secure storage has been initialized for a new scope
- Clear keychain items when initializing secure storage for a new scope
- Rename default scope name for Auth Cogntio from "auth" to "awsCognitoAuthPlugin"


packages/secure_storage/amplify_secure_storage/pigeons/messages.macos.dart
```diff
@@ -0,0 +1,37 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// To regenerate, run `make pigeons`.
+
+import 'package:pigeon/pigeon.dart';
+
+@ConfigurePigeon(
```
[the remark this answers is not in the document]
Jordan-Nelson: I considered just copying the output in the Makefile command, but there is a minor difference between the two files since `TaskQueue` isn't supported for MacOS.

packages/secure_storage/amplify_secure_storage/Makefile
```diff
@@ -1,4 +1,21 @@
...
+pigeons_ios:
+	flutter pub run pigeon --input pigeons/messages.ios.dart
+	flutter format --fix lib/src/messages.cupertino.g.dart
+
+pigeons_macos:
+	flutter pub run pigeon --input pigeons/messages.macos.dart
+	flutter format --fix lib/src/messages.cupertino.g.dart
+	# TODO(Jordan-Nelson): Remove when swift codegen in pigeon is no longer experimental.
+	# See: https://github.com/flutter/flutter/issues/73738#issuecomment-903927725
+	sed -i '' -e 's/\<Flutter\/Flutter\.h\>/\<FlutterMacOS\/FlutterMacOS\.h\>/' macos/Classes/Messages.m
```
Jordan-Nelson, line 21: *Note:* Pigeon doesn't officially support using objective-c for MacOS because flutter doesn't officially support using it for MacOS plugins. However, it works just fine with this one import update. Pigeon recently added support for swift codegen for MacOS, but swift codgen in pigeon is considered experimental.

I think it makes sense to stick with objective-c until the swift code generator is no longer "experimental". At that point we can switch iOS and MacOS over to using swift.

packages/secure_storage/amplify_secure_storage_dart/lib/src/types/macos_secure_storage_options.dart
```diff
@@ -26,29 +27,41 @@ abstract class MacOSSecureStorageOptions
...
   ///
   /// **Note:** [useDataProtection] must be set to true when specifying an [accessGroup].
   ///
-  /// **Note:** When [useDataProtection] is set to true, the app must be added to one
+  /// **Note:** Unless [useDataProtection] is set to false, the app must be added to one
   /// or more Keychain Access Groups in Xcode. For more info, see the
   /// [Platform Setup docs](https://docs.amplify.aws/lib/project-setup/platform-setup/q/platform/flutter/).
   factory MacOSSecureStorageOptions({
-    bool useDataProtection = true,
+    @visibleForTesting bool useDataProtection = true,
```
dnys1, line 64: If it's visiblefortesting only should it even be documented publicly?

I've been thinking about this generally, too. What about `bool useDataProtection = zReleaseMode` which would allow easier onboarding since even testing Auth apps locally is a pain having to configure Xcode and remember not to commit those changes.

dnys1 (approved):
Could be fun to migrate NS* APIs to ffigen at some point since it supports ObjC/Swift interop now. Looks good, though!

packages/secure_storage/amplify_secure_storage_dart/lib/src/types/macos_secure_storage_options.dart
Jordan-Nelson: Re: should it be documented - Maybe not. I think it would be valid for a customer to use it in a test if they wanted, but not sure we need to document it.

It would allow easier onboarding, but potentially lead to confusion down the line. I could see a customer not realizing there is a difference in release build and being pretty confused why their first release build has an issue. I have also stumbled upon some odd differences in keychain behavior when this is true/false (See [this comment](https://github.com/aws-amplify/amplify-flutter/pull/2118/files#diff-0454077cd10a395b9edaaca51c76b71f0504991655de69ed05f185dfd94aced5R178-R182)). That is why I added the `@visibleForTesting`. I want to really discourage customers from using it given the differences in keychain behavior. I actually want to update our tests to have this set to `true`. I have been running the tests locally with it set to true, but would like to get that in CI given the differences in behavior I have seen.

dnys1: Bummer, okay. That makes sense.

dnys1: I think eventually, then, we should consider removing the parameter altogether.

PR #2119, opened 2022-09-13
chore(api): Refactor http client & cancelable op

Equartey:
*Description of changes:*
Implements necessary changes to reconcile `feat/api-next` with `next`'s http client changes.


packages/api/amplify_api/lib/src/method_channel_api.dart
```diff
@@ -333,106 +333,130 @@ class AmplifyAPIMethodChannel extends AmplifyAPI {
...
+    return AWSHttpOperation(
+      _restFunctionHelper(
+        methodName: 'get',
+        path: path,
+        headers: headers,
+        queryParameters: queryParameters,
+        apiName: apiName,
+      ),
+      requestProgress: const Stream.empty(),
+      responseProgress: const Stream.empty(),
```
Equartey, line 351: The function changes in this file all include these lines. They are not correct and should be replaced with the proper streams.

packages/amplify_core/lib/src/plugin/amplify_api_plugin_interface.dart
```diff
@@ -53,7 +53,7 @@ abstract class APIPluginInterface extends AmplifyPluginInterface {
   void registerAuthProvider(APIAuthProvider authProvider);
 
   // ====== RestAPI ======
-  CancelableOperation<AWSStreamedHttpResponse> delete(
```
dnys1, line 56: Realizing that GraphQL operations should have their own type too probably, extending from AWSOperation @ragingsquirrel3 but that can probably be in a separated PR

packages/api/amplify_api/lib/src/method_channel_api.dart
dnys1: This file will be deleted soon so it's not a big deal, but yeah this is how you could handle it, or probably better would be `Stream.error` since `Stream.empty` just creates a Stream which immediately completes which could be misleading in this case since that is also how progress completion is conveyed typically.

dnys1 (approved):
Looks like tests still need to be update to pass but otherwise looks good

packages/api/amplify_api/lib/src/api_plugin_impl.dart
```diff
@@ -116,7 +115,7 @@ class AmplifyAPIDart extends AmplifyAPI {
   ///
   /// Use [apiName] if there are multiple endpoints of the same type.
   @visibleForTesting
-  http.Client getHttpClient(EndpointType type, {String? apiName}) {
+  AWSHttpClient getHttpClient(EndpointType type, {String? apiName}) {
```
ragingsquirrel3, line 118: I was getting a lot of errors in android w http, so would suggest we make the http clients set to http/1 for now. 

This got it working for me:

```
return _clientPool[endpoint.name] ??= AmplifyHttpClient(
      baseClient: AmplifyAuthorizationRestClient(
        endpointConfig: endpoint.config,
        baseClient: _baseHttpClient,
        authProviderRepo: _authProviderRepo,
      ),
    )..supportedProtocols = SupportedProtocols.http1;
```

packages/api/amplify_api/lib/src/amplify_authorization_rest_client.dart
```diff
@@ -14,47 +14,39 @@
...
   @override
-  Future<http.StreamedResponse> send(http.BaseRequest request) async =>
-      _baseClient.send(await authorizeHttpRequest(
-        request,
-        endpointConfig: endpointConfig,
-        authProviderRepo: authProviderRepo,
-      ));
+  final AWSHttpClient baseClient;
 
+  /// Implementation of [send] that authorizes any request without "Authorization"
```
ragingsquirrel3, line 42: This is sweet. Love how this file got simplified.

ragingsquirrel3 (approved):
awesome, thanks again for jumping on this! super helpful

PR #2123, opened 2022-09-14
chore(authenticator): simplify use of generics

Jordan-Nelson:
Due to ongoing issues with DDC which rear their head in certain unavoidable cases (like using zapp.run which compiles apps as a single DDC module)


dnys1 (approved):
Awesome, thanks! 🎉

ragingsquirrel3 (approved):

PR #2132, opened 2022-09-16
Analytics Dart First Initial Commit

fjnoyp:
*Description of changes:*

- delete old analytics files
- dart / flutter directory separation
- event caching
- all api methods implemented

TODO:
use secure_storage for saving global properties
upgrade code for old analytics
testing and bug fixing
redo unit tests/integration tests?


packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/app_lifecycle_provider.dart
```diff
@@ -0,0 +1,17 @@
+abstract class AppLifecycleProvider {
+  bool _isObserving = false;
+
+  AppLifecycleProvider();
+
+  void setOnForegroundListener(Function onForeground);
+  void setOnBackgroundListener(Function onBackground);
+
+  void startObserving() {
```
Jordan-Nelson, line 9: You might want to call `@mustCallSuper` here to enforce that these are called. Either that or just remove the behavior of `startObserving` & `stopObserving`.

 You re-implemented the `_isObserving` logic in `FlutterAppLifecycleProvider` instead of calling super. If this behavior is specific to `FlutterAppLifecycleProvider` then remove it from this class. If it should be part of all AppLifecycleProvider implementations, then add `@mustCallSuper` and call super from `FlutterAppLifecycleProvider`.

packages/analytics/amplify_analytics_pinpoint/lib/src/flutter_app_lifecycle_provider.dart
```diff
@@ -0,0 +1,57 @@
...
+    _isObserving = false;
+    WidgetsBinding.instance.removeObserver(this);
+  }
+
+  @override
+  void didChangeAppLifecycleState(AppLifecycleState state) {
+    switch (state) {
+      case AppLifecycleState.resumed:
+        _onForeground?.call();
+        // startSession
```
Jordan-Nelson, line 38: Are these comments supposed to be TODOs?

packages/analytics/amplify_analytics_pinpoint/lib/src/flutter_device_context_info_provider.dart
```diff
@@ -0,0 +1,71 @@
...
+  late final String? _make;
+  late final String? _model;
+  late final String? _modelVersion;
+  late final String? _platform;
+  late final String? _platformVersion;
+
+  FlutterDeviceContextInfoProvider();
+
+  @override
+  Future<void> getDeviceInfoDetails() async {
```
Jordan-Nelson, line 57: This pattern where all data is null until `getDeviceInfoDetails()` is called seems error prone to me. The consumer of this class has to know to call this before they try to access properties. Seems like an easy thing to mess up when making changes inside the flutter package in the future.

I would suggest something like:

```dart
abstract class DeviceContextInfoProvider {
  Future<DeviceContextInfo> getDeviceInfoDetails();
}

class DeviceContextInfo {
  const DeviceContextInfo({
    required this.countryCode,
    required this.locale,
    required this.timezone,
    required this.appName,
    required this.appPackageName,
    required this.appVersion,
    required this.make,
    required this.model,
    required this.modelVersion,
    required this.platform,
  });

  final String countryCode;
  final String locale;
  final String timezone;
  final String appName;
  final String appPackageName;
  final String appVersion;
  final String make;
  final String model;
  final String modelVersion;
  final String platform;
}

class FlutterDeviceContextInfoProvider implements DeviceContextInfoProvider {

  const FlutterDeviceContextInfoProvider();

  @override
  Future<DeviceContextInfo> getDeviceInfoDetails() async {
    var appInfo = await AppInfoProvider.getAppData();
    var deviceInfo = await DeviceInfoProvider.getDeviceData();
    return DeviceContextInfo(
      countryCode: WidgetsBinding.instance.window.locale.countryCode,
      locale: WidgetsBinding.instance.window.locale.toString(),
      timezone: DateTime.now().timeZoneName,
      appName: appInfo.appName,
      appPackageName: appInfo.packageName,
      appVersion: appInfo.version,
      make: deviceInfo.make,
      model: deviceInfo.model,
      modelVersion: deviceInfo.modelVersion,
      platform: deviceInfo.platform,
    );
  }
}
```

packages/analytics/amplify_analytics_pinpoint/pubspec.yaml
```diff
@@ -17,6 +17,9 @@ dependencies:
   flutter:
     sdk: flutter
   meta: ^1.7.0
+  plugin_platform_interface: ^2.0.0
```
Jordan-Nelson, line 20: Is this used?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -0,0 +1,125 @@
...
+import 'impl/analytics_client/shared_prefs.dart';
+import 'impl/device_context_info_provider.dart';
+
+// TODO - what are campaignattributes android? (setCampaignAttributes)
+
+/// Public facing Plugin
+/// Validate and parse inputs
+/// Before delegating to AnalyticsClient
+class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
+  late AnalyticsClient? _analyticsClient;
```
Jordan-Nelson, line 33: Why does this need to be a later initializer? Why does it need to be nullable?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/graveyard/event_field_manager.txt
```diff
@@ -0,0 +1,92 @@
+import 'dart:collection';
```
Jordan-Nelson, line 1: Was this intended to be committed?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/analytics_client.dart
```diff
@@ -0,0 +1,92 @@
...
+
+class AnalyticsClient {
+  // Should be managed by SessionTracker
+  late final SessionManager _sessionManager; // = SessionManager();
+
+  late final EventClient _eventManager;
+
+  late final StoppableTimer _autoEventSubmitter;
+
+  late final EventCreator _eventClient;
```
Jordan-Nelson, line 23: The names here are hard to follow. `_eventManager` is of type `EventClient`, and `_eventClient` is of type `EventCreator`. I would suggest renaming these.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_client.dart
```diff
@@ -0,0 +1,154 @@
...
+    }
+    if (copyFromProfile.email != null) {
+      addAttribute('email', [copyFromProfile.email!]);
+    }
+    if (copyFromProfile.plan != null) {
+      addAttribute('plan', [copyFromProfile.plan!]);
+    }
+
+    if (copyFromProfile.location != null) {
+      var newLoc = copyFromProfile.location!;
```
Jordan-Nelson, line 85: Please update this throughout the PR.
```suggestion
      final newLoc = copyFromProfile.location!;
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_global_fields_manager.dart
```diff
@@ -0,0 +1,110 @@
+import 'dart:collection';
+import 'dart:convert';
+import '../shared_prefs.dart';
+
+class EndpointGlobalFieldsManager {
```
Jordan-Nelson, line 5: Can you add doc comments throughout? I know this is still a draft, but it makes reviewing it much easier to read a comment and get a quick understanding of what this class does.

Jordan-Nelson (commented):
I didn't quite make it through everything, but here is some initial feedback

packages/analytics/amplify_analytics_pinpoint/example/pubspec.yaml
```diff
@@ -26,7 +26,8 @@ dependencies:
     sdk: flutter
 
 dev_dependencies:
-  amplify_lints: ^2.0.0
+  amplify_lints:
```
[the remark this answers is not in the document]
fjnoyp: Thanks updated

packages/analytics/amplify_analytics_pinpoint/lib/src/device_info_provider.dart
```diff
@@ -0,0 +1,64 @@
...
+
+import 'package:device_info_plus/device_info_plus.dart';
+import 'package:flutter/foundation.dart';
+import 'package:flutter/services.dart';
+
+class DeviceInfo {
+  String? make; // manufacturer
+  String? model; // device model?
+  String? modelVersion;
+  String? platform; // web/ios/android
```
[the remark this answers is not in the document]
fjnoyp: switched to enum

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/app_lifecycle_provider.dart
fjnoyp: Thanks removed the abstract class method implementations.

packages/analytics/amplify_analytics_pinpoint/lib/src/flutter_device_context_info_provider.dart
fjnoyp: Thanks added the changes

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
fjnoyp: late => because it is only initialized in configure call.

nullable => in case user forgets to call configure, need way to check if analyticsClient is not enabled.  I have just updated code in this class to check nullable value and throw exception otherwise.

PR #2143, opened 2022-09-20
feat(api): authorizationMode property for GraphQLRequest

ragingsquirrel3:
Adds the `authorizationMode` parameter to `GraphQLRequest` which will override the default auth mode of the endpoint for the request. Works for query, mutation and subscription. Has a lot of unit/integ tests.

There is also a small change to expose `baseHttpClient` property of the high-level plugin.


packages/api/amplify_api/example/integration_test/util.dart
```diff
@@ -165,50 +165,28 @@ Future<Post> addPostAndBlog(
...
-  final headers = {AWSHeaders.authorization: accessToken};
-
+GraphQLRequest<T> authorizeRequestForUserPools<T>(GraphQLRequest<T> request) {
   return GraphQLRequest<T>(
     document: request.document,
     variables: request.variables,
     modelType: request.modelType,
     decodePath: request.decodePath,
-    headers: headers,
+    authorizationMode: APIAuthorizationType.userPools,
```
dnys1, line 174: Is APIAuthorizationType the correct name? I just realized we call it `mode` almost everywhere else

packages/api/amplify_api/lib/src/decorators/authorize_http_request.dart
```diff
@@ -24,11 +25,13 @@ Future<AWSBaseHttpRequest> authorizeHttpRequest(
   AWSBaseHttpRequest request, {
   required AWSApiConfig endpointConfig,
   required AmplifyAuthProviderRepository authProviderRepo,
+  APIAuthorizationType? authorizationMode,
 }) async {
-  if (request.headers.containsKey(AWSHeaders.authorization)) {
+  if (request.headers.containsKey(AWSHeaders.authorization) ||
+      request.headers.containsKey(xApiKey)) {
```
dnys1, line 31: Is this correct? What if the user's lambda wants to repurpose this header? Probably need to check that `authType == api key` as well.

packages/api/amplify_api/test/ws/web_socket_auth_utils_test.dart
```diff
@@ -75,11 +80,38 @@ void main() {
...
 
-      _assertBasicSubscriptionPayloadHeaders(payload);
+      assertBasicSubscriptionPayloadHeaders(payload);
       expect(
         payload.authorizationHeaders[xApiKey],
         testApiKeyConfig.apiKey,
       );
     });
+
+    test('should generate an authorized message with custom authorizationMode',
```
dnys1, line 90: This shows user pools unless I'm missing something

ragingsquirrel3: Yes, authorizes w user pools where the config's default mode is api key.

packages/api/amplify_api/lib/src/decorators/authorize_http_request.dart
ragingsquirrel3: I will add check for `== apiKey`.

packages/api/amplify_api/example/integration_test/util.dart
ragingsquirrel3: Are suggesting maybe renaming the enum to `APIAuthorizationMode`?  That might be better as I believe usage in gql request is the most prominent (kinda only) place we are exposing publicly so might be better if matched. On the other hand, I think "Type" is ok and does not have to match the property name in this case.

packages/amplify_core/lib/src/types/api/graphql/graphql_request.dart
```diff
@@ -19,17 +19,21 @@ import 'package:amplify_core/amplify_core.dart';
...
+  /// The body of the request, starting with the operation type and operation name.
+  ///
+  /// See https://graphql.org/learn/queries/#operation-name for examples and more information.
+  final String document;
+
   /// Only required if your backend has multiple GraphQL endpoints in the amplifyconfiguration.dart file. This parameter is then needed to specify which one to use for this request.
   final String? apiName;
 
+  /// Authorization type to use for this request. If not supplied, request will use
+  /// default endpoint mode. Note: this is an enum already defined in core.
```
Jordan-Nelson, line 31: First line should be a single sentence (https://dart.dev/guides/language/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph).

I'm not sure that `Note: this is an enum already defined in core.` should be included. Is that intended to improve discoverability for the customer? Maybe we re-export export that enum api to make it more discoverable?

```suggestion
  /// Authorization type to use for this request.
  ///
  /// If not supplied, the request will use the default endpoint mode.
```

packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
```diff
@@ -27,28 +24,15 @@ void main({bool useExistingTestUser = false}) {
...
     });
 
     group('queries', () {
       testWidgets('should query authorized model', (WidgetTester tester) async {
-        final req = authorizeRequestForApiKey(
-          ModelQueries.list<Blog>(Blog.classType),
-          apiKey,
-        );
+        final req =
+            authorizeRequestForApiKey(ModelQueries.list<Blog>(Blog.classType));
```
Jordan-Nelson, line 35: *nit:* Might want to update the lint rules if "library" from amplify_lints isn't already being used.
```suggestion
        final req = authorizeRequestForApiKey(
            ModelQueries.list<Blog>(Blog.classType),
        );
```

Jordan-Nelson (approved):

packages/amplify_core/lib/src/types/api/graphql/graphql_request.dart
ragingsquirrel3: good catch, thanks, this was bad copy pasta from design doc

packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
ragingsquirrel3: plan to do as separate PR bc anticipating lots of related changes

packages/api/amplify_api/example/integration_test/util.dart
ragingsquirrel3: I just pushed a commit that renamed this everywhere and re-generated the .g.dart files that get affected.

packages/amplify_core/lib/src/config/api/appsync/api_config.dart
```diff
@@ -22,7 +22,7 @@ class AWSApiConfig with AWSEquatable<AWSApiConfig>, AWSSerializable {
   final EndpointType endpointType;
   final String endpoint;
   final String region;
-  final APIAuthorizationType authorizationType;
+  final APIAuthorizationMode authorizationType;
```
dnys1, line 25: Oh this is why it was called authorization type - I think we should just align with what other SDKs are doing

ragingsquirrel3: discussed offline but changed it back

packages/api/amplify_api/example/integration_test/util.dart
ragingsquirrel3: and reverted this

dnys1 (approved):

PR #2147, opened 2022-09-21
chore(version): Bump version

dnys1:


packages/amplify_datastore/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.0.0-next.0+1
+
+- Minor bug fixes and improvements
```
Jordan-Nelson, line 3: *nit:* I had a similar comment the automated version bump PR. In this situation, "Internal dependencies bumped to latest version" would make more sense to me. "Minor bug fixes and improvements" isn't necessarily incorrect, but since the changes were not in this package it doesn't seems like the most appropriate message.

Jordan-Nelson (approved):

packages/amplify_datastore/CHANGELOG.md
dnys1: Hmm, not sure it's worth the effort distinguishing these scenarios. This message appears in one of three cases:
- The DataStore package's list of commits consisted purely of chores
- The DataStore package's list of commits consisted purely of cascading updates in dependencies
- Some mix of the two

It would be probably more effort than it's worth to suss out which of the three actually occurred. "Minor bug fixes and improvements" seems to cast a wider net and is applicable in all three scenarios. In this case, the DataStore bump is due to a bump in its dependencies plus a change in the README, so both messages would really be needed to describe the full range of changes.

Maybe instead of an either/or, we could add "Internal dependencies bumped to latest version" in the case of cascading updates. This could be done with some sort of marker commit.

packages/aft/pubspec.yaml
```diff
@@ -9,14 +9,14 @@ environment:
...
-  aws_common: ^0.2.0
+  aws_common: ">=0.2.3 <0.3.0"
   checked_yaml: ^2.0.0
   cli_util: ^0.3.5
   collection: ^1.16.0
   flutter_tools:
     git:
       url: https://github.com/flutter/flutter.git
-      ref: f3d182f4b4e3c5e5cde322f5a14e49e9a50cc083
+      ref: 0df0e2ea1f7fe5021c3593d61903e70bdbda8107
```
HuiSF, line 19: Curious how do we determine what `ref` to use?

packages/auth/amplify_auth_cognito/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: amplify_auth_cognito
 description: The Amplify Flutter Auth category plugin using the AWS Cognito provider.
-version: 1.0.0-next.0+4
+version: 1.0.0-next.0+5
```
HuiSF, line 3: In the changelogs we mentioned `feat(*)`, should be version be `1.0.0-next.1`?

HuiSF (approved):

packages/auth/amplify_auth_cognito/pubspec.yaml
dnys1: It could be either. I was saving `.` bumps for category additions, so `1.0.0-next.1` would be for API/storage/analytics.

packages/aft/pubspec.yaml
dnys1: Trial and error unfortunately. It's finding the latest versions that don't have conflicting dependencies + a bit of tweaking dependency_overrides to get it all to align.

PR #2148, opened 2022-09-21
chore(api,core): change API types

ragingsquirrel3:
This PR makes some changes to API category user-facing API (hate saying that) now that API review is finished. In summary:
* brings back `GraphQLOperation` for query/mutate but based on `CancelableOperation` 
* brings back `RestOperation` with a factory to create from `AWSHttpOperation` and ensure `.response` not streamed type
* brings back `RestException` and throws in REST methods when response not successful.
* pass custom headers to GraphQL subscription registration methods
* some formatting changes in the files I touched here

Also changes tests and makes a little change to `AWSHttpClient` to properly handle exceptions during `transformResponse`.


packages/aws_common/lib/src/http/aws_http_client.dart
```diff
@@ -131,8 +131,12 @@ abstract class AWSBaseHttpClient extends AWSCustomHttpClient {
     responseProgressCompleter.setSourceStream(operation.responseProgress);
     operation.operation.then(
       (resp) async {
-        resp = await transformResponse(resp);
-        completer.complete(resp);
+        try {
```
ragingsquirrel3, line 134: @dnys1 double-checking this change makes sense? Without this, exceptions thrown in `transformResponse` weren't getting caught as expecting when await operation `.response`.

packages/api/amplify_api/lib/src/decorators/web_socket_auth_utils.dart
```diff
@@ -109,7 +113,10 @@ Future<Map<String, String>> _generateAuthorizationHeaders(
   final maybeConnect = isConnectionInit ? '/connect' : '';
   final canonicalHttpRequest = AWSStreamedHttpRequest.post(
     Uri.parse('${config.endpoint}$maybeConnect'),
-    headers: _requiredHeaders,
+    headers: {
+      ..._requiredHeaders,
+      ...(customHeaders ?? {}),
```
dnys1, line 118: I would switch the order so that custom headers cannot overwrite required headers

```suggestion
      ...?customHeaders,
      ..._requiredHeaders,
```

packages/aws_common/lib/src/http/aws_http_client.dart
```diff
@@ -131,8 +131,12 @@ abstract class AWSBaseHttpClient extends AWSCustomHttpClient {
...
+          resp = await transformResponse(resp);
+          completer.complete(resp);
+        } on Exception catch (e, st) {
```
dnys1, line 137: Yeah that's fine, but let's add some tests since I didn't understand that about `CancelableOperation.then`

Also, let's handle all errors and exceptions this way so that errors don't get lost either.

```suggestion
        } on Object catch (e, st) {
```

packages/api/amplify_api/lib/src/amplify_authorization_rest_client.dart
```diff
@@ -30,30 +39,42 @@ class AmplifyAuthorizationRestClient extends AWSBaseHttpClient {
...
     );
   }
+
+  @override
+  Future<AWSBaseHttpResponse> transformResponse(
+    AWSBaseHttpResponse response,
+  ) async {
+    // For REST endpoints, throw [RestException] on non-successful responses.
+    if (endpointConfig.endpointType == EndpointType.rest &&
+        (response.statusCode < 200 || response.statusCode >= 300)) {
```
Jordan-Nelson, line 69: Are we throwing for 3xx?

Jordan-Nelson (approved):

packages/api/amplify_api/lib/src/amplify_authorization_rest_client.dart
ragingsquirrel3: Yes. 1) to be consistent w current behavior 2) though I guess you could make the case for following redirects that's not a change to behavior that was in API review.

Jordan-Nelson: Got it. Did not know that was the current behavior.

packages/aws_common/lib/src/http/aws_http_client.dart
ragingsquirrel3: Made this change, plus added some unit tests in aws_common to test this behavior in request/response transform (though the REST tests in API category would fail as well without this change).

dnys1 (approved):

PR #2154, opened 2022-09-22
test(analytics): Integration test backend

dnys1:
Currently, there are no integration tests for Analytics. Part of the problem is that the Analytics API consists entirely of "fire-and-forget" methods. Even `flushEvents` does not throw an error when it fails - all events are meant to be async.

In order to test that the category is working as expected, we employ a method similar to what's done in the Auth category for capturing SMS messages. Namely, we route all Pinpoint events to an AppSync subscription we can listen on. This is done with the help of a Kinesis stream (to which Pinpoint can natively export) and a Lambda function (which can be triggered by a Kinesis stream).

```mermaid
flowchart LR
    F[Flutter] -->|Analytics| P[Pinpoint];
    P -->|Event Stream| K[Kinesis Data Stream];
    K -->|Consumer| L[Lambda];
    L -->|GraphQL| A[AppSync];
    A -->|Subscription| F;
```


packages/analytics/amplify_analytics_pinpoint/example/backend/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "version": "0.1.0",
+  "bin": {
+    "backend": "bin/backend.js"
+  },
+  "scripts": {
+    "build": "tsc",
+    "watch": "tsc -w",
+    "test": "jest",
+    "cdk": "cdk",
+    "deploy": "cdk deploy --outputs-file=outputs.json && tool/create_config.dart"
```
ragingsquirrel3, line 12: I got an error first time I ran this bc needed to run `cdk boostrap`. Can that be added here?

packages/analytics/amplify_analytics_pinpoint/example/backend/bin/backend.ts
```diff
@@ -0,0 +1,20 @@
...
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+import 'source-map-support/register';
+import * as cdk from 'aws-cdk-lib';
+import { BackendStack } from '../lib/backend-stack';
+
+const app = new cdk.App();
+new BackendStack(app, 'BackendStack');
```
ragingsquirrel3, line 20: request: Could you give this a more descriptive name so we can find it in cloudformation?

packages/analytics/amplify_analytics_pinpoint/example/tool/provision_integration_test_resources.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -e
+
```
ragingsquirrel3, line 4: request: Could this script take an `AWS_PROFILE` env variable like the other provision scripts? I believe `cdk` commands will take that param downstream.

packages/analytics/amplify_analytics_pinpoint/example/backend/lib/backend-stack.ts
```diff
@@ -0,0 +1,244 @@
...
+import * as iam from "aws-cdk-lib/aws-iam";
+import * as lambda from "aws-cdk-lib/aws-lambda";
+import * as lambda_nodejs from "aws-cdk-lib/aws-lambda-nodejs";
+import * as lambda_events from "aws-cdk-lib/aws-lambda-event-sources";
+import * as appsync from "@aws-cdk/aws-appsync-alpha";
+import * as dynamodb from "aws-cdk-lib/aws-dynamodb";
+import * as path from "path";
+import { CfnOutput, Duration, Expiration, RemovalPolicy } from "aws-cdk-lib";
+
+export class BackendStack extends cdk.Stack {
```
ragingsquirrel3, line 28: request: Could this stack also include an s3 bucket with name that matches this pattern https://github.com/aws-amplify/amplify-ci-support/blob/main/src/integ_test_resources/flutter/amplify/cloudformation_template.yaml#L32 (starts with "amplify-test") and then have the provision script upload the amplifyconfig file it creates to that bucket? That way, someone else (or the CI) can pull it down as long as we later add a pull script. If the bucket has that name pattern then existing gh action role should have no issue. I'm also open to other ideas to solve that problem but would like for it to be part of this stack.

ragingsquirrel3 (commented):
Super innovative, nice! Worked for me after running `cdk bootstrap`. Left some comments around making the stack/script more useful.

packages/analytics/amplify_analytics_pinpoint/example/backend/package.json
dnys1: Maybe I can add it to the README? `cdk bootstrap` only needs to be run once per account in order to use CDK in general. It's not project-specific, though, and I don't think calling it multiple times is best.

packages/analytics/amplify_analytics_pinpoint/example/backend/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "backend",
+  "version": "0.1.0",
```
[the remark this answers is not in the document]
dnys1: Yeah, sounds good

packages/analytics/amplify_analytics_pinpoint/example/backend/package.json
dnys1: Actually, I take that back. Looks like bootstrap can, and maybe even should, be run multiple times, since the resources it deploys need to be updated occasionally. I've added a separate npm script to do this though so it's only called when running the provisioning script or manually.

packages/analytics/amplify_analytics_pinpoint/example/tool/provision_integration_test_resources.sh
dnys1: It looks like npm scripts can manage this so I've updated the scripts to use the AWS_PROFILE env variable

packages/analytics/amplify_analytics_pinpoint/example/backend/lib/backend-stack.ts
dnys1: Sure - would I just write the amplifyconfiguration.dart to the root of the bucket?

ragingsquirrel3: That's what I was thinking. The pull script (not needed in this PR) can just use aws CLI command to download.

packages/analytics/amplify_analytics_pinpoint/example/backend/lib/backend-stack.ts
```diff
@@ -0,0 +1,244 @@
...
+          appsync.PrimaryKey.partition("id").is("input.id"),
+          appsync.Values.projecting("input")
+        ),
+        responseMappingTemplate: appsync.MappingTemplate.dynamoDbResultItem(),
+      });
+
+    // Create the Kinesis consumer Lambda which will capture events from the
+    // Kinesis Data Stream and forward them to AppSync.
+
+    const kinesisConumer = new lambda_nodejs.NodejsFunction(
```
fjnoyp: minor nit: typo consumer

packages/analytics/amplify_analytics_pinpoint/example/backend/lib/schema.graphql
```diff
@@ -0,0 +1,22 @@
+type Query {
```
fjnoyp, line 1: This is really great work ... 

Want to make a note here for future work to possibly test the 'updateEndpoint' call being made to Pinpoint as well.  Test should validate that it's called before events are being sent as otherwise Pinpoint won't register those events.  It is the responsibility of the Analytics plugin to automatically call `updateEndpoint` on session start. 

This call will change in the future though as we switch to the new User Centric Model API from Pinpoint.  Regardless, events will need to be associated to either a Pinpoint Endpoint or User object so we'll still want to monitor update calls to Endpoint/User.

fjnoyp (approved):
Really great work here ... Thanks for tackling this!

Left a minor comment about testing for calls to Pinpoint updateEndpoint. 

While I reviewed the ts code, I'm new to ts and cdk so might want to find someone else to take another look at the code.  From what I saw, the code seems to be logically organized and written.

packages/analytics/amplify_analytics_pinpoint/example/integration_test/events_test.dart
```diff
@@ -0,0 +1,132 @@
...
+        if (data == null) {
+          return const <String, Object?>{};
+        }
+        final json = jsonDecode(data) as Map<String, Object?>;
+        return jsonDecode((json['onCreateRecord'] as Map)['payload'] as String)
+            as Map<String, Object?>;
+      }).asBroadcastStream();
+
+      eventsStream.listen((event) {
+        print('Got event: ${jsonEncode(event)}');
```
ragingsquirrel3, line 75: important request: There is some info logged here that I don't think we will want in logs once this hits CI (e.g. account ID). Would suggest either not logging this at all or picking off some specific values from the object.

packages/analytics/amplify_analytics_pinpoint/example/integration_test/events_test.dart
```diff
@@ -0,0 +1,132 @@
...
+              // There is some lossiness behavior to doubles
+              Map.fromEntries([lossyDoubleProperty, intProperty]),
+            ),
+          ])),
+        );
+
+        await Amplify.Analytics.recordEvent(event: customEvent);
+        await Amplify.Analytics.flushEvents();
+      },
+      timeout: const Timeout(Duration(minutes: 1)),
```
ragingsquirrel3, line 129: nit/suggestion: I got some timeouts when trying this and raising to 2 seemed to help 😢 .

packages/analytics/amplify_analytics_pinpoint/example/integration_test/events_test.dart
```diff
@@ -0,0 +1,132 @@
...
+  group('recordEvent', () {
+    const boolProperty = MapEntry('boolProperty', true);
+    // Boolean properties are recorded as strings
+    const stringifiedBoolProperty = MapEntry('boolProperty', 'true');
+
+    const doubleProperty = MapEntry('doubleProperty', pi);
+    // There is some lossiness behavior to doubles
+    const lossyDoubleProperty = MapEntry('doubleProperty', 3.141592653589793);
+
+    const intProperty = MapEntry('intProperty', 42);
```
ragingsquirrel3, line 37: question/suggestion: Should one of the properties of the event be random (like a UUID) so we can ensure received event came from this invocation and not from some other client?

packages/analytics/amplify_analytics_pinpoint/example/integration_test/events_test.dart
```diff
@@ -0,0 +1,132 @@
...
+
+      eventsStream = Amplify.API
+          .subscribe(
+        GraphQLRequest<String>(document: '''
+subscription {
+  onCreateRecord {
+    id
+    payload
+  }
+}
```
ragingsquirrel3, line 58: question: Should this `setUpAll` block wait for `onEstablished` to be called to ensure the subscription setup before we start sending events to Amplify? Seems unlikely but checking.

packages/analytics/amplify_analytics_pinpoint/example/tool/provision_integration_test_resources.sh
```diff
@@ -0,0 +1,12 @@
...
+pushd backend
+npm ci
+npm run setup
+npm run deploy
+popd
+
+BUCKET=$(jq -r .AnalyticsIntegrationTestStack.BucketName backend/outputs.json)
+aws --profile=${AWS_PROFILE:=default} s3 cp lib/amplifyconfiguration.dart s3://$BUCKET/amplifyconfiguration.dart
```
ragingsquirrel3, line 12: nice, love the changes to this script. Also seemed to work well when I provided non-default AWS_PROFILE.

ragingsquirrel3 (approved):
nice! 🚢

packages/analytics/amplify_analytics_pinpoint/example/backend/lib/schema.graphql
fjnoyp: Nice that's really great!

fjnoyp (approved):

PR #2155, opened 2022-09-22
chore: add `.gitattributes` for generated files

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Update pigeon generated code to use consistent naming and/or directories for generated code
- Add `.gitattributes` for code generated by Flutter create in sample apps, pigeon, sdk, and other generated dart code

**Note:** You should see files such as `NativeAuthPluginBindings.java` automatically collapses when reviewing this PR in GitHub. These files were not automatically collapsed previously.


.gitattributes
```diff
@@ -0,0 +1,25 @@
...
+.github/workflows/native_test.yml linguist-generated=true
+tool/ci.sh linguist-generated=true
+
+## Generated by Pigeon
+**/pigeons/*.java linguist-generated=true
+**/pigeons/*.h linguist-generated=true
+**/pigeons/*.m linguist-generated=true
+
+## Generated SDK files
+packages/**/src/lib/sdk/src linguist-generated=true
```
dnys1, line 22:
```suggestion
packages/**/lib/src/sdk/ linguist-generated=true
```

dnys1 (approved):
Can we add `package-lock.json`?

.gitattributes
Jordan-Nelson: Oops. Mixed up `lib` and `src` here. Will update.

Re: trailing `src` - Some files under `sdk` are not generated like `sdk_exception.dart`. Maybe we move those to a new folder (`/src/bridge`?) which I think would allow us to exclude them.

HuiSF (approved):

dnys1:
Hey sorry can add goldens too? They follow a different structure than the other Smithy SDKs.

```
packages/smithy/goldens/
```

PR #2166, opened 2022-09-23
feat(smithy): `@httpChecksum` trait

dnys1:
Depends on #2124, #2151.

Implements the runtime and code generation components of the `@httpChecksum` trait which is required by some S3 operations.

https://awslabs.github.io/smithy/2.0/aws/aws-core.html#aws-protocols-httpchecksum-trait

### DIFF
--- a/packages/smithy/smithy_aws/lib/src/http/interceptors/with_checksum.dart
+++ b/packages/smithy/smithy_aws/lib/src/http/interceptors/with_checksum.dart
@@ -1,0 +1,108 @@
+// Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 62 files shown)


packages/smithy/smithy_aws/lib/src/http/interceptors/with_checksum.dart
```diff
@@ -0,0 +1,108 @@
...
+/// {@template smithy_aws.interceptors.with_checksum}
+/// Intercepts HTTP requests to include a content hash in the headers.
+///
+/// See: https://awslabs.github.io/smithy/2.0/aws/aws-core.html#aws-protocols-httpchecksum-trait
+/// {@endtemplate}
+class WithChecksum extends HttpRequestInterceptor {
+  /// {@macro smithy_aws.interceptors.with_checksum}
+  const WithChecksum([String? algorithm]) : _algorithm = algorithm ?? 'MD5';
+
+  final String _algorithm;
```
HuiSF, line 70: Can the type of this field be `ChecksumAlgorithm` enum ?

HuiSF (approved):

packages/smithy/smithy_aws/lib/src/http/interceptors/with_checksum.dart
dnys1: Unfortunately, no, with the way Smithy works, since the checksumAlgorithm needs to be part of the input and the input has to be fully defined in Smithy and is unique to that service.

The only way around it would be for Smithy to define a `ChecksumAlgorithm` type as a core type which could be used here, but it doesn't do that.

Jordan-Nelson (approved):

PR #2170, opened 2022-09-23
chore(datastore): Migrate `ModelAssociation`

dnys1:
Migrates `ModelAssociation` and `ModelAssociationEnum` to the new MIPR format.

Internally breaking changes:
- Renames `ModelAssociationEnum` types to camelCase
- `ModelAssociationEnum` has been renamed to `ModelAssociationType`

---

**Stack**:
- #2187
- #2186
- #2181
- #2180
- #2178
- #2177
- #2170 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #2177, opened 2022-09-26
chore(datastore): Migrate `ModelFieldType`

dnys1:
Migrates `ModelFieldType`, replacing with new `SchemaType` from MIPR format.

Internally breaking changes:
- Replaces `ModelFieldType` with `SchemaType` in public APIs (although internally, `ModelFieldType` is still retained for serialization behavior)

---

**Stack**:
- #2187
- #2186
- #2181
- #2180
- #2178
- #2177 ⬅
- #2170


packages/amplify_core/lib/src/types/models/mipr/schema_type.dart
```diff
@@ -0,0 +1,644 @@
...
+  /// A custom structure with no `@model` directive.
+  ///
+  /// ```graphql
+  /// # `MyNonModel` is a non-model type.
+  /// type MyNonModel {
+  ///   name: String!
+  /// }
+  /// ```
+  /// {@endtemplate}
+  const factory SchemaType.nonModel(
```
HuiSF, line 258: I couldn't find non-model references from both legacy and transformer v2 backed data modeling documentation...
So I'd recommend to name this type `customType` following the name used by Amplify Studio which is a public facing interface. 

image

packages/amplify_core/lib/src/types/models/mipr/schema_type.dart
```diff
@@ -0,0 +1,644 @@
...
+/// {@template amplify_core.models.schema_type}
+/// An Amplify schema type, either a scalar, model, non-model, or enum.
+///
+/// - [ScalarType]
+/// - [ModelType]
+/// - [NonModelType]
+/// - [EnumType]
+/// - [ListType]
+/// {@endtemplate}
+abstract class SchemaType with AWSSerializable<Map<String, Object?>> {
```
HuiSF, line 208: Not sure if `SchemaType` is more accurate name than `FiledType`. To me "schema" is much more broad covering object specification, query/mutation selection set etc.

packages/amplify_core/lib/src/types/models/mipr/schema_type.dart
dnys1: NonModel is how they call it in amplify-codegen. Also, I think customType is kind of misleading - all model, non-model, and enum types are custom types.

packages/amplify_core/lib/src/types/models/mipr/schema_type.dart
dnys1: FieldType is not fully accurate though. These types will be used in other places beyond a field context.

packages/amplify_core/lib/src/types/models/mipr/schema_type.dart
dnys1: For public APIs we can keep it customType, but this would be an internal API.

packages/amplify_core/lib/src/types/models/mipr/schema_type.dart
dnys1: SchemaType = a type defined in the schema

dnys1: Would it make more sense to remove `isRequired` from `SchemaType` and wrap it in a `ModelFieldType` like it was before. I started with this but was trying to consolidate:

dnys1: Actually the problem was that describing recursive types was very difficult. For example, it wasn't possible before to distinguish `[String]!` from `[String!]` or `[String!]!`

HuiSF: >Would it make more sense to remove isRequired from SchemaType and wrap it in a ModelFieldType like it was before? I started with this but was trying to consolidate.

Yes I think that would make it clearer, not mixing a property with a `field`, so "SchemaType = a type defined in the schema" would make more sense 👍

packages/amplify_core/lib/src/types/models/mipr/schema_type.dart
HuiSF: Non-model could be type other than model including scalar, enum ha! confusing either way. I agree we only need to keep the naming consistent on public APIs. (as long as it's not `embedded` 😅)

dnys1: Haha true, it's not much better. Only in the sense that it would align with codegen at this layer.

packages/amplify_core/lib/src/types/models/mipr/schema_type.dart
dnys1: Okay, I mean I don't think there's anything field-specific here. Maybe the confusion is just that it's called `isRequired`. If it was `nonNull` or something, it might be easier to understand. But I do think this is the best way to represent a type in any context.

dnys1: I can try again with separating the two. I guess that might work since lists are only ever found in a field context.

dnys1: It would basically require wrapping SchemaType like this, removing `isRequired` and `ListType` from `SchemaType`. So `SchemaType` would represent just the name and kind of the type basically.

Thoughts?

```dart
abstract class ModelFieldType {
  const ModelFieldType();

  bool get isRequired;
}

class ModelFieldBaseType extends ModelFieldType {
  const ModelFieldBaseType(this.schemaType, {required this.isRequired});

  final SchemaType schemaType;

  @override
  final bool isRequired;
}

class ModelFieldListType extends ModelFieldType {
  const ModelFieldListType(this.elementType, {required this.isRequired});

  final ModelFieldType elementType;

  @override
  final bool isRequired;
}
```

HuiSF (approved):

packages/amplify_core/lib/src/types/models/mipr/schema_type.dart
dnys1: Resolved offline - decided to stay with SchemaType as-is

Jordan-Nelson (approved):

PR #2179, opened 2022-09-26
fix(api): correct subscription error handling

ragingsquirrel3:
This PR fixes error handling of a simple invalid GraphQL subscription (like malformed request document). It handles by creating a `GraphQLResponse` with an error and sending in the stream data event (not the error stream). You could make a case for sending as error, not normal event. However, this approach is consistent with stable on Android (iOS just throws a general exception not in stream which I don't think is correct). I also noticed there was an error with canceling these subscriptions so I fixed that, too.


packages/api/amplify_api/example/integration_test/graphql/iam_test.dart
```diff
@@ -369,6 +369,35 @@ void main({bool useExistingTestUser = false}) {
...
+          await expectLater(
+            stream,
+            emits(predicate<GraphQLResponse<String>>(
+              (GraphQLResponse<String> response) =>
+                  response.hasErrors && response.data == null,
+              'Has GraphQL Errors',
+            )),
+          );
+          // Cleanup.
+          await stream.listen((_) {}).cancel();
```
dnys1, line 399: This is typically how you'll see it

```suggestion
          await stream.listen(null).cancel();
```

Does this test the issue though? I thought the issue was calling cancel and then receiving an error? This will wait for the error and then call cancel. Should it be like this instead?

```dart
          expect(
            stream,
            emits(predicate<GraphQLResponse<String>>(
              (GraphQLResponse<String> response) =>
                  response.hasErrors && response.data == null,
              'Has GraphQL Errors',
            )),
          );
          // Cleanup.
          await stream.listen(null).cancel();
```

ragingsquirrel3: Thanks, that helps. I found to repro the issue I had to also add a few second delay after this to give the channel time to throw an error. Otherwise, the error gets thrown in the internal stream and throws error after the test is completed.

packages/api/amplify_api/test/util.dart
```diff
@@ -150,21 +150,39 @@ class MockWebSocketConnection extends WebSocketConnection {
...
-              ),
-            )
-            ..add(
-              WebSocketMessage(
-                messageType: MessageType.data,
-                id: messageFromEvent.id,
-                payload:
-                    const SubscriptionDataPayload(mockSubscriptionData, null),
-              ),
+          if (!event.contains('Invalid')) {
```
Equartey, line 153: This logic is going to need refactoring when reconnection gets merged. This whole function block will no longer exist. Mock messages now get put directly onto the channel stream. An example of how I replaced response mocking is [here](https://github.com/Equartey/amplify-flutter/blob/b1f162b1ea9c1cd3e73d08682bdb83f0ddb01b31/packages/api/amplify_api/test/util.dart#L164)

packages/api/amplify_api/test/ws/web_socket_types_test.dart
```diff
@@ -0,0 +1,89 @@
...
+              'message': 'Unknown operation id abc-456'
+            }
+          ]
+        }
+      },
+      MessageType.error,
+    ),
+  ];
+
+  group('WebSocketMessage should create expected messages from JSON', () {
```
Equartey, line 81: I like this test, this will be a useful reference when working with subscriptions. Nice work 🔥

packages/api/amplify_api/test/dart_graphql_test.dart
```diff
@@ -375,5 +375,31 @@ void main() {
       await operation.operation.valueOrCancellation();
       expect(operation.operation.isCanceled, isTrue);
     });
+
+    test('stream should emit response with error when subscription fails',
```
Equartey, line 379: Similar comment from the utils file, test will need slight refactoring to get the response from the web socket you're looking for here.

Equartey (commented):
Nice work, really like the improvements. Just a couple considerations...

packages/api/amplify_api/test/dart_graphql_test.dart
ragingsquirrel3: I think it would be tough to mock this behavior so I just left out the unit tests here and tried to make the integration test better.

packages/api/amplify_api/test/util.dart
ragingsquirrel3: Couldn't see a good way to mock this w new mock impl so just did it with integration test.

Equartey (approved):
Nice! 🚢

packages/api/amplify_api/example/integration_test/graphql/iam_test.dart
```diff
@@ -369,6 +370,52 @@ void main({bool useExistingTestUser = false}) {
...
+            )),
+          );
+          // Cancel subscription that had an error.
+          await streamWithError.listen(null).cancel();
+          // Give AppSync a few seconds to send an error, which happens when
+          // canceling a failed subscription and throws if not handled correctly.
+          // Needs to be on a canceled error subscription with an open connection.
+          await Future<void>.delayed(const Duration(seconds: 3));
+          // Cancel the first subscription, which will close the WebSocket connection.
+          await firstStream.listen(null).cancel();
```
dnys1, line 417: I don't get why this is needed

dnys1 (approved):

packages/api/amplify_api/example/integration_test/graphql/iam_test.dart
ragingsquirrel3: The error happens when you cancel a failed subscription, then appsync sends an error message in response to the cancelation message (sends a stream event on a closed stream). Now w reconnect in place, if you only have 1 subscription on the connection, then canceling that subscription closes the connection before that error message causes a client error. In order to keep the connection open to get the error message about canceling the failed subscription, this keeps another subscription open so that we don't close the connection.

PR #2182, opened 2022-09-27
chore(datastore): run integ tests in CI

ragingsquirrel3:
This PR enables subset of datastore integ tests in CI. In summary:
* changes to GH action config to run datastore tests in CI
* adds "small" parameter for integ tests although datastore is far only test to differ logic here. For datastore, it will only run the main_test.dart (lots of offline tests) and one basic test for cloud sync. This is necessary because running all the datastore tests same way as other tests will cause increase in flakiness/resource consumption that we don't want for current test execution environment. Eventually, might be good to have a "large" suite that runs tests like this once a week or something like that.
* tweak a sorting test that was failing in CI as written because 2 `AWSDate`s are the same and get into different orders.

### DIFF
--- a/.github/workflows/amplify_integration_tests.yaml
+++ b/.github/workflows/amplify_integration_tests.yaml
@@ -20,7 +20,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        scope: ["amplify_api_example", "amplify_storage_s3_example", "amplify_auth_cognito_example"]
+        # TEMP
+        scope: ["amplify_datastore_example"]
+        # scope: ["amplify_api_example", "amplify_storage_s3_example", "amplify_auth_cognito_example"]
     steps:
       - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
         with:
@@ -52,11 +54,11 @@ jobs:
 
       - name: Run Android integration tests
         uses: reactivecircus/android-emulator-runner@e790971012b979513b4e2fe70d4079bc0ca8a1ae # 2.24.0
-        timeout-minutes: 25
+        timeout-minutes: 45
         with:
           # API levels 30+ too slow https://github.com/ReactiveCircus/android-emulator-runner/issues/222
           api-level: 29
-          script: melos exec -c 1 --scope ${{ matrix.scope }} -- "deviceId=emulator-5554 retries=1 \$MELOS_ROOT_PATH/build-support/integ_test_android.sh"
+          script: melos exec -c 1 --scope ${{ matrix.scope }} -- "deviceId=emulator-5554 retries=1 small=true \$MELOS_ROOT_PATH/build-support/integ_test_android.sh"
 
   ios:
     runs-on: macos-latest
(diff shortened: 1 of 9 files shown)


.github/workflows/amplify_integration_tests.yaml
```diff
@@ -52,11 +56,11 @@ jobs:
 
       - name: Run Android integration tests
         uses: reactivecircus/android-emulator-runner@e790971012b979513b4e2fe70d4079bc0ca8a1ae # 2.24.0
-        timeout-minutes: 25
+        timeout-minutes: 45
         with:
           # API levels 30+ too slow https://github.com/ReactiveCircus/android-emulator-runner/issues/222
           api-level: 29
-          script: melos exec -c 1 --scope ${{ matrix.scope }} -- "deviceId=emulator-5554 retries=1 \$MELOS_ROOT_PATH/build-support/integ_test_android.sh"
+          script: melos exec -c 1 --scope ${{ matrix.scope }} -- "deviceId=emulator-5554 retries=1 small=true \$MELOS_ROOT_PATH/build-support/integ_test_android.sh"
```
Jordan-Nelson, line 63: Could we make use of [tags](https://pub.dev/packages/test#tagging-tests) instead of the small/large flag?

packages/amplify_datastore/example/integration_test/query_test/sort_order_test.dart
```diff
@@ -104,7 +104,7 @@ void main() {
         DateTime(0000, 01, 01, 00, 00, 00),
         DateTime(1970, 01, 01, 00, 00, 00),
         DateTime(2020, 01, 01, 00, 00, 00),
-        DateTime(2020, 01, 01, 23, 59, 59),
+        DateTime(2020, 01, 02, 23, 59, 59),
```
Jordan-Nelson, line 107: Was the test failing without this change? Might be a bug?

ragingsquirrel3: Yes was failing, although not bug IMO. Discussed offline w @HuiSF but was failing consistently in CI on android/ios but not locally. These 2 values get converted to the same date and their order would get switched. Not sure why their order gets switched as I understand return in save in order otherwise? though not clear if that's a requirement. However, this test should just test the date order not the order when date equals.

If expected behavior is indeed that should be in some order even when the same date, there should be a separate test for that IMO, and it's a bug.

.github/workflows/amplify_integration_tests.yaml
ragingsquirrel3: Looked into this and had some issues, will ping you offline.

packages/amplify_datastore/example/integration_test/query_test/sort_order_test.dart
Jordan-Nelson: I see. Thanks for the explanation. I thought this test was for `AWSDateTime`. I see it is for `AWSDate`. It is interesting that it was only failing in CI, but I agree there is no expectation for order in this case.

Jordan-Nelson (approved):

HuiSF (approved):

PR #2184, opened 2022-09-27
fix(auth, secure_storage): update generated java pigeon code

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Update generated java code to be in the main package to allow for accessing package-private members in tests.


.gitattributes
```diff
@@ -14,15 +14,16 @@
 tool/ci.sh linguist-generated=true
 
 ## Generated by Pigeon
-**/pigeons/*.java linguist-generated=true
+*Pigeon.java linguist-generated=true
+*Messages.java linguist-generated=true
```
dnys1, line 18: I think this needs to be:

```suggestion
**/*Pigeon.java linguist-generated=true
**/*Messages.java linguist-generated=true
```

since `*` matches everything except a slash, this would only catch top-level files.

dnys1 (approved):

HuiSF (approved):

PR #2197, opened 2022-09-30
feat(analytics): Dart implementation

fjnoyp:
### Analytics Dart First Plugin 

All features of Analytics have been tested on iOS, Android, Web, and MacOS on the original Analytics Flutter Example app. 

Further testing is needed as well as initial tests for Linux and Windows

The following remaining features will be merged as separate PRs to prevent slowing review of main PR and to reduce complexity of PR review:

- Add Dart example app? 
- Unit Tests 
- How to support injection of Drift web sql dependency into web/index.html 
- Migration support for older versions of Analytics (non dart first) 
- On user sign in - auto pass sign in credentials to Pinpoint Endpoint 
- Expose pinpoint endpoint fields in a PinpointAnalyticsUserProfile subtype 

**PR Review Tips:**

Most files are autogenerated and can be disregarded. 

The main files to review are in `packages/analytics_amplify_analytics_pinpoint` and `packages/amplify_analytics_pinpoint_dart`.  

The files in `packages/amplify_analytics_pinpoint_dart/lib/src/sdk` can be ignored as they are the auto generated AWS Dart SDK analytics pinpoint files. 


.gitattributes
```diff
@@ -1,10 +1,10 @@
 ## Platform files generated by Flutter during `flutter create`
-**/example/android/ linguist-generated=true
-**/example/ios/ linguist-generated=true
-**/example/linux/ linguist-generated=true
-**/example/macos/ linguist-generated=true
-**/example/web/ linguist-generated=true
-**/example/windows/ linguist-generated=true
+**/example/android/** linguist-generated=true
```
Jordan-Nelson, line 2: I am guessing you pulled this commit in from next to get the unit tests passing. Can you also pull it into `feat/next/analytics`? I am not sure if there are other commits in this PR that are the same. You might want to rebase this branch and `feat/next/analytics`.

packages/analytics/amplify_analytics_pinpoint/example/web/index.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta charset="UTF-8">
+  <title>amplify_analytics_pinpoint_example_app</title>
+</head>
+
+<body>
+<script defer src="sql-wasm.js"></script>
```
Jordan-Nelson, line 10: *nit:* I might just add a comment to point to the drift docs in case someone wasn't sure why this was added.

packages/analytics/amplify_analytics_pinpoint/lib/src/analytics_plugin_impl.dart
```diff
@@ -0,0 +1,45 @@
...
+
+import 'package:amplify_analytics_pinpoint/src/flutter_path_provider/flutter_path_provider.dart';
+import 'package:amplify_analytics_pinpoint_dart/amplify_analytics_pinpoint_dart.dart';
+import 'package:amplify_secure_storage/amplify_secure_storage.dart';
+
+import 'device_context_info_provider/flutter_device_context_info_provider.dart';
+import 'flutter_app_lifecycle_provider.dart';
+
+/// {@template amplify_analytics_pinpoint.amplify_analytics_pinpoint}
+/// The Amplify Analytics Pinpoint implementation of the Amplify Analytics category.
```
Jordan-Nelson, line 23:
```suggestion
/// The AWS Pinpoint implementation of the Amplify Analytics category.
```

packages/analytics/amplify_analytics_pinpoint/lib/src/analytics_plugin_impl.dart
```diff
@@ -0,0 +1,45 @@
...
+/// {@endtemplate}
+class AmplifyAnalyticsPinpoint extends AmplifyAnalyticsPinpointDart {
+  /// {@macro amplify_analytics_pinpoint.amplify_analytics_pinpoint}
+  ///
+  /// The parameters provide Flutter related functionality to the underlying
+  /// Dart only implementation.
+  /// [keyValueStorage] is a [AmplifySecureStorage] instance used for storing global properties and unique id
+  /// [pathProvider] is used on non web platforms to provide a proper storage location for cached analytics events
+  /// [appLifecycleProvider] provides app lifecycle events for auto session tracking
+  /// [deviceContextInfoProvider] provides device information sent to Pinpoint Endpoints
```
Jordan-Nelson, line 33: I don't think these comments should be part of the public doc comments since these are not exposed on `AmplifyAnalyticsPinpoint`.  You coudl move them to `AmplifyAnalyticsPinpointDart`.

```suggestion
  /// {@macro amplify_analytics_pinpoint.amplify_analytics_pinpoint} 
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -0,0 +1,159 @@
...
+import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/key_value_store.dart';
+import 'package:amplify_analytics_pinpoint_dart/src/sdk/pinpoint.dart';
+import 'package:amplify_core/amplify_core.dart';
+import 'package:amplify_secure_storage_dart/amplify_secure_storage_dart.dart';
+import 'package:aws_signature_v4/aws_signature_v4.dart';
+
+/// Public facing Plugin
+/// Validate and parse inputs
+/// Receive and provide external Flutter Provider implementations
+/// Delegate work to PinpointClient
```
Jordan-Nelson, line 28: `AmplifyAnalyticsPinpoint` is the public facing package. This is also technically public, but AmplifyAnalyticsPinpoint is what will be used by customers.

```suggestion
/// The Dart implementation of AmplifyAnalyticsPinpoint.
```

packages/analytics/amplify_analytics_pinpoint/lib/src/device_context_info_provider/app_info_provider.dart
```diff
@@ -0,0 +1,44 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'package:package_info_plus/package_info_plus.dart';
+
+/// Data representation of AppInfo required for Pinpoint EndpointDemographic:
+/// https://docs.aws.amazon.com/pinpoint/latest/apireference/apps-application-id-endpoints.html
+class AppInfo {
+  /// The app name. `CFBundleDisplayName` on iOS, `application/label` on Android.
```
Jordan-Nelson, line 20: https://dart.dev/guides/language/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph

```suggestion
  /// The app name. 
  ///
  /// `CFBundleDisplayName` on iOS, `application/label` on Android.
```

Jordan-Nelson: If this will be null on all other platforms, you could add that to the comment.

packages/analytics/amplify_analytics_pinpoint/lib/src/device_context_info_provider/app_info_provider.dart
```diff
@@ -0,0 +1,44 @@
...
+  String? appName;
+
+  /// The package name. `bundleIdentifier` on iOS, `getPackageName` on Android.
```
Jordan-Nelson, line 23:
```suggestion
  /// The package name.
  /// 
  /// `bundleIdentifier` on iOS, `getPackageName` on Android.
```

packages/analytics/amplify_analytics_pinpoint/lib/src/device_context_info_provider/app_info_provider.dart
```diff
@@ -0,0 +1,44 @@
...
+  String? packageName;
+
+  /// The package version. `CFBundleShortVersionString` on iOS, `versionName` on Android.
```
Jordan-Nelson, line 26:
```suggestion
  /// The package version.
  /// 
  /// `CFBundleShortVersionString` on iOS, `versionName` on Android.
```

packages/analytics/amplify_analytics_pinpoint/lib/src/device_context_info_provider/app_info_provider.dart
```diff
@@ -0,0 +1,44 @@
...
+  String? version;
+
+  /// Constructs AppInfo data object with provided information
+  AppInfo(this.appName, this.packageName, this.version);
+}
+
+/// Provides AppInfo from Flutter -> Dart
```
Jordan-Nelson, line 33:
```suggestion
/// Provides AppInfo from Flutter -> Dart.
```

packages/analytics/amplify_analytics_pinpoint/lib/src/device_context_info_provider/device_info_provider/device_info_provider_html.dart
```diff
@@ -0,0 +1,41 @@
...
+    try {
+      final webInfo = await deviceInfo.webBrowserInfo;
+      return DeviceInfo(
+        make: webInfo.vendor, // vendor of the browser
+        model: webInfo.browserName.toString(),
+        modelVersion: webInfo.appVersion, // version of browser
+        platform: DevicePlatform.web,
+        platformVersion: webInfo.platform, // version of browser?
+      );
+    } on PlatformException {
```
Jordan-Nelson, line 35: Are there known scenarios in which we would expect this to fail? Do we want the the fallback below to be the fallback in the case of any error?

packages/analytics/amplify_analytics_pinpoint/lib/src/device_context_info_provider/flutter_device_context_info_provider.dart
```diff
@@ -0,0 +1,51 @@
...
+
+import 'package:amplify_analytics_pinpoint/src/device_context_info_provider/device_info_provider/device_info_provider.dart';
+
+import 'package:amplify_analytics_pinpoint_dart/amplify_analytics_pinpoint_dart.dart';
+import 'package:flutter/material.dart';
+
+/// Provide information required for Pinpoint EndpointDemographic object
+/// https://docs.aws.amazon.com/pinpoint/latest/apireference/apps-application-id-endpoints.html
+class FlutterDeviceContextInfoProvider implements DeviceContextInfoProvider {
+  /// No work is done when instantiating.
```
Jordan-Nelson, line 25: Can probably just omit this

packages/analytics/amplify_analytics_pinpoint/lib/src/device_context_info_provider/flutter_device_context_info_provider.dart
```diff
@@ -0,0 +1,51 @@
...
+  const FlutterDeviceContextInfoProvider();
+
+  @override
+  Future<DeviceContextInfo> getDeviceInfoDetails() async {
+    var appInfo = await AppInfoProvider.getAppInfo();
```
Jordan-Nelson, line 30:
```suggestion
    final appInfo = await AppInfoProvider.getAppInfo();
```

packages/analytics/amplify_analytics_pinpoint/lib/src/device_context_info_provider/flutter_device_context_info_provider.dart
```diff
@@ -0,0 +1,51 @@
...
+    var deviceInfo = await DeviceInfoProvider().getDeviceInfo();
```
Jordan-Nelson, line 31:
```suggestion
    final deviceInfo = await DeviceInfoProvider().getDeviceInfo();
```

packages/analytics/amplify_analytics_pinpoint/lib/src/device_context_info_provider/flutter_device_context_info_provider.dart
```diff
@@ -0,0 +1,51 @@
...
+    return DeviceContextInfo(
+      countryCode: WidgetsBinding.instance.window.locale.countryCode,
+      locale: WidgetsBinding.instance.window.locale.toString(),
+      // https://github.com/dart-lang/sdk/issues/21758
+      // Flutter timezone.name provides time dependent names ie. PDT/PST
+      // which are not accepted by Pinpoint
+      // Other packages require location objects (tz database) or don't work on all platforms
+      // NOTE: all times sent to Pinpoint must be in UTC ... (DateTime.now().toUtc()...)
```
Jordan-Nelson, line 39: I'm a bit confused by the comment. The issue appears to be a feature request to get the current timezone but the comment notes all times must be UTC. Is there an action that we need to take? Is there something the customer needs to be aware of?

packages/analytics/amplify_analytics_pinpoint/lib/src/flutter_app_lifecycle_provider.dart
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  @override
+  void stopObserving() {
+    _isObserving = false;
+    WidgetsBinding.instance.removeObserver(this);
+  }
+
+  // Current implementation based on Android
+  // TODO - consider supporting session pause like iOS
```
Jordan-Nelson, line 48:
```suggestion
  // TODO(fjnoyp): consider supporting session pause like iOS
```

Not sure if we strictly follow this, but definitely a good pattern: https://dart-lang.github.io/linter/lints/flutter_style_todos.html

packages/analytics/amplify_analytics_pinpoint/lib/src/flutter_path_provider/flutter_path_provider_html.dart
```diff
@@ -0,0 +1,24 @@
...
+// limitations under the License.
+
+import 'package:amplify_analytics_pinpoint_dart/amplify_analytics_pinpoint_dart.dart';
+
+/// Storage location is not needed on Web
+/// As we used IndexedDB
+class FlutterPathProvider extends CachedEventsPathProvider {
+  @override
+  Future<String> getApplicationSupportPath() async {
+    throw UnsupportedError('Do not call FlutterPathProvider on web');
```
Jordan-Nelson, line 22: Do you need to split this out if it will just throw unimplemented? I imagine there are separate code paths from web and io where this is used.

If you are doing it just to maintain a consistent pattern, I think that is fine. But I am not sure you really need to.

packages/analytics/amplify_analytics_pinpoint/lib/src/flutter_path_provider/flutter_path_provider.dart
```diff
@@ -0,0 +1,19 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// Multi platform class that provides storage location path
+/// Requires Flutter specific dependencies
+export 'flutter_path_provider_none.dart'
+    if (dart.library.html) 'flutter_path_provider_html.dart'
+    if (dart.library.io) 'flutter_path_provider_io.dart';
```
Jordan-Nelson, line 19:
```suggestion
export 'flutter_path_provider_none.dart'
    if (dart.library.io) 'flutter_path_provider_io.dart';
```

You could return the stub on web given that it throws unimplemented anyway.

Jordan-Nelson (commented):
I still need to go through the code in the _dart package, which is the bulk of the PR, but here some some comments from a first pass

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/key_value_store.dart
```diff
@@ -0,0 +1,75 @@
...
+    } else {
+      storage = storage;
+    }
+
+    // Retrieve Unique ID
+    var storedUniqueId = await storage.read(key: uniqueIdKey);
+    if (storedUniqueId == null) {
+      storedUniqueId = const Uuid().v1();
+      await storage.write(key: uniqueIdKey, value: storedUniqueId);
+    }
```
Jordan-Nelson, line 57: No need to re-use `storedUniqueId`. Make it final and create a new final var if it doesn't exist.

```suggestion
    final storedUniqueId = await storage.read(key: uniqueIdKey);
    if (storedUniqueId == null) {
      final uniqueId = const Uuid().v1();
      await storage.write(key: uniqueIdKey, value: uniqueId);
    }
```

Jordan-Nelson (commented):
I still have a bit left to review in the dart package, but wanted to leave comments for what I have reviewed so far.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/drift/connection/connection_html.dart
```diff
@@ -0,0 +1,21 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'package:amplify_analytics_pinpoint_dart/src/impl/flutter_provider_interfaces/path_provider.dart';
+import 'package:drift/backends.dart';
+import 'package:drift/web.dart';
+
+QueryExecutor connect(CachedEventsPathProvider? pathProvider) {
+  return WebDatabase('db');
```
HuiSF, line 20: Should the value of the `name` parameter be more elaborated other than `'db'`? Maybe `amplify_analytics_cached_events`?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/events_storage_adapter.dart
```diff
@@ -0,0 +1,104 @@
...
+import 'package:amplify_analytics_pinpoint_dart/src/sdk/pinpoint.dart';
+import 'package:amplify_analytics_pinpoint_dart/src/sdk/src/pinpoint/common/serializers.dart';
+import 'package:amplify_core/amplify_core.dart';
+import 'package:built_value/serializer.dart';
+
+/// Interface with underlying device storage using the [Drift] package
+/// Present interface for saving/retrieving PinpointEvents
+class EventStorageAdapter {
+  factory EventStorageAdapter(CachedEventsPathProvider? pathProvider) {
+    final db = DriftDatabaseJsonStrings(pathProvider);
```
HuiSF, line 29: If I read the code correctly, the db writes happens only within `saveEvent` call, should open the connection within the function call, and also close when finish saving, instead of keep the connection open all the time?

packages/analytics/amplify_analytics_pinpoint/example/lib/main.dart
```diff
@@ -68,59 +69,62 @@ class _MyAppState extends State<MyApp> {
...
+    final event = AnalyticsEvent(_uniqueId);
 
     event.properties.addBoolProperty('boolKey', true);
-    event.properties.addDoubleProperty('doubleKey', 10.0);
+    event.properties.addDoubleProperty('doubleKey', 10);
     event.properties.addIntProperty('intKey', 10);
     event.properties.addStringProperty('stringKey', 'stringValue');
 
-    Amplify.Analytics.recordEvent(event: event);
+    await Amplify.Analytics.recordEvent(event: event);
```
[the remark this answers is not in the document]
fjnoyp: No public API signature changed.  This was added now to ensure that Future propogates up

packages/analytics/amplify_analytics_pinpoint/example/lib/main.dart
```diff
@@ -68,59 +69,62 @@ class _MyAppState extends State<MyApp> {
   }
 
   void _recordEvent() async {
-    AnalyticsEvent event = AnalyticsEvent(_uniqueId);
+    final event = AnalyticsEvent(_uniqueId);
 
     event.properties.addBoolProperty('boolKey', true);
-    event.properties.addDoubleProperty('doubleKey', 10.0);
+    event.properties.addDoubleProperty('doubleKey', 10);
```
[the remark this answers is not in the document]
fjnoyp: Linter is okay here actually.

packages/analytics/amplify_analytics_pinpoint/lib/src/device_context_info_provider/device_info_provider/device_info_provider_html.dart
fjnoyp: Good point.  I haven't tested on different web browsers so I added that in case a strange or outdated version of web doesn't work.  In that case I would want to behavior to continue running instead of breaking the code execution.

packages/analytics/amplify_analytics_pinpoint/lib/src/device_context_info_provider/flutter_device_context_info_provider.dart
fjnoyp: Hey I think we discussed this more in our meeting last week. 

I think we agreed to just not send timezone because of issues with the tz database.

packages/analytics/amplify_analytics_pinpoint/lib/src/flutter_path_provider/flutter_path_provider.dart
fjnoyp: I throw different exceptions depending on the platform.  I think the difference can be helpful as the context in which the exception could be thrown is different.  In one case the user is trying to use a platform that just doesn't work, in the other case an implementation is wrong and the method is being called on web when it shouldn't be. 

throw UnsupportedError('Do not call FlutterPathProvider on web');

vs

throw UnsupportedError(
      'No suitable implementation was found on this platform.',
    );

packages/analytics/amplify_analytics_pinpoint_dart/README.md
```diff
@@ -0,0 +1 @@
+# Amplify Analytics Pinpoint for Dart
```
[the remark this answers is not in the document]
fjnoyp: Nice thank you!

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_global_fields_manager.dart
```diff
@@ -0,0 +1,154 @@
...
+  // External dependencies
+  final KeyValueStore _keyValueStore;
+
+  static Future<EndpointGlobalFieldsManager> getInstance(
+    KeyValueStore sharedPrefs,
+  ) async {
+    /// Retrieve stored GlobalAttributes
+    Map<String, dynamic> decodedGlobalAttributesJson = jsonDecode(
+      await sharedPrefs.getJson(KeyValueStore.endpointGlobalAttrsKey) ?? '{}',
+    );
```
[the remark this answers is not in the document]
fjnoyp: yeah weird it had compiled before but then stopped.  that is all resolved now

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_global_fields_manager.dart
```diff
@@ -0,0 +1,154 @@
...
+class EndpointGlobalFieldsManager {
+  EndpointGlobalFieldsManager._getInstance(
+    this._keyValueStore,
+    this._globalAttributes,
+    this._globalMetrics,
+  );
+  final int _maxKeyLength = 50;
+  final int _maxAttributeValues = 50;
+  final int _maxAttributes = 20;
+  final int _maxAttributeValueLength = 100;
```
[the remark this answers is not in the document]
fjnoyp: Changed to static const.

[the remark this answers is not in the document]
fjnoyp: added documentation

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -0,0 +1,159 @@
...
+  }) async {
+    _checkConfigured();
+    await _analyticsClient?.identifyUser(userId, userProfile);
+  }
+
+  void _checkConfigured() {
+    if (_analyticsClient == null) {
+      throw const AnalyticsException(
+        'Analytics not configured',
+        recoverySuggestion: 'Please call Amplify.configure(amplifyconfig)',
```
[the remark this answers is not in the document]
fjnoyp: Good point updated code

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -0,0 +1,159 @@
...
+      _pathProvider,
+      _appLifecycleProvider,
+      deviceContextInfo,
+    );
+  }
+
+  @override
+  Future<void> enable() async {
+    _checkConfigured();
+    _analyticsClient?.enable();
```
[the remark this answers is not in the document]
fjnoyp: thanks good call implemented

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/analytics_client.dart
```diff
@@ -0,0 +1,143 @@
...
+import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_creator/event_creator.dart';
+import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/key_value_store.dart';
+import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/session_manager.dart';
+import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/stoppable_timer.dart';
+import 'package:amplify_analytics_pinpoint_dart/src/sdk/pinpoint.dart';
+import 'package:amplify_core/amplify_core.dart';
+
+/// Top level client for executing all behaviors of the Analytics Plugin
+/// by managing behaviors and inter dependencies of sub clients and helper classes
+class AnalyticsClient {
```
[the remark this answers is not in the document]
fjnoyp: Interesting point, let's keep this open.  

I prefer keeping them separate as the focus on the analytics_plugin_impl is to: 

/// - Validates and parses inputs
/// - Receives and provides external Flutter Provider implementations
/// - Delegates work to the [PinpointClient]

While the analytics_client does:

/// Top level client for executing all behaviors of the Analytics Plugin.
///
/// Manages behaviors and inter dependencies of sub clients and helper classes.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/analytics_client.dart
```diff
@@ -0,0 +1,143 @@
...
+      endpointClient,
+    );
+  }
+
+  Future<void> flushEvents() async {
+    await _eventClient.flushEvents();
+  }
+
+  /// Create a [Event] object from [AnalyticsEvent] received from the public API
+  /// Received events are NEVER sent immediately but cached to be sent in a batch
```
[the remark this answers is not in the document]
fjnoyp: Thanks let's also keep this open for now in case other people encounter this confusion. 

Updated comments to :

  /// Create a Pinpoint [Event] object from [AnalyticsEvent] received from the public API.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -0,0 +1,159 @@
...
+      keyValueStore,
+      pinpointClient,
+      _pathProvider,
+      _appLifecycleProvider,
+      deviceContextInfo,
+    );
+  }
+
+  @override
+  Future<void> enable() async {
```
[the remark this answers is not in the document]
fjnoyp: This is consistent with the public API from before. 

I think it's best to keep it in case it does change a future is actually needed.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/analytics_client.dart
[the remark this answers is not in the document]
fjnoyp: Yeah agreed, it's surprising how we don't have comments on that.

HuiSF:
[packages/analytics/amplify_analytics_pinpoint_dart/lib.zip](https://github.com/aws-amplify/amplify-flutter/pull/2197/files#diff-a1575ea09db3820736f571c8241e7e4c780799fda11fecf6d5955a2862bc1960) should not be committed 😅

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_client.dart
```diff
@@ -0,0 +1,181 @@
...
+    if (copyFromProfile.email != null) {
+      addAttribute('email', [copyFromProfile.email!]);
+    }
+    if (copyFromProfile.plan != null) {
+      addAttribute('plan', [copyFromProfile.plan!]);
+    }
+
+    if (copyFromProfile.location != null) {
+      final newLoc = copyFromProfile.location!;
+      final lb = _endpointBuilder.location;
```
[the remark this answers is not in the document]
fjnoyp: I agree with you, typically I avoid abbreviations like this.

They were abbreviated to make the following code section more concise.  I still prefer the current approach instead of having the code become a lot more wordy. 

Maybe we can just add comments on what they are 

ex: 

// Location builder 
final newLoc = .....

CURRENT: 

```
      if (newLoc.latitude != null) lb.latitude = newLoc.latitude;
      if (newLoc.longitude != null) lb.longitude = newLoc.longitude;
      if (newLoc.postalCode != null) lb.postalCode = newLoc.postalCode;
      if (newLoc.city != null) lb.city = newLoc.city;
      if (newLoc.region != null) lb.region = newLoc.region;
      if (newLoc.country != null) lb.country = newLoc.country;
```
      
YOUR SUGGESTION: 

1) It must become 2 lines
2) Because it is 2 lines it must have curly braces

```
      if (newLocation.latitude != null) {
        locationBuilder.latitude = newLocation.latitude;
      }
      if (newLocation.longitude != null) {
        locationBuilder.longitude = newLocation.longitude;
      }
      if (newLocation.postalCode != null) {
        locationBuilder.postalCode = newLocation.postalCode;
      }
      if (newLocation.city != null) {
        locationBuilder.city = newLocation.city;
      }
      if (newLocation.region != null) {
        locationBuilder.region = newLocation.region;
      }
      if (newLocation.country != null) {
        locationBuilder.country = newLocation.country;
      }
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_client.dart
```diff
@@ -0,0 +1,181 @@
...
+  void removeAttribute(String name) =>
+      _globalFieldsManager.removeAttribute(name);
+  void addMetric(String name, double value) =>
+      _globalFieldsManager.addMetric(name, value);
+  void removeMetric(String name) => _globalFieldsManager.removeMetric(name);
+
+  /// Update the UserProfile object on the EndpointProfile
+  Future<void> setUser(
+    String userId,
+    AnalyticsUserProfile copyFromProfile,
```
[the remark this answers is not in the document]
fjnoyp: We are copying from this profile. 

I can add more comments to make that more clear.  

Similar to event, the AnalyticsUserProfile is the public facing object but must be converted into a EndpointUserBuilder and PublicEndpointBuilder

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/key_value_store.dart
```diff
@@ -0,0 +1,75 @@
...
+
+    return KeyValueStore._(storedUniqueId, storage);
+  }
+
+  Future<void> saveJson(String key, String json) async {
```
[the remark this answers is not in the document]
fjnoyp: Yeah good call, updated to be saveString instead of saveJson

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/drift/connection/connection_html.dart
fjnoyp: Yeah good call, updated to the suggested name.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/events_storage_adapter.dart
fjnoyp: I think db writes will also be done on deleteEvents

Yeah perhaps I am a bit unfamiliar with runtime performance here.  Is it expensive to keep it open vs open/close it on use?

Typically we will be saving events every 10 seconds (auto submission of events).

HuiSF: I was think that each time when we need to write the db, to invoke `DriftDatabaseJsonStrings()` which open the connection, and after write, invoke `db.close()`.

But Drift actually doesn't recommend creating multiple instances of DB which duplicates the `QueryExecutor` which may lead to race condition. So single instance (connection) through out an App session seems reasonable. 

Probably need to test in mobile, when minimize App and reopen, will the connection still work? (same for me to test in Storage use cases.)

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/stoppable_timer.dart
```diff
@@ -0,0 +1,46 @@
...
+  void start() {
+    if (_timer.isActive) return;
+    _timer = Timer.periodic(_duration, (Timer t) => _function());
+  }
+
+  /// Stop the timer
+  void stop() {
+    _timer.cancel();
+  }
+}
```
Jordan-Nelson, line 46: I would avoid naming a variable `function`. `callback` is what the `Timer` class uses.

Worth adding a comment to the constructor to explain what the args are.

```suggestion
class StoppableTimer {
  /// {@macro amplify_analytics_pinpoint_dart.stoppable_timer}
  ///
  /// The [callback] is invoked repeatedly with [duration] intervals until
  /// stopped with the [stop] function.
  StoppableTimer({
    required Duration duration,
    required void Function() callback,
  })  : _duration = duration,
        _callback = callback,
        _timer = Timer.periodic(duration, (Timer t) => callback());

  Timer _timer;
  final Duration _duration;
  final void Function() _callback;

  /// Start the timer
  void start() {
    if (_timer.isActive) return;
    _timer = Timer.periodic(_duration, (Timer t) => _callback());
  }

  /// Stop the timer
  void stop() {
    _timer.cancel();
  }
}
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/stoppable_timer.dart
```diff
@@ -0,0 +1,46 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'dart:async';
+
+/// {@template amplify_analytics_pinpoint_dart.stoppable_timer}
+/// A Timer that can be stopped and started again.
+///
+/// Used for auto flush of events
+/// Timer must be stoppable to support the public API's disable method
```
Jordan-Nelson, line 21: I would focus the comment on what the code does, rather than how it will be used internally.

> The doc comment should focus on what the code _does_.
- https://dart.dev/guides/language/effective-dart/documentation#prefer-starting-function-or-method-comments-with-third-person-verbs

```suggestion
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/drift/drift_tables.dart
```diff
@@ -0,0 +1,84 @@
...
+  static DriftDatabaseJsonStrings getInstance({
+    CachedEventsPathProvider? pathProvider,
+  }) {
+    return _instance ??= DriftDatabaseJsonStrings(
+      pathProvider: pathProvider,
+    );
+  }
+
+  // you should bump this number whenever you change or add a table definition.
+  // Migrations are covered later in the documentation.
```
Jordan-Nelson, line 54: Update or remove

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/drift/drift_tables.dart
```diff
@@ -0,0 +1,84 @@
...
+// limitations under the License.
+
+import 'package:amplify_analytics_pinpoint_dart/src/impl/drift/connection/connection.dart'
+    as impl;
+import 'package:amplify_analytics_pinpoint_dart/src/impl/flutter_provider_interfaces/cached_events_path_provider.dart';
+import 'package:drift/drift.dart';
+
+part 'drift_tables.g.dart';
+
+/// {@template amplify_analytics_pinpoint_dart.drift_json_strings}
```
Jordan-Nelson, line 22: I am glad you are using templates, but you don't need to use a template if the comment it not reused. I don't see this template referenced anywhere. I think there are a handful of other templates in this PR that are also not used anywhere.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/flutter_provider_interfaces/device_context_info_provider.dart
```diff
@@ -0,0 +1,187 @@
...
+
+/// {@template amplify_analytics_pinpoint_dart.device_context_info.sent_to_pinpoint_event}
+/// Sent as a field to [Pinpoint Event](https://docs.aws.amazon.com/pinpoint/latest/apireference/apps-application-id-events.html).
+/// {@endtemplate}
+
+/// {@template amplify_analytics_pinpoint_dart.device_context_info}
+/// Data representation of all device information sent to Pinpoint EndpointDemographic
+///
+/// For more details see [Pinpoint Endpoint](https://docs.aws.amazon.com/pinpoint/latest/apireference/apps-application-id-endpoints.html)online spec.
+/// {@endtemplate}
```
Jordan-Nelson, line 39: I'm not sure what the intention here is, but all of these comments are included in the doc comment for `DeviceContextInfo`. The first three templates all become one paragraph because there is no white space. Is that what you want?
Screen Shot 2022-10-13 at 3 15 24 PM

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/flutter_provider_interfaces/device_context_info_provider.dart
```diff
@@ -0,0 +1,187 @@
...
+      'field truncated to: $newField '
+      'to ensure max length of $_maxFieldLength',
+    );
+    return newField;
+  }
+}
+
+/// Device Platforms
+enum DevicePlatform {
+  ///
```
Jordan-Nelson, line 167: Update or remove the comment.

packages/analytics/amplify_analytics_pinpoint_dart/mono_pkg.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - unit_test:
+      - group:
+          - test: --coverage=coverage
+          - test: -p chrome --coverage=coverage
+          - command: ../../../tool/coverage.sh
+  - unit_test_native:
+      - test:
+        os:
+          - windows
+          - macos
```
Jordan-Nelson, line 20: Is linux intentionally left out?

packages/analytics/amplify_analytics_pinpoint_dart/pubspec.yaml
```diff
@@ -0,0 +1,39 @@
...
+  built_collection: ^5.0.0
+  built_value: ^8.4.0
+  meta: ^1.7.0
+  smithy: ^0.1.0
+  smithy_aws: ^0.1.0
+  uuid: ^3.0.0
+  path: ^1.8.1 # path manipulation
+  drift: ^2.1.0 # local database
+
+  intl: ^0.17.0 # for time format
```
Jordan-Nelson, line 25: Sort these. No need to include the comments.

packages/analytics/amplify_analytics_pinpoint_dart/pubspec.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+dev_dependencies:
+  amplify_lints:
+    path: ../../amplify_lints
+  drift_dev: ^2.1.0
+  build_runner: ^2.2.0
+  build_verify: ^3.0.0
+  build_version: ^2.0.0
+  built_value_generator: 8.4.1
+  test: ^1.16.0
```
Jordan-Nelson, line 35: Sort

packages/analytics/amplify_analytics_pinpoint_dart/pubspec.yaml
Jordan-Nelson: We have a lint for this. If it is not showing in your IDE, see if you can configure it to show.

packages/analytics/amplify_analytics_pinpoint_dart/pubspec.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+  # TODO - use cipher libraries for encrypted cached Analytics Events
+  #sqlcipher_flutter_libs: ^0.5.1 # for encoding sql
+  #sqlite3_flutter_libs: ^0.5.0 # FLUTTER ONLY
```
Jordan-Nelson, line 39: Is this something we are planning to do?

packages/analytics/amplify_analytics_pinpoint_dart/test/analytics_test.dart
```diff
@@ -0,0 +1,19 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'package:test/test.dart';
+
+void main() {
+  test('', () {});
+}
```
Jordan-Nelson, line 19: Are you planning on adding tests in this PR?

packages/analytics/amplify_analytics_pinpoint_dart/tool/patch/user-models.patch
```diff
@@ -0,0 +1,787 @@
+From 173f432f1c7916d2a591394eda35f044a6bcad08 Mon Sep 17 00:00:00 2001
```
Jordan-Nelson, line 1: What is the purpose of this?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/flutter_provider_interfaces/device_context_info_provider.dart
```diff
@@ -0,0 +1,187 @@
...
+    this.appVersion,
+    this.make,
+    this.model,
+    this.modelVersion,
+    this.platform,
+    this.platformVersion,
+  });
+
+  static final AmplifyLogger _logger =
+      AmplifyLogger.category(Category.analytics);
```
dnys1, line 85:
```suggestion
      AmplifyLogger.category(Category.analytics).createChild('DeviceContextInfoProvider');
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/flutter_provider_interfaces/device_context_info_provider.dart
```diff
@@ -0,0 +1,187 @@
...
+  ///
+  /// {@macro amplify_analytics_pinpoint_dart.device_context_info.sent_to_pinpoint_endpoint}
+  final DevicePlatform? platform;
+
+  static String? _sanitize(String? field) {
+    if (field == null || field.length <= _maxFieldLength) return field;
+
+    final newField = field.substring(0, _maxFieldLength);
+    _logger.warn(
+      'Warning - DeviceContextInfoProvider '
```
dnys1, line 157: Can remove this line. Warning is already a given since you're calling `.warn` and I left a comment above to put `DeviceContextInfoProvider` in the logger name

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/events_storage_adapter.dart
fjnoyp: I mis spoke above.  There are 2 cases when a db write will occur: 

1) When user calls `sendEvent` - instead of sending the event we will write to the db. 

2) Every 10 seconds, we auto send events.  This means we will delete events from the db which I assume is a write as well I assume.

fjnoyp: Yeah good point, good to capture that in an integration test. 

In manual tests on iOS when the App is minimized and put to foreground again, I am still able to read from the data

packages/analytics/amplify_analytics_pinpoint/lib/src/analytics_plugin_impl.dart
```diff
@@ -0,0 +1,37 @@
...
+/// {@template amplify_analytics_pinpoint.analytics_plugin_impl}
+/// The AWS Pinpoint implementation of the Amplify Analytics category.
+/// {@endtemplate}
+class AmplifyAnalyticsPinpoint extends AmplifyAnalyticsPinpointDart {
+  /// {@macro amplify_analytics_pinpoint.analytics_plugin_impl}
+  AmplifyAnalyticsPinpoint()
+      : super(
+          keyValueStorage: AmplifySecureStorage(
+            config: AmplifySecureStorageConfig(
+              scope: 'analytics',
```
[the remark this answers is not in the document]
fjnoyp: Good call updated scope string

packages/analytics/amplify_analytics_pinpoint/lib/src/device_context_info_provider/device_info_provider/device_info_provider_io.dart
```diff
@@ -0,0 +1,81 @@
...
+          make: null,
+          model: macInfo.model,
+          modelVersion: null,
+          platform: DevicePlatform.macOS,
+          platformVersion: macInfo.osRelease,
+        );
+      } else if (Platform.isWindows) {
+        return const DeviceInfo(
+          make: null,
+          model: null,
```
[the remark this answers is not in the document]
fjnoyp: Looked into other packages and options here, there don't seem to be many. 

Fortunately device info plus has updated their windows component to include a lot of fields.  However the fields aren't really relevant, so I updated to include platformVersion now but that's it:

```
final windowsInfo = await deviceInfo.windowsInfo;
        return DeviceInfo(
          make: null,
          model: null,
          modelVersion: null,
          platform: DevicePlatform.windows,
          /// For example: `Windows 10 Home Single Language`.
          platformVersion: windowsInfo.productName,
        );
```

packages/analytics/amplify_analytics_pinpoint/lib/src/device_context_info_provider/device_info_provider/device_info_provider_io.dart
```diff
@@ -0,0 +1,81 @@
...
+          modelVersion: null,
+          platform: DevicePlatform.windows,
+          platformVersion: null,
+        );
+      }
+    } on PlatformException {
+      // fall through
```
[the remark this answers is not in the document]
fjnoyp: Yeah good point.  The idea was to prevent an uncaught exception bubbling up and breaking the execution.  So we should at least print a warning and return an empty case here. 

on PlatformException catch (e) {
      _logger.warn('Failed to retrieve DeviceInfo: $e');
      return const DeviceInfo(
        make: null,
        model: null,
        modelVersion: null,
        platform: null,
        platformVersion: 'unknown',
      );
    }

packages/analytics/amplify_analytics_pinpoint/lib/src/device_context_info_provider/device_info_provider/device_info_provider.dart
```diff
@@ -0,0 +1,19 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// Multi platform class that provides DeviceInfo
+/// Requires Flutter specific dependencies
+export 'device_info_provider_none.dart'
```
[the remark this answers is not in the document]
fjnoyp: Sounds good always down for consistency.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/drift/connection/connection_io.dart
```diff
@@ -0,0 +1,90 @@
...
+///
+/// The [NativeDatabase] from drift will synchronously use sqlite3's C APIs.
+/// To move synchronous database work off the main thread, we use a
+/// [DriftIsolate], which can run queries in a background isolate under the
+/// hood.
+/// {@endtemplate}
+QueryExecutor connect(CachedEventsPathProvider? pathProvider) {
+  return DatabaseConnection.delayed(
+    Future.sync(() async {
+      var dir = '';
```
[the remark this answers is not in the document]
fjnoyp: Yeah this is a tricky case..

It's only nullable because web doesn't need this class but it's required otherwise.  

Feasibly if there was a web only implementation, it doesn't make any sense to provide pathProvider that's why it's nullable.

I think the best approach would be to throw an exception.  If someone doesn't pass the pathProvider it'll throw and they'll know to fix it.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -0,0 +1,159 @@
...
+    }
+    return __analyticsClient!;
+  }
+
+  final SecureStorageInterface? _keyValueSecureStorage;
+
+  /// External Flutter Provider implementations
+  final CachedEventsPathProvider? _pathProvider;
+  final AppLifecycleProvider? _appLifecycleProvider;
+  final DeviceContextInfoProvider? _deviceContextInfoProvider;
```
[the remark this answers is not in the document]
fjnoyp: Yeah, null does not interfere with basic analytics behavior. 

Here's why: 

**pathProvider** - not required for web, but otherwise required
**appLifecycleProvider** - if null, sessions are not automatically tracked 
**deviceContextInfoProvider** - if null, no device context info is sent when Pinpoint Endpoint is registered. 
**keyValueStore** - if null, defaults to AmplifySecureStorageWorker

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -0,0 +1,159 @@
...
+        'No AWSIamAmplifyAuthProvider available.  Is Auth category added and configured?',
+      );
+    }
+
+    final credentials = await authProvider.retrieve();
+
+    final pinpointClient = PinpointClient(
+      region: region,
+      credentialsProvider: AWSCredentialsProvider(credentials),
+    );
```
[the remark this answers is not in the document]
fjnoyp: Damn that could have been bad ... Thank you guys for catching this.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -0,0 +1,159 @@
...
+
+    /// Prepare AnalyticsClient
```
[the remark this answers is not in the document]
fjnoyp: When looking through the code, I find it helpful to have those comments to logically group what is a rather big block of initialization code.  Changed to //

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_global_fields_manager.dart
```diff
@@ -0,0 +1,201 @@
...
+
+    _instance = EndpointGlobalFieldsManager(
+      sharedPrefs,
+      globalAttributes,
+      globalMetrics,
+    );
+    return _instance!;
+  }
+
+  // External dependencies
```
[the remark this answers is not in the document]
fjnoyp: Sounds good thanks.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
```diff
@@ -0,0 +1,215 @@
...
+      });
+    } on BadRequestException catch (e) {
+      _handleRecoverableException(e, eventIdsToDelete);
+    } on InternalServerErrorException catch (e) {
+      _handleRecoverableException(e, eventIdsToDelete);
+    } on NotFoundException catch (e) {
+      _handleRecoverableException(e, eventIdsToDelete);
+    } on TooManyRequestsException catch (e) {
+      _handleRecoverableException(e, eventIdsToDelete);
+    } on PayloadTooLargeException catch (e) {
```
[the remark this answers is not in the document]
fjnoyp: I defined recoverable based on what Amplify Android does.  I did not test each case though.  

Please look at the code below though!  PayloadTooLargeException is NOT treated as recoverable: 

```
on PayloadTooLargeException catch (e) {
      _logger
        ..warn('putEvents - exception encountered: ${e.toString}')
        ..warn(
            'Pinpoint event batch limits exceeded: 100 events / 4 mb total size / 1 mb max size per event \n '
            'Reduce your event size or change number of events in a batch');
      eventIdsToDelete.clear();
    } 
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/session_manager.dart
```diff
@@ -0,0 +1,197 @@
...
+
+  /// App is running (foregrounded)
+  active,
+
+  /// App is paused
+  paused,
+}
+
+/// Function that listens to Session lifecycle updates
+typedef SessionFunc = void Function(Session);
```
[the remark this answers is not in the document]
fjnoyp: Thanks for the docs link, as this pattern is used many times throughout the code I think it makes more sense to have a typedef here to have a single source of truth for that function declaration instead of rewriting throughout. 

Updated name to be better: 
OnSessionUpdated

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/session_manager.dart
```diff
@@ -0,0 +1,197 @@
...
+
+/// {@template amplify_analytics_pinpoint_dart.session_manager}
```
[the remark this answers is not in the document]
fjnoyp: Yes thanks meant to use in constructor

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/session_manager.dart
```diff
@@ -0,0 +1,197 @@
...
+/// Manage creation and deletion of current Session.
+///
+/// Updates session based on App foreground/background events
+/// Used for provisioning Events sent to Pinpoint with a Session.
+/// {@endtemplate}
+class SessionManager {
+  /// {@template amplify_analytics_pinpoint_dart.session_manager_constructor}
+  ///
```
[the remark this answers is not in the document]
fjnoyp: Hmm I added that deliberately to make the comments appear on new line

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/drift/drift_tables.dart
fjnoyp: Yeah I made sure to use them.  Did you see that below? 

**/// {@template amplify_analytics_pinpoint_dart.drift_database_json_strings}**
/// Drift class for managing stored [DriftJsonStrings]
/// {@endtemplate}
@DriftDatabase(tables: [DriftJsonStrings])
class DriftDatabaseJsonStrings extends _$DriftDatabaseJsonStrings {
  **/// {@macro amplify_analytics_pinpoint_dart.drift_database_json_strings}**
  DriftDatabaseJsonStrings({CachedEventsPathProvider? pathProvider})
      : super(impl.connect(pathProvider));

  static DriftDatabaseJsonStrings? _instance;

  **/// {@macro amplify_analytics_pinpoint_dart.drift_database_json_strings}**
  static DriftDatabaseJsonStrings getInstance({
    CachedEventsPathProvider? pathProvider,
  }) {
    return _instance ??= DriftDatabaseJsonStrings(
      pathProvider: pathProvider,
    );
  }

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/flutter_provider_interfaces/device_context_info_provider.dart
fjnoyp: No thanks for catching that.  Moved declaration to first time use

packages/analytics/amplify_analytics_pinpoint_dart/mono_pkg.yaml
fjnoyp: Think I need to rerun the command

fjnoyp: No I had forgotten to update this file.

packages/analytics/amplify_analytics_pinpoint_dart/pubspec.yaml
fjnoyp: why would we remove comments?  Perhaps not for path, but the inclusion of intl isn't clear.

packages/analytics/amplify_analytics_pinpoint_dart/pubspec.yaml
fjnoyp: In the future, comment was a bit too long though so I shortened it to just the todo

packages/analytics/amplify_analytics_pinpoint_dart/test/analytics_test.dart
fjnoyp: This PR is a bit too big, I will add in separate followup PR

packages/analytics/amplify_analytics_pinpoint_dart/tool/patch/user-models.patch
fjnoyp: Not sure, looking at the full file it seems like it was related to a codegen update that Dillon had made: 

From: Dillon Nys <nydillon@amazon.com>
Date: Mon, 18 Jul 2022 13:13:49 -0700
Subject: [PATCH] Add user profile models

I think it was accidentally committed so I'm going to delete it.

.circleci/config.yml
```diff
@@ -231,7 +231,6 @@ workflows:
             parameters:
               plugin:
                 [
-                  "amplify_analytics_pinpoint",
```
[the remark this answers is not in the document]
fjnoyp: I can add this back in the future PR to add unit tests.  Right now there are no flutter unit tests though and all original Dart unit tests are invalidated due to new implementation.

packages/analytics/amplify_analytics_pinpoint/lib/src/flutter_app_lifecycle_provider.dart
```diff
@@ -0,0 +1,73 @@
...
+        _onBackground?.call();
+        break;
+      // backgrounded
+      case AppLifecycleState.paused:
+        _onBackground?.call();
+        break;
+      // flutter engine running without view
+      case AppLifecycleState.detached:
+        break;
+      default:
```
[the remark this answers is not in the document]
fjnoyp: Thanks, cases are already exhaustive so removed

packages/analytics/amplify_analytics_pinpoint/lib/src/flutter_path_provider/flutter_path_provider_io.dart
```diff
@@ -0,0 +1,26 @@
...
+import 'package:amplify_analytics_pinpoint_dart/amplify_analytics_pinpoint_dart.dart';
+import 'package:path_provider/path_provider.dart';
+
+/// {@template amplify_analytics_pinpoint.flutter_path_provider_io}
+/// Provides storage location from Flutter -> Dart.
+/// {@endtemplate}
+class FlutterPathProvider extends CachedEventsPathProvider {
+  @override
+  Future<String> getApplicationSupportPath() async {
+    return (await getApplicationSupportDirectory()).path;
```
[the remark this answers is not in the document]
fjnoyp: Yes good call, updated to just send the String? value instead

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/session_manager.dart
```diff
@@ -0,0 +1,197 @@
...
+
+    _sessionBuilder
+      ..stopTimestamp = curTime.toUtc().toIso8601String()
+      ..duration = curTime.millisecondsSinceEpoch - _startTimeMilliseconds;
+  }
+
+  static String _generateSessionId(String fixedEndpointId) {
+    var id = fixedEndpointId;
+    id = id.padLeft(_maxIdLength, '_');
+    id = id.substring(0, _maxIdLength);
```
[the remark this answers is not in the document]
fjnoyp: Thanks code logic wasn't proper, here is the update: 

```
if (id.length > _maxIdLength) {
      id = id.substring(0, _maxIdLength);
    } else {
      id = id.padLeft(_maxIdLength, '_');
    }
```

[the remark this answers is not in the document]
dnys1: This does the same thing really. I’m just wondering when fixedEndpointId would ever be more than maxIdLength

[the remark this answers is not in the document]
fjnoyp: In case we change how the fixedEndpointId is generated and decided to give a longer string, we don't want to send a longer id in that case.

.github/workflows/build.yml
```diff
@@ -1,4 +1,4 @@
-# Created with package:mono_repo v6.3.0
+# Created with package:mono_repo v6.4.1
```
dnys1, line 1: Can you use `6.3.0`? I will make a separate PR for updating mono_repo for all the packages.

`dart pub global activate mono_repo 6.3.0`

packages/analytics/amplify_analytics_pinpoint/lib/src/device_context_info_provider/app_info.dart
```diff
@@ -0,0 +1,38 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// {@template amplify_analytics_pinpoint.app_info}
+/// Data representation of AppInfo required for Pinpoint EndpointDemographic.
+///
+/// For more details see [Pinpoint Endpoint](https://docs.aws.amazon.com/pinpoint/latest/apireference/apps-application-id-endpoints.html)online spec.
```
dnys1, line 18: Please update for all of these:

```suggestion
/// For more details see the Pinpoint [Endpoint](https://docs.aws.amazon.com/pinpoint/latest/apireference/apps-application-id-endpoints.html) online spec.
```

packages/analytics/amplify_analytics_pinpoint_dart/analysis_options.yaml
```diff
@@ -0,0 +1,7 @@
+include: package:amplify_lints/library.yaml
+
+analyzer:
+  exclude:
+    - lib/**/*.g.dart
+    - lib/src/sdk/src/**
+    - lib/src/native_auth_plugin.dart
```
dnys1, line 7: Remove

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
```diff
@@ -0,0 +1,214 @@
...
+    // Unless a retryable exception has been received (see above)
+    finally {
+      await _storageAdapter.deleteEvents(eventIdsToDelete.values);
+    }
+  }
+
+  // If exception is recoverable, do not delete eventIds from local cache
+  void _handleRecoverableException(
+    SmithyHttpException e,
+    HashMap<String, int> eventIdsToDelete,
```
dnys1, line 196:
```suggestion
    Map<String, int> eventIdsToDelete,
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
```diff
@@ -0,0 +1,214 @@
...
+      _logger
+        ..warn('putEvents - exception encountered: ', e)
+        ..warn(
+            'Pinpoint event batch limits exceeded: 100 events / 4 mb total size / 1 mb max size per event \n '
+            'Reduce your event size or change number of events in a batch')
+        ..warn('Unrecoverable issue, deleting cache for local event batch');
+    } on Exception catch (e) {
+      _logger
+        ..warn('putEvents - exception encountered: ', e)
+        ..warn('Unrecoverable issue, deleting cache for local event batch');
```
dnys1, line 184: Why are we deleting events for some exceptions? I feel like this should be the opposite.. we should not be deleting IDs by default and only deleting events for very particular cases.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/session_manager.dart
```diff
@@ -0,0 +1,203 @@
...
+    AppLifecycleProvider? appLifecycleProvider,
+    required OnSessionUpdated onSessionStart,
+    required OnSessionUpdated onSessionEnd,
+  })  : _fixedEndpointId = fixedEndpointId,
+        _appLifecycleProvider = appLifecycleProvider,
+        _onSessionStart = onSessionStart,
+        _onSessionEnd = onSessionEnd {
+    // Start session by _executeStart()
+    // By the time we start observing on lifecycleObserver,
+    // we have missed the first onForeground event
```
dnys1, line 61: What is the implication of this? It's not clear why this is mentioned here.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/session_manager.dart
```diff
@@ -0,0 +1,203 @@
...
+    _sessionBuilder
+      ..stopTimestamp = curTime.toUtc().toIso8601String()
+      ..duration = curTime.millisecondsSinceEpoch - _startTimeMilliseconds;
+  }
+
+  static String _generateSessionId(String fixedEndpointId) {
+    var id = fixedEndpointId;
+
+    if (id.length > _maxIdLength) {
+      id = id.substring(0, _maxIdLength);
```
dnys1, line 190: Should we just be trimming the ID? Why is that appropriate? I feel like this should be an assert statement instead.

```dart
assert(id.length <= _maxIdLength, 'Max ID length exceeded');
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/drift/connection/connection_io.dart
```diff
@@ -0,0 +1,92 @@
...
+      if (driftStoragePath == null) {
+        throw const AnalyticsException(
+          'driftStoragePath is null for a non web platform',
+          recoverySuggestion:
+              'Make sure the pathProvider sent in the constructor of AmplifyAnalyticsPinpointDart '
+              'is returning nonnull and returning a value. ',
+        );
+      } else {
```
dnys1, line 48: else not needed here

dnys1: Which means that the `var dir` is not needed above. Just use `driftStoragePath` directly

tool/LICENSE
```diff
@@ -173,3 +173,30 @@
       defend, and hold each Contributor harmless for any liability
       incurred by, or claims asserted against, such Contributor by reason
       of your accepting any such warranty or additional liability.
+
+   END OF TERMS AND CONDITIONS
```
dnys1, line 177: Why was this added?

dnys1 (commented):
There are some files in `/tool` being committed which should not be:
- tool/addlicense_1.0.0_macOS_x86_64.tar.gz
- tool/checksums.txt

Please get CI checks working as well

packages/analytics/amplify_analytics_pinpoint/lib/src/flutter_path_provider/flutter_path_provider_io.dart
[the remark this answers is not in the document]
Jordan-Nelson: I don't think this value is being cached though. If `getApplicationSupportDirectory()` will be called each time `getApplicationSupportPath()` is called.

You can use an [AsyncMemoizer](https://pub.dev/documentation/async/latest/async/AsyncMemoizer-class.html), which will return the initial result after it is fetched once.

If we don't expect the result of `getApplicationSupportDirectory()` to change, we should cache it.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/session_manager.dart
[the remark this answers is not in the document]
Jordan-Nelson: Oh, I'm not sure if I had missed the comment on line 46 or what. You can ignore this.

Jordan-Nelson (approved):
There are some comments that @dnys1 made that should be addressed, but otherwise this looks great 👍

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
fjnoyp: Some exceptions are not recoverable.  If the exception is recoverable we do not delete those events (by clearing eventIdsToDelete) in _handleRecoversableException. 

The logic is that future requests will continue to fail due to what is contained within the event batch.  For example PayloadTooLargeException or any other Exception type that is not recoverable.

fjnoyp: The following exception types are defined in the Pinpoint SDK: 

Treated as recoverable: 
BadRequestException 
InternalServerErrorException
NotFoundException
TooManyRequestsException

Treated as non recoverable (events are deleted): 
ForbiddenException 
MethodNotAllowedException
PayloadTooLargeException

dnys1: Okay, so my point is we should be catching those errors and only deleting on those exceptions.

Why is `ForbiddenException` unrecoverable? Couldn't it potentially be solved by signing in (e.g. if unauthenticated identities are not allowed)?

I don't see why deleting events would help with a `MethodNotAllowedException`.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/session_manager.dart
fjnoyp: Removed and added comment to the actual method

tool/LICENSE
fjnoyp: I think this was auto generated possibly due to the upgrade of monorepo

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/session_manager.dart
fjnoyp: The id passed is generated via uuid.v1() which is longer than 8.  The prefix isn't as important as the remainder of the session id is a timestamp at millisecond precision.

dnys1 (approved):

PR #2204, opened 2022-10-03
chore(datastore): run integ tests in CI on next

ragingsquirrel3:
Cherry picks 7307233838d9acd323b614242477c74596429e3f off main and adds another commit to skip tests on android for tests related to `AWSTime` which are failing due to https://github.com/aws-amplify/amplify-flutter/issues/2214

### DIFF
--- a/packages/amplify_datastore/example/integration_test/model_type_test.dart
+++ b/packages/amplify_datastore/example/integration_test/model_type_test.dart
@@ -166,6 +166,7 @@ void main() {
       testModelOperations(
         models: models,
         skips: {
+          // Skip bc AWSTime issue on Android https://github.com/aws-amplify/amplify-flutter/issues/2214
           DataStoreOperation.save: Platform.isAndroid,
           DataStoreOperation.query: Platform.isAndroid,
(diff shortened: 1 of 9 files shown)


packages/amplify_datastore/example/integration_test/model_type_test.dart
```diff
@@ -162,7 +163,15 @@ void main() {
           .map((value) =>
               ModelWithAppsyncScalarTypes(awsTimeValue: TemporalTime(value)))
           .toList();
-      testModelOperations(models: models);
+      testModelOperations(
+        models: models,
+        skips: {
```
Jordan-Nelson, line 168: Can you add a TODO with a link the issue that is requiring that these be skipped to each test that is skipped?

Jordan-Nelson (approved):

packages/amplify_datastore/example/integration_test/model_type_test.dart
ragingsquirrel3: just added

HuiSF (approved):

PR #2205, opened 2022-10-04
Chore(API): GraphQL Reconnection Unit Tests

Equartey:
*Description of changes:*
Adds unit test coverage for reconnection workflows: wifi on/off, and failed pings. 

### DIFF
--- a/packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
+++ b/packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
@@ -378,18 +378,18 @@ class WebSocketConnection implements Closeable {
         closeStatus == _defaultCloseStatus ? 'client closed' : 'unknown';
 
     await _channel?.sink.close(closeStatus, reason);
-    await _channel?.sink.done.whenComplete(() {
-      _network?.cancel();
-      _subscription?.cancel();
-      _timeoutTimer?.cancel();
-      _pingTimer?.cancel();
+    await _channel?.sink.done;
 
-      _channel = null;
-      _network = null;
-      _subscription = null;
-      _timeoutTimer = null;
-      _pingTimer = null;
-    });
+    await _network?.cancel();
+    await _subscription?.cancel();
+    _timeoutTimer?.cancel();
+    _pingTimer?.cancel();
+
+    _channel = null;
+    _network = null;
+    _subscription = null;
+    _timeoutTimer = null;
+    _pingTimer = null;
   }
 
   /// Initializes the connection.
@@ -400,7 +400,7 @@ class WebSocketConnection implements Closeable {
   Future<void> init() => _initMemo.runOnce(
         () {
           _hubEventsController.add(SubscriptionHubEvent.connecting());
-          _init();
+          return _init();
         },
       );
 
--- a/packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
+++ b/packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
@@ -159,7 +159,7 @@ class WebSocketConnection implements Closeable {
   Completer<void> _connectionPending = Completer<void>();
   Completer<void> _connectionReady = Completer<void>();
   Completer<void> _connectionStart = Completer<void>();
-  final Completer<void> _reconnectPending = Completer<void>();
+  Completer<void> _reconnectPending = Completer<void>();
 
   /// Fires when connection is waiting for the first `connection_ack` message.
   @visibleForTesting
@@ -214,7 +214,8 @@ class WebSocketConnection implements Closeable {
   StreamSubscription<ConnectivityResult> getStreamNetwork() {
     return (_connectivityOverride ?? Connectivity())
         .onConnectivityChanged
-        .listen((ConnectivityResult connectivityResult) async {
+        .asyncExpand<ConnectivityResult>(
+            (ConnectivityResult connectivityResult) async* {
       switch (connectivityResult) {
         case ConnectivityResult.ethernet:
         case ConnectivityResult.mobile:
@@ -238,7 +239,7 @@ class WebSocketConnection implements Closeable {
         default:
           break;
       }
-    });
+    }).listen(null);
   }
 
   /// Connects WebSocket _channel to _subscription stream but does not send connection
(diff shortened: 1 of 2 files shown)


packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
```diff
@@ -373,23 +385,27 @@ class WebSocketConnection implements Closeable {
...
   }
 
   /// Initializes the connection.
   ///
   /// Connects to WebSocket, sends connection message and resolves future once
   /// connection_ack message received from server. If the connection was previously
   /// established then will return previously completed future.
-  Future<void> init() => _initMemo.runOnce(_init);
+  Future<void> init() => _initMemo.runOnce(
+        (() {
```
dnys1, line 396: Remove extra parens and trailing comma

dnys1 (commented):
Logic is getting pretty complicated here. Can we add some more tests around retry behavior? Thinking one with multiple pings before a success (but still within retry territory), and then one where retry fails and then a new subscription request comes in

dnys1:
Also, what about if the wifi goes in and out repeatedly? Let's get tricky here and try to break this thing!

packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
```diff
@@ -373,23 +385,27 @@ class WebSocketConnection implements Closeable {
...
+          _hubEventsController.add(SubscriptionHubEvent.connecting());
```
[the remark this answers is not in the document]
Equartey: To answer your question, its so the inital connection state fires a hub event and then in the other situations it should fire happen more promptly than having it soley in the `_init` function. For example, when wifi gets turned off, `connecting` is emitted from the network listener. I don't love the scatteredness and hope to clean this up in the pending refactor.

packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
```diff
@@ -357,11 +371,14 @@ class WebSocketConnection implements Closeable {
...
-  void _resetConnectionVariables([int closeStatus = _defaultCloseStatus]) {
+  Future<void> _resetConnectionVariables([
+    int closeStatus = _defaultCloseStatus,
+  ]) async {
     final reason =
         closeStatus == _defaultCloseStatus ? 'client closed' : 'unknown';
 
-    _channel?.sink.done.whenComplete(() {
+    await _channel?.sink.close(closeStatus, reason);
+    await _channel?.sink.done.whenComplete(() {
```
ragingsquirrel3, line 381: question: why await and call whenComplete w callback? I would think you could just `await _channel?.sink.done;` on this line, then call the other stuff on next line with no callback bc you are awaiting done.

packages/api/amplify_api/test/ws/web_socket_connection_test.dart
```diff
@@ -137,42 +149,109 @@ void main() {
...
+      await expectLater(pingFailed.future, completes);
+
+      await Future<void>.delayed(const Duration(seconds: 15));
+
+      pleaseFail = false;
+
+      await expectLater(connection.reconnectPending, completes);
+
+      await assertWebSocketConnected(connection, subscriptionRequest.id);
+    });
```
ragingsquirrel3, line 254: nice job w all these tests!

packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
```diff
@@ -325,15 +333,21 @@ class WebSocketConnection implements Closeable {
 
     try {
       _hasConnectionBroken = false;
-      _hubEventsController.add(SubscriptionHubEvent.connecting());
 
       await retryStrategy.retry(
         // Make a GET request
         () => _getPollRequest().timeout(_defaultRetryTimeout),
       );
+      _resetConnectionInit();
```
ragingsquirrel3, line 341: Why is rest of init/variables needed here? Is this needed for new testing variables or something else? Calling _reconnectPending.complete() makes enough sense.

ragingsquirrel3 (commented):
Tests look great and love the situations tested for and mocking. Overall just want to understand changes to actual implementation and make sure integ tests pass.

packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
Equartey: The intent has been to start every new connection with a clean/blank slate. Nothing gets preserved except a list of existing subscriptions (so new listeners can be setup). Here, it's a little verbose by clearing the `_initMemo`, but the other completers need to be cleared for the new connection.

packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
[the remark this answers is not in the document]
dnys1: Just noticed this doesn't `return _init()` so the returned future will complete early.

packages/api/amplify_api/test/ws/web_socket_connection_test.dart
```diff
@@ -137,42 +149,109 @@ void main() {
...
+      );
+
+      await assertWebSocketConnected(connection, subscriptionRequest.id);
+      fakePlatform.controller.sink.add(ConnectivityResult.wifi);
+
+      pleaseFail = true;
+
+      await expectLater(pingFailed.future, completes);
+
+      await Future<void>.delayed(const Duration(seconds: 15));
```
dnys1, line 247: Not necessary here but might be fun to explore fake_async in the future to avoid literal delays https://pub.dev/packages/fake_async

dnys1 (approved):
Great job!

ragingsquirrel3 (approved):
Awesome job here! I tested this pretty hard locally and connection was pretty resilient and dispatched expected events, tests are great, too. I left a comment about a way to improve reconnect logic in the future but should not be in this PR.

packages/api/amplify_api/lib/src/graphql/ws/web_socket_connection.dart
```diff
@@ -325,15 +333,21 @@ class WebSocketConnection implements Closeable {
 
     try {
       _hasConnectionBroken = false;
-      _hubEventsController.add(SubscriptionHubEvent.connecting());
 
       await retryStrategy.retry(
```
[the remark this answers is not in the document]
Equartey: Great idea. We'll try to include in the pending rewrite :)

PR #2218, opened 2022-10-08
test(auth): Migrate backend to CDK

dnys1:
- Migrates backend to CDK
- Cleans up integration tests
- Adds SMS MFA integration tests


packages/amplify_test/lib/src/integration_test_utils/auth_cognito/integration_test_auth_utils.dart
```diff
@@ -92,121 +87,107 @@ Future<void> adminCreateUser(
...
-      url!,
+      api.endpoint,
       defaultHeaders: {
-        'x-api-key': key!,
+        'x-api-key': api.apiKey!,
       },
     ),
   );
 
+  final createUserParams = <String, dynamic>{
```
haverchuck, line 101: nice 👍

haverchuck (approved):
Approving, assuming it's within the boundary of what we got approved by appsec for ci/cd (@ragingsquirrel3?), or assuming it's for local use only for now.

packages/amplify_test/lib/src/integration_test_utils/auth_cognito/integration_test_auth_utils.dart
```diff
@@ -92,121 +87,107 @@ Future<void> adminCreateUser(
...
+              orElse: () => const AuthUserAttribute(
+                  userAttributeKey: CognitoUserAttributeKey.phoneNumber,
+                  value: '+15555555'))
+          .value,
+      'username': username,
+      'verifyAttributes': verifyAttributes,
+    },
+  };
+
+  _logger.debug('Creating user "$username" with values: $createUserParams');
```
ragingsquirrel3, line 140: request: Could we not log this params object here as it puts passwords in the logs? Or otherwise redact that?

dnys1: Hmm what is the risk of the password leaking?

ragingsquirrel3 (approved):
nice! As mentioned offline I got an exception running this that is likely similar to other PR so approved assuming that same fix can be applied or not needed bc other PR will fix it.

To @haverchuck 's question this is fine to use in CI (though will need other PRs to integrate) bc uses same credential strategy and nothing about that changes.

PR #2243, opened 2022-10-13
feat(http): Downgrade to HTTP/1.1 when possible

dnys1:
Improve usability of the HTTP client by downgrading to HTTP/1.1 when HTTP/2 is unavailable. This allows setting the default to a more reasonable 1/2/3 setting so that things will _generally_ work as expected.


Equartey (approved):
LGTM

ragingsquirrel3 (approved):

PR #2263, opened 2022-10-19
feat(smithy)!: Cancelable operations

dnys1:
Replaces custom `HttpClient` class with `AWSHttpClient` and returns a cancelable `SmithyOperation<Output>` instead of `Future<Output>` for all HTTP operations.


HuiSF (approved):

packages/aws_common/lib/src/http/aws_http_operation.dart
```diff
@@ -43,16 +42,10 @@ class AWSHttpOperation<ResponseType extends AWSBaseHttpResponse>
     return result;
   }
 
-  /// The operation's progress sending the request, in bytes.
-  ///
-  /// If [cancel] is called before the request has been fully sent, the stream
-  /// will be closed and its `onDone` callback will fire.
+  @override
   final Stream<int> requestProgress;
```
fjnoyp, line 46: Does this need to be here still since it's provided in the mixin?  Is there any benefit to overriding it again here,  perhaps just for clarity?

fjnoyp (approved):
Looks good, just 2 minor comments.

packages/auth/amplify_auth_cognito_dart/lib/src/sdk/sdk_bridge.dart
```diff
@@ -173,405 +177,493 @@ class WrappedCognitoIdentityProviderClient
...
+    return SmithyOperation(
+      operation.operation.then(
+        (res) => res,
+        onError: (e, st) {
+          Error.throwWithStackTrace(transformSdkException(e), st);
+        },
+      ),
+      operationName: operation.runtimeTypeName,
+      requestProgress: operation.requestProgress,
+      responseProgress: operation.responseProgress,
```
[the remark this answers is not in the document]
dnys1: That's a good suggestion. I will consider adding it in a separate PR.

packages/aws_common/lib/src/http/aws_http_operation.dart
dnys1: The mixin just adds the abstract getter to the class. It still needs to be implemented. In this case, it's implemented as a field rather than a getter.

PR #2266, opened 2022-10-20
test(storage): Integration test backend

dnys1:
Adds a backend for running integration tests against in the Storage category.


HuiSF (approved):

packages/storage/amplify_storage_s3/example/backend/lib/stack.ts
```diff
@@ -105,6 +105,27 @@ class StorageIntgrationTestEnvironment extends cdk.NestedStack {
...
+          ],
+          allowedHeaders: ["*"],
+          allowedOrigins: ["*"],
+          exposedHeaders: [
+            "x-amz-server-side-encryption",
+            "x-amz-request-id",
+            "x-amz-id-2",
+            "ETag",
+          ],
+          maxAge: 300,
```
HuiSF, line 126: Should this be `MaxAgeSeconds: 3000`?

HuiSF (approved):

packages/storage/amplify_storage_s3/example/backend/lib/stack.ts
dnys1: Yep - good catch. Fixed

packages/storage/amplify_storage_s3/example/backend/bin/backend.ts
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+
+import "source-map-support/register";
+import * as cdk from "aws-cdk-lib";
+import { StorageIntgrationTestStack } from "../lib/stack";
+
+const app = new cdk.App();
+
+new StorageIntgrationTestStack(app, [
```
HuiSF, line 24: nit:
```suggestion
new StorageIntegrationTestStack(app, [
```

packages/storage/amplify_storage_s3/example/backend/lib/stack.ts
```diff
@@ -0,0 +1,309 @@
...
+      "read-auth": new iam.PolicyDocument({
+        statements: [
+          new iam.PolicyStatement({
+            actions: ["s3:ListBucket"],
+            conditions: {
+              StringLike: {
+                "s3:prefix": [
+                  `${prefixes[StorageAccessLevel.private]}`,
+                  `${prefixes[StorageAccessLevel.private]}*`,
+                ],
```
HuiSF, line 256: Oops, wasn't aware until actually tested... looks like we are missing read permission of `protected/` and `protected/*` here.

I noticed that the IAM policies created after deploy the stack are slightly different from the IAM policies created by Amplify CLI. It may be worth it to precisely replicate the Amplify CLI created IAM policy to ensure the tests results are accurate?

dnys1: Updated. I believe it should match now 👍

HuiSF: TY for the quick fix, will continue adding integration tests based on this new change.

HuiSF (approved):
Test runs succeed, I think this is a good shape now! Thanks.

ragingsquirrel3 (approved):
approved assuming we remove the no-prefix stack bc my understanding is we are not testing that and that it causes exceptions in other PR for test execution.

Thanks!

PR #2271, opened 2022-10-20
feat(db_common): Add Flutter library

dnys1:
Adds a Flutter library to replace `sqlite3_flutter_libs` in the short-term until we remove our dependency on iOS DataStore.


.github/workflows/lint.yml
```diff
@@ -133,30 +133,69 @@ jobs:
...
-        working-directory: packages/auth/amplify_auth_cognito
+      - name: "packages/analytics/amplify_analytics_pinpoint/example/ios/.symlinks/plugins/amplify_secure_storage; dart format --output=none --set-exit-if-changed ."
+        if: "always() && steps.packages_analytics_amplify_analytics_pinpoint_example_ios_.symlinks_plugins_amplify_secure_storage_pub_upgrade.conclusion == 'success'"
+        working-directory: packages/analytics/amplify_analytics_pinpoint/example/ios/.symlinks/plugins/amplify_secure_storage
         run: "dart format --output=none --set-exit-if-changed ."
-      - name: "packages/auth/amplify_auth_cognito; flutter analyze --fatal-infos lib"
-        if: "always() && steps.packages_auth_amplify_auth_cognito_pub_upgrade.conclusion == 'success'"
-        working-directory: packages/auth/amplify_auth_cognito
-        run: flutter analyze --fatal-infos lib
+      - name: "packages/analytics/amplify_analytics_pinpoint/example/ios/.symlinks/plugins/amplify_secure_storage; flutter analyze --fatal-infos lib pigeons"
```
fjnoyp, line 145: Seems weird we're adding all these changes for just adding flutter library?  This seems to be unrelated to the PR, perhaps we could add a comment explaining what additional changes were made here as well.

packages/common/amplify_db_common/LICENSE
```diff
@@ -1,175 +1 @@
-
```
fjnoyp, line 1: Shouldn't this LICENSE be the same as well?

packages/common/amplify_db_common/android/.gitignore
```diff
@@ -0,0 +1,9 @@
+*.iml
+.gradle
```
fjnoyp, line 2: Why are we adding android, ios, linux, macos, windows folders outside of the example sub directory?

packages/common/amplify_db_common/example/lib/db.dart
```diff
@@ -0,0 +1,61 @@
+// Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
fjnoyp, line 1: Not a major issue, but is there a way to change your commit so that this is a rename?  These files were created by Jordan but are just showing up as new.

packages/common/amplify_db_common/example/test/widget_test.dart
```diff
@@ -0,0 +1,41 @@
+// Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
```
fjnoyp, line 2: Is this file necessary?

packages/common/amplify_db_common/pubspec.yaml
```diff
@@ -1,29 +1,35 @@
...
+flutter:
+  plugin:
+    platforms:
+      android:
+        package: com.amazonaws.amplify.amplify_db_common
+        pluginClass: AmplifyDbCommonPlugin
+      linux:
+        pluginClass: AmplifyDbCommonPlugin
+      windows:
+        pluginClass: AmplifyDbCommonPlugin
```
fjnoyp, line 35: what about MacOS?

packages/common/amplify_db_common/README.md
```diff
@@ -1,15 +1,15 @@
-## amplify_db_common
+# amplify_db_common
```
fjnoyp, line 2: Can we at least mention this is temporary here to keep track.  In case the team forgets this is a temporary fix.

fjnoyp: This top level Flutter library will replace our import of sqlite3_flutter_libs in the short term until we remove our dependency on iOS DataStore.  When that happens we will switch to using sqlite3_flutter_libs and the amplify_db_common_dart package instead.

packages/common/amplify_db_common/android/.gitignore
fjnoyp: I think they were autogenerated but have no use in our context here right?  If so can we remove it to reduce the # of files being added in the PR.

packages/common/amplify_db_common/ios/amplify_db_common.podspec
```diff
@@ -0,0 +1,20 @@
...
+  s.name             = 'amplify_db_common'
+  s.version          = '0.0.1'
+  s.summary          = 'A new Flutter plugin project.'
+  s.description      = <<-DESC
+A new Flutter plugin project.
+                       DESC
+  s.homepage         = 'http://example.com'
+  s.license          = { :file => '../LICENSE' }
+  s.author           = { 'Your Company' => 'email@example.com' }
+
```
HuiSF, line 15: Should these fields be updated?

packages/common/amplify_db_common/macos/amplify_db_common.podspec
```diff
@@ -0,0 +1,22 @@
...
+Pod::Spec.new do |s|
+  s.name             = 'amplify_db_common'
+  s.version          = '0.0.1'
+  s.summary          = 'A new Flutter plugin project.'
+  s.description      = <<-DESC
+A new Flutter plugin project.
+                       DESC
+  s.homepage         = 'http://example.com'
+  s.license          = { :file => '../LICENSE' }
+  s.author           = { 'Your Company' => 'email@example.com' }
```
HuiSF, line 14: ditto

HuiSF (approved):

packages/common/amplify_db_common/lib/src/connect_io.dart
```diff
@@ -12,55 +12,21 @@
...
-/// See [Using drift in a background isolate](https://drift.simonbinder.eu/docs/advanced-features/isolates/#initialization-on-the-main-thread)
-/// and [the drift native example](https://github.com/simolus3/drift/blob/3253cd7ead51d7b163542929eb0fc8daee573fce/examples/app/lib/database/connection/native.dart)
-/// for more info.
 QueryExecutor connect({
   required String name,
   String? path,
 }) {
-  assert(path != null, 'path cannot be null on vm.');
-
-  return DatabaseConnection.delayed(
```
fjnoyp, line 35: I was able to get drift working on iOS with this same code.  Is it necessary that we have the amplify_db_common and amplify_db_common_dart?  

Couldn't we just have amplify_sqlite3_flutter_libs (which would be sqlite3_flutter_libs without iOS) and then the existing amplify_db_common?

fjnoyp (commented):
Thanks again for tackling this.  Added some comments.

packages/common/amplify_db_common/ios/amplify_db_common.podspec
fjnoyp: I think the git change was weird, the file was actually moved but git registered it as modified.

packages/common/amplify_db_common/pubspec.yaml
dnys1: Not needed for iOS/macOS - we just want the podspec. Probably not the others either, but I'm not sure how to remove them yet.

packages/common/amplify_db_common/README.md
dnys1: I'd rather not publish something saying it's a temporary fix. However, this package will be needed even once we decide to use the flutter_libs package. It incorporates path_provider into the `connect` function.

packages/common/amplify_db_common/lib/src/connect_io.dart
dnys1: This adds path_provider to the connect function. The separation is nice if we ever need it for other things as well. A bit of overhead but I don't think it's too bad.

We don't want to bring in DataStore as a dependency though? We just want to match the SQLite version used which does require a separate package.

packages/common/amplify_db_common/macos/amplify_db_common.podspec
dnys1: Updated

packages/common/amplify_db_common/ios/amplify_db_common.podspec
dnys1: Updated.

packages/common/amplify_db_common/example/test/widget_test.dart
dnys1: Nope. Removed.

.github/workflows/lint.yml
dnys1: Didn't mean to - it's mono_repo being silly

fjnoyp (approved):
Looks good, approved.

PR #2273, opened 2022-10-21
chore: integration test improvements

ragingsquirrel3:
A few fixes for integration tests:
* cherry-picks https://github.com/aws-amplify/amplify-flutter/commit/66542af4792618fe586e805bad63a48ac3b622b9 off main to start using secrets from AWS secrets manager
* changes auth integration test pull script to get new auth backend config from s3
* runs auth tests on web in CI

### DIFF
--- a/.github/workflows/amplify_integration_tests.yaml
+++ b/.github/workflows/amplify_integration_tests.yaml
@@ -41,10 +41,7 @@ jobs:
           role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
           aws-region: ${{ secrets.AWS_REGION }}
           scope: ${{ matrix.scope }}
-          api-app-id: ${{ secrets.API_APP_ID }}
-          auth-app-id: ${{ secrets.AUTH_APP_ID }}
-          datastore-app-id: ${{ secrets.DATASTORE_APP_ID }}
-          storage-app-id: ${{ secrets.STORAGE_APP_ID }}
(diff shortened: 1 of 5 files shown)


.github/workflows/amplify_integration_tests.yaml
```diff
@@ -98,3 +92,36 @@ jobs:
...
+
+      - name: Fetch Amplify backend configurations
+        uses: ./.github/composite_actions/fetch_backends
+        with:
+          role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
+          aws-region: ${{ secrets.AWS_REGION }}
+          scope: ${{ matrix.scope }}
+          secret-identifier: ${{ secrets.AWS_SECRET_IDENTIFIER }}
+
+      - uses: nanasess/setup-chromedriver@7cbd35794f8ab317f778c3172fb86c1e9b2853f7 # 1.1.0
```
dnys1, line 122: Seems fine but it doesn't do much - would rather just own a script locally we can run ourselves. Maybe we can add this to aft.

dnys1 (approved):

Equartey (approved):

.github/workflows/amplify_integration_tests.yaml
ragingsquirrel3: Don't have strong feelings about this, agree this community action doesn't do too much.

PR #2290, opened 2022-10-28
chore(analytics): integration tests

fjnoyp:
*Issue #, if available:*

*Description of changes:*

Draft - still getting tests to work after refactoring them. 

### DIFF
--- a/packages/analytics/amplify_analytics_pinpoint/example/integration_test/auto_session_tracking_test.dart
+++ b/packages/analytics/amplify_analytics_pinpoint/example/integration_test/auto_session_tracking_test.dart
@@ -1,0 +1,91 @@
+// Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// You may not use this file except in compliance with the License.
+// A copy of the License is located at
+//
+//  http://aws.amazon.com/apache2.0
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+import 'dart:async';
+
+import 'package:amplify_flutter/amplify_flutter.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:integration_test/integration_test.dart';
+
+import 'utils/mock_lifecycle_provider.dart';
+import 'utils/setup_utils.dart';
+
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  group('auto session tracking', () {
+    final mockLifecycleObserver = MockLifecycleProvider();
+
+    late final Stream<Map<String, Object?>> eventsStream;
+
+    setUpAll(() async {
+      eventsStream = await configureAnalytics(
+        appLifecycleProvider: mockLifecycleObserver,
+      );
+    });
+
+    tearDownAll(Amplify.reset);
+
+    test(
+      'manual trigger of onBackground/onForeground triggers session start/end events ',
+      () async {
+        mockLifecycleObserver.triggerOnBackgroundListener();
+
+        Map<String, String>? originalSession;
+
+        // Verify new session has newer values than old session
+        final streamSubscription = eventsStream.listen((event) {
+          expect(event['session'], isNotEmpty);
+
+          final session = event['session'] as Map<String, String>?;
+          if (originalSession == null) {
+            originalSession = session;
+          } else {
+            final newSession = session;
+
+            expect(originalSession!['id'], isNot(newSession!['id']));
+
+            final stopTimestamp =
+                DateTime.parse(originalSession!['stopTimestamp']!);
+
+            final startTimestamp =
+                DateTime.parse(newSession['startTimestamp']!);
+
+            expect(startTimestamp.isAfter(stopTimestamp), isTrue);
+          }
+        });
+
+        await expectLater(
+          eventsStream,
+          emits(
+            containsPair('event_type', '_session.stop'),
+          ),
+        );
+
+        mockLifecycleObserver.triggerOnForegroundListener();
+
+        expect(
+          eventsStream,
+          emitsThrough(
+            containsPair('event_type', '_session.start'),
+          ),
+        );
+
+        await Amplify.Analytics.flushEvents();
+
+        await streamSubscription.cancel();
+      },
+      timeout: const Timeout(Duration(minutes: 2)),
+    );
+  });
+}
--- a/packages/analytics/amplify_analytics_pinpoint/example/integration_test/enable_disable_test.dart
+++ b/packages/analytics/amplify_analytics_pinpoint/example/integration_test/enable_disable_test.dart
@@ -1,0 +1,62 @@
+// Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// You may not use this file except in compliance with the License.
+// A copy of the License is located at
+//
+//  http://aws.amazon.com/apache2.0
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+import 'dart:async';
+
+import 'package:amplify_flutter/amplify_flutter.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:integration_test/integration_test.dart';
+
+import 'utils/mock_lifecycle_provider.dart';
+import 'utils/setup_utils.dart';
+
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  group('enable/disable', () {
+    final mockLifecycleObserver = MockLifecycleProvider();
+
+    late final Stream<Map<String, Object?>> eventsStream;
+
+    setUpAll(() async {
+      eventsStream = await configureAnalytics(
+        appLifecycleProvider: mockLifecycleObserver,
+      );
+    });
+
+    tearDownAll(Amplify.reset);
+
+    test(
+      'disable prevents events from being auto flushed and sessions from being auto tracked',
+      () async {
+        await Amplify.Analytics.disable();
+
+        // Wait for all previous events to pass through
+        final streamSubscription = eventsStream.listen((event) {
+          fail('Analytics disabled but events were still sent!');
+        });
+
+        const customEventName = 'my custom event type name';
+        final customEvent = AnalyticsEvent(customEventName);
+
+        await Amplify.Analytics.recordEvent(event: customEvent);
+
+        mockLifecycleObserver.triggerOnBackgroundListener();
+        mockLifecycleObserver.triggerOnForegroundListener();
+
+        streamSubscription.cancel();
+      },
+      timeout: const Timeout(Duration(minutes: 2)),
+    );
+  });
+}
--- a/packages/analytics/amplify_analytics_pinpoint/example/integration_test/events_test.dart
+++ b/packages/analytics/amplify_analytics_pinpoint/example/integration_test/events_test.dart
@@ -12,89 +12,49 @@
 // permissions and limitations under the License.
 
 import 'dart:async';
-import 'dart:convert';
-import 'dart:math';
 
-import 'package:amplify_analytics_pinpoint/amplify_analytics_pinpoint.dart';
-import 'package:amplify_analytics_pinpoint_example/amplifyconfiguration.dart';
-import 'package:amplify_api/amplify_api.dart';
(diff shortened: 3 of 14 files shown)


fjnoyp:
2 issues to resolve today: 

1) When running the identify user test more than 15 times, it will fail from AWS Pinpoint due to max endpoints reached.  I need to confirm with Travis perhaps if new devices / backends are always provisioned for CI integration tests, else the tests will keep on failing after 15 times.  

This isn't a major issue as we will switch to the new Pinpoint API that does not have these limits. 

2) Need some help with expecting proper values in unit tests.  I want to check for string values in a map but get fails here: 

```
// expected:
                    // {'boolProperty': [true], 'name': ['name'], 'plan': ['plan'], 'email': ['email']}
                    // actual:
                    // {boolProperty: [true], name: [name], plan: [plan], email: [email]}
```

packages/analytics/amplify_analytics_pinpoint/lib/src/analytics_plugin_impl.dart
```diff
@@ -23,15 +23,22 @@ import 'package:amplify_secure_storage/amplify_secure_storage.dart';
...
-            ),
-          ),
-          pathProvider: FlutterPathProvider(),
-          appLifecycleProvider: FlutterAppLifecycleProvider(),
-          deviceContextInfoProvider: const FlutterDeviceContextInfoProvider(),
+  AmplifyAnalyticsPinpoint({
+    AmplifySecureStorageInterface? keyValueStore,
+    CachedEventsPathProvider? pathProvider,
+    AppLifecycleProvider? appLifecycleProvider,
+    DeviceContextInfoProvider? deviceContextInfoProvider,
```
dnys1, line 30: Should these be overridable? Maybe add `@visibleForTesting` for these.

packages/analytics/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
```diff
@@ -16,4 +16,12 @@
 /// Amplify Analytics Pinpoint
 library amplify_analytics_pinpoint;
 
+/// Overridable Flutter injected dependencies
+export 'package:amplify_analytics_pinpoint_dart/amplify_analytics_pinpoint_dart.dart'
+    show
+        AppLifecycleProvider,
+        CachedEventsPathProvider,
+        DeviceContextInfoProvider;
```
dnys1, line 24: I don't know if we should allow overriding these in dev-preview. Let's discuss since it will change the public API.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_client.dart
```diff
@@ -180,7 +180,16 @@ class EndpointClient {
...
     _endpointBuilder.user = newUserBuilder;
 
-    await updateEndpoint();
+    try {
+      await updateEndpoint();
+    } on BadRequestException catch (e) {
+      throw AnalyticsException(
+        e.message ?? 'BadRequestException',
+        recoverySuggestion:
+            'Only call identifyUser() once per device.  Calling identifyUser() multiple times will associate multiple endpoints to the device.',
```
dnys1, line 189: Every `BadRequestException` is due to this? I find that hard to believe.

packages/analytics/amplify_analytics_pinpoint/example/integration_test/utils/setup_utils.dart
```diff
@@ -0,0 +1,85 @@
...
+                  payload
+                }
+              }
+              ''',
+    ),
+    onEstablished: subscriptionEstablished.complete,
+  )
+      .map((event) {
+    if (event.errors.isNotEmpty) {
+      return {'errors': event.errors};
```
dnys1, line 57: `throw`? Are we ever expecting errors here?

packages/analytics/amplify_analytics_pinpoint/example/integration_test/enable_disable_test.dart
```diff
@@ -0,0 +1,66 @@
...
+
+        await Amplify.Analytics.recordEvent(event: customEvent);
+
+        mockLifecycleObserver.triggerOnBackgroundListener();
+        mockLifecycleObserver.triggerOnForegroundListener();
+
+        streamSubscription.cancel();
+
+        await Amplify.Analytics.enable();
+        await Amplify.Analytics.flushEvents();
```
dnys1, line 61: What is the point of this if you're not testing anything afterwards? Shouldn't you test to see that events are now received?

packages/analytics/amplify_analytics_pinpoint/example/integration_test/enable_disable_test.dart
```diff
@@ -0,0 +1,66 @@
...
+
+    test(
+      'disable prevents events from being auto flushed and sessions from being auto tracked',
+      () async {
+        await Amplify.Analytics.disable();
+
+        // Wait for all previous events to pass through
+        final streamSubscription = eventsStream.listen((event) {
+          expect(false, isTrue,
+              reason: 'Analytics disabled but events were still sent!');
```
dnys1, line 47:
```suggestion
          fail('Analytics disabled but events were still sent!');
```

packages/analytics/amplify_analytics_pinpoint/example/integration_test/auto_session_tracking_test.dart
```diff
@@ -0,0 +1,91 @@
...
+        final streamSubscription = eventsStream.listen((event) {
+          expect(event['session'], isNotEmpty);
+
+          final session = event['session'] as Map<String, String>?;
+          if (originalSession == null) {
+            originalSession = session;
+          } else {
+            final newSession = session;
+
+            expect(originalSession!['id'] == newSession!['id'], isFalse);
```
dnys1, line 56:
```suggestion
            expect(originalSession!['id'], isNot(newSession!['id']));
```

packages/analytics/amplify_analytics_pinpoint/example/integration_test/auto_session_tracking_test.dart
```diff
@@ -0,0 +1,91 @@
...
+                DateTime.parse(originalSession!['stopTimestamp']!);
+
+            final startTimestamp =
+                DateTime.parse(newSession['startTimestamp']!);
+
+            expect(startTimestamp.isAfter(stopTimestamp), isTrue);
+          }
+        });
+
+        expect(
```
dnys1, line 68: You want this expectation to complete before you trigger the onForeground listener.

```suggestion
        await expectLater(
```

packages/analytics/amplify_analytics_pinpoint/example/integration_test/events_test.dart
```diff
@@ -12,96 +12,57 @@
...
-        final customEventId = uuid();
-        final customEvent = AnalyticsEvent(customEventId);
+        const customEventName = 'my custom event type name';
+        final customEvent = AnalyticsEvent(customEventName);
+
+        expect(
+          eventsStream,
+          emits(allOf([
+            containsPair('event_type', customEventName),
+          ])),
```
dnys1, line 45:
```suggestion
          emits(
            containsPair('event_type', customEventName),
          ),
```

packages/analytics/amplify_analytics_pinpoint/example/integration_test/events_test.dart
```diff
@@ -135,11 +97,129 @@ void main() {
...
+            containsPair(
+                'endpoint',
+                allOf(
+                  containsPair('EndpointStatus', 'ACTIVE'),
+                  containsPair('OptOut', 'ALL'),
+                  containsPair(
+                    'Location',
+                    allOf(
+                      containsPair('Latitude', 0.0),
+                      containsPair('Longitude', 0.0),
```
dnys1, line 110: Why would these be 0s?

packages/analytics/amplify_analytics_pinpoint/example/integration_test/events_test.dart
```diff
@@ -135,11 +97,129 @@ void main() {
...
+          secondIntProperty.key,
+          secondIntProperty.value,
+        );
+
+        // Add attribute global property types
+        final attributeGlobalProperties = AnalyticsProperties()
+          ..addBoolProperty(boolProperty.key, boolProperty.value)
+          ..addStringProperty(stringProperty.key, stringProperty.value);
+
+        await Amplify.Analytics.registerGlobalProperties(
```
dnys1, line 146: Will global properties persist during tests? If so, should you be calling `Amplify.reset` after each test or should this go in its own group?

packages/analytics/amplify_analytics_pinpoint/example/integration_test/events_test.dart
```diff
@@ -135,11 +97,129 @@ void main() {
...
+            containsPair('event_type', customEventName),
+            containsPair(
+              'attributes',
+              // Boolean properties are recorded as strings
+              Map.fromEntries([
+                stringifiedBoolProperty,
+                stringProperty,
+                secondStringProperty
+              ]),
+            ),
```
dnys1, line 165: Add check for secondInt

packages/analytics/amplify_analytics_pinpoint/example/integration_test/identify_user_test.dart
```diff
@@ -0,0 +1,144 @@
...
+    late final Stream<Map<String, Object?>> eventsStream;
+
+    setUpAll(() async {
+      eventsStream = await configureAnalytics();
+    });
+
+    tearDownAll(Amplify.reset);
+
+    // TODO(fjnoyp): Running this locally too many times will cause max endpoint error of 15 from Pinpoint
+    // We need to delete/reset Analytics backend somehow
```
dnys1, line 36: Is it because you're using the same userId below? The pinpoint backend doesn't need to be reset each time, you just need to present as a different user, either by changing `userId` or making sure Auth is signedOut on teardowns.

packages/analytics/amplify_analytics_pinpoint/example/integration_test/identify_user_test.dart
```diff
@@ -0,0 +1,144 @@
...
+                    ),
+                    containsPair(
+                      'User',
+                      containsPair(
+                        'UserId',
+                        userId,
+                      ),
+                    ),
+                    // TODO(fjnoyp): Don't know why this is failing!
+                    // expected:
```
dnys1, line 110: Are strings being JSON-encoded somewhere?

packages/analytics/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
fjnoyp: Sure I can bring this up in our team meeting to see what consensus we get.  

Not sure how to run integration tests for AppLifecycleProvider if it's not exposed though.

packages/analytics/amplify_analytics_pinpoint/example/integration_test/utils/setup_utils.dart
fjnoyp: We are not expecting any errors here.  Errors should be returned in the client public API as bad request or too many request exceptions instead of coming through kinesis -> graphql streams.

packages/analytics/amplify_analytics_pinpoint/example/integration_test/enable_disable_test.dart
fjnoyp: Thanks yeah there is no point now.  For context added that when all tests were together to avoid event from one polluting another but this is no longer necessary anymore.

packages/analytics/amplify_analytics_pinpoint/example/integration_test/events_test.dart
fjnoyp: That's the default value, but you're right we shouldn't need to be checking them.

packages/analytics/amplify_analytics_pinpoint/example/integration_test/events_test.dart
fjnoyp: in this test, I ensure no global properties are persisted by calling: 
        await Amplify.Analytics.unregisterGlobalProperties();

Which unregisters all global properties.

packages/analytics/amplify_analytics_pinpoint/example/integration_test/events_test.dart
fjnoyp: thanks will do

packages/analytics/amplify_analytics_pinpoint/example/integration_test/identify_user_test.dart
fjnoyp: Resolved it now, yeah it's because of that and I needed to use your stringifiedboolproperty!

packages/analytics/amplify_analytics_pinpoint/example/integration_test/identify_user_test.dart
fjnoyp: thanks! That seems to fix it by changing userId.

fjnoyp:
Going to close this as I created a new PR based off of next

PR #2293, opened 2022-10-28
feat(storage): leverage amplify_db_common packages integrating drift

HuiSF:
*Issue #, if available:*

*Description of changes:*

See the list of commits.


packages/common/amplify_db_common/lib/src/connect_io.dart
```diff
@@ -20,10 +22,11 @@ import 'package:path_provider/path_provider.dart';
 /// {@macro amplify_db_common.connect}
 QueryExecutor connect({
   required String name,
-  String? path,
+  FutureOr<String>? path,
 }) {
   return LazyDatabase(() async {
-    if (path == null) {
+    final resolvedPath = await path;
```
fjnoyp, line 28: It seems the future itself is null so can we check if it's not null first before awaiting on it?

fjnoyp (approved):
Looks good one minor comment.  Thanks for tackling this!

packages/common/amplify_db_common/lib/src/connect_io.dart
dnys1, line 28:
```suggestion
    final resolvedPath = await (path ?? getApplicationSupportDirectory());
    return base.connect(name: name, path: resolvedPath);
```

Right now not passing a path here results in p.join being called twice (probably a miss on my part)

dnys1 (approved):

packages/common/amplify_db_common_dart/lib/src/connect.dart
```diff
@@ -12,6 +12,16 @@
...
+import 'dart:async';
+
+import 'package:drift/backends.dart';
+
 export 'connect_stub.dart'
     if (dart.library.html) 'connect_html.dart'
     if (dart.library.io) 'connect_io.dart';
+
+/// Interface of the [connect].
+typedef Connect = QueryExecutor Function({
```
fjnoyp, line 24: Since the connect function already takes a path parameter do we need to keep the path provider injected dependency as well?

HuiSF: Yea I think so, the result of path provide may vary between packages for different use cases.

packages/common/amplify_db_common/lib/src/connect_io.dart
HuiSF: `await null` is legit.

packages/common/amplify_db_common_dart/lib/src/connect.dart
fjnoyp: Perhaps the Connect function should already have the path provided?  It appears one needs a path anyway to get a queryExecutor anyway.

fjnoyp: So it's just exposed with the name parameter.

packages/common/amplify_db_common/lib/src/connect_io.dart
HuiSF: Should be ?
```dart
    final resolvedPath =
        await path ?? (await getApplicationSupportDirectory()).path;
```

packages/storage/amplify_storage_s3_dart/pubspec.yaml
```diff
@@ -10,12 +10,13 @@ environment:
...
 dependencies:
   amplify_core: ^1.0.0-next.0
+  amplify_db_common_dart: ^0.1.0
   async: ^2.8.2
   aws_common: ^0.2.2
   aws_signature_v4: ^0.2.0
   built_collection: ^5.0.0
   built_value: ">=8.4.0 <8.5.0"
-  drift: ^2.1.0
+  drift: any
```
fjnoyp, line 19: Are we sure we want to do any?  This appears like a new approach that we haven't done before so perhaps we should comment the motivations for doing so at least.

fjnoyp (approved):

packages/common/amplify_db_common/lib/src/connect_io.dart
fjnoyp: alright thanks guys!

dnys1: Yeah

packages/storage/amplify_storage_s3_dart/pubspec.yaml
dnys1: I think a comment here is fine- shouldn't be in the pubspec, though IMO. `drift` is listed here since the package directly depends on it. `any` means that you impose no constraints on the version you receive. Since we are also directly depending on amplify_db_common (which also depends on `drift`), we could either keep those version constraints in sync or use `any` to defer constraints elsewhere (in this case to `amplify_db_common` and any other transitive dependencies which depend on `drift`).

PR #2296, opened 2022-10-31
fix(authenticator): Username form validation Cognito requirements

### LINKED ISSUE #2188: Authenticator Username form validation does not match Cognito's requirements
### Description

Cognito uses the regex `[\p{L}\p{M}\p{S}\p{N}\p{P}]` for validate usernames. The Authenticator allows for any non-empty string.

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [ ] Auth
- [X] Authenticator
- [ ] DataStore
- [ ] Storage

### Steps to Reproduce

1. Set up an app with username/password based auth.
2. Try to sign up with a username that has a white space in the middle
3. **Actual behavior:** Request is sent to cognito, cognito error message is displayed.
4. **Expected behavior:** Form validation prevents the request from being sent to Cognito and displays a form validation message.

### Screenshots

_No response_

### Platforms

- [X] iOS
- [X] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Android Device/Emulator API Level

_No response_

### Environment

```bash
N/A
```

### Dependencies

```bash
N/A
```

### Device

N/A

### OS

N/A

### Deployment Method

Amplify CLI

### CLI Version

_No response_

### Additional Context

_No response_

### Amplify Config

N/A

Lorenzohidalgo:
*Issue #, if available:* resolves #2188

*Description of changes:* 
- Created new validator `usernameValidator`
- Modified `input_resolver.dart` to handle the case where the provided username does not match the basic requirements.
- Added the new string `usernameRequirements` to `inputs_en.arb`

Additional comments:
- I've updated the provided regex expression provided in the issue to avoid usernames with whitespace.
- The current regex expression for the email validator does accept inputs like `my@email`, new expression was proposed in the referenced issue.

I've not been able to generate the `l10n` files and therefore the linter is showing one error regarding the new string. Executing the `generate_l10n.sh` file did not generate the needed code for the new string. Let me know if there is any additional change/update I should apply for the script to work correctly.

### DIFF
--- a/packages/amplify_authenticator/lib/src/l10n/input_resolver.dart
+++ b/packages/amplify_authenticator/lib/src/l10n/input_resolver.dart
@@ -534,7 +534,7 @@ class InputResolver extends Resolver<InputResolverKey> {
       case InputResolverKeyType.empty:
         return empty(context, key.field);
       case InputResolverKeyType.usernameRequirements:
-        return passwordRequires(context, key.unmetPasswordRequirements!);
+        return usernameRequires(context);
       case InputResolverKeyType.passwordRequirements:
         return passwordRequires(context, key.unmetPasswordRequirements!);
       case InputResolverKeyType.mismatch:
--- a/packages/amplify_authenticator/lib/src/l10n/input_resolver.dart
+++ b/packages/amplify_authenticator/lib/src/l10n/input_resolver.dart
@@ -52,6 +52,7 @@ enum InputResolverKeyType {
   hint,
   confirmHint,
   empty,
+  usernameRequirements,
   passwordRequirements,
   format,
   mismatch
@@ -78,6 +79,10 @@ class InputResolverKey {
   );
   static const usernameEmpty = InputResolverKey._(
     InputResolverKeyType.empty,
+    field: InputField.username,
+  );
+  static const usernameRequirementsUnmet = InputResolverKey._(
+    InputResolverKeyType.usernameRequirements,
     field: InputField.username,
   );
   static const passwordTitle = InputResolverKey._(
@@ -477,6 +482,11 @@ class InputResolver extends Resolver<InputResolverKey> {
         .warnInvalidFormat(title(context, field).toLowerCase());
   }
 
+  /// Returns the text displayed when the username requirements are not met
+  String usernameRequires(BuildContext context) {
+    return AuthenticatorLocalizations.inputsOf(context).usernameRequirements;
+  }
+
   /// Returns the text displayed when a password input does match the password requirements
   /// defined in the amplify configuration.
   String passwordRequires(
@@ -523,6 +533,8 @@ class InputResolver extends Resolver<InputResolverKey> {
         return confirmHint(context, key.field);
       case InputResolverKeyType.empty:
         return empty(context, key.field);
+      case InputResolverKeyType.usernameRequirements:
+        return passwordRequires(context, key.unmetPasswordRequirements!);
       case InputResolverKeyType.passwordRequirements:
         return passwordRequires(context, key.unmetPasswordRequirements!);
       case InputResolverKeyType.mismatch:
--- a/packages/amplify_authenticator/lib/src/utils/validators.dart
+++ b/packages/amplify_authenticator/lib/src/utils/validators.dart
@@ -17,6 +17,7 @@ import 'package:amplify_authenticator/amplify_authenticator.dart';
 import 'package:amplify_flutter/amplify_flutter.dart';
 import 'package:flutter/material.dart';
 
+final usernameRegex = RegExp(r'[\p{L}\p{M}\p{S}\p{N}\p{P}]+');
 final emailRegex = RegExp(r'^\S+@\S+$');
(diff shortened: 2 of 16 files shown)


packages/amplify_authenticator/lib/src/l10n/input_resolver.dart
```diff
@@ -523,6 +533,8 @@ class InputResolver extends Resolver<InputResolverKey> {
         return confirmHint(context, key.field);
       case InputResolverKeyType.empty:
         return empty(context, key.field);
+      case InputResolverKeyType.usernameRequirements:
+        return passwordRequires(context, key.unmetPasswordRequirements!);
```
dnys1, line 537:
```suggestion
        return usernameRequires(context);
```

dnys1 (commented):
Minor comment, but otherwise LGTM! Thanks @Lorenzohidalgo!

Lorenzohidalgo:
Thanks for the catch @dnys1, I've committed your suggestion.
I've reviewed the failing tests, they are regarding automatically generated files:
```logs
flutter format --dry-run --set-exit-if-changed .
Changed lib\src\l10n\generated\country_localizations_en.dart
Changed lib\src\l10n\generated\message_localizations_en.dart
Changed lib\src\l10n\generated\title_localizations_en.dart
Formatted 129 files (3 changed) in 1.01 seconds.
Formatting failed: 1
```

Should I modify them manually so the PR passes all tests?

packages/amplify_authenticator/lib/src/utils/validators.dart
```diff
@@ -17,6 +17,7 @@ import 'package:amplify_authenticator/amplify_authenticator.dart';
 import 'package:amplify_flutter/amplify_flutter.dart';
 import 'package:flutter/material.dart';
 
+final usernameRegex = RegExp(r'[\p{L}\p{M}\p{S}\p{N}\p{P}]+');
```
Jordan-Nelson, line 20: Unfortunately it seems that the flavor of regex that Dart uses does not support `\p{L}`, `\p{M}`, etc. `\p{L}` is treated two separate matches - `\p` (literal "p" character) and `{L}` ("L" character). So in Dart, `[\p{L}\p{M}\p{S}\p{N}\p{P}]+` will match and string that has one of the follow characters in it - pLMSNP.

You can see the difference between common flavors on https://regex101.com/. PCRE2 allows for these characters. Ecmascript, which I believe is what dart follows, does not.

I propose we use `RegExp(r'^\S+$');` in this PR, and I can open a separate issue to look into a more complete solution. `RegExp(r'^\S+$');` will allow for any non white space character, which will be an improvement on the current validator which allows for usernames with white space. `RegExp(r'^\S+$');` will still allow for some characters which Cognito will reject, such as emojis, but this is still a step in the right direction.

In order to align this completely with Cognito, we will either need an exhaustive list of all the characters that are included in `[\p{L}\p{M}\p{S}\p{N}\p{P}]` in regex flavors that allow for this, or possibly a 3rd party lib.

This may apply to other validators besides username. We use `RegExp(r'^\S+@\S+$')` for email validation, which is probably also allowing for some characters that Cognito will reject, such as emojis.

@dnys1 - any objections to using `RegExp(r'^\S+$');` for now?

dnys1: Good catch. 

> any objections to using RegExp(r'^\S+$'); for now?
Works for me - maybe with a `TODO` comment as well.

Lorenzohidalgo: I've updated the regex expression to match the one you agreed on and added the required `TODO`.
It now looks like this:
```dart
/* TODO: update `usernameRegex` expression to match Cognito requirements
* Proposed expression `[\p{L}\p{M}\p{S}\p{N}\p{P}]+` does not work as 
* expected due to a mismatch in Regex flavor used by dart
*/
final usernameRegex = RegExp(r'^\S+$');
```

Jordan-Nelson (commented):
The new error message will require more than 1 line on many phones. Authenticator form fields default to having a errorMaxLines of 1, as seen here:

https://github.com/aws-amplify/amplify-flutter/blob/83e52981a1ccdab0e374bcb8791705bc1008de37/packages/amplify_authenticator/lib/src/widgets/form_field.dart#L172

I think this was more restrictive than it needed to be. Can you increase this to `2` as part of this PR to allow all error messages to have a max of two lines?

Here is a screenshot on an iPhone 14 showing the message is cut off

Lorenzohidalgo:
Sorry I missed that @Jordan-Nelson,  I've applied the requested change.

Jordan-Nelson (approved):

dnys1 (approved):

Jordan-Nelson:
Thanks for the contribution @Lorenzohidalgo! This will be included in our next release.

dnys1:
This has been published in version 0.6.10 of the library. Thanks again, @Lorenzohidalgo!

PR #2304, opened 2022-10-31
chore(aft): Speed up bootstrap

dnys1:
`build_runner` only needs to run for packages which require assets built for development. The other packages only need it run on publish.


HuiSF (approved):
Are other generated files, e.g. generated drift source files included in this step? or they should always be generated by manually running build_runner and committed into the codebase?

dnys1:
Should be committed or we can expand the check here if there's a good reason not to commit the files. But all generated files besides JS are currently checked in.

Jordan-Nelson (approved):

PR #2305, opened 2022-10-31
feat(storage): add listAll option of the list API

HuiSF:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/storage/amplify_storage_s3_dart/lib/src/model/s3_list_options.dart
+++ b/packages/storage/amplify_storage_s3_dart/lib/src/model/s3_list_options.dart
@@ -37,12 +37,13 @@ class S3ListOptions extends StorageListOptions {
     super.nextToken,
     this.targetIdentityId,
     this.excludeSubPaths = false,
+    this.listAll = false,
   });
 
   /// {@macro storage.amplify_storage_s3.list_options}
   ///
-  /// Use this when list objects that belongs to other user (identified by
-  /// [targetIdentityId]) rather than the currently signed user.
+  /// Use this to list objects that belongs to other user (identified by
+  /// [targetIdentityId]) rather than the currently signed in user.
   const S3ListOptions.forIdentity(
     String targetIdentityId, {
     int pageSize = 1000,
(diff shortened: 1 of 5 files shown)


packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
```diff
@@ -110,20 +110,55 @@ class StorageS3Service {
...
+          prefixToDrop: resolvedPrefix,
+        );
+      } on smithy.UnknownSmithyHttpException catch (error) {
+        // S3Client.headObject may return 403 error
+        throw S3Exception.fromUnknownSmithyHttpException(error);
+      }
+    }
+
+    late smithy.PaginatedResult<s3.ListObjectsV2Output, int, String> listResult;
+    late S3ListResult recursiveResult;
```
dnys1, line 135: Why do these need to be late? They can be `var`s when instantiated I think

packages/storage/amplify_storage_s3_dart/lib/src/model/s3_list_options.dart
```diff
@@ -56,6 +57,32 @@ class S3ListOptions extends StorageListOptions {
           excludeSubPaths: excludeSubPaths,
         );
 
+  /// {@macro storage.amplify_storage_s3.list_options}
+  ///
+  /// Use this when list all objects without pagination.
+  const S3ListOptions.listAll({
```
dnys1, line 63: Is listAll the primary consideration for a list operation? Just wondering why these are factories instead of options?

dnys1 (approved):

packages/storage/amplify_storage_s3_dart/lib/src/model/s3_list_options.dart
HuiSF: Alternative considered:
```dart
const S3ListOptions({
  StorageAccessLevel accessLevel = StorageAccessLevel.guest,
  int pageSize = 1000,
  bool excludeSubPaths = false,
  bool listAll = false,
  String? nextToken,
})
```
Was thinking `pageSize` and `listAll` are both params of the options constructor could be confusing, as if `listAll` is set to `true`, `pageSize` becomes ineffective. Open for suggestions :)

dnys1: Ah okay that makes sense then 👍

Jordan-Nelson (approved):

packages/storage/amplify_storage_s3_dart/lib/src/model/s3_list_options.dart
```diff
@@ -65,4 +92,13 @@ class S3ListOptions extends StorageListOptions {
...
   /// default value is `false`.
   final bool excludeSubPaths;
+
+  /// Whether to list all objects under a given path without pagination. The
+  /// default value is `false`.
+  ///
+  /// This can be set by calling the [S3ListOptions.listAll] or
+  /// [S3ListOptions.listAllForIdentity].
+  ///
+  /// Use with caution if numerous objects are under the given path.
```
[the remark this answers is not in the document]
HuiSF: Each list object request should comes back on time, but if there is a million objects to list (extreme case) it may cause memory leak and crash.

PR #2306, opened 2022-11-01
feat(authenticator): Add AutoFill Capabilities

### LINKED ISSUE #2189: [Feature Request] Autofill for SignInForm and SignUpForm
### Description

As a user, when login in to an app on an Amplify Authenticator sign-in form, I want to have my AutoFill service to fill in username and password for me.

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [ ] Auth
- [X] Authenticator
- [ ] DataStore
- [ ] Storage

### Steps to Reproduce

_No response_

### Screenshots

_No response_

### Platforms

- [ ] iOS
- [ ] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Android Device/Emulator API Level

_No response_

### Environment

```bash
N/A
```

### Dependencies

```bash
N/A
```

### Device

N/A

### OS

N/A

### Deployment Method

Amplify CLI

### CLI Version

_No response_

### Additional Context

_No response_

### Amplify Config

N/A

Lorenzohidalgo:
*Issue #, if available:* resolves #2189

*Description of changes:*
* Updated `AuthenticatorFormFieldState` to have a getter method to return the AutoFillHints `Iterable<String>? get autofillHints => null;`
* Updated the `build` method of `AuthenticatorUsernameField`, `AuthenticatorTextField`, and `AuthenticatorDateField` to pass the `autofillHints` to the created `TextFormField`.
* Updated the `showCountryDialog` method of `AuthenticatorPhoneFieldMixin` to provide `AutofillHints.countryName` to the search `TextField`.
* Updated `_SignUpFormFieldState`, `_SignUpDateFieldState`, `_SignInFormFieldState`, `_ResetPasswordFormFieldState` and `_ConfirmSignUpFormFieldState` to override the AutoFillHints getter and return the corresponding AutoFillHints list.
* Updated the `build` method of `AuthenticatorFormState` to wrap the `Column` inside the `Form` with an `AutofillGroup`
* Updated `AuthenticatorState` to call `TextInput.finishAutofillContext(shouldSave: true);` every time the form has been validated. This prompts the user to save or automatically save the input field content.
* Updated the `README.md` to reference the official Flutter documentation for further configurations and autofill troubleshooting.

Additional comments:
Adding this feature would benefit from an update to the package documentation, since:
* The [official documentation](https://api.flutter.dev/flutter/material/TextField/autofillHints.html) mentions the need to configure the associated domains to provide the best user experience to iOS users.
* Troubleshooting `AutoFill` can be tricky and there are some good-to-know steps and configurations mentioned in the official documentation.


packages/amplify_authenticator/lib/src/widgets/form_fields/sign_up_form_field.dart
```diff
@@ -739,4 +800,9 @@ class _SignUpDateFieldState extends _SignUpFormFieldState<String>
       isOptional: isOptional,
     );
   }
+
+  @override
+  Iterable<String>? get autofillHints => const [
+        AutofillHints.birthday,
+      ];
```
Jordan-Nelson, line 807: I think you can remove this. Setting it in `_SignUpFormFieldState` should be enough since `_SignUpDateFieldState` extends `_SignUpFormFieldState`.

Jordan-Nelson (commented):
Thanks for opening the PR!

This looks pretty good. Can you add autofill to `confirm_sign_in_form_field.dart` and `verify_user_form_field.dart` as well?

`confirm_sign_in_form_field.dart` has a few different purposes, such as MFA, but as far as autofill goes, it shouldn't be too different than the screens you have already updated in this PR. You will have to add cases for every type of `ConfirmSignInField`. Let me know if you have any questions.

The only text field shown in `verify_user_form_field.dart` is a confirmation code, so you can simply add to following to _VerifyUserTextFieldState.

```dart
  @override
  Iterable<String>? get autofillHints => [AutofillHints.oneTimeCode];
```

dnys1 (commented):
Thank you!

packages/amplify_authenticator/lib/src/widgets/form_fields/sign_up_form_field.dart
```diff
@@ -405,6 +405,67 @@ abstract class _SignUpFormFieldState<FieldValue>
...
+  Iterable<String>? get autofillHints {
+    switch (widget.field) {
+      case SignUpField.password:
+      case SignUpField.passwordConfirmation:
+        return const [
+          AutofillHints.newPassword,
+        ];
+      case SignUpField.address:
+        return const [
+          AutofillHints.postalAddress,
```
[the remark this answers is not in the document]
Lorenzohidalgo: Thanks for the catch @dnys1! I've applied your suggestion and added a table in the `README` file stating the current platform support for each `AutoFillHint` used in this project.

Lorenzohidalgo:
Hey there @Jordan-Nelson  and @dnys1,

I've applied your suggestions, could you please verify that my proposed changes fit your requests?

packages/amplify_authenticator/lib/src/widgets/form_fields/confirm_sign_in_form_field.dart
```diff
@@ -90,26 +97,30 @@ abstract class ConfirmSignInFormField<FieldValue>
...
         field: ConfirmSignInField.customChallenge,
         validator: validator,
+        autofillHints: autofillHints,
       );
 
   /// Creates a verification code component.
   static ConfirmSignInFormField verificationCode({
     Key? key,
     FormFieldValidator<String>? validator,
+    Iterable<String>? overrideAutofillHints,
```
dnys1, line 107: I would rather name these all the same `autofillHints` since we may add defaults on other fields in the future and the other parameters to these functions already override defaults.

dnys1 (commented):
One small comment but otherwise looks good!

packages/amplify_authenticator/lib/src/widgets/form_fields/confirm_sign_in_form_field.dart
Lorenzohidalgo: Thanks Dillon, I'll apply the change to name all inputs `autofillHints` instead of `overrideAutofillHints`.

dnys1 (dismissed):
LGTM!

packages/amplify_authenticator/lib/src/mixins/authenticator_username_field.dart
```diff
@@ -260,6 +260,7 @@ mixin AuthenticatorUsernameField<FieldType,
       ),
       keyboardType: keyboardType,
       obscureText: false,
+      autofillHints: autofillHints,
```
Jordan-Nelson, line 263: Further up in this file there is a widget `AuthenticatorPhoneField()`, which is used when phone number is set as the username. The auto fill groups will need to be passed there as well. Right now I am getting nothing when phone number is used as a sign up alias (username).

If you want to test that out, you can update the `usernameAttributes` value in your amplify config temporarily.

```
"usernameAttributes": [
    "PHONE_NUMBER"
],
```

packages/amplify_authenticator/lib/src/widgets/form.dart
```diff
@@ -174,27 +174,33 @@ class AuthenticatorFormState<T extends AuthenticatorForm>
...
-              if (widget.actions.isNotEmpty) const SizedBox(height: 4),
-              ...widget.actions,
-              if (runtimeActions.isNotEmpty) ...[
-                const Divider(),
-                ...runtimeActions,
-              ]
-            ].spacedBy(const SizedBox(height: 12)),
-          ),
-        ],
+      child: AutofillGroup(
```
Jordan-Nelson, line 187: I am not that familiar with auto fill groups, but I don't think this is how `AutofillGroup` should be used. My understanding is that it should be used when all the fields are related. For example, a multi field address, a multi field credit card input, or possibly a username/password group. The use of `AutofillGroup` around the entire form seems to lead to some odd behavior, at least on iOS.

I noted a change above that will have to be made to get auto fill working for phone number as a sign up alias. When the change noted above is made, the sign up form has some strange behavior. It suggests using an email for the phone number field, a username for the password, etc. Removing this `AutofillGroup` resolves that (phone shows a phone number, password shows a password, etc.).

I'm not sure if removing it all together is the correct solution. It might provide some utility. I'll have to look into it a bit more. If this is something you are familiar with, please feel free to provide some insight.

Jordan-Nelson (commented):
The changes you made look good. I noted one other.

As I noted in a comment, I need to look into AutofillGroup a bit, because it seems to be leading to some strange behavior (at least on iOS).

packages/amplify_authenticator/lib/src/mixins/authenticator_username_field.dart
Lorenzohidalgo: Hi @Jordan-Nelson, I've added the autofill hints to the AuthenticatorPhoneField but it's now triggering an overflow error.

I'm pretty busy these days, will try to debug it once I have a bit more time.

packages/amplify_authenticator/lib/src/widgets/form.dart
Lorenzohidalgo: I have reviewed the official documentation and, as you mention, my use of `AutoFillGroup` was not correct. I've removed it.

It might be useful to update the code in the future to group the fields automatically into logical `AutoFillGroup`s

packages/amplify_authenticator/lib/src/mixins/authenticator_username_field.dart
Jordan-Nelson: Do you mean that the UI is overflowing the screen? What device/simulator are you testing on? Can you share a screenshot if you have the chance? I'm not sure how this change would have impacted that. It could be unrelated. If you have a screenshot or can share what device you are seeing that on I can take a look when I get a chance.

Lorenzohidalgo: @Jordan-Nelson , the textformfield for the phonenumber does overflow when autofillhints are added. I've used a OnePlus 5T for debugging

Jordan-Nelson (commented):
I added some small suggestions. I will push changes for those

packages/amplify_authenticator/lib/src/mixins/authenticator_username_field.dart
Jordan-Nelson: I think the overflow you were seeing was a stack overflow due to a recursive call. I pushed a fix. I am going to resolve this comment.

Jordan-Nelson (approved):

dnys1 (approved):

packages/amplify_authenticator/lib/src/widgets/form.dart
0xAPPA: Did you test the autofill functionality after removing AutoFillGroup with email and password?
From what I've tested (iOS device), autofilling email/password or auto-generating a password via keychain **only** works with AutofillGroup. However, telephoneNumber an givenName works without AutoFillGroup.

PR #2307, opened 2022-11-01
fix: HTTP cancellation

dnys1:
Various fixes and QoL improvements for HTTP cancellation and stream forwarding in the smithy and aws_common libraries.


packages/aws_common/lib/src/http/aws_http_client_io.dart
```diff
@@ -441,7 +443,7 @@ class AWSHttpClientImpl extends AWSHttpClient {
   }) {
     final requestProgressController = StreamController<int>.broadcast();
     final responseProgressController = StreamController<int>.broadcast();
-    final cancelTrigger = Completer<void>();
+    final cancelTrigger = Completer<void>.sync();
```
ragingsquirrel3, line 446: nit: Could we have a little comment explaining why it's safe to use `sync` constructor here?

ragingsquirrel3 (approved):

packages/aws_common/lib/src/http/aws_http_client_io.dart
dnys1: Done 👍

PR #2331, opened 2022-11-08
ci(repo): New CI pipeline

dnys1:
- Removes `mono_repo` dependency throughout
- Adds custom workflows for Dart+Flutter packages
- Adds `aft generate workflows` command for automatically generating workflows


packages/aft/lib/src/models.dart
```diff
@@ -80,11 +140,20 @@ class PackageInfo
             falsePositiveExamples.contains(name));
   }
 
+  /// Skip package checks for Flutter packages when running in CI without
+  /// Flutter, which may happen when testing Dart-only packages or specific
+  /// Dart versions.
```
Jordan-Nelson, line 145: I don't think I follow this. Would this be a dart-only package that has flutter listed as a dependency for some reason?

packages/aft/lib/src/commands/generate/generate_workflows_command.dart
```diff
@@ -0,0 +1,143 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'dart:io';
+
+import 'package:aft/aft.dart';
+import 'package:path/path.dart' as p;
+
+/// Command for generating associated items in the repo.
```
Jordan-Nelson, line 20: Update comment

packages/aft/lib/src/models.dart
```diff
@@ -49,26 +68,67 @@ class PackageInfo
...
-  final bool usesMonoRepo;
-
   /// The package's pubspec.
   final PubspecInfo pubspecInfo;
 
   /// The package flavor, e.g. Dart or Flutter.
   final PackageFlavor flavor;
 
+  /// The integration test directory within the enclosing directory, if any
+  Directory? get integTestDirectory {
```
Jordan-Nelson, line 103: This isn't used currently, correct? Are you workflows for integ tests could be generated as well?

packages/aft/README.md
```diff
@@ -7,7 +7,8 @@ A CLI tool for managing the Amplify Flutter repository.
 - `bootstrap`/`bs`: Sets up repo for development work
 - `clean`: Cleans temporary files and build artifacts for all packages
 - `deps check`: Checks dependencies of all packages against `aft.yaml`
-- `generate-sdk`: Generates the AWS SDK for a given package
+- `generate`: Generates various repo items
+  - `workflows`: Generates GitHub actions workflows for all packages in the repo
```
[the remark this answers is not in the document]
dnys1: I marked the `sdk` command as hidden for right now. It only works for me.

packages/aft/lib/src/models.dart
dnys1: It's not currently used. Needed for `aft test` though. I can remove it here.

packages/aft/lib/src/models.dart
dnys1: This is a roundabout way of saying "don't try to run `flutter pub upgrade` when `flutter` is not available"

.github/workflows/flutter_vm.yaml
```diff
@@ -0,0 +1,67 @@
...
+      matrix:
+        sdk:
+          - stable
+          - master
+    steps:
+      - name: Git Checkout
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # 3.1.0
+
+      - name: Setup Flutter
+        uses: subosito/flutter-action@dbf1fa04f4d2e52c33185153d06cdb5443aa189d # 2.7.0
```
ragingsquirrel3, line 27: uber nit: isn't this 2.8.0 https://github.com/subosito/flutter-action/commit/dbf1fa04f4d2e52c33185153d06cdb5443aa189d?

.github/workflows/amplify_datastore.yaml
```diff
@@ -0,0 +1,26 @@
+name: amplify_datastore
+on:
+  push:
+    branches:
+      - main
+      - stable
+      - next
+  pull_request:
+    paths:
+      - 'packages/amplify_datastore/**/*.dart'
```
ragingsquirrel3, line 10: question: Would it make sense to also include interface paths to this pattern so that changes to the interface execute these tests?

.github/workflows/amplify_api.yaml
```diff
@@ -0,0 +1,26 @@
+name: amplify_api
+on:
+  push:
+    branches:
+      - main
+      - stable
+      - next
+  pull_request:
+    paths:
+      - 'packages/api/amplify_api/**/*.dart'
```
ragingsquirrel3, line 10: question: Is the reason for having the category flutter vm tests all be their own workflow that is seems difficult to enforce this pr/pattern thing any other way? Absent that, seems like a matrix would work nicely but not sure how you would do this via matrix. Seems to make sense, just trying to make sure I understand why.

.github/workflows/dart_native.yaml
```diff
@@ -0,0 +1,64 @@
...
+    name: Native Test
+    timeout-minutes: 60
+    strategy:
+      # Allows other matrix items to run if one fails
+      fail-fast: false
+      matrix:
+        os:
+          # macos-latest currently points to macos-11
+          # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on
+          - macos-12
```
ragingsquirrel3, line 22: callout (no change needed inthis PR): This is going to consume lots of macos compute across its consumers, so just might be something to keep an eye on if we hit a limit at some point. Also dk if people run workflows on their fork like I do where you can only run 5 mac jobs at once.

packages/aft/lib/src/commands/generate/generate_workflows_command.dart
```diff
@@ -0,0 +1,147 @@
...
+        'amplify_secure_storage_test',
+      ];
+      final needsWebTest =
+          package.pubspecInfo.pubspec.devDependencies.containsKey('build_test');
+      final testDdcStable =
+          needsWebTest && !failingDdcStable.contains(package.name);
+      final isDartPackage = package.flavor == PackageFlavor.dart;
+      final workflowContents = StringBuffer(
+        '''
+name: ${package.name}
```
ragingsquirrel3, line 71: nit/question: Should this include a comment that they are autogenerated in the output? I had no idea until I got here.

packages/auth/amplify_auth_cognito_test/test/flows/srp/srp_password_verifier_worker_test.dart
```diff
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// TODO(dnys1): Investigate DDC failures
+@Tags(['no-ddc'])
```
ragingsquirrel3, line 16: nit: Would we ever consider keeping tags in a centralized place as consts/enums? Like maybe in amplify_test? That way they're easier to keep track of/not repeat, etc...

packages/common/amplify_db_common/pubspec.yaml
```diff
@@ -11,7 +11,7 @@ environment:
 
 dependencies:
   amplify_db_common_dart: ^0.1.0
-  drift: ^2.2.0
+  drift: ^2.0.0
```
ragingsquirrel3, line 14: why is this changed here/other places?

ragingsquirrel3 (approved):
Approved w some small nit/questions/callouts that are non-blocking.

One general question: When would it make sense to remove the tests from circle that are redundant here? Specifically thinking of flutter unit tests. I know you mentioned it's on the roadmap to remove circle runners when Dart impls finished but dk if we want to specifically remove the same tests being run in 2 places.

.github/workflows/amplify_core.yaml
```diff
@@ -0,0 +1,47 @@
+name: amplify_core
+on:
+  push:
+    branches:
+      - main
+      - stable
+      - next
+  pull_request:
+    paths:
+      - 'packages/amplify_core/**/*.dart'
```
ragingsquirrel3, line 10: Fore these patterns, I dk if would make more sense to specify what should NOT run the tests, like `!**/*README.md` than trying to state what SHOULD and leave this open to errors of omission, but I do not feel strongly one way or the other.

.github/workflows/dart_native.yaml
dnys1: Good point. My thinking with these is that they will only run on pushes (not pull requests) so they can run asynchronously without blocking any real work. Open to further discussion, though (maybe run them on a schedule instead?)

.github/workflows/amplify_datastore.yaml
dnys1: Yes, in the future this will be possible to do, specifically when #2068 is merged since it has the semantic understanding of the repo (which packages depend on which).

.github/workflows/amplify_api.yaml
dnys1: Not sure I understand the question. Each package gets its own workflow. Which workflows run depends on the structure of the package. The `flutter_vm` workflow is chosen for any package which depends on Flutter.

packages/auth/amplify_auth_cognito_test/test/flows/srp/srp_password_verifier_worker_test.dart
dnys1: This is definitely temporary. At some point, if we have more permanent tags used throughout, that could definitely make sense.

packages/common/amplify_db_common/pubspec.yaml
dnys1: The changes are from this commit https://github.com/aws-amplify/amplify-flutter/pull/2331/commits/be12ace6c3b71009b9df523c9727558313979b33

.github/workflows/amplify_core.yaml
dnys1: My feeling is that CI checks in PRs should do the absolute minimum amount of work to verify the change. This is best accomplished through an "opt-in" rule vs "opt-out". On push, CI checks run for _any_ change and this is a better mechanism for catching things which weren't immediately obvious.

packages/aft/lib/src/commands/generate/generate_workflows_command.dart
dnys1: Good point. Updated 👍

.github/workflows/flutter_vm.yaml
dnys1: Whoops. Updated 👍

.github/workflows/amplify_api.yaml
ragingsquirrel3: My bad for confusing phrasing (I also saw this bit/left this comment before I saw the autogeneration code). My point was that the similar workflow logic is running/repeated across these categories. You could run them via a matrix to save repitition like the integration tests but I'm not sure how you would stop one category from running on pushes to certain files with that pattern, so I thought that was why you did it this way. Knowing they are autogenerated would also be another reason.

dnys1: Yeah, since paths affect the workflow as a whole and there isn't a way to control this at any lower level from what I've seen. Since it's autogenerated, there's no real maintenance risk to 1 workflow vs. 50. It's also why all the workflow logic is centralized in `flutter_vm`, so all these files do is control the `paths` to run on, essentially.

We also have more information about the packages in `aft` than we do once the workflow is running which is another reason we generate them.

Jordan-Nelson (approved):

PR #2333, opened 2022-11-09
feat(api): Web Socket - Hub Events

Equartey:
*Description of changes:*
Adds Hub Events to Web Socket Bloc. State change event are emitted to the Amplify Hub.

Also renamed test folder from `packages/api/amplify_api/test/ws` to `packages/api/amplify_api/test/web_socket`


packages/api/amplify_api/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart
```diff
@@ -343,6 +350,25 @@ class WebSocketBloc with AWSDebuggable, AmplifyLoggerMixin {
...
+      _hubEventsController.add(SubscriptionHubEvent.connecting());
+    } else if (state is ConnectedState) {
+      _hubEventsController.add(SubscriptionHubEvent.connected());
+    } else if (state is PendingDisconnect) {
+      _hubEventsController.add(SubscriptionHubEvent.pendingDisconnect());
+    } else if (state is DisconnectedState) {
+      // Once disconnected we throw away the bloc and need to close hub controller
+      _hubEventsController
+        ..add(SubscriptionHubEvent.disconnected())
+        ..close();
```
dnys1, line 366: Let's keep this method single-purpose and only close the controller when the rest of them are closed

packages/api/amplify_api/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart
```diff
@@ -38,6 +38,9 @@ class WebSocketBloc with AWSDebuggable, AmplifyLoggerMixin {
     required AmplifyAuthProviderRepository authProviderRepo,
     required WebSocketService wsService,
   }) {
+    // Establish HubEvents Stream
+    Amplify.Hub.addChannel(HubChannel.Api, _hubEventsController.stream);
```
dnys1, line 42: I would move all this logic up a layer to `AmplifyAPI`. Reason being that hub events are not really the concern of the bloc, but of the plugin. In Auth, we do the same, so the state machine has no knowledge of hub events; the plugin listens to the state machine and translates its state changes to hub events.

packages/api/amplify_api/lib/src/api_plugin_impl.dart
```diff
@@ -166,11 +190,7 @@ class AmplifyAPIDart extends AmplifyAPI {
     )..supportedProtocols = SupportedProtocols.http1;
   }
 
-  /// Returns the websocket bloc to use for a given endpoint.
-  ///
-  /// Use [apiName] if there are multiple endpoints.
-  @visibleForTesting
-  WebSocketBloc getWebSocketBloc({String? apiName}) {
+  WebSocketBloc _webSocketBloc({String? apiName}) {
```
dnys1, line 193: I would reorganize this part a little bit:

```dart
  WebSocketBloc _webSocketBloc({String? apiName}) {
    final endpoint = _apiConfig.getEndpoint(
      type: EndpointType.graphQL,
      apiName: apiName,
    );

    return _webSocketBlocPool[endpoint.name] ??= getWebSocketBloc(endpoint);
  }

  /// Returns the websocket bloc to use for a given endpoint.
  ///
  /// Use [endpoint] if there are multiple endpoints.
  @visibleForTesting
  WebSocketBloc getWebSocketBloc(EndpointConfig endpoint) {
    return WebSocketBloc(
      config: endpoint.config,
      authProviderRepo: _authProviderRepo,
      wsService: AmplifyWebSocketService(),
    )..stream.listen((event) {
        _emitHubEvent(event);

        if (event is PendingDisconnect) {
          _webSocketBlocPool.remove(endpoint.name);
        }
      });
  }
```

packages/api/amplify_api/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart
```diff
@@ -55,6 +55,10 @@ class WebSocketBloc with AWSDebuggable, AmplifyLoggerMixin {
   @override
   String get runtimeTypeName => 'WebSocketBloc';
 
+  /// Indicates if the bloc has finished closing
+  @visibleForTesting
+  final closed = Completer();
```
dnys1, line 60: Can we call it `done` to match `Sink.done` and mark it `<void>`

```suggestion
  final done = Completer<void>();
```

Instead of awaiting it in the test, can we await it in `ApiPlugin.reset`? And then in the test, we can await `Amplify.reset` to get the expected behavior.

ragingsquirrel3 (commented):
Looks good with one nit, just suggesting we add an integration test for hub events.

packages/api/amplify_api/lib/src/api_plugin_impl.dart
```diff
@@ -60,9 +62,17 @@ class AmplifyAPIDart extends AmplifyAPI {
...
 
+  final StreamController<ApiHubEvent> _hubEventController =
+      StreamController<ApiHubEvent>.broadcast();
+
   /// Subscription options
   final GraphQLSubscriptionOptions? subscriptionOptions;
 
+  @override
+  Future<void> reset() async {
+    await _hubEventController.close();
```
dnys1, line 73: I would do something like this to ensure this works correctly given what we saw in the tests

```suggestion
    for (final bloc in _webSocketBlocPool.values) {
       bloc.add(ShutdownEvent());
    }
    await _Future.wait(_webSocketBlocPool.values.map((bloc) => bloc.done));
    await _hubEventController.close();
```

packages/api/amplify_api/lib/src/api_plugin_impl.dart
```diff
@@ -140,6 +150,20 @@ class AmplifyAPIDart extends AmplifyAPI {
     }
   }
 
+  void _emitHubEvent(WebSocketState state) {
```
dnys1, line 153: At some point we should add `apiName` to the events so they can be distinguished when multiple blocs are running. Can you add a TODO for now?

dnys1 (approved):

packages/api/amplify_api/test/graphql_test.dart
```diff
@@ -161,13 +162,16 @@ class MockAmplifyAPI extends AmplifyAPIDart {
   });
 
   @override
-  MockWebSocketBloc getWebSocketBloc({String? apiName}) {
+  MockWebSocketBloc getWebSocketBloc(EndpointConfig endpoint) {
```
ragingsquirrel3, line 165: I think unit tests are failing bc this needs to be renamed to match renamed util.

ragingsquirrel3 (approved):
LGTM. Approved assuming small comment addressed so that flutter unit tests pass.

PR #2366, opened 2022-11-16

dnys1:
Updates dependencies throughout the repo. `pub` does not approve of setting version constraints to `any` as we were doing before. `aft.yaml` is leveraged to keep `drift` dependency in sync.

`dev_dependencies` versions are conflicting with `dependencies` versions on publish. New rule is don't list a version for repo packages in `dev_dependencies` since it will be handled by `pubspec_overrides` anyway.

### DIFF
--- a/packages/amplify/amplify_flutter/pubspec.yaml
+++ b/packages/amplify/amplify_flutter/pubspec.yaml
@@ -32,12 +32,12 @@ dependencies:
   plugin_platform_interface: ^2.0.0
 
 dev_dependencies:
-  amplify_analytics_pinpoint: 1.0.0-next.0
-  amplify_api: 1.0.0-next.0
-  amplify_auth_cognito: 1.0.0-next.0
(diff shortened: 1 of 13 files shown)


HuiSF (approved):

packages/amplify/amplify_flutter/pubspec.yaml
```diff
@@ -32,12 +32,12 @@ dependencies:
   plugin_platform_interface: ^2.0.0
 
 dev_dependencies:
-  amplify_analytics_pinpoint: 1.0.0-next.0
-  amplify_api: 1.0.0-next.0
-  amplify_auth_cognito: 1.0.0-next.0
-  amplify_datastore: 1.0.0-next.0
+  amplify_analytics_pinpoint:
```
ragingsquirrel3, line 35: add values here?

Jordan-Nelson (approved):

PR #2388, opened 2022-11-18
chore(analytics): integ tests in CI

ragingsquirrel3:
Cherry picks 026cc9daa4582a4ecc9fc1d0117e9beddde83a07 from main to enable analytics tests in CI for Android/iOS and then adds another commit to enable in mac/web/linux.

### DIFF
--- a/packages/analytics/amplify_analytics_pinpoint/lib/src/analytics_plugin_impl.dart
+++ b/packages/analytics/amplify_analytics_pinpoint/lib/src/analytics_plugin_impl.dart
@@ -18,19 +18,23 @@ import 'package:amplify_analytics_pinpoint/src/flutter_path_provider/flutter_pat
 import 'package:amplify_analytics_pinpoint_dart/amplify_analytics_pinpoint_dart.dart';
 import 'package:amplify_db_common/amplify_db_common.dart' as db_common;
 import 'package:amplify_secure_storage/amplify_secure_storage.dart';
+import 'package:meta/meta.dart';
 
 /// {@template amplify_analytics_pinpoint.analytics_plugin_impl}
 /// The AWS Pinpoint implementation of the Amplify Analytics category.
 /// {@endtemplate}
 class AmplifyAnalyticsPinpoint extends AmplifyAnalyticsPinpointDart {
   /// {@macro amplify_analytics_pinpoint.analytics_plugin_impl}
-  AmplifyAnalyticsPinpoint()
-      : super(
-          keyValueStore: AmplifySecureStorage(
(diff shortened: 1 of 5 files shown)


Jordan-Nelson (approved):

packages/analytics/amplify_analytics_pinpoint/lib/src/analytics_plugin_impl.dart
```diff
@@ -21,16 +21,23 @@ import 'package:amplify_secure_storage/amplify_secure_storage.dart';
...
 class AmplifyAnalyticsPinpoint extends AmplifyAnalyticsPinpointDart {
   /// {@macro amplify_analytics_pinpoint.analytics_plugin_impl}
-  AmplifyAnalyticsPinpoint()
-      : super(
-          keyValueStore: AmplifySecureStorage(
-            config: AmplifySecureStorageConfig(
-              scope: 'analyticsPinpoint',
-            ),
-          ),
+  AmplifyAnalyticsPinpoint({
```
ragingsquirrel3, line 32: @fjnoyp @Jordan-Nelson thoughts on this? This was needed for my testing use case but wasn't sure if others would want an escape hatch like auth. The alternative would be bringing in the dart plugin in the integ test and overriding that way I suppose.

Jordan-Nelson: Can you make the following changes:
- Make this `@visibleForTesting`. Exposing this for overriding within an app probably deserves more discussion. Allowing it to be overridden for testing seems reasonable though.
- add a TODO to rename this. `keyValueStore` doesn't describe this well. Probably want a better name if it will be exposed (even if just for testing).


```dart
// TODO(fjnoyp): Rename to reflect the data that is stored.
@visibleForTesting SecureStorageInterface? keyValueStore,
```

ragingsquirrel3: just made this change, thanks, leaving thread open for Kyle to comment

dnys1 (approved):

packages/analytics/amplify_analytics_pinpoint/lib/src/analytics_plugin_impl.dart
fjnoyp: Yeah had the same problem when developing, the annotation visible for testing looks good.  Thanks for keeping open for me to see.

fjnoyp (approved):

PR #2396, opened 2022-11-21
chore(repo): Align example app secure storage settings

dnys1:
- Aligns all example apps to share the same secure storage settings and warnings suggesting alternate usage in production cases.
- Aligns macOS Xcode settings to have Keychain Sharing disabled, Network permissions enabled, and deployment target set to 10.15.


ragingsquirrel3 (approved):
Thanks for this!

Issue #2398, opened 2022-11-21
When working offline, id/access/refresh token expiration time is not honored.


heikeschmitz:
### Description

I set the expiration time for the ID and the Access tokens to 1 day and the Refresh token to 360 days.
But when I then go and work offline, I am asked to sign back in already after 1 hour. 

Note:
- I verified that the times were set correctly before and after the error occurred. 
(see screen shot attached:
![Cognito Screen Shot 2022-11-21 at 10 42 04 AM](https://user-images.githubusercontent.com/109546239/203138254-d9d2beaf-28d1-4875-afcb-560bfff3294d.png)
)

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [X] Auth
- [X] Authenticator
- [ ] DataStore
- [ ] Storage

### Steps to Reproduce

1. Set the expiration of the tokens in AWS Cognito.
    1.1 Open AWS Cognito -> User Pools
    1.2 Click on your user pool
    1.3 Select the App Integration tab.
    1.4 Scroll down to the list of app clients
    1.5 Click on your app client 
    1.6 Under 'App client information', click the Edit button.
    1.7 Enter 1 day for the expiration of both, the ID token and the Access token.
    1.8 Enter 360 days for the expiration of the Refresh token.
    1.9 Click the 'Save changes' button.
2. While online, freshly sign-in to your app (if necessary sign out of it before).
3. Close the app (without signing out.)
4. Go offline (e.g., switch on Airplane mode)
5. Wait 1 hour 5 minutes.
6. While still offline, restart the app.
7. The bug: You are asked to sign in again.
    What should happen: Since only 1 hour 5 minutes passed (not 1 day or more), you should not need to sign in again.
    Why this is a problem: At this point you are offline and cannot sign in again so you are unable to work with the app offline. 
    The app works fine up to roughly an hour after the sign in and only then asks for a new sign in, if offline.

### Platforms

- [X] iOS
- [X] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Environment

```bash
heike@Heikes-MacBook-Pro-2 eplant_mobile_app % flutter doctor
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.3.6, on macOS 11.4 20F71 darwin-x64, locale en-US)
[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.0)
[...]

PR #2404, opened 2022-11-22
fix(authenticator)!: use `isSignedIn` to determine auth state on launch

Jordan-Nelson:
DO NOT MERGE. SEE BELOW FOR DETAILS.

*Issue #, if available:* #2398

*Description of changes:*
Use is `isSignedIn` to determine auth state on launch, rather than the presence of Cognito User Pool Tokens.

This is a breaking change since it changes the behavior on app launch for an end user that had authenticated and then went into an offline state for a period longer than the token expiration.

Do not merge until the change is behavior is confirmed to be desired.


dnys1 (approved):

Jordan-Nelson:
Closing since this introduces other issues. Will need to find an alternate solution.

PR #2407, opened 2022-11-23
chore(repo): use dart pub instead of flutter pub to activate melos

NikaHsn:
*Issue #, if available:*

*Description of changes:*
use `dart pub global` instead of `flutter pub global` to activate melos.
running `flutter pub global run melos bootstrap` returns `no active package melos` in response.
we should use `dart pub` for dart package management and use `flutter pub` in flutter projects.


CONTRIBUTING.md
```diff
@@ -64,11 +64,11 @@ See [invertase/melos](https://github.com/invertase/melos) for more instructions
 ```
 $ git clone git@github.com:[username]/amplify-flutter.git
 $ cd amplify-flutter
-$ flutter pub global activate melos
+$ dart pub global activate melos
 $ melos bootstrap
 ```
 
-> Note: If you don't include `melos` on your path, you may execute `flutter pub global run melos bootstrap` instead of the last command above.
+> Note: If you don't include `melos` on your path, you may execute `dart pub global run melos bootstrap` instead of the last command above.
```
dnys1, line 71:
```suggestion
> Note: If you're not able to run `melos`, follow [these instructions](https://dart.dev/tools/pub/cmd/pub-global#running-a-script-from-your-path) for updating your PATH to include Dart's package cache. Alternatively, you may execute `dart pub global run melos bootstrap` instead of `melos bootstrap`.
```

dnys1 (approved):
One small comment - I think we should include instructions for updating PATH. Otherwise LGTM!

CONTRIBUTING.md
NikaHsn: `dart pub global activate melos` returns a warning message on how to update the PATH as needed. so we may skip adding it into this doc. wdyt?
this is the dart pub warning message on how to update PATH:
```
Warning: Pub installs executables into $HOME/.pub-cache/bin, which is not on your path.
You can fix that by adding this to your shell's config file (.bashrc, .bash_profile, etc.):

  export PATH="$PATH":"$HOME/.pub-cache/bin"
```

dnys1: Brilliant! Love that they've added that. Works for me to delete this warning then 👍

dnys1 (approved):

HuiSF (approved):

Issue #2352, opened 2022-11-11
Model Helpers should allow setting `GraphQLRequest` parameters


dnys1:
Currently, model helpers do not allow setting `GraphQLRequest` parameters like `apiName`, `authorizationMode`, `headers`, etc. This prevents them from being used in multi-auth settings.

**Current**
```dart
static GraphQLRequest<T> create<T extends Model>(T model);
```

**Expected**
```dart
static GraphQLRequest<T> create<T extends Model>(
  T model, {
  String? apiName,
  APIAuthorizationType? authorizationMode,
  Map<String, String>? headers,
});
```

ragingsquirrel3:
released w https://pub.dev/packages/amplify_api/versions/1.0.0-next.1+1

PR #2423, opened 2022-11-28
fix(api): allow setting extra GQL request params from model helpers

ragingsquirrel3:
Fix for [Issue #2352](https://github.com/aws-amplify/amplify-flutter/issues/2352).

Adds support for setting extra properties of `GraphQLRequest`: `apiName`, `authorizationMode` and `headers`. These are things that can be done on manual request but cannot be done w helpers which makes them less useful in apps that require these properties on requests (multi auth mode use cases).


packages/amplify_core/lib/src/types/api/graphql/graphql_helpers.dart
```diff
@@ -17,34 +17,82 @@ import 'package:amplify_core/amplify_core.dart';
...
+    Map<String, String>? headers,
+    APIAuthorizationType? authorizationMode,
+  });
 
   // DeleteById
   GraphQLRequest<T> deleteById<T extends Model>(
-      ModelType<T> modelType, String id);
+    ModelType<T> modelType,
+    String id, {
+    QueryPredicate? where,
```
ragingsquirrel3, line 70: `where` was missing as well here

dnys1 (approved):

Equartey (approved):

PR #2427, opened 2022-11-29
chore: Regenerate AWS SDK

dnys1:
- Suppresses waiter generation in Smithy
- Removes unused user model APIs in Analytics
- Updates SDKs for Analytics, Storage, Auth, and Smoke Test SDK


packages/auth/amplify_auth_cognito_dart/lib/src/sdk/src/cognito_identity/cognito_identity_client.dart
```diff
@@ -1,18 +1,4 @@
-// Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
Jordan-Nelson, line 1: Do we want/need a license header?

Jordan-Nelson (approved):

packages/auth/amplify_auth_cognito_dart/lib/src/sdk/src/cognito_identity/cognito_identity_client.dart
dnys1: I think we decided for generated code to not have it since it would just be overwritten and complicated the generation process.

This was just a remnant from before that decision I think.

HuiSF (approved):

PR #2431, opened 2022-11-29
chore(auth): Rename `UserAttributeKey` to `AuthUserAttributeKey`

dnys1:
For consistency with other platforms

### DIFF
--- a/packages/amplify_core/lib/src/types/auth/attribute/auth_user_attribute_key.dart
+++ b/packages/amplify_core/lib/src/types/auth/attribute/auth_user_attribute_key.dart
@@ -16,15 +16,18 @@
 import 'package:amplify_core/amplify_core.dart';
 import 'package:meta/meta.dart';
 
-/// {@template amplify_core.user_attribute_key}
+@Deprecated('Use AuthUserAttributeKey instead')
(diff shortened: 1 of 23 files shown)


packages/amplify_core/lib/src/types/auth/attribute/auth_user_attribute_key.dart
```diff
@@ -16,15 +16,18 @@
 import 'package:amplify_core/amplify_core.dart';
 import 'package:meta/meta.dart';
 
-/// {@template amplify_core.user_attribute_key}
+@Deprecated('Use AuthUserAttributeKey instead')
```
Jordan-Nelson, line 19: Should we add a TODO to remove this before GA?

Jordan-Nelson (approved):

packages/amplify_core/lib/src/types/auth/attribute/auth_user_attribute_key.dart
dnys1: I can. I think all `Deprecated` tags should be removed for GA, so I've kind of been using them for that.

PR #2434, opened 2022-11-29
chore(auth): Deprecate customAuth flow type

dnys1:
In favor of more precise `customAuthWithSrp` and `customAuthWithoutSrp`. Retain old behavior for `customAuth` since the new values do not have corresponding `amplifyconfiguration.dart` values.


Jordan-Nelson (approved):

ragingsquirrel3 (approved):
Approved assuming checks get fixed, at first glance just looked like some deprecation errors needed to be ignored but I did not look at all of them.

PR #2437, opened 2022-11-29
feat(auth): Add user ID to `CognitoSignUpResult`

dnys1:
Return the user ID for calls to `Amplify.Auth.signUp`.


Jordan-Nelson (approved):
Approved, knowing we will discuss adding a new type for confirmSignUp.

ragingsquirrel3 (approved):

PR #2438, opened 2022-11-29
chore(aft): fix temp model dir structure for using local models

NikaHsn:
to create and use flattened model files for local copy of models same as what is done for models’ git repo.


dnys1 (approved):
Thanks! Small comments about docs+naming

Jordan-Nelson (approved):

PR #2442, opened 2022-11-30
chore(auth): Fix SignUpResult interface

dnys1:
Moves `userId` to the category-level result.

### DIFF
--- a/packages/auth/amplify_auth_cognito_dart/lib/src/model/signup/cognito_sign_up_result.dart
+++ b/packages/auth/amplify_auth_cognito_dart/lib/src/model/signup/cognito_sign_up_result.dart
@@ -23,11 +23,17 @@ class CognitoSignUpResult extends SignUpResult {
   const CognitoSignUpResult({
     required super.nextStep,
     required super.isSignUpComplete,
-    this.userId,
-  });
+    super.userId,
+  }) : _userId = userId;
+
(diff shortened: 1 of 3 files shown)


packages/auth/amplify_auth_cognito_dart/lib/src/model/signup/cognito_sign_up_result.dart
```diff
@@ -23,11 +23,9 @@ class CognitoSignUpResult extends SignUpResult {
   const CognitoSignUpResult({
     required super.nextStep,
     required super.isSignUpComplete,
-    this.userId,
+    super.userId,
   });
 
-  /// The user ID of the signed-up user.
-  ///
-  /// Will not be present for `Amplify.Auth.confirmSignUp` calls.
```
dnys1, line 31: Removed this comment - we'll need to add it to docs

Jordan-Nelson: Its a pretty goofy pattern, but you could override the field to change the comment for this class. Not sure of any other way to do it. I'm fine either way.

Screen Shot 2022-11-30 at 11 11 12 AM

Jordan-Nelson (approved):

ragingsquirrel3 (approved):

PR #2443, opened 2022-11-30
chore(auth): reduce race conditions in integ test GQL subscriptions

ragingsquirrel3:
There is some flakiness in CI with auth integ tests in web where they timeout. Although it's tough to tell exactly where/why they timeout in CI bc the output is opaque, I have been able to reproduce a timeout locally with those tests (likely due to my high latency network). While I'm not certain this PR will fix the CI flakiness, it has a few things to help:
1) adds establish completers to GQL subscriptions related to OTP codes to ensure subscription established before calling the auth method that will cause the code to be sent. (This eliminated the timeouts from my local env.)
2) Uses the integ test script for web in CI so that it will retry like the other platforms.

### DIFF
--- a/.github/workflows/amplify_integration_tests.yaml
+++ b/.github/workflows/amplify_integration_tests.yaml
@@ -130,7 +130,7 @@ jobs:
       - name: Run integration tests
         run: |
           chromedriver --port=4444 &
-          melos exec -c 1 --scope ${{ matrix.scope }} -- flutter drive --driver=test_driver/integration_test.dart --target=integration_test/main_test.dart -d web-server
+          melos exec -c 1 --scope ${{ matrix.scope }} -- \$MELOS_ROOT_PATH/build-support/integ_test.sh -d web-server --retries 1
 
   macos:
     runs-on: macos-latest
--- a/packages/amplify_test/lib/src/integration_test_utils/auth_cognito/integration_test_auth_utils.dart
+++ b/packages/amplify_test/lib/src/integration_test_utils/auth_cognito/integration_test_auth_utils.dart
@@ -156,7 +156,10 @@ Future<void> adminCreateUser(
 
 /// Returns the OTP code for [username]. Must be called before the network call
 /// generating the OTP code.
-Future<String> getOtpCode(String username) async {
+Future<String> getOtpCode(
+  String username, {
+  void Function()? onEstablished,
+}) async {
   const subscriptionDocument = r'''
     subscription OnCreateMFACode($username: String!) {
       onCreateMFACode(username: $username) {
@@ -172,7 +175,10 @@ Future<String> getOtpCode(String username) async {
         'username': username,
       },
     ),
-    onEstablished: () => _logger.debug('Established connection'),
+    onEstablished: () {
+      onEstablished?.call();
+      _logger.debug('Established connection');
+    },
   );
 
   // Collect code delivered via Lambda
@@ -196,7 +202,7 @@ Future<String> getOtpCode(String username) async {
 /// Returns the stream of all OTP codes broadcast by Cognito.
 ///
 /// This is useful with aliases when the username is not known ahead of time.
-Stream<String> getOtpCodes() {
+Stream<String> getOtpCodes({void Function()? onEstablished}) {
   const subscriptionDocument = r'''
     subscription OnCreateMFACode {
       onCreateMFACode {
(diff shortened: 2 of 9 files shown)


dnys1 (approved):

.github/workflows/amplify_integration_tests.yaml
```diff
@@ -130,7 +130,7 @@ jobs:
       - name: Run integration tests
         run: |
           chromedriver --port=4444 &
-          melos exec -c 1 --scope ${{ matrix.scope }} -- flutter drive --driver=test_driver/integration_test.dart --target=integration_test/main_test.dart -d web-server
+          melos exec -c 1 --scope ${{ matrix.scope }} -- \$MELOS_ROOT_PATH/build-support/integ_test.sh -d web-server --retries 1
```
Jordan-Nelson, line 133: Is this considered temporary? I understand retries are necessary for platforms where flakiness is out of our control (emulators not starting for example), but if this issue might be in our control I would rather see it fail.

I'm fine with this for now, but wondering if we should add a task to investigate more.

packages/amplify_test/lib/src/integration_test_utils/auth_cognito/integration_test_auth_utils.dart
```diff
@@ -156,7 +156,10 @@ Future<void> adminCreateUser(
 
 /// Returns the OTP code for [username]. Must be called before the network call
 /// generating the OTP code.
-Future<String> getOtpCode(String username) async {
+Future<String> getOtpCode(
```
Jordan-Nelson, line 159: There seems to be a bit of duplication everywhere this is consumed (creating the completer, passing it in, awaiting the future). Is there a reason to pass in `onEstablished` or could this be done internally?

You could change the response to `Future<OtpResult>` where `OtpResult` contains a member `Future<String> code`.

Instead of:
```dart
final establishedCompleter = Completer<void>();
final code = getOtpCode(
  username,
  onEstablished: establishedCompleter.complete,
);
await establishedCompleter.future;
...
await code
```

Consuming this would look like:
```dart
final otpResult = await getOtpCode(username);
...
await otpResult.code;
```

The implementation would look something like this. I haven't tested this, but I believe this would work.
```dart
class OtpResult {
  OtpResult({required this.code});
  Future<String> code;
}

/// Returns a result, which containts a future for getting the OTP code for [username].
///
/// The [OtpResult] future should complete before the network call
/// generating the OTP code is made.
Future<OtpResult> getOtpCode(String username) async {
  const subscriptionDocument = r'''
    subscription OnCreateMFACode($username: String!) {
      onCreateMFACode(username: $username) {
        username
        code
      }
    }''';

  final Completer<void> established = Completer<void>();

  final Stream<GraphQLResponse<String>> operation = Amplify.API.subscribe(
    GraphQLRequest<String>(
      document: subscriptionDocument,
      variables: {
        'username': username,
      },
    ),
    onEstablished: () {
      _logger.debug('Established connection');
      established.complete();
    },
  );

  final code = operation
      .tap(
        (event) => _logger.debug(
          'Got event: ${event.data}, errors: ${event.errors}',
        ),
      )
      .map((event) {
        if (event.hasErrors) {
          throw Exception(event.errors);
        }
        final json = jsonDecode(event.data!)['onCreateMFACode'] as Map;
        return CreateMFACodeResponse.fromJson(json.cast());
      })
      .map((event) => event.code)
      .first;

  await established.future;

  return OtpResult(code: code);
}
```

Jordan-Nelson (approved):

packages/amplify_test/lib/src/integration_test_utils/auth_cognito/integration_test_auth_utils.dart
dnys1: Good idea

.github/workflows/amplify_integration_tests.yaml
ragingsquirrel3: I think a task to investigate more is appropriate, and we could try turning off retries on all non-android platforms to see how noisy it gets.  However, I do think there is some flakiness outside our control that will come and go with new versions of flutter ,etc...

While letting it fail might make sense and raise visibility of errors that need attention, my experience has always been that despite best intentions, people tend to retry failed tests manually without always investigating, so I would prefer to go the direction of logging some metrics on failed tests and retries so we can look for trends while reducing ops load of retrying tests manually.

packages/amplify_test/lib/src/integration_test_utils/auth_cognito/integration_test_auth_utils.dart
ragingsquirrel3: agree this is nice, added this

PR #2445, opened 2022-11-30
fix(aws-common): add missing contentType getter implementation

HuiSF:
*Issue #, if available:*

*Description of changes:*

Fix:
* When use `AWSFile.fromPath` constructor, storage upload function may not extract `contentType` correctly.
* Add optional parameter `contentType` to `AWSFile.fromPath` constructor as well to allow providing the value when `contentType` cannot be inferred.


packages/aws_common/lib/src/io/aws_file_platform_html.dart
```diff
@@ -81,10 +82,18 @@ class AWSFilePlatform extends AWSFile {
...
 
   @override
-  String? get contentType => _contentType ?? super.contentType;
+  String? get contentType {
+    final file = _inputFile ?? _inputBlob;
+    if (file != null) {
+      return file.type;
+    }
+
+    return _contentType ?? super.contentType;
```
dnys1, line 95: `package:mime` works on web - why not use that here?

packages/aws_common/lib/src/io/aws_file_platform_io.dart
```diff
@@ -104,4 +106,14 @@ class AWSFilePlatform extends AWSFile {
...
   }
+
+  @override
+  String? get contentType {
+    final file = _inputFile;
+    if (file != null) {
+      return mime.lookupMimeType(file.path);
+    }
+
+    return super.contentType;
```
dnys1, line 117: If `super.contentType` is not `null`, then it's been manually set by the user, right? If that's true, we should not be deducing the content type, we should just be returning this.

dnys1: Same Q for web, too

packages/aws_common/lib/src/io/aws_file_platform_io.dart
```diff
@@ -104,4 +106,14 @@ class AWSFilePlatform extends AWSFile {
     // the constructors ensures it won't reach to this point, but just in case
     throw const InvalidFileException();
   }
+
+  @override
+  String? get contentType {
```
dnys1, line 111: Can we make this `late final` so it's cached the first time if called multiple times?

dnys1: Same Q for web

packages/aws_common/lib/src/io/aws_file_platform_io.dart
HuiSF: Yea in that case we should return user defined contentType, will update

packages/aws_common/lib/src/io/aws_file_platform_html.dart
HuiSF: In Web, the path is a url that can resolve a blob from, `lookupMimeType` returns `null` in this case, as its implementation is extracting the extension name from the path string to determine the mime type...

packages/aws_common/lib/src/io/aws_file_platform_html.dart
```diff
@@ -81,10 +82,37 @@ class AWSFilePlatform extends AWSFile {
   final Stream<List<int>>? _stream;
   int? _size;
   Blob? _resolvedBlobFromPath;
-  String? _contentType;
+  String? _resolvedContentType;
+  bool _hasResolvedContentType = false;
 
   @override
-  String? get contentType => _contentType ?? super.contentType;
+  Future<String?> get contentType async {
```
HuiSF, line 89: In Web file type can be determined via `Blob.type` hence not using `mime` package here.

packages/aws_common/lib/src/io/aws_file.dart
```diff
@@ -89,18 +89,23 @@ abstract class AWSFile {
   /// Creates an [AWSFile] from a path, optionally specifying a file [name].
   ///
   /// On VM platforms, the [path] is the absolute path to a file in the file
-  /// system.
+  /// system. When [contentType] is not specified, it leverages
+  /// [mime](https://pub.dev/packages/mime) to determine the value of
```
dnys1, line 93:
```suggestion
  /// system. When [contentType] is not specified, the
  /// [mime](https://pub.dev/packages/mime) package is used to determine the value of
```

packages/aws_common/lib/src/io/aws_file.dart
```diff
@@ -89,18 +89,23 @@ abstract class AWSFile {
...
+  /// [contentType].
   ///
   /// On Web, the [path] can be an object url created by
   /// [`URL.createObjectURL()`](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL)
   /// . When passing an object url, it's recommended to set a value for [name].
+  /// When [contentType] is not specified, it gets the value of [contentType]
+  /// from [`Blob.type`](https://developer.mozilla.org/en-US/docs/Web/API/Blob/type).
```
dnys1, line 100:
```suggestion
  /// When [contentType] is not specified, value is retrieved using
  /// from [`Blob.type`](https://developer.mozilla.org/en-US/docs/Web/API/Blob/type).
```

packages/aws_common/lib/src/io/aws_file_platform_io.dart
```diff
@@ -104,4 +108,34 @@ class AWSFilePlatform extends AWSFile {
     // the constructors ensures it won't reach to this point, but just in case
     throw const InvalidFileException();
   }
+
+  @override
+  Future<String?> get contentType async {
+    if (_hasResolvedContentType) {
+      return _resolvedContentType;
+    }
```
dnys1, line 116: You don't need the extra variable

```suggestion
    final resolvedContentType = _resolvedContentType;
    if (resolvedContentType != null) {
      return resolvedContentType;
    }
```

packages/aws_common/lib/src/io/aws_file_platform_io.dart
```diff
@@ -104,4 +108,34 @@ class AWSFilePlatform extends AWSFile {
     // the constructors ensures it won't reach to this point, but just in case
     throw const InvalidFileException();
   }
+
+  @override
+  Future<String?> get contentType async {
```
dnys1, line 113: Should be wrapped in an `AsyncMemoizer` or made `late final` since this should only run once.

packages/aws_common/lib/src/io/aws_file_platform_io.dart
```diff
@@ -104,4 +108,34 @@ class AWSFilePlatform extends AWSFile {
...
+    if (_hasResolvedContentType) {
+      return _resolvedContentType;
+    }
+
+    final externalContentType = await super.contentType;
+    if (externalContentType != null) {
+      _resolvedContentType = externalContentType;
```
dnys1, line 120: And remove `else`

```suggestion
      return _resolvedContentType = externalContentType;
```

packages/aws_common/lib/src/io/aws_file_platform_io.dart
```diff
@@ -104,4 +108,34 @@ class AWSFilePlatform extends AWSFile {
...
+        final headerBytes =
+            (await file.openRead(0, mime.defaultMagicNumbersMaxLength).toList())
+                .expand((e) => e)
+                .toList();
+        // Mime type will be null if the header bytes doesn't match the
+        // predefined magic numbers in the mime package.
+        // https://github.com/dart-lang/mime/blob/master/lib/src/magic_number.dart
+        // Then it falls back to determine mime type by looking up the file
+        // extension name. If none works, the it falls to null.
+        _resolvedContentType =
```
dnys1, line 133:
```suggestion
        return _resolvedContentType =
```

packages/aws_common/lib/src/io/aws_file_platform_io.dart
```diff
@@ -104,4 +108,34 @@ class AWSFilePlatform extends AWSFile {
...
+            mime.lookupMimeType(file.path, headerBytes: headerBytes);
+      }
+    }
+
+    _hasResolvedContentType = true;
+    return _resolvedContentType;
```
dnys1, line 139: `.fromData` and `.fromStream` cases aren't handled.

dnys1: For `.fromStream` you can use a `StreamSplitter`

packages/aws_common/lib/src/io/aws_file_platform_io.dart
dnys1: Same for web

packages/aws_common/lib/src/io/aws_file_platform_io.dart
dnys1: Same for web

packages/aws_common/lib/src/io/aws_file_platform_io.dart
dnys1: If you use `AsyncMemoizer` or `late final` you don't need `_resolvedContentType` either

packages/aws_common/lib/src/io/aws_file_platform_html.dart
dnys1: Does `Blob.type` just check the path or also the header bytes? Seems like there are cases below where it can't determine it.

packages/aws_common/test/io/aws_file_html_test.dart
```diff
@@ -53,15 +60,17 @@ void main() {
       test('should return ChunkedStreamReader over a file path', () async {
         final awsFile = AWSFile.fromPath(testFilePath);
 
+        expect(await awsFile.contentType, 'text/plain');
```
dnys1, line 63: Probably need another test where the content-type is not embedded in the URL - can add a TODO for now since it would involve spinning up an HTTP server.

packages/aws_common/test/io/aws_file_io_test.dart
```diff
@@ -30,15 +31,19 @@ void main() {
...
     final testBytes = utf8.encode(testFileContent);
 
     final testFile = io.File(testFilePath);
+    final testFileWithWrongExtension = io.File(testFileWithWrongExtensionPath);
 
     group('getChunkedStreamReader() API', () {
       test('should return ChunkedStreamReader over io File', () async {
         final awsFile = AWSFilePlatform.fromFile(testFile);
 
+        expect(await awsFile.contentType, lookupMimeType(testFile.path));
```
dnys1, line 46: We should assert it's resolving to `text/plain`.

If this is testing `expect(null, null)` we wouldn't know and that wouldn't be very helpful.

packages/aws_common/test/io/aws_file_io_test.dart
```diff
@@ -48,15 +53,20 @@ void main() {
...
           await collectBytesFromChunkedReader(awsFile.getChunkedStreamReader()),
           equals(testBytes),
         );
       });
 
       test('should return ChunkedStreamReader over bytes data', () async {
-        final awsFile = AWSFile.fromData(testBytes);
+        final awsFile = AWSFile.fromData(
+          testBytes,
+          contentType: lookupMimeType(testFilePath),
```
dnys1, line 66: This should be tested without passing `contentType` as well.

packages/aws_common/test/io/aws_file_io_test.dart
```diff
@@ -66,14 +76,24 @@ void main() {
       test('should return ChunkedStreamReader over bytes stream', () async {
         final awsFile = AWSFile.fromStream(
           Stream.value(testBytes),
+          contentType: lookupMimeType(testFilePath),
```
dnys1, line 79: This should be tested without passing contentType as well.

PR #2453, opened 2022-12-01
fix(http): Synchronous cancellations on JS

dnys1:
Fixes issues with multiple synchronous calls to `cancel` on JS

### DIFF
--- a/packages/aws_common/test/http/cancellation_test.dart
+++ b/packages/aws_common/test/http/cancellation_test.dart
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import 'dart:async';
+
 import 'package:aws_common/aws_common.dart';
 import 'package:test/test.dart';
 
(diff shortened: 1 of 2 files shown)


packages/aws_common/test/http/cancellation_test.dart
```diff
@@ -77,6 +79,14 @@ void main() {
           expect(response.body, emits(equals([1, 2, 3, 4, 5])));
         },
       );
+
+      test('can call cancel multiple times synchronously', () async {
+        final client_ = client();
+        final request = AWSHttpRequest.get(createUri('/body'));
+        final operation = client_.send(request);
+        unawaited(operation.operation.cancel());
+        await client_.close();
```
ragingsquirrel3, line 88: nit/question: Should this have an `expect` like `await expectLater(client_.close(), completes);` ?

ragingsquirrel3 (approved):
approved w minor question

packages/aws_common/test/http/cancellation_test.dart
dnys1: Yeah, I can add it for clarity. Good thinking.

PR #2454, opened 2022-12-01
fix(storage): use correct fallback contentType for upload

HuiSF:
*Issue #, if available:*

*Description of changes:*


packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_upload_task.dart
```diff
@@ -302,7 +306,7 @@ class S3UploadTask {
       builder
         ..bucket = _bucket
         ..body = body
-        ..contentType = body.contentType
+        ..contentType = body.contentType ?? fallbackContentType
```
dnys1, line 309: But if I pass `AWSFile.fromPath('myfile.txt')` wouldn't this get a content type of `application/octet-stream`?

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_upload_task_test.dart
```diff
@@ -139,6 +140,58 @@ void main() {
...
+            options: testUploadDataOptions,
+            logger: logger,
+            transferDatabase: transferDatabase,
+          );
+
+          unawaited(uploadDataTask.start());
+
+          final result = await uploadDataTask.result;
+
+          expect(result.key, testKey);
```
ragingsquirrel3, line 178: nit/question: Is it necessary to repeat the assertions from previous test when this test is checking something totally different (for which it has a new assertion which is great)?

packages/aws_common/test/io/aws_file_io_test.dart
```diff
@@ -31,32 +31,45 @@ void main() {
...
     const testFileContent = 'I ❤️ Amplify, œ 小新';
+    const testContentType = 'text/plain';
     final testBytes = utf8.encode(testFileContent);
 
     final testFile = io.File(testFilePath);
 
     group('getChunkedStreamReader() API', () {
       test('should return ChunkedStreamReader over io File', () async {
-        final awsFile = AWSFilePlatform.fromFile(testFile);
+        final awsFile = AWSFilePlatform.fromFile(
```
ragingsquirrel3, line 41: question: Nice assertion here, but would it make sense to add similar tests for other common file formats like images? same question applies to _html version.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_upload_task.dart
ragingsquirrel3: I had similar question, from description of "when content type is not determinable" I was expecting to see some automatic determination of type which I don't see, though very possible I'm missing something, or not sure if automatic determination was the intention of flutter plugin caller?

ragingsquirrel3 (commented):
Left a few small comments, but mostly I wasn't sure if we were expecting this to automatically determine content type? I might be missing intention here.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_upload_task.dart
HuiSF: > But if I pass AWSFile.fromPath('myfile.txt') wouldn't this get a content type of application/octet-stream?

Yes that's correct.

>from description of "when content type is not determinable" I was expecting to see some automatic determination of type which I don't see, hough very possible I'm missing something, or not sure if automatic determination was the intention of flutter plugin caller?

My description is confusing sorry. Amplify libraries don't provide functionality to auto detect the mime type of a file to be uploaded. If library consumer doesn't specify `contentType` calling upload APIs, the `contentType` will be `application/octet-stream` as arbitrary binary data. Note that this won't change the file content, when proper extension name is provided, the file consumer can still interpret the file content correctly, including use in a file system, or rendering on Web (svg, jpg, png etc.).

Auto mime type detection is tricky to implement as there is not a "standard" to guide the implementation, I'm happy to sync up offline regarding this topic.

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_upload_task_test.dart
HuiSF: Ahh yea this assertion is not needed here, could be removed. 
This test focuses on the fallback `contentType` value.

dnys1 (approved):
Can you add an integ test where the contentType is not specified but correctly inferred?

dnys1 (approved):

PR #2459, opened 2022-12-02
chore(analytics): remove static instances in analytics

fjnoyp:
*Issue #, if available:*

*Description of changes:*

Remove usage of static instances for client helpers within Analytics.  

Static instances causes integration tests to fail as state is carried over between tests. 


dnys1:
Merged into #2447

PR #2469, opened 2022-12-08
chore(api): refactor ConnectivityPlus usage to prep for Dart package

ragingsquirrel3:
This PR has some prep work for moving the non-Flutter Dart implementation of the API category into its own package. Actually moving the code to another package will be a separate PR but this will make it simpler by handing more complicated bits:
* Create a separate class for the Flutter implementation that has the Flutter requirements: native pigeons and ConnectivityPlus plugin usage. The file naming here is intentionally funky and will be changed when the files are actually copied to a new package.
* Removes ConnectivityPlus usage from the Dart implementation of GQL subscriptions and instead relies on an injected Stream of booleans which represent the presence/absence of network connection. The Flutter part uses ConnectivityPlus and translates that Stream to booleans.

### DIFF
--- a/packages/api/amplify_api/lib/src/api_plugin_impl.dart
+++ b/packages/api/amplify_api/lib/src/api_plugin_impl.dart
@@ -23,7 +23,7 @@ import 'package:amplify_api/src/graphql/providers/oidc_function_api_auth_provide
 import 'package:amplify_api/src/graphql/web_socket/blocs/web_socket_bloc.dart';
 import 'package:amplify_api/src/graphql/web_socket/services/web_socket_service.dart';
 import 'package:amplify_api/src/graphql/web_socket/state/web_socket_state.dart';
-import 'package:amplify_api/src/graphql/web_socket/types/connectivity_status.dart';
+import 'package:amplify_api/src/graphql/web_socket/types/connectivity_platform.dart';
 import 'package:amplify_api/src/util/amplify_api_config.dart';
 import 'package:amplify_api/src/util/amplify_authorization_rest_client.dart';
 import 'package:amplify_core/amplify_core.dart';
@@ -32,7 +32,7 @@ import 'package:meta/meta.dart';
 export 'package:amplify_api/src/graphql/model_helpers/model_mutations.dart';
 export 'package:amplify_api/src/graphql/model_helpers/model_queries.dart';
 export 'package:amplify_api/src/graphql/model_helpers/model_subscriptions.dart';
-export 'package:amplify_api/src/graphql/web_socket/types/connectivity_status.dart';
+export 'package:amplify_api/src/graphql/web_socket/types/connectivity_platform.dart';
 export 'package:amplify_core/src/types/api/api_types.dart';
 
 /// {@template amplify_api_dart.amplify_api_dart}
@@ -42,11 +42,12 @@ class AmplifyAPIDart extends APIPluginInterface with AWSDebuggable {
   /// {@macro amplify_api_dart.amplify_api_dart}
   AmplifyAPIDart({
     List<APIAuthProvider> authProviders = const [],
+    ConnectivityPlatform connectivity = const ConnectivityPlatform(),
     AWSHttpClient? baseHttpClient,
     this.modelProvider,
     this.subscriptionOptions,
-    this.connectivityStreamCreator,
-  }) : _baseHttpClient = baseHttpClient {
+  })  : _baseHttpClient = baseHttpClient,
+        _connectivity = connectivity {
     authProviders.forEach(registerAuthProvider);
     Amplify.Hub.addChannel(HubChannel.Api, _hubEventController.stream);
   }
@@ -56,7 +57,7 @@ class AmplifyAPIDart extends APIPluginInterface with AWSDebuggable {
   late final AmplifyAuthProviderRepository _authProviderRepo;
 
   /// Creates a stream representing network connectivity at the hardware level.
-  ConnectivityInterface? connectivityStreamCreator;
+  final ConnectivityPlatform _connectivity;
 
   /// A map of the keys from the Amplify API config with auth modes to HTTP clients
   /// to use for requests to that endpoint/auth mode. e.g. { "myEndpoint.AWS_IAM": AWSHttpClient}
@@ -210,7 +211,7 @@ class AmplifyAPIDart extends APIPluginInterface with AWSDebuggable {
       wsService: AmplifyWebSocketService(),
       subscriptionOptions:
           subscriptionOptions ?? const GraphQLSubscriptionOptions(),
-      connectivityStreamCreator: connectivityStreamCreator,
+      connectivity: _connectivity,
     );
   }
 
--- a/packages/api/amplify_api/lib/src/api_plugin_impl.dart
+++ b/packages/api/amplify_api/lib/src/api_plugin_impl.dart
@@ -12,37 +12,39 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-library amplify_api;
+/// Amplify API for Dart
+library amplify_api_dart;
 
 import 'dart:async';
-import 'dart:io';
 
-import 'package:amplify_api/amplify_api.dart';
 import 'package:amplify_api/src/graphql/helpers/send_graphql_request.dart';
 import 'package:amplify_api/src/graphql/providers/app_sync_api_key_auth_provider.dart';
 import 'package:amplify_api/src/graphql/providers/oidc_function_api_auth_provider.dart';
 import 'package:amplify_api/src/graphql/web_socket/blocs/web_socket_bloc.dart';
 import 'package:amplify_api/src/graphql/web_socket/services/web_socket_service.dart';
 import 'package:amplify_api/src/graphql/web_socket/state/web_socket_state.dart';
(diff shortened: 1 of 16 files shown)


packages/api/amplify_api/lib/src/api_plugin_impl_flutter.dart
```diff
@@ -0,0 +1,119 @@
...
+  }) : super(
+          authProviders: authProviders,
+          networkStreamGenerator: () => Connectivity()
+              .onConnectivityChanged
+              .map((ConnectivityResult connectivityResult) {
+            switch (connectivityResult) {
+              case ConnectivityResult.none:
+                return false;
+              default:
+                return true;
```
ragingsquirrel3, line 42: Callout, there were 2 enums not being used before: VPN and bluetooth. I figured simpler to do this way, although one could argue they might introduce some value in the future that should map to `false` so I could also be convinced to list them out again.

packages/api/amplify_api/lib/src/api_plugin_impl.dart
```diff
@@ -51,6 +53,10 @@ class AmplifyAPIDart extends AmplifyAPI {
   final AWSHttpClient? _baseHttpClient;
   late final AmplifyAuthProviderRepository _authProviderRepo;
 
+  /// A function which creates a stream of booleans representing network connectivity
+  /// at the hardware level.
+  Stream<bool> Function()? networkStreamGenerator;
```
dnys1, line 58: I would create an interface:

```dart
enum ConnectivityStatus {
  connected,
  disconnected,
}

abstract class ConnectivityInterface {
  Stream<ConnectivityResult> get onConnectivityChanged;
}
```

and use an enum instead of a boolean. A boolean locks us into two values and is not very descriptive.

You could even add a third `unknown` value, if that's helpful, and make that the default e.g.

```dart
enum ConnectivityStatus {
  unknown,
  connected,
  disconnected,
}

class ConnectivityInterface {
  const ConnectivityInterface();  

  Stream<ConnectivityResult> get onConnectivityChanged => const Stream.value(unknown);
}
```

packages/api/amplify_api/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart
```diff
@@ -462,9 +461,9 @@ class WebSocketBloc with AWSDebuggable, AmplifyLoggerMixin {
     }
     _currentState.service.close();
 
+    // TODO(equartey): https://github.com/fluttercommunity/plus_plugins/issues/1382
+    if (zIsWeb || !Platform.isWindows) await _networkSubscription?.cancel();
     await Future.wait<void>([
```
dnys1, line 466:
```suggestion
    await Future.wait<void>([
    // TODO(equartey): https://github.com/fluttercommunity/plus_plugins/issues/1382
    if (zIsWeb || !Platform.isWindows) Future.value(_networkSubscription?.cancel()),
```

packages/api/amplify_api/lib/src/graphql/web_socket/types/connectivity_status.dart
```diff
@@ -0,0 +1,33 @@
...
+
+/// {@template amplify_api_dart.connectivity_interface}
+/// Used to create a stream representing network connectivity at the hardware level.
+/// {@endtemplate}
+abstract class ConnectivityInterface {
+  /// {@macro amplify_api_dart.connectivity_interface}
+  const ConnectivityInterface();
+
+  /// Generates a new stream of [ConnectivityStatus].
+  Stream<ConnectivityStatus> get onConnectivityChanged;
```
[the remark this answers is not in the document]
dnys1: Since it's not really an interface if you do that - maybe `ConnectivityPlatform` is more appropriate? I'm trying to avoid a `null` value for this anywhere.

packages/api/amplify_api/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart
```diff
@@ -86,7 +85,10 @@ class WebSocketBloc with AWSDebuggable, AmplifyLoggerMixin {
   /// Captures events added to the bloc and forwards them to the [_eventTransformer].
   late final Stream<WebSocketEvent> _wsEventStream = _wsEventController.stream;
   late final StreamSubscription<WebSocketState> _stateSubscription;
-  late final StreamSubscription<ConnectivityResult> _networkSubscription;
+  late final StreamSubscription<ConnectivityStatus>? _networkSubscription;
+
+  /// Creates a stream representing network connectivity at the hardware level.
+  ConnectivityInterface? connectivityStreamCreator;
```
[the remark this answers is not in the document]
Equartey: Also, when would this be null?

packages/api/amplify_api/lib/src/graphql/web_socket/types/connectivity_status.dart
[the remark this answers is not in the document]
Equartey: +1

packages/api/amplify_api/test/web_socket/web_socket_bloc_test.dart
```diff
@@ -179,9 +182,9 @@ void main() {
         await dataCompleter.future;
         dataCompleter = Completer<String>();
 
-        fakePlatform.controller.sink.add(ConnectivityResult.none);
-
-        fakePlatform.controller.sink.add(ConnectivityResult.mobile);
+        mockNetworkStreamController
+          ..add(ConnectivityStatus.disconnected)
+          ..add(ConnectivityStatus.connected);
```
Equartey, line 187: I like this more than the previous mock setup, good job! Is it possible to add the actual network result to the stream and let the .map() transform it into a `ConnectivityStatus`? Doesn't look like we have any tests around that branch yet.

packages/api/amplify_api/test/graphql_test.dart
```diff
@@ -309,18 +305,12 @@ void main() {
...
-      final connectivity = Connectivity();
-      mockClient = MockPollClient();
-
       mockWebSocketBloc = MockWebSocketBloc(
         config: testApiKeyConfig,
         authProviderRepo: getTestAuthProviderRepo(),
         wsService: mockWebSocketService!,
         subscriptionOptions: subscriptionOptions,
         pollClientOverride: mockClient.client,
-        connectivityOverride: connectivity,
```
Equartey, line 323: I guess this answers my question as to when `connectivityStreamCreator` i null. Although I would prefer to not have it nullable as we wouldn't use the bloc realistically without it.

packages/api/amplify_api/test/web_socket/web_socket_bloc_test.dart
```diff
@@ -17,25 +17,32 @@ import 'dart:convert';
...
+late StreamController<ConnectivityStatus> mockNetworkStreamController;
+
+class MockConnectivity extends ConnectivityInterface {
+  const MockConnectivity();
+
+  @override
+  Stream<ConnectivityStatus> get onConnectivityChanged =>
+      mockNetworkStreamController.stream;
+}
+
```
Equartey, line 39: Let's move this to the utils file and we can reuse in the other tests when `connectivityStreamCreator` becomes non-null.

packages/api/amplify_api/lib/src/graphql/web_socket/types/connectivity_status.dart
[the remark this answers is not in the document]
ragingsquirrel3: `ConnectivityPlatform` seems like a fine name. I just pushed a commit with this change and other comments addressed.

packages/api/amplify_api/test/graphql_test.dart
ragingsquirrel3: I made this non-null but I think there's an important note here that the bloc can be used without connectivity info as we would expect a non-Flutter user to not necessarily provide that info. Although the "no info mode" is now a default empty stream rather than a null provider as it was originally, the idea is still the same. For that reason I've intentionally passed the default provider in some of the tests to ensure it works without network connectivity info.

packages/api/amplify_api/test/web_socket/web_socket_bloc_test.dart
ragingsquirrel3: moved it to util, but not used anywhere else, yet

Equartey (approved):
Crushed it! 🚢-it!

packages/api/amplify_api/lib/src/api_plugin_impl_flutter.dart
```diff
@@ -0,0 +1,110 @@
...
+class AmplifyAPI extends AmplifyAPIDart with AWSDebuggable {
+  /// {@macro amplify_api.amplify_api}
+  AmplifyAPI({
+    List<APIAuthProvider> authProviders = const [],
+    super.baseHttpClient,
+    super.modelProvider,
+    super.subscriptionOptions,
+  }) : super(
+          authProviders: authProviders,
+          connectivity: const ConnectivityPlusStreamCreator(),
```
dnys1, line 35:
```suggestion
          connectivity: const ConnectivityPlusPlatform(),
```

dnys1 (approved):
Small name change requested. Otherwise, looks good 👍

PR #2472, opened 2022-12-09
refactor(core)!: Align error types

dnys1:
Aligns exception types with other platforms. Adds SDK generation plugin to automate this effort for Auth service exceptions.

- Removes direct instantiation of `AmplifyException`
- Migrates some unrecoverable `Exception` types to `Error`
- Removes legacy `fromMap` methods
- Generates `ServiceException` types for Auth SDK exceptions.
- Removes `SrpException` in favor of `SrpError`.


packages/amplify_core/lib/src/types/query/query_field_operators.dart
```diff
@@ -220,7 +220,7 @@ class ContainsQueryOperator extends QueryFieldOperatorSingleValue<String> {
     } else if (other is List<String>) {
       return other.contains(value);
     } else {
-      throw const AmplifyException(
+      throw const DataStoreException(
```
ragingsquirrel3, line 223: Query fields can be used outside of Datastore in model helpers for API category. Is there another exception type that could be used here?

dnys1: Good catch! Created a new `ModelQueryException` type which extends `AmplifyException`

HuiSF (approved):

PR #2483, opened 2022-12-13
fix(api): SubscriptionDataPayload error decoding type fix

Equartey:
*Description of changes:*
This provides a fix for an error I saw when errors were present in the SubscriptionDataPayload fromJson method. Errors come from AppSync as a List not a Map.

### DIFF
--- a/packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
+++ b/packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
@@ -116,6 +116,27 @@ void main({bool useExistingTestUser = false}) {
 
             expect(blogFromEvent?.name, equals(name));
           });
+
+          testWidgets('should parse errors within a web socket data message',
+              (WidgetTester tester) async {
+            final name =
+                'Integration Test Blog - subscription create ${uuid()}';
+            const error =
+                'Cannot return null for non-nullable type: \'AWSDateTime\' within parent \'Blog\' (/onCreateBlog/createdAt)';
+            final subscriptionRequest = ModelSubscriptions.onCreate(
+              Blog.classType,
+              authorizationMode: APIAuthorizationType.apiKey,
+            );
+
+            final eventResponse = await establishSubscriptionAndMutate(
+              subscriptionRequest,
+              () => addBlogBadMutation(name),
+              canFail: true,
+            );
+            final dataErrors = eventResponse.errors;
+
+            expect(dataErrors.first.message, equals(error));
+          });
         },
       );
     },
--- a/packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
+++ b/packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
@@ -122,7 +122,8 @@ void main({bool useExistingTestUser = false}) {
             final name =
                 'Integration Test Blog - subscription create ${uuid()}';
             const error =
-                'Cannot return null for non-nullable type: \'AWSDateTime\' within parent \'Blog\' (/onCreateBlog/createdAt)';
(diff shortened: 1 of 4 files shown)


packages/api/amplify_api/test/web_socket/web_socket_types_test.dart
```diff
@@ -85,5 +85,26 @@ void main() {
...
+        'payload': {'data': null, 'errors': errors}
+      };
+      final message = WebSocketMessage.fromJson(entry);
+      expect(message.messageType, MessageType.data);
+      expect(
+        message.payload!.toJson()['errors'],
+        errors,
+      );
+    });
   });
```
ragingsquirrel3, line 109: request: Could you add either another test or another assertion here that the graphql response decoder will accept the JSON produced by this websocket message? Basically, something like:

```
final response = GraphQLResponseDecoder.instance.decode<T>(
    request: request, // you will have to add a request for this which can be basic subscription request
    response: message.payload.toJson(),
 );
expect(response.errors.first.message, errorMessage); // should make the errorMessage a const
```

I think part of the issue here isn't just that it's the wrong type, but that we're not matching downstream behavior of how we actually convert the ws message to the stream event we give to the user and the decoder is the part where it will eventually get integrated. Ideally there would also be a test where we mock a message like this on the mocked stream and assert it all works as expected but not required.

ragingsquirrel3 (approved):
Good catch! approved w request for additional test

packages/api/amplify_api/test/web_socket/web_socket_types_test.dart
dnys1: I agree - it surprised me we didn't have a test which caught this before. No worries at all, though, this is the perfect opportunity to add it 👍

Equartey: Good call, that assertion has been added!

packages/api/amplify_api/example/integration_test/util.dart
```diff
@@ -146,6 +146,24 @@ Future<Blog> addBlog(String name) async {
   return blog;
 }
 
+// Run a mutation with an incorrect document
+Future<GraphQLResponse<String>> addBlogBadMutation(String name) async {
```
dnys1, line 150:
```suggestion
/// Run a mutation on [Blog] with a partial selection set.
///
/// This is used to trigger an error on subscriptions listening for the
/// full selection set.
Future<GraphQLResponse<String>> runPartialMutation(String name) async {
```

packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
```diff
@@ -116,6 +116,27 @@ void main({bool useExistingTestUser = false}) {
 
             expect(blogFromEvent?.name, equals(name));
           });
+
+          testWidgets('should parse errors within a web socket data message',
+              (WidgetTester tester) async {
+            final name =
+                'Integration Test Blog - subscription create ${uuid()}';
+            const error =
+                'Cannot return null for non-nullable type: \'AWSDateTime\' within parent \'Blog\' (/onCreateBlog/createdAt)';
```
dnys1, line 125: Rules for strings are:
- Use single quotes by default
- Use double-quotes when the string contains single-quotes
- Prefer raw strings `r` instead of escapes (`\`) 
- Wrap strings if they break the 80-char limit

Any two touching strings in Dart are concatenated, e.g. `'Hello, ' 'world'`, so you can do:

```suggestion
                "Cannot return null for non-nullable type: 'AWSDateTime' within "
                "parent 'Blog' (/onCreateBlog/createdAt)";
```

dnys1 (approved):

PR #2489, opened 2022-12-14
feat(analytics): Legacy data migration of Pinpoint Endpoint ID

fjnoyp:
*Issue #, if available:*

*Description of changes:*

Add code to extract the previous Analytics Pinpoint Endpoint ID stored in Android or iOS. 

Users upgrading to our next release won't have their Pinpoint Endpoint regenerated and can keep pinpoint session on a given user device between the upgrade. 

*Testing*

Before, when upgrading Analytics, a new Pinpoint ID would be assigned.  This causes the AWS Pinpoint console to register a new endpoint and therefore increment the 'daily active endpoint' in the analytics dashboard.  This is problematic because Pinpoint is now detecting a new user. 

With this change, when upgrading and running the same Analytics app on the same device, the 'daily active endpoint' number should NOT increment. 

You can test this by modifying the Analytics example app pubspec.yaml to use 0.6.1 and to remove the local path dependencies.  You will also need to delete the pubspec_overrides.yaml generated by aft bootstrap.  Then modify example app's main.dart to simply add the analytics/auth plugins (in dart first there are new initialization options).  Run the app and verify the AWS console registers your session and 'daily active endpoint' increases by 1.  Afterwards, undo the previous changes and run the Analytics dart first version.  The code will now retrieve the old pinpoint id, and thus 'daily active endpoint' will not increment. 


fjnoyp:
I based this on the other PR for integration tests because that PR refactors several parts of Analytics that this PR relies on (primarily with configure of the plugin).

packages/analytics/amplify_analytics_pinpoint/lib/src/legacy_native_data_provider/flutter_legacy_native_data_provider_io.dart
```diff
@@ -0,0 +1,43 @@
...
+import 'dart:io' show Platform;
+
+import 'package:amplify_analytics_pinpoint/src/legacy_native_data_provider/data_provider.android.dart';
+import 'package:amplify_analytics_pinpoint/src/legacy_native_data_provider/data_provider.ios.dart';
+import 'package:amplify_analytics_pinpoint_dart/amplify_analytics_pinpoint_dart.dart';
+
+/// {@macro amplify_analytics_pinpoint.flutter_legacy_native_data_provider}
+class FlutterLegacyNativeDataProvider implements LegacyNativeDataProvider {
+  /// {@macro amplify_analytics_pinpoint.flutter_legacy_native_data_provider}
+  factory FlutterLegacyNativeDataProvider() {
```
dnys1, line 25: This can just be a single class for both platforms

packages/analytics/amplify_analytics_pinpoint/pigeons/pigeon_config_android.dart
```diff
@@ -0,0 +1,40 @@
...
+    javaOptions: JavaOptions(
+      className: 'Messages',
+      package:
+          'com.amazonaws.amplify.amplify_analytics_pinpoint.amplify_analytics_pinpoint',
+    ),
+    copyrightHeader: 'pigeons/copyright.txt',
+  ),
+)
+@HostApi()
+abstract class PigeonDataProvider {
```
dnys1, line 34: Can we name this more descriptively? e.g. something like `LegacyDataProvider`.

packages/analytics/amplify_analytics_pinpoint/pigeons/pigeon_config_ios.dart
```diff
@@ -0,0 +1,40 @@
...
+    ),
+    objcHeaderOut:
+        '../amplify_analytics_pinpoint/ios/Classes/pigeons/Messages.h',
+    objcSourceOut:
+        '../amplify_analytics_pinpoint/ios/Classes/pigeons/Messages.m',
+    copyrightHeader: 'pigeons/copyright.txt',
+  ),
+)
+@HostApi()
+abstract class PigeonDataProvider {
```
dnys1, line 34: There only needs to be one of these defined for use with both platforms.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/flutter_provider_interfaces/legacy_native_data_provider.dart
```diff
@@ -0,0 +1,10 @@
+import 'dart:async';
```
dnys1, line 1: License header

packages/analytics/amplify_analytics_pinpoint/ios/Classes/SwiftAmplifyAnalyticsPinpointPlugin.swift
```diff
@@ -0,0 +1,62 @@
...
+
+        var result: AnyObject?
+        let status = SecItemCopyMatching(query, &result)
+
+        guard status != errSecItemNotFound else {
+            return nil
+        }
+
+        guard status == errSecSuccess else {
+            return nil
```
dnys1, line 57: Is this what iOS does? I thought there were errors worth surfacing here, e.g. keychain access issues due to permissions.

packages/analytics/amplify_analytics_pinpoint/ios/Classes/SwiftAmplifyAnalyticsPinpointPlugin.swift
```diff
@@ -0,0 +1,62 @@
+import Flutter
+import UIKit
+
+public class SwiftAmplifyAnalyticsPinpointPlugin: NSObject, FlutterPlugin, PigeonDataProvider {
+    public func getEndpointId(completion: @escaping (String?, FlutterError?) -> Void) {
+        let keychain = KeychainReader()
+        guard let uniqueId = keychain.getEndpointId() else {
+            completion (nil, nil)
```
dnys1, line 8:
```suggestion
            completion(nil, nil)
```

packages/analytics/amplify_analytics_pinpoint/ios/Classes/SwiftAmplifyAnalyticsPinpointPlugin.swift
```diff
@@ -0,0 +1,62 @@
+import Flutter
```
dnys1, line 1: License header

packages/analytics/amplify_analytics_pinpoint/android/src/main/kotlin/com/amazonaws/amplify/amplify_analytics_pinpoint/amplify_analytics_pinpoint/AmplifyAnalyticsPinpointPlugin.kt
```diff
@@ -0,0 +1,40 @@
+
+package com.amazonaws.amplify.amplify_analytics_pinpoint.amplify_analytics_pinpoint
```
dnys1, line 2: License header

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -150,14 +157,26 @@ class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
     );
 
     // Retrieve Unique ID
-    final savedFixedEndpointId =
-        await _keyValueStore.read(key: endpointIdStorageKey);
-    final fixedEndpointId = savedFixedEndpointId ?? const Uuid().v1();
-    if (savedFixedEndpointId == null) {
+    final endpointInformationVersion =
+        await _keyValueStore.read(key: _endpointInformationVersionKey);
+    String? fixedEndpointId;
```
dnys1, line 162: Should be non-null

packages/analytics/amplify_analytics_pinpoint/lib/src/legacy_native_data_provider/data_provider.ios.dart
```diff
@@ -0,0 +1,36 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'dart:async';
+
+import 'package:amplify_analytics_pinpoint/src/legacy_native_data_provider/pigeon_data_provider.ios.g.dart';
+import 'package:amplify_analytics_pinpoint_dart/amplify_analytics_pinpoint_dart.dart';
+
+/// {@macro amplify_analytics_pinpoint.flutter_legacy_native_data_provider}
+class DataProviderIos implements LegacyNativeDataProvider {
```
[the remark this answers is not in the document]
fjnoyp: They use the same class name but the implementations are different.  See: 


import 'package:amplify_analytics_pinpoint/src/legacy_native_data_provider/pigeon_data_provider.android.g.dart';

import 'package:amplify_analytics_pinpoint/src/legacy_native_data_provider/pigeon_data_provider.ios.g.dart';

packages/analytics/amplify_analytics_pinpoint/lib/src/legacy_native_data_provider/flutter_legacy_native_data_provider_io.dart
fjnoyp: Not sure I understand here, there already is a single class for Android and iOS (data_provider_android.dart and data_provider_ios.dart) 

This class is what is exported via flutter_legacy_native_data_provider.

packages/analytics/amplify_analytics_pinpoint/ios/Classes/SwiftAmplifyAnalyticsPinpointPlugin.swift
fjnoyp: I believe iOS just logs the issue, we can do the same.  I don't think we should throw an exception and disrupt code flow over an initialization issue here.

fjnoyp: guard status == errSecSuccess else {
            print("Read for \(key) failed")
            return nil
        }

packages/analytics/amplify_analytics_pinpoint/lib/src/analytics_plugin_impl.dart
```diff
@@ -25,16 +27,25 @@ import 'package:meta/meta.dart';
...
   /// {@macro amplify_analytics_pinpoint.analytics_plugin_impl}
   AmplifyAnalyticsPinpoint({
-    @visibleForTesting super.endpointInfoStore,
+    @visibleForTesting SecureStorageInterface? endpointInfoStore,
     @visibleForTesting CachedEventsPathProvider? pathProvider,
     @visibleForTesting AppLifecycleProvider? appLifecycleProvider,
     @visibleForTesting DeviceContextInfoProvider? deviceContextInfoProvider,
+    @visibleForTesting LegacyNativeDataProvider? legacyNativeDataProvider,
   }) : super(
+          keyValueStore: keyValueStore ??
```
Jordan-Nelson, line 36: Should this be `endpointInfoStore`? It seems like this was renamed in some places and not in others.

packages/analytics/amplify_analytics_pinpoint/lib/src/legacy_native_data_provider/data_provider.ios.dart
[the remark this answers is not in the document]
Jordan-Nelson: `pigeon_data_provider.android.g.dart` and `pigeon_data_provider.ios.g.dart` seem to have the exact same code though. You can combine the following:
- `pigeon_config_ios.dart` and `pigeon_config_android.dart`
- `data_provider.ios.dart` and `data_provider.android.dart`
- All `flutter_legacy_native_data_provider*` files

Typically when pigeon is being used, we want to expose the same APIs from each native platform (`initialize` and `getEndpointId` in this case). The dart code can be platform agnostic, and the swift/java code can be specific to the platform.

We have some rare cases where we expose a completely different set of APIs on each platform. You may have been following one of these as an example. For example, in secure_storage we have the need to interact with encrypted storage on android, whereas on iOS we only need to set/read a bool value from unecrypted storage. Since these are totally different use cases that require a different set of APIs, there are .ios and .android files for pigeon. This is not one of those rare examples where the dart code also needs to be different on each platform.

packages/analytics/amplify_analytics_pinpoint/lib/src/legacy_native_data_provider/flutter_legacy_native_data_provider_io.dart
Jordan-Nelson: See comment above on `data_provider.ios.dart`.

packages/analytics/amplify_analytics_pinpoint/pigeons/pigeon_config_android.dart
Jordan-Nelson: This was marked as resolved but I think still applies

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -80,9 +82,13 @@ class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
   /// Storage key for the static Pinpoint endpoint id
   @visibleForTesting
   static const String endpointIdStorageKey = 'UniqueId';
+  static const String _endpointInformationVersionKey =
+      'EndpointInformationVersionKey';
   static const String _endpointGlobalAttrsKey = 'EndpointGlobalAttributesKey';
   static const String _endpointGlobalMetricsKey = 'EndpointGlobalMetricsKey';
 
+  static const String _endpointInformationCurrentVersionValue = '0';
```
Jordan-Nelson, line 90: What is the purpose of this? Can you add a comment? Would an enum make more sense here?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -149,15 +156,29 @@ class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
...
+    if (endpointInformationVersion == null) {
+      _endpointInfoStore.write(
+        key: _endpointInformationVersionKey,
+        value: _endpointInformationCurrentVersionValue,
+      );
+
+      await _legacyNativeDataProvider?.initialize(appId);
+      retrievedEndpointId = await _legacyNativeDataProvider?.getEndpointId();
+
+      retrievedEndpointId ??= const Uuid().v1();
```
ragingsquirrel3, line 172: question: Why use this version of uuid instead of `uuid()` from aws_common? If there is specific requirement for that could we have a comment here saying why?

packages/analytics/amplify_analytics_pinpoint/lib/src/legacy_native_data_provider/flutter_legacy_native_data_provider_io.dart
```diff
@@ -0,0 +1,43 @@
...
+    final provider = Platform.isIOS ? DataProviderIos() : DataProviderAndroid();
```
ragingsquirrel3, line 26: question: Maybe I'm missing something here, but is this using `DataProviderAndroid` for all non-ios io platforms (desktop windows, macos, linux)? When I ran example app on macos I got this error:


```
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: PlatformException(channel-error, Unable to establish connection on channel., null, null)
#0      PigeonDataProvider.initialize (package:amplify_analytics_pinpoint/src/legacy_native_data_provider/pigeon_data_provider.android.g.dart:40:7)
<asynchronous suspension>
#1      AmplifyAnalyticsPinpointDart.configure (package:amplify_analytics_pinpoint_dart/src/analytics_plugin_impl.dart:169:7)
<asynchronous suspension>
#2      Future.wait.<anonymous closure> (dart:async/future.dart:522:21)
<asynchronous suspension>
#3      AmplifyClassImpl.configurePlatform (package:amplify_core/src/amplify_class_impl.dart:73:5)
<asynchronous suspension>
#4      AmplifyClass.configure (package:amplify_core/src/amplify_class.dart:107:7)
<asynchronous suspension>
#5      _MyAppState._configureAmplify (package:amplify_analytics_pinpoint_example/main.dart:70:7)
<asynchronous suspension>
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -149,15 +156,29 @@ class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
...
+    final endpointInformationVersion =
+        await _endpointInfoStore.read(key: _endpointInformationVersionKey);
+    String? retrievedEndpointId;
+
+    if (endpointInformationVersion == null) {
+      _endpointInfoStore.write(
+        key: _endpointInformationVersionKey,
+        value: _endpointInformationCurrentVersionValue,
+      );
+
```
ragingsquirrel3, line 168: question: Should this block only execute on ios/android like:

```
if (!zIsWeb && (Platform.isAndroid || Platform.isIOS)) {
}
```

I mentioned in another comment I got an error from here on macos but I wonder if it would simplify things just to stop this logic far upstream (here) to simplify non ios/android platforms as much as possible.

ragingsquirrel3 (commented):
I agree with other comments that would be good to see platform/pigeon stuff consolidated as well as comment about secure storage detail before reviewing again. Biggest question was how non android/ios platforms should avoid the legacy stuff as it did not seem correct when I tested on MacOS (I did not try other platforms).

packages/analytics/amplify_analytics_pinpoint/lib/src/legacy_native_data_provider/flutter_legacy_native_data_provider_io.dart
fjnoyp: Thanks good catch, I'll need to fix this then

packages/analytics/amplify_analytics_pinpoint/lib/src/legacy_native_data_provider/data_provider.ios.dart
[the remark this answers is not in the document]
fjnoyp: Thanks for clarifying that Jordan.  Yes I followed the secure storage example and should have combined them. 

However now that secure storage is working I have removed the ios pigeon part.

packages/analytics/amplify_analytics_pinpoint/pigeons/pigeon_config_android.dart
fjnoyp: My bad, I thought I had renamed it but was clearly mistaken.  Thanks for catching!

packages/analytics/amplify_analytics_pinpoint/ios/Classes/SwiftAmplifyAnalyticsPinpointPlugin.swift
fjnoyp: not relevant anymore, ios code removed

packages/analytics/amplify_analytics_pinpoint/android/src/main/kotlin/com/amazonaws/amplify/amplify_analytics_pinpoint/amplify_analytics_pinpoint/AmplifyAnalyticsPinpointPlugin.kt
fjnoyp: I think I made some changes but didn't commit them hence the issues that were resolved but not actually fixed here ... 

Likely due to some rebase work I did I might have missed a commit here

packages/analytics/amplify_analytics_pinpoint/ios/Classes/SwiftAmplifyAnalyticsPinpointPlugin.swift
```diff
@@ -0,0 +1,65 @@
...
+          kSecAttrAccount: key,
+          kSecReturnData: true
+        ] as CFDictionary
+
+        var result: AnyObject?
+        let status = SecItemCopyMatching(query, &result)
+
+        guard status != errSecItemNotFound else {
+            return nil
+        }
```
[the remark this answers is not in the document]
fjnoyp: Thanks for helping here, looks like this was resolved.

packages/analytics/amplify_analytics_pinpoint/lib/src/analytics_plugin_impl.dart
fjnoyp: Thanks resolved.  I did a rebase and appear to have missed a commit somehow ...

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
fjnoyp:
```
  /// Version of metadata stored into the `endpointInfoStore`
  /// If stored data changes, increment this value
  static const String _endpointInformationCurrentVersionValue = '0';

```

Added the above comments.  The initial idea is to track if we have read legacy data from native.  

If so, we store this value into secure storage to avoid further reads from legacy data.  

In case of future changes, if we need to change legacy read behavior to read other values we would increment this value and store that instead.  Having this value stored before allows us to quickly know what the storage state is (either no value, "0", "1", etc)

packages/analytics/amplify_analytics_pinpoint/.metadata
```diff
@@ -0,0 +1,45 @@
+# This file tracks properties of this Flutter project.
+# Used by Flutter tool to assess capabilities and perform upgrades etc.
+#
+# This file should be version controlled.
+
+version:
+  revision: 52b3dc25f6471c27b2144594abb11c741cb88f57
+  channel: stable
+
+project_type: app
```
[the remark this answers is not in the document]
fjnoyp: Thanks good catch there.  It appears this was created on the flutter create . command not sure how/why it was modified since then though.  It should be project_type plugin so I changed it.  Looks to be used by flutter when performing upgrades so making this change should be fine. 

Used this as a resource in case you're curious to learn more about what metadata is: https://dev.to/swimmingkiim/what-is-metadata-file-in-flutter-project-38m3

packages/analytics/amplify_analytics_pinpoint/ios/amplify_analytics_pinpoint.podspec
```diff
@@ -0,0 +1,23 @@
...
+  s.license          = { :file => '../LICENSE' }
+  s.author           = { 'Your Company' => 'email@example.com' }
+  s.source           = { :path => '.' }
+  s.source_files = 'Classes/**/*'
+  s.dependency 'Flutter'
+  s.platform = :ios, '13.0'
+
+  # Flutter.framework does not contain a i386 slice.
+  s.pod_target_xcconfig = { 'DEFINES_MODULE' => 'YES', 'EXCLUDED_ARCHS[sdk=iphonesimulator*]' => 'i386' }
+  s.swift_version = '5.0'
```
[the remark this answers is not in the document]
fjnoyp: Thanks good catch again, since iOS is removed this is removed too.

packages/analytics/amplify_analytics_pinpoint/ios/amplify_analytics_pinpoint.podspec
```diff
@@ -0,0 +1,23 @@
...
+#
+Pod::Spec.new do |s|
+  s.name             = 'amplify_analytics_pinpoint'
+  s.version          = '0.0.1'
+  s.summary          = 'A new Flutter plugin project.'
+  s.description      = <<-DESC
+A new Flutter plugin project.
+                       DESC
+  s.homepage         = 'http://example.com'
+  s.license          = { :file => '../LICENSE' }
```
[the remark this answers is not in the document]
fjnoyp: Thanks good catch again, since iOS is removed this is removed too.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
fjnoyp: Thanks there is no reason, wasn't aware of this method in core package.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
fjnoyp: Thanks yes updated the logic

fjnoyp:
Do not merge until PR fixes to secure storage are in: 

https://github.com/aws-amplify/amplify-flutter/pull/2516/commits/bfeae2a0d4ef9414f867215983fcf2dc651119b3

That PR allows us to properly extract legacy ios data using the secure storage plugin instead of custom ios code.

packages/analytics/amplify_analytics_pinpoint/pubspec.yaml
```diff
@@ -35,3 +35,12 @@ dev_dependencies:
   amplify_lints: ^2.0.0
   flutter_test:
     sdk: flutter
+  pigeon: ^4.2.5
+
+# The following section is specific to Flutter.
```
Jordan-Nelson, line 40: You can probably remove the comment. We have this comment in a few other files. I think it is part of the auto generated comments from `flutter create` and was never removed from those files, but probably should be.

packages/analytics/amplify_analytics_pinpoint/android/src/main/kotlin/com/amazonaws/amplify/amplify_analytics_pinpoint/amplify_analytics_pinpoint/AmplifyAnalyticsPinpointPlugin.kt
```diff
@@ -0,0 +1,54 @@
...
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
```
Jordan-Nelson, line 14: Do you mind updating the license headers in this PR to the new header added in https://github.com/aws-amplify/amplify-flutter/pull/2501?

```suggestion
// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
// SPDX-License-Identifier: Apache-2.0
```

packages/analytics/amplify_analytics_pinpoint/analysis_options.yaml
```diff
@@ -1 +1,5 @@
 include: package:amplify_lints/library.yaml
+
+analyzer:
+  exclude:
+    - "**/**/**/*.g.dart"
```
Jordan-Nelson, line 5: `**/**/**/` should be reducible to `**/`
```suggestion
    - "**/*.g.dart"
```

packages/analytics/amplify_analytics_pinpoint/lib/src/analytics_plugin_impl.dart
```diff
@@ -14,16 +16,25 @@ import 'package:meta/meta.dart';
 class AmplifyAnalyticsPinpoint extends AmplifyAnalyticsPinpointDart {
   /// {@macro amplify_analytics_pinpoint.analytics_plugin_impl}
   AmplifyAnalyticsPinpoint({
-    @visibleForTesting super.endpointInfoStore,
+    @visibleForTesting SecureStorageInterface? endpointInfoStore,
     @visibleForTesting CachedEventsPathProvider? pathProvider,
     @visibleForTesting AppLifecycleProvider? appLifecycleProvider,
     @visibleForTesting DeviceContextInfoProvider? deviceContextInfoProvider,
+    @visibleForTesting LegacyNativeDataProvider? legacyNativeDataProvider,
```
Jordan-Nelson, line 23: I don't see this used in tests. Is there any reason to expose it?

You could add some unit tests that mock this, and then assert that certain behavior does/doesn't happen when the mock does/doesn't provide data, but that should be doable from within amplify_analytics_pinpoint_dart. I don't see a reason to expose it here.

On the subject of tests, It would be a good idea to add tests if possible. We don't have a good way to integ test this, but it seems like some unit tests here would be fairly straightforward and valuable.

Jordan-Nelson: This is out of scope of this PR, but is there a reason to expose `pathProvider` or `deviceContextInfoProvider`? I don't see those used in tests either.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
Jordan-Nelson: Okay. I would suggest using an enum for this as we did in Auth. It provides a nice way to document this.

https://github.com/aws-amplify/amplify-flutter/blob/834a81d0192c0e2a6700dd9452e750308da60e65/packages/auth/amplify_auth_cognito_dart/lib/src/credentials/credential_store_keys.dart#L10-L20

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/flutter_provider_interfaces/legacy_native_data_provider.dart
Jordan-Nelson: This is still missing a header.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -138,15 +146,33 @@ class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
     );
 
     // Retrieve Unique ID
-    final savedFixedEndpointId =
-        await _endpointInfoStore.read(key: endpointIdStorageKey);
-    final fixedEndpointId = savedFixedEndpointId ?? const Uuid().v1();
-    if (savedFixedEndpointId == null) {
+    final endpointInformationVersion =
+        await _endpointInfoStore.read(key: _endpointInformationVersionKey);
+    String? retrievedEndpointId;
```
Jordan-Nelson, line 151: I think you can likely make this final an non-nullable. See other comments below.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -138,15 +146,33 @@ class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
...
+    print('retrieved id: ${retrievedEndpointId ?? 'null'}');
+
+    if (endpointInformationVersion == null) {
+      _endpointInfoStore.write(
+        key: _endpointInformationVersionKey,
+        value: _endpointInformationCurrentVersionValue,
+      );
+
+      await _legacyNativeDataProvider?.initialize(appId);
+      retrievedEndpointId = await _legacyNativeDataProvider?.getEndpointId();
```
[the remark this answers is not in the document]
Jordan-Nelson: I am having a bit of a hard time following this. I believe you want to achieve something like:
- Check version
- If version is null, attempt migration (one time event)
  - read ID from _legacy_ storage
  - if non-null, write to _new_ storage
  - update version to `'0'` to prevent migration on future app inits
- Read version from storage
- If null, generate a new one
  - generate new ID
  - write to storage

This is an example of a flow that I think is easier to follow.
```dart
    // Attempt migration if version is null.
    if (endpointInformationVersion == null) {
      await _legacyNativeDataProvider?.initialize(appId);
      final legacyEndpointId = await _legacyNativeDataProvider?.getEndpointId();
      // Migrate legacy data if it is non-null
      if (legacyEndpointId != null) {
        await _endpointInfoStore.write(
          key: endpointIdStorageKey,
          value: legacyEndpointId,
        );
      }
      // Update the version to prevent future legacy data migrations.
      await _endpointInfoStore.write(
        key: _endpointInformationVersionKey,
        value: _endpointInformationCurrentVersionValue,
      );
    }

    // Read the existing ID.
    final currentEndpointId = await _endpointInfoStore.read(
      key: endpointIdStorageKey,
    );

    final String fixedEndpointId;

    // Generate a new ID if one does not exist.
    if (currentEndpointId == null) {
      fixedEndpointId = uuid();
      await _endpointInfoStore.write(
        key: endpointIdStorageKey,
        value: fixedEndpointId,
      );
    } else {
      fixedEndpointId = currentEndpointId;
    }
```

This will result in reading the ID twice in the scenario where there is data to migrate (first from legacy storage, and then from the new storage), but since it is a one time event I think that is fine.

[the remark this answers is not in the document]
fjnoyp: Alright it seems like a bit longer of a solution and possibly more complex because there are now 2 conditionals but I see how some of the code before was implicit.  I think it's good that the read call no longer requires a force cast, so if something goes wrong and currentEndpointId is null we no everything will still work and no crash will occur.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -323,4 +321,53 @@ class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
     await _eventClient.close();
     await _eventStorageAdapter.close();
   }
+
+  @visibleForTesting
+  /// Retrieve the stored pinpoint endpoint id
```
Jordan-Nelson, line 326: Annotations should always be directly above the function, member, or class that they annotate. That means that they should be below doc comments.

This is likely failing a formatting check. If you run formatting on this file, a space will be added below the annotation because the formatter doesn't recognize what it is annotating.

```suggestion
  /// Retrieve the stored pinpoint endpoint id
  @visibleForTesting
```

packages/analytics/amplify_analytics_pinpoint/lib/src/analytics_plugin_impl.dart
fjnoyp: Only exposing necessary fields now.  Other fields that aren't necessary for unit tests have had their @visibleForTesting annotation removed and are no longer present as parameters.

packages/analytics/amplify_analytics_pinpoint_dart/test/analytics_legacy_native_data_test.dart
```diff
@@ -0,0 +1,177 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:async';
+
+import 'package:amplify_analytics_pinpoint_dart/amplify_analytics_pinpoint_dart.dart';
+import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/endpoint_client/endpoint_store_keys.dart';
+import 'package:amplify_secure_storage_dart/amplify_secure_storage_dart.dart';
+import 'package:test/test.dart';
+
+void main() {
```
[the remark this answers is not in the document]
fjnoyp: Thanks ... Knowing about Mocktail now is very helpful as I was manually doing what can easily be captured in Mocktail instead now.  I took most of your changes, just made minor edits as I made additional API changes.

packages/analytics/amplify_analytics_pinpoint_dart/test/analytics_legacy_native_data_test.dart
```diff
@@ -0,0 +1,120 @@
...
+        store: store,
+        legacyDataProvider: legacyDataProvider,
+      );
+
+      final migratedEndpointId = await store.read(
+        key: AmplifyAnalyticsPinpointDart.endpointIdStorageKey,
+      );
+      expect(migratedEndpointId, endpointId);
+
+      verifyNever(() => legacyDataProvider.getEndpointId(captureAny()));
```
Jordan-Nelson, line 94: *nit:* `captureAny()` is intended to match any argument _and_ save the value of the argument. Since you don't need to save the value, you can use `any()` which matches any argument, but doesn't save it.

Example use of `captureAny()` from the docs:
```dart
when(() => cat.likes('fish')).thenReturn(true);
expect(cat.likes('fish'), isTrue);
final captured = verify(() => cat.likes(captureAny())).captured;
expect(captured.last, equals(['fish']))
``` 

```suggestion
      verifyNever(() => legacyDataProvider.getEndpointId(any()));
```

Jordan-Nelson (approved):

packages/analytics/amplify_analytics_pinpoint/lib/src/legacy_native_data_provider/flutter_legacy_native_data_provider.dart
```diff
@@ -0,0 +1,37 @@
...
+
+/// {@template amplify_analytics_pinpoint.flutter_legacy_native_data_provider}
+/// Provides legacy data stored by Amplify Analytics iOS and Android.
+/// {@endtemplate}
+class FlutterLegacyNativeDataProvider implements LegacyNativeDataProvider {
+  /// {@macro amplify_analytics_pinpoint.flutter_legacy_native_data_provider}
+  factory FlutterLegacyNativeDataProvider() {
+    LegacyNativeDataProvider? provider;
+
+    if (Platform.isIOS) {
```
ragingsquirrel3, line 19: This line crashed on web for me.

suggestion: `if (!zIsWeb && Platform.isIOS) `

packages/analytics/amplify_analytics_pinpoint/lib/src/legacy_native_data_provider/flutter_legacy_native_data_provider.dart
```diff
@@ -0,0 +1,37 @@
...
+      provider = DataProviderIos();
+    } else if (Platform.isAndroid) {
```
ragingsquirrel3, line 21: same here

packages/analytics/amplify_analytics_pinpoint_dart/test/analytics_legacy_native_data_test.dart
```diff
@@ -0,0 +1,120 @@
...
+
+      verify(() => legacyDataProvider.getEndpointId(appId)).called(1);
+    });
+
+    test('Second app load, legacy data is ignored', () async {
+      const legacyEndpointId = 'legacy-endpointId';
+      when(() => legacyDataProvider.getEndpointId(appId))
+          .thenAnswer((_) => Future.value(legacyEndpointId));
+
+      final endpointId = const Uuid().v1();
```
ragingsquirrel3, line 77: nit: `uuid()` from aws_common

ragingsquirrel3 (approved):
approved assuming web crash fixed per comments regarding `!zIsWeb`.

packages/analytics/amplify_analytics_pinpoint/analysis_options.yaml
dnys1: Merging with next was unsuccessful. Please fix and address Jordan's comment:

```yaml
analyzer: 
  errors:
    implementation_imports: error #TODO(equartey): Remove when lint is enforced project-wide    
  exclude:
    - "**/*.g.dart"
```

packages/analytics/amplify_analytics_pinpoint/lib/src/legacy_native_data_provider/flutter_legacy_native_data_provider.dart
```diff
@@ -0,0 +1,40 @@
...
+    } else if (Platform.isIOS) {
+      provider = DataProviderIos();
+    } else if (Platform.isAndroid) {
+      provider = DataProviderAndroid();
+    }
+
+    return FlutterLegacyNativeDataProvider._(provider);
+  }
+
+  FlutterLegacyNativeDataProvider._(this._nativeDataProvider);
```
dnys1, line 31:
```suggestion
  const FlutterLegacyNativeDataProvider._([this._nativeDataProvider]);
```

packages/analytics/amplify_analytics_pinpoint/lib/src/legacy_native_data_provider/flutter_legacy_native_data_provider.dart
```diff
@@ -0,0 +1,40 @@
...
+import 'package:amplify_analytics_pinpoint_dart/amplify_analytics_pinpoint_dart.dart';
+import 'package:aws_common/aws_common.dart';
+
+/// {@template amplify_analytics_pinpoint.flutter_legacy_native_data_provider}
+/// Provides legacy data stored by Amplify Analytics iOS and Android.
+/// {@endtemplate}
+class FlutterLegacyNativeDataProvider implements LegacyNativeDataProvider {
+  /// {@macro amplify_analytics_pinpoint.flutter_legacy_native_data_provider}
+  factory FlutterLegacyNativeDataProvider() {
+    LegacyNativeDataProvider? provider;
```
dnys1, line 18: I would prefer to be more explicit in this method. It's currently a bit confusing to read.

```suggestion
    if (zIsWeb || !(Platform.isIOS || Platform.isAndroid)) {
      return const FlutterLegacyNativeDataProvider._();
    }

    final LegacyNativeDataProvider provider;
```

packages/analytics/amplify_analytics_pinpoint/lib/src/legacy_native_data_provider/flutter_legacy_native_data_provider.dart
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  FlutterLegacyNativeDataProvider._(this._nativeDataProvider);
+
+  final LegacyNativeDataProvider? _nativeDataProvider;
+
+  @override
+  Future<String?> getEndpointId(String pinpointAppId) async {
+    if (_nativeDataProvider == null) return null;
+    return _nativeDataProvider!.getEndpointId(pinpointAppId);
```
dnys1, line 38:
```suggestion
    return _nativeDataProvider?.getEndpointId(pinpointAppId);
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -323,4 +321,54 @@ class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
...
+      fixedEndpointId = uuid();
+      await store.write(
+        key: endpointIdStorageKey,
+        value: fixedEndpointId,
+      );
+    } else {
+      fixedEndpointId = currentEndpointId;
+    }
+
+    return fixedEndpointId;
```
dnys1, line 372: Since the goal is to arrive at a fixed endpoint ID, use a single variable to convey that journey. This also cleans up the logic branches and saves you a read from the DB in some cases.

```suggestion
    String? fixedEndpointId;
    if (endpointInformationVersion == null) {
      final legacyEndpointId =
          await legacyDataProvider?.getEndpointId(pinpointAppId);
      // Migrate legacy data if it is non-null
      if (legacyEndpointId != null) {
        fixedEndpointId = legacyEndpointId;
        await store.write(
          key: endpointIdStorageKey,
          value: legacyEndpointId,
        );
      }
      // Update the version to prevent future legacy data migrations.
      await store.write(
        key: EndpointStoreKey.version.name,
        value: EndpointStoreVersion.v1.name,
      );
    }

    // Read the existing ID.
    fixedEndpointId ??= await store.read(
      key: endpointIdStorageKey,
    );

    // Generate a new ID if one does not exist.
    if (fixedEndpointId == null) {
      fixedEndpointId = uuid();
      await store.write(
        key: endpointIdStorageKey,
        value: fixedEndpointId,
      );
    }

    return fixedEndpointId;
```

packages/analytics/amplify_analytics_pinpoint_dart/test/analytics_legacy_native_data_test.dart
```diff
@@ -0,0 +1,120 @@
...
+    const legacyEndpointId = 'legacy-endpointId';
+
+    setUp(() {
+      legacyDataProvider = MockLegacyNativeDataProvider();
+      store = MockSecureStorage();
+    });
+
+    test('First app load, no legacy data, writes proper values', () async {
+      when(() => legacyDataProvider.getEndpointId(appId))
+          .thenAnswer((_) => Future.value());
```
dnys1, line 28: Explicitly show that null is being returned

```suggestion
          .thenAnswer((_) => Future.value(null));
```

Or better yet:

```suggestion
          .thenAnswer((_) async => null);
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -323,4 +321,54 @@ class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
     await _eventClient.close();
     await _eventStorageAdapter.close();
   }
+
+  /// Retrieve the stored pinpoint endpoint id
+  @visibleForTesting
+  static Future<String> retrieveEndpointId({
```
dnys1, line 327: This should be an instance method and only the app ID should be passed in (or also saved as a property on the instance). Having a static method doesn't make much sense since the method is dependent on the secure storage + legacy data provider instance variables.

packages/analytics/amplify_analytics_pinpoint_dart/test/analytics_legacy_native_data_test.dart
```diff
@@ -0,0 +1,120 @@
...
+
+      verify(() => legacyDataProvider.getEndpointId(appId)).called(1);
+    });
+
+    test('Second app load, legacy data is ignored', () async {
+      const legacyEndpointId = 'legacy-endpointId';
+      when(() => legacyDataProvider.getEndpointId(appId))
+          .thenAnswer((_) => Future.value(legacyEndpointId));
+
+      final endpointId = UUID.getUUID();
```
dnys1, line 77:
```suggestion
      final endpointId = uuid();
```

packages/analytics/amplify_analytics_pinpoint_dart/test/analytics_legacy_native_data_test.dart
```diff
@@ -0,0 +1,120 @@
...
+      final migratedEndpointId = await store.read(
+        key: AmplifyAnalyticsPinpointDart.endpointIdStorageKey,
+      );
+      expect(migratedEndpointId, legacyEndpointId);
+
+      verify(() => legacyDataProvider.getEndpointId(appId)).called(1);
+    });
+
+    test('Second app load, legacy data is ignored', () async {
+      const legacyEndpointId = 'legacy-endpointId';
```
dnys1, line 73: This is already defined above

packages/analytics/amplify_analytics_pinpoint/lib/src/legacy_native_data_provider/flutter_legacy_native_data_provider.dart
fjnoyp: Can't make provider final because due of compile error variable could be uninitialized.

packages/analytics/amplify_analytics_pinpoint_dart/test/analytics_legacy_native_data_test.dart
fjnoyp: Alright thanks!

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
fjnoyp: Alright, refactored the code to be an instance method instead.

packages/analytics/amplify_analytics_pinpoint_dart/test/analytics_legacy_native_data_test.dart
fjnoyp: oh gosh my bad, thanks

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
dnys1: My comments are not in conflict with Jordan's. He was confused about the duplication of logic and unnecessary reads/writes.

dnys1 (approved):

PR #2494, opened 2022-12-15
fix(api): prevent GQL subscription race condition error after shutdown

ragingsquirrel3:
This PR prevents a race condition error in GraphQL subscriptions. Sometimes, when a bloc is closing (which closes the underlying websocket) there will be a brief period of time where the bloc's internal event stream controller is closed but it still gets events from the websocket channel. This throws an error bc the stream controller is already closed.

This PR fixes it by ignoring events when the controller is closed and shutdown is pending/complete.

### DIFF
--- a/packages/api/amplify_api/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart
+++ b/packages/api/amplify_api/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart
@@ -116,6 +116,13 @@ class WebSocketBloc with AWSDebuggable, AmplifyLoggerMixin {
 
   /// Adds an event to the Bloc.
   void add(WebSocketEvent event) {
+    if (_wsEventController.isClosed &&
+        (_currentState is DisconnectedState ||
+            _currentState is PendingDisconnect)) {
+      // In some cases, the service will keep getting messages during/after
+      // disconnect. Ignore those messages.
+      return;
+    }
     _wsEventController.add(event);
   }
 
--- a/packages/api/amplify_api/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart
+++ b/packages/api/amplify_api/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart
@@ -116,13 +116,6 @@ class WebSocketBloc with AWSDebuggable, AmplifyLoggerMixin {
 
   /// Adds an event to the Bloc.
   void add(WebSocketEvent event) {
-    if (_wsEventController.isClosed &&
-        (_currentState is DisconnectedState ||
-            _currentState is PendingDisconnect)) {
-      // In some cases, the service will keep getting messages during/after
-      // disconnect. Ignore those messages.
-      return;
-    }
     _wsEventController.add(event);
   }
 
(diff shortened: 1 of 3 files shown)


packages/api/amplify_api/test/util.dart
```diff
@@ -185,7 +185,10 @@ class MockWebSocketSink extends DelegatingStreamSink<dynamic>
   MockWebSocketSink(super.sink);
 
   @override
-  Future<void> close([int? closeCode, String? closeReason]) => super.close();
+  Future<void> close([int? closeCode, String? closeReason]) async {
+    await Future<void>.delayed(const Duration(milliseconds: 10));
```
dnys1, line 189: Can you add a comment why?

packages/api/amplify_api/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart
```diff
@@ -116,6 +116,13 @@ class WebSocketBloc with AWSDebuggable, AmplifyLoggerMixin {
 
   /// Adds an event to the Bloc.
   void add(WebSocketEvent event) {
+    if (_wsEventController.isClosed &&
+        (_currentState is DisconnectedState ||
+            _currentState is PendingDisconnect)) {
+      // In some cases, the service will keep getting messages during/after
+      // disconnect. Ignore those messages.
+      return;
```
dnys1, line 124: While this is true, we can't know the origin of `event` here. It's better to handle this at the source so that we can ensure events that erroneously end up here from other places are reported/caught by tests and not discarded.

Ultimately `add` should only ever be called if we are sure the bloc is still open. To call `add` at any other point is incorrect.

ragingsquirrel3: Makes sense. In this case, is coming from the service, so would you recommend changing [this line ](https://github.com/aws-amplify/amplify-flutter/blob/next/packages/api/amplify_api/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart#L274) to ?:

```dart
    _currentState.service.init(_currentState).listen(
      (event) {
        if (_currentState is! DisconnectedState &&
            _currentState is! PendingDisconnect) {
          add(event);
        }
      },
      onError: (Object error, StackTrace st) {
        _emit(_currentState.failed(error, st));
      },
    );
```

Equartey: I know I had a similar idea to the original, but after Dillon's call out I like this approach. Assuming it still resolves our race condition, I support it.

dnys1: Yeah, sounds good. Although this logic is used elsewhere so may be a good opportunity to add a helper method. Also, checking `isClosed` on the controller is relatively cheap compared to an `is` type check so I would keep that first.

```dart
void _checkAdd(Event event) {
  if (_wsEventController.isClosed || _currentState is DisconnectedState || _currentState is PendingDisconnect) {
    return;  
  }
  add(event);
}
```

Equartey: Also could still be checking for closed controller. Might be useful to break the logic check into a method and reuse within `_poll()` as well.

ragingsquirrel3: Just made this change (and used in `_poll()`).

Equartey (approved):
Stellar! 🚀

dnys1 (approved):

PR #2502, opened 2022-12-20
chore(api): reduce integ test flakiness

ragingsquirrel3:
This PR modifies a few API integ tests to reduce flakiness:
* no longer looks for specific blog entries in a list request response because there is no guarantee first page of responses will have newly created entries if the backend already has a lot of data. Just checking that the list has some blogs is sufficient.
* Require all subscription tests that use a util to pass a filter callback for the events. If this is not provided, it causes issues with getting events from other clients using the backend at the same time (like tests on another platform in CI). This came up in a recently added test for errors but should be required for all similar tests in the future.

### DIFF
--- a/packages/api/amplify_api/example/integration_test/graphql/iam_test.dart
+++ b/packages/api/amplify_api/example/integration_test/graphql/iam_test.dart
@@ -103,19 +103,13 @@ void main({bool useExistingTestUser = false}) {
 
       testWidgets('should LIST blogs with Model helper',
           (WidgetTester tester) async {
-        const blog1Name = 'Integration Test Blog 1';
-        const blog2Name = 'Integration Test Blog 2';
-        const blog3Name = 'Integration Test Blog 3';
-        final blog1 = await addBlog(blog1Name);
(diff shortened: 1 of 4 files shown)


packages/api/amplify_api/example/integration_test/graphql/iam_test.dart
```diff
@@ -103,19 +103,13 @@ void main({bool useExistingTestUser = false}) {
...
-        final blog3 = await addBlog(blog3Name);
+        await addBlog('Integration Test Blog 1');
         final req = ModelQueries.list<Blog>(Blog.classType);
         final operation = Amplify.API.query(request: req);
         final res = await operation.response;
         final data = res.data;
-        final blogs = [blog1, blog2, blog3];
         expect(res, hasNoGraphQLErrors);
-        expect(data?.items, containsAll(blogs));
+        expect(data?.items.length, greaterThan(0));
```
HuiSF, line 112:
```suggestion
        expect(data?.items, isNotEmpty);
```

HuiSF (approved):

Equartey (approved):

Issue #2492, opened 2022-12-15
[api] Model helpers should not pass `null` fields


dnys1:
### Description

AppSync interprets explicit `null` values different than implicit ones in the `variables` map of a request. Model helpers explicitly setting these values can lead to unexpected errors when these fields are read-only or otherwise protected.

For example, consider the schema:

```graphql
type Todo @model @auth(rules: [{ allow: public }, { allow: owner, ownerField: "owners" }]) {
  id: ID!
  name: String!
  owners: [String]
  private: String @auth(rules: [{ allow: owner, ownerField: "owners" }])
}
```

And a mutation to create a `Todo`:

```graphql
mutation CreateTodo($input: CreateTodoInput!) {
  createTodo(input: $input) {
    id
    name
    owners
    private
    createdAt
    updatedAt
  }
}
```

The following input allows creation with no issues: 

```json
{
  "input": {
    "name": "test",
    "private": "secret"
  }
}
```

However, model helpers generates the following input (notice the inclusion of `owners` which cannot be written to):

```json
{
  "input": {
    "name": "test",
    "owners": null,
    "private": "secret"
  }
}
```

This leads to the following error:

```json
{
  "data": {
    "createTodo": null
  },
  "errors": [
    {
      "path": [
        "createTodo"
      ],
      "data": null,
      "errorType": "Unauthorized",
      "errorInfo": null,
      "locations": [
        {
          "line": 2,
          "column": 3,
          "sourceName": null
        }
      ],
      "message": "Not Authorized to access createTodo on type Todo"
    }
  ]
}
```

### Categories

- [ ] Analytics
- [ ] API (REST)
- [X] API (GraphQL)
- [ ] Auth
- [ ] Authenticator
- [ ] DataStore
- [ ] Storage

### Platforms

- [ ] iOS
- [ ] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Environment

```bash
N/A
```

### Dependencies

```bash
vNext
```

### Device

N/A

### OS

N/A

### Deployment Method

Amplify CLI

### Amplify Config

N/A

ragingsquirrel3:
I made a PR for this. Noted in the PR description but IMO we should not remove all null values from mutation input (plenty of valid use cases for doing that), only null values when they are owner fields.

okhomin:
I faced the same issue for my model:
```
type UserInfo @model @auth(rules: [{allow: owner}]) {
  id: ID!
[...]

PR #2504, opened 2022-12-21
fix(api): do not include null values in ModelMutations.create

ragingsquirrel3:
Addresses https://github.com/aws-amplify/amplify-flutter/issues/2492

GraphQL model helpers have mutation helpers that generate JSON input for mutation variables based on models' `toJson()` methods. In the case of the model having a null value for an owner field, that null value gets passed to AppSync which will reject the mutation. This PR removes null fields from input variables when calling `ModelMutations.create()`. That avoids the error for the case of owner fields and generally simplifies client logic, allowing the server to set null values if omitted.

Note the original issue states "Model helpers should not pass null fields" but I believe that in general, there are valid cases of setting null values in mutations, like unsetting a nullable field that had previously been set to something non-null. For that reason, this PR only scopes the fix to create operations, which does not have any need to explicitly send null values.


okhomin (approved):

HuiSF (approved):

dnys1 (changes requested):
What about `update`? I'm guessing that would suffer the same problem.

You're correct that we should be setting `null` in cases where the user passed `null` but it's possible to distinguish these cases precisely. In the `variables` map, the presence of that field's key indicates the user passed a value for it. You need to check that instead of the value.

```dart
final variables = {
  'setNull': null,
};

// Doing this gives no insight about the user's intentions.
print(variables['setNull'] == null); // true
print(variables['doNotSetNull'] == null); // true

// Here we see what their intentions were.
print(variables.containsKey('setNull')); // true
print(variables.containsKey('doNotSetNull')); // false
```

Issue #2499, opened 2022-12-18
[Web] ModelSubscriptions.onUpdate callback not fired when GraphQL mutation update occurs


fvisticot:
### Description

- I subscribe to a model update like in the code bellow.
- When update occurs from an update in the Flutter App, callback is called successfully
- When I try to update the data from AppSync AWS console with a mutation, the callback is not fired.

Any idea why it does not work ?

```
final subscriptionRequest = ModelSubscriptions.onUpdate(Device.classType);
    _subscription = Amplify.API
        .subscribe(
          subscriptionRequest,
          onEstablished: () => print('Subscription established'),
        )
        .take(5)
        .handleError((error) {
      print(error);
    }).listen((event) {
      print(event);
    });
```

### Categories

- [ ] Analytics
- [ ] API (REST)
- [X] API (GraphQL)
- [ ] Auth
- [ ] Authenticator
- [ ] DataStore
- [ ] Storage

### Platforms

- [ ] iOS
- [ ] Android
- [X] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Environment

```bash
[✓] Flutter (Channel stable, 3.3.9, on macOS 12.6.1 21G217 darwin-arm, locale fr-FR)
[!] Android toolchain - develop for Android devices (Android SDK version 33.0.1)
    ✗ cmdline-tools component is missing
      Run `path/to/sdkmanager --install "cmdline-tools;latest"`
      See https://developer.android.com/studio/command-line for more details.
    ✗ Android license status unknown.
      Run `flutter doctor --android-licenses` to accept the SDK licenses.
      See https://flutter.dev/docs/get-started/install/macos#android-setup for more details.
[✓] Xcode - develop for iOS and macOS (Xcode 14.2)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2021.3)
[✓] Connected device (2 available)
[✓] HTTP Host Availability
```

### Dependencies

```bash
dependencies:
- ai_barcode 3.2.4 [flutter ai_barcode_platform_interface ai_barcode_web]
- amplify_analytics_pinpoint 1.0.0-next.2 [amplify_analytics_pinpoint_dart amplify_core amplify_db_common amplify_secure_storage aws_common device_info_plus flutter meta package_info_plus path_provider]
[...]

PR #2507, opened 2022-12-21
fix(api): improve GQL subscription error visibility/recovery

ragingsquirrel3:
Fixes some issues I discovered while debugging https://github.com/aws-amplify/amplify-flutter/issues/2499.

While the underlying issues in above issue are 1) subscription selection set was not a subset of mutation selection set which causes AppSync error 2) error not parsed correctly, fixed in https://github.com/aws-amplify/amplify-flutter/pull/2483, I also noticed that any similar error would get completely swallowed making it difficult to debug. Further, the `WebSocketBloc` gets stuck in an error state so that user cannot make a new subscription.

This PR fixes that by calling an error utility already used other places. That error utility passes down the errors to subscription streams so that users can see them. It also disconnects the bloc (and underlying WebSocket) and cleans it up. That way, future subscriptions get a fresh bloc/WebSocket connection.


Equartey (approved):
Good call, this will be a good improvement 🎉

HuiSF (approved):

PR #2513, opened 2022-12-22
chore(analytics): Pub Score Improvements

Equartey:
*Description of changes:*

- Enforces `implementation_imports` lint rule
- Fixes pubspec.yaml Repo URL
- Fixes missing example in dart package by adding `packages/analytics/amplify_analytics_pinpoint_dart/example/readme.md`


packages/analytics/amplify_analytics_pinpoint/analysis_options.yaml
```diff
@@ -1 +1,5 @@
 include: package:amplify_lints/library.yaml
+
+analyzer: 
+  errors:
+    implementation_imports: error
```
Jordan-Nelson, line 5: If this is leading to reduced pub score, we should considering changing it in amplify_lints.

https://github.com/aws-amplify/amplify-flutter/blob/b9d6d08c069571496211b9a462b383932fc5a831/packages/amplify_lints/lib/library.yaml#L10

If changing it in amplify_lints results in CI failures in other packages, you could leave amplify_lints as is for now add a TODO here to remove this once it has been updated in in amplify_lints. Once you have fixed this in other packages, you can come back and update amplify_lints and update all the TODOs.

Equartey: Yeah, the plan is override it locally per package as we go and then circle back and clean up the overrides. Good call on the TODOs.

HuiSF (approved):

packages/analytics/amplify_analytics_pinpoint/analysis_options.yaml
dnys1: We can still update `amplify_lints` so it will at least call it out when we're developing.

Equartey: Unfortunately, Pana doesn't respect ignores when running the pub score analysis. Here is a small discussion about [this](https://github.com/dart-lang/pana/issues/1029). If we want to increase the pub score we have to adhere to their lint rules too.

dnys1: This is why generated code doesn't throw up all kinds of alarm bells
Screen Shot 2023-01-04 at 9 26 01 AM

Equartey: Ah I see that now. However, part of this task is ensuring no regressions happen with the pub score (when possible). If we moved away from a lint rule, is it possible to provide a mechanism beyond manual inspection?

dnys1: Yes, I'm suggesting we enable the lint in `amplify_lints`. We can also have a CI step which runs pana (@Jordan-Nelson had started this effort at one point and can likely port it to the current setup very easily).

Equartey: Understood. I'll adjust to `ignore_for_file` and add the lint rule (per package at first to avoid build failures). I'll defer to @Jordan-Nelson about adding pana to CI. It sounded like it was problematic before.

Jordan-Nelson (approved):

PR #2516, opened 2022-12-27
fix(auth): legacy data migration for iOS

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*

This PR fixes iOS legacy data migration which was broken when https://github.com/aws-amplify/amplify-flutter/pull/2023 was merged, which added a default `accessible` value for Keychain queries. The PR changes the default iOS/MacOS Keychain options to all be `null` when using the `byNameSpace` constructor so that keychain defaults are used. This aligns with the iOS SDK.


ragingsquirrel3 (approved):

fjnoyp (approved):
Thanks I cherry picked this into my branch for data migration and was able to confirm iOS legacy data is now being loaded.  

I'll wait for this PR to be merged in before merging in data migration for Analytics now

PR #2517, opened 2022-12-27
chore(analytics): support Pinpoint User Attributes

fjnoyp:
*Issue #, if available:*

*Description of changes:*

Allow users to call `identifyUser` with an  AWSPinpointUserProfile that has an additional user attributes field so attributes can be related to users and not the endpoint.  This is to have feature parity with Amplify Android which does the same thing.  

### DIFF
--- a/packages/amplify_core/lib/src/types/analytics/analytics/analytics_user_profile.dart
+++ b/packages/amplify_core/lib/src/types/analytics/analytics/analytics_user_profile.dart
@@ -18,14 +18,4 @@ class AnalyticsUserProfile {
     this.location,
     AnalyticsProperties? analyticsProperties,
   }) : properties = analyticsProperties;
-
-  Map<String, Object?> getAllProperties() => {
-        if (name != null) 'name': name,
-        if (email != null) 'email': email,
-        if (plan != null) 'plan': plan,
-        if (location != null) 'location': location!.getAllProperties(),
-        if (properties != null) 'propertiesMap': properties!.getAllProperties(),
-        if (properties != null)
-          'propertiesTypesMap': properties!.getAllPropertiesTypes(),
-      };
 }
--- a/packages/analytics/amplify_analytics_pinpoint/example/integration_test/identify_user_test.dart
+++ b/packages/analytics/amplify_analytics_pinpoint/example/integration_test/identify_user_test.dart
@@ -99,7 +99,7 @@ void main() {
                 .having((e) => e.endpoint.user?.userId, 'UserId', userId)
                 .having(
                   (e) => e.endpoint.user?.userAttributes?.toMap() ?? const {},
-                  'Attributes',
+                  'UserAttributes',
                   equals({
                     boolProperty.key: [stringifiedBoolProperty.value],
                     stringProperty.key: [stringProperty.value],
--- a/packages/analytics/amplify_analytics_pinpoint/example/integration_test/identify_user_test.dart
+++ b/packages/analytics/amplify_analytics_pinpoint/example/integration_test/identify_user_test.dart
@@ -3,6 +3,7 @@
 
 import 'dart:async';
 
+import 'package:amplify_analytics_pinpoint/amplify_analytics_pinpoint.dart';
 import 'package:amplify_analytics_pinpoint_dart/src/sdk/pinpoint.dart'
     show EndpointLocation;
 import 'package:amplify_flutter/amplify_flutter.dart';
@@ -56,12 +57,13 @@ void main() {
 
         await Amplify.Analytics.identifyUser(
           userId: userId,
-          userProfile: AnalyticsUserProfile(
+          userProfile: AWSPinpointUserProfile(
             name: name,
             email: email,
             plan: plan,
             location: location,
             analyticsProperties: properties,
+            userAttributes: properties,
           ),
         );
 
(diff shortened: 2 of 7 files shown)


packages/amplify_core/lib/src/types/analytics/analytics/analytics_user_profile.dart
```diff
@@ -30,14 +30,4 @@ class AnalyticsUserProfile {
     this.location,
     AnalyticsProperties? analyticsProperties,
   }) : properties = analyticsProperties;
-
```
fjnoyp, line 33: Removing non used code

Equartey (approved):

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/aws_pinpoint_user_profile.dart
```diff
@@ -0,0 +1,35 @@
...
+
+import 'package:amplify_core/amplify_core.dart';
+
+/// {@template amplify_analytics_pinpoint_dart.aws_pinpoint_user_profile}
+/// Extends the category-defined UserProfile class to
+/// include features supported relevant to Pinpoint only.
+/// {@endtemplate}
+class AWSPinpointUserProfile extends AnalyticsUserProfile {
+  /// {@macro amplify_analytics_pinpoint_dart.aws_pinpoint_user_profile}
+  AWSPinpointUserProfile({
```
dnys1, line 24: `const`

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_client.dart
```diff
@@ -108,6 +109,14 @@ class EndpointClient {
       await _globalFieldsManager.addMetrics(userProfile.properties!.metrics);
     }
 
+    if (userProfile is AWSPinpointUserProfile) {
+      final attributes = <String, List<String>>{};
+      userProfile.userAttributes.getAllProperties().forEach((key, value) {
+        attributes[key] = [value.toString()];
+      });
+      newUserBuilder.userAttributes = ListMultimapBuilder(attributes);
```
dnys1, line 117:
```suggestion
      userProfile.userAttributes.getAllProperties().forEach(
            (key, value) => newUserBuilder.userAttributes.add(
              key,
              value.toString(),
            ),
          );
```

dnys1: This brings up a question - should `userAttributes` allow setting multiple values per key? Right now, only one value is allowed but it seems like Pinpoint allows multiple.

fjnoyp: Great question and catch there.  That is correct Pinpoint allows that. 

Unfortunately Android and iOS only expose it as a single value and not as a list.  This is true for the Endpoint global attributes and for the user attributes as well.  

For consistency I believe we should keep it the same as they do.  That's the general direction I'm aware of from our PM given how we are looking to possibly gut and change all of Analytics anyway in the future ..

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/aws_pinpoint_user_profile.dart
```diff
@@ -0,0 +1,35 @@
...
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
```
Jordan-Nelson, line 14:
```suggestion
// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
// SPDX-License-Identifier: Apache-2.0
```

packages/analytics/amplify_analytics_pinpoint/example/integration_test/identify_user_test.dart
```diff
@@ -105,6 +107,16 @@ void main() {
                   ),
                 )
                 .having((e) => e.endpoint.user?.userId, 'UserId', userId)
+                .having(
+                  (e) => e.endpoint.user?.userAttributes?.toMap() ?? const {},
+                  'Attributes',
```
dnys1, line 112:
```suggestion
                  'UserAttributes',
```

dnys1: Although this is confusing - why does `Attributes` below have `name` and `email`, but then these "user attributes" have random values. Shouldn't these be things like phone number, birthday, etc.?

fjnoyp: It's because Analytics organization is a bit strange now ... Abdallah says we'll be revamping Analytics particularly because of these legacy behaviors that shouldn't be supported into the future :) 

We call this code: 

```
await Amplify.Analytics.identifyUser(
          userId: userId,
          userProfile: AWSPinpointUserProfile(
            name: name,
            email: email,
            plan: plan,
            location: location,
            analyticsProperties: properties,
            userAttributes: properties,
          ),
        );
```

The AwsPinpointUserProfile name, email, plan fields are mapped into Pinpoint Endpoint attributes. 

However, the AwsPinpointUserProfile userAttributes are mapped into Pinpoint Endpoint user attributes. 

Thus the integration test checks for that, but if you see the triggering API call it doesn't match at all.  It's great how integration tests can show how a public API isn't good, when the code expectation doesn't match what's actually being verified.

dnys1 (approved):

PR #2518, opened 2022-12-27
chore(analytics): expose pinpoint id in hub

fjnoyp:
*Issue #, if available:*

*Description of changes:*

Send the Pinpoint ID in Analytics Hub Events so that Auth can register and send it on their side to Cognito.  iOS and Android already do this in their own way (sending Pinpoint ID to Cognito so it can be registered there). 

### DIFF
--- a/packages/amplify_core/lib/src/types/analytics/hub/analytics_hub_event.dart
+++ b/packages/amplify_core/lib/src/types/analytics/hub/analytics_hub_event.dart
@@ -1,0 +1,59 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'package:amplify_core/amplify_core.dart';
+
+/// {@template amplify_common.hub.analytics_hub_event_type}
+/// Hub Event types for the Analytics category.
+/// {@endtemplate}
+enum AnalyticsHubEventType {
+  /// {@template amplify_common.hub.analytics_hub_event_endpoint_configured}
+  /// Emitted when the user's endpoint is configured during application start.
+  /// {@endtemplate}
+  endpointConfigured('ENDPOINT_CONFIGURED');
+
+  /// The event name.
+  final String eventName;
+
+  /// {@macro amplify_common.hub.analytics_hub_event_type}
+  const AnalyticsHubEventType(this.eventName);
+}
+
+/// The base class for hub events of the Analytics category.
+///
+/// One of [AnalyticsHubEventType].
+class AnalyticsHubEvent extends HubEvent<AnalyticsHubEventPayload>
+    with AWSEquatable<AnalyticsHubEvent>, AWSDebuggable {
+  AnalyticsHubEvent._(
+    this.type, {
+    AnalyticsHubEventPayload? payload,
+  }) : super(type.eventName, payload: payload);
+
+  /// {@macro amplify_common.hub.analytics_hub_event_type}
+  final AnalyticsHubEventType type;
+
+  /// {@macro amplify_common.hub.analytics_hub_event_endpoint_configured}
+  AnalyticsHubEvent.endpointConfigured(String endpointId)
+      : this._(
+          AnalyticsHubEventType.endpointConfigured,
+          payload: EndpointConfiguredPayload(endpointId),
+        );
+
+  @override
+  List<Object?> get props => [type, payload];
+
+  @override
+  String get runtimeTypeName => 'AnalyticsHubEvent';
+}
+
+abstract class AnalyticsHubEventPayload {
+  const AnalyticsHubEventPayload();
+}
+
+/// {@macro amplify_common.hub.analytics_hub_event_endpoint_configured}
+class EndpointConfiguredPayload extends AnalyticsHubEventPayload {
+  const EndpointConfiguredPayload(this.endpointId);
+
+  /// The unique ID of the user's endpoint.
+  final String endpointId;
+}
--- a/packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
+++ b/packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
@@ -34,7 +34,8 @@ const zSessionStopEventType = '_session.stop';
 /// - Validates and parses inputs
 /// - Receives and provides external Flutter Provider implementations
 /// {@endtemplate}
-class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
+class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface
+    implements Closeable {
   /// {@macro amplify_analytics_pinpoint_dart.amplify_analytics_pinpoint_dart}
   AmplifyAnalyticsPinpointDart({
     SecureStorageInterface? endpointInfoStore,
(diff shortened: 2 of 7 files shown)


Equartey (approved):
LGTM

HuiSF (approved):

packages/amplify_core/lib/src/types/analytics/hub/analytics_hub_event.dart
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+import 'package:amplify_core/amplify_core.dart';
+
+/// {@template amplify_common.hub.analytics_hub_event_type}
+/// Hub Event types for the Analytics category.
+/// {@endtemplate}
+enum AnalyticsHubEventType {
+  /// {@macro amplify_common.hub.analytics_hub_event_pinpoint_endpoint_id_set}
+  pinpointEndpointIdSet('PINPOINT_ENDPOINT_ID_SET');
```
dnys1, line 23: Not in love with this name - it implies a non-fixed ID and the existence of an `unset` event as well.. what about just `configured` or something since it will be fired every startup after configuration completes?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -197,6 +208,11 @@ class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
     unawaited(
       Amplify.asyncConfig.then((_) {
         _endpointClient.updateEndpoint();
+        _hubEventController.add(
+          AnalyticsHubEvent.pinpointEndpointIdSet(
+            AnalyticsHubEventPayload(pinpointEndpointId: fixedEndpointId),
```
dnys1, line 213: If `updateEndpoint` fails, is it correct to fire this event?

packages/amplify_core/lib/src/types/analytics/hub/analytics_hub_event.dart
fjnoyp: Sounds good!  Made the change to -> PINPOINT_ENDPOINT_ID_CONFIGURED

I think ID is a bit wordy but it conveys that we're talking specifically about ID being configured here and nothing else as the scope is quite focused on the endpoint id being configured.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
fjnoyp: No as it can fail and no pinpoint id would exist, thanks fixed and added a check.

packages/amplify_core/lib/src/types/analytics/hub/analytics_hub_event.dart
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+import 'package:amplify_core/amplify_core.dart';
+
+/// {@template amplify_common.hub.analytics_hub_event_type}
+/// Hub Event types for the Analytics category.
+/// {@endtemplate}
+enum AnalyticsHubEventType {
+  /// {@macro amplify_common.hub.analytics_hub_event_pinpoint_endpoint_id_set}
+  pinpointEndpointIdSet('PINPOINT_ENDPOINT_ID_CONFIGURED');
```
dnys1, line 23: You've only updated the string value from "set" -> "configured". More importantly, though, I was suggesting something shorter and more generic like just `configured`, `endpointConfigured`, or `endpointUpdated`. Although, I think `endpointConfigured` here works best.

```suggestion
  endpointConfigured('ENDPOINT_CONFIGURED'),
```

Hub events should not mention AWS services since they are generic to the category. Further, the event that triggers this hub event is the successful creation of an "endpoint" which happens to include an ID. Since we already have a category API called `updateEndpoint`, piggy backing on that verbiage would seem to be the most appropriate.

An `endpointUpdated` event could also make sense which is called after each `updateEndpoint` call and includes the full endpoint instead of just the ID.

packages/amplify_core/lib/src/types/analytics/hub/analytics_hub_event.dart
```diff
@@ -0,0 +1,60 @@
...
+  List<Object?> get props => [type, payload];
+
+  @override
+  String get runtimeTypeName => 'AnalyticsHubEvent';
+}
+
+class AnalyticsHubEventPayload {
+  const AnalyticsHubEventPayload({required this.pinpointEndpointId});
+
+  final String pinpointEndpointId;
```
dnys1, line 59: Hub events should not mention AWS services

```suggestion
  final String endpointId;
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -195,8 +206,15 @@ class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
       endpoint.toBuilder(),
     );
     unawaited(
-      Amplify.asyncConfig.then((_) {
-        _endpointClient.updateEndpoint();
+      Amplify.asyncConfig.then((_) async {
+        final result = await _endpointClient.updateEndpoint();
+        if (result.messageBody.message == 'Accepted') {
```
dnys1, line 211: Not needed - if `updateEndpoint` completes, then the request succeeded. It will throw an exception otherwise.

If this fails, though... we have a pretty big problem, right?

packages/amplify_core/lib/src/types/analytics/hub/analytics_hub_event.dart
fjnoyp: Right thanks for clarifying that.  I think it would be good to mention endpoint to specify the exact nature of the configuration here.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
fjnoyp: Indeed, Analytics wouldn't be working properly.  I updated the code to catch and throw the exception. 

```
unawaited(
      Amplify.asyncConfig.then((_) async {
        try {
          await _endpointClient.updateEndpoint();
          _hubEventController.add(
            AnalyticsHubEvent.pinpointEndpointIdSet(
              AnalyticsHubEventPayload(pinpointEndpointId: fixedEndpointId),
            ),
          );
        } on Exception catch (e) {
          throw AnalyticsException(
            'Error updating Pinpoint Endpoint',
            underlyingException: e,
          );
        }
      }),
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -195,8 +206,18 @@ class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
...
-      Amplify.asyncConfig.then((_) {
-        _endpointClient.updateEndpoint();
+      Amplify.asyncConfig.then((_) async {
+        try {
+          await _endpointClient.updateEndpoint();
+          _hubEventController.add(
+            AnalyticsHubEvent.endpointConfigured(fixedEndpointId),
+          );
+        } on Exception catch (e) {
+          throw AnalyticsException(
```
dnys1, line 216: This will result in an unhandled exception - there is no way for the user to catch this or see it in their code.

There are two solutions to this issue.

1. Instead of waiting for `Amplify.asyncConfig` we can add category-specific waiters, so that dependencies like this are easier to control, e.g. `Amplify.Auth.asyncConfig`
2. Instead of throwing the error, store it locally, then throw it whenever the first `Amplify.Analytics.*` method is called so that it is available to the customer.

fjnoyp: Got it thanks. 

I missed that you can't catch exceptions in unawaited futures. 

For 1) I feel that would add coupling between the categories we were trying to avoid via the use of hub events. 

I've implemented your suggestion 2).

fjnoyp:
Based on talks with Amplify native teams, the pinpoint id should be generated by auth and analytics independent of each other.

PR #2530, opened 2023-01-04
fix(analytics): Secure storage on Android

dnys1:
The secure storage worker is not supported on Android which is why Flutter clients use the `AmplifySecureStorage` wrapper instead.


fjnoyp (approved):
Thanks for getting this!

fjnoyp:
Was wondering how this made it through without being fixed.  

I actually made this fix in another PR that's been in review for a while, so if it gets merged in before this shouldn't be necessary anymore: 

https://github.com/aws-amplify/amplify-flutter/pull/2489

Jordan-Nelson (approved):

PR #2541, opened 2023-01-05
ci(repo): Use custom bash for `geckodriver`

dnys1:
Since the setup action does not authenticate GitHub requests which leads to constant rate limiting issues.


Equartey (approved):
LGTM

HuiSF (approved):

PR #2545, opened 2023-01-06
fix(http): Cancel streamed responses

dnys1:
Allow for the ability to cancel streamed responses via the body `Stream` only (vs. the operation, which will have already completed).

### DIFF
--- a/packages/aws_common/test/http/http_common.dart
+++ b/packages/aws_common/test/http/http_common.dart
@@ -56,8 +56,10 @@ void clientTest(
           client = debugClient..supportedProtocols = supportedProtocols;
         });
         tearDown(() async {
-          httpServerChannel.sink.add(null);
-          // await Future<void>.delayed(const Duration(seconds: 30));
+          httpServerChannel.sink.rejectErrors().add(null);
+        });
+        tearDownAll(() async {
+          // await Future<void>.delayed(const Duration(seconds: 60));
         });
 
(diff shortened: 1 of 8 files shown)


packages/aws_common/test/http/http_common.dart
```diff
@@ -56,8 +56,10 @@ void clientTest(
           client = debugClient..supportedProtocols = supportedProtocols;
         });
         tearDown(() async {
-          httpServerChannel.sink.add(null);
-          // await Future<void>.delayed(const Duration(seconds: 30));
+          httpServerChannel.sink.rejectErrors().add(null);
+        });
+        tearDownAll(() async {
+          // await Future<void>.delayed(const Duration(seconds: 60));
```
HuiSF, line 62: Is the `tearDownAll` needed?

HuiSF (approved):

packages/aws_common/test/http/http_common.dart
dnys1: It's the best way to debug browser tests, since the browser console logs does not output to the Dart console. I end up needing it every time I update HTTP code so I'd rather keep it 😅

HuiSF: lol OK, I used the same thing making tests not to finish.

Jordan-Nelson (approved):

Issue #1816, opened 2022-06-24
App crash when using datastore with api sync and internet is offline


yazoonic:
### Description

using datastore with api sync for cloud, when you disconnect internet and add entries to data store, it will work momentary then app crash 

I/amplify:aws-datastore(27749): Orchestrator transitioning from SYNC_VIA_API to LOCAL_ONLY
I/amplify:aws-datastore(27749): Setting currentState to LOCAL_ONLY
I/amplify:aws-datastore(27749): Stopping subscription processor.
I/amplify:aws-datastore(27749): Stopped subscription processor.
W/amplify:aws-datastore(27749): API sync failed - transitioning to LOCAL_ONLY.
W/amplify:aws-datastore(27749): DataStoreException{message=Error during subscription., cause=ApiException{message=Interrupted waiting for Cognito Userpools token., cause=java.lang.InterruptedException, recoverySuggestion=Sorry, we don't have a suggested fix for this error yet.}, recoverySuggestion=Evaluate details.}
W/amplify:aws-datastore(27749): 	at com.amplifyframework.datastore.appsync.AppSyncClient.lambda$subscription$3(AppSyncClient.java:331)
W/amplify:aws-datastore(27749): 	at com.amplifyframework.datastore.appsync.AppSyncClient$$ExternalSyntheticLambda1.accept(Unknown Source:4)
W/amplify:aws-datastore(27749): 	at com.amplifyframework.api.aws.SubscriptionOperation.lambda$null$1$com-amplifyframework-api-aws-SubscriptionOperation(SubscriptionOperation.java:87)
W/amplify:aws-datastore(27749): 	at com.amplifyframework.api.aws.SubscriptionOperation$$ExternalSyntheticLambda0.accept(Unknown Source:4)
W/amplify:aws-datastore(27749): 	at com.amplifyframework.api.aws.SubscriptionEndpoint.requestSubscription(SubscriptionEndpoint.java:131)
W/amplify:aws-datastore(27749): 	at com.amplifyframework.api.aws.SubscriptionOperation.lambda$start$2$com-amplifyframework-api-aws-SubscriptionOperation(SubscriptionOperation.java:77)
W/amplify:aws-datastore(27749): 	at com.amplifyframework.api.aws.SubscriptionOperation$$ExternalSyntheticLambda3.run(Unknown Source:2)
W/amplify:aws-datastore(27749): 	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:462)
[...]

PR #2549, opened 2023-01-06
fix(datastore): prevent unhandled exception crashing App rebuilding s…

HuiSF:
…ync expression

*Issue #, if available:*

Related: https://github.com/aws-amplify/amplify-flutter/issues/1816

*Description of changes:*

There is reported that a unhandled exception causes App to crash on Android while DataStore building sync expression for amplify-android implementation from Dart implementation. I haven't figure out why the latch await would be interrupted, at least we can handle this exception thrown from `latch.await()` to prevent App from crashing. 

### DIFF
--- a/packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
+++ b/packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
@@ -537,7 +537,14 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
                             }
                         )
                     }
-                    latch.await()
+                    try {
+                        latch.await()
+                    } catch (e: InterruptedException) {
+                        LOG.error(
+                            "Failed to resolve query predicate due to ${e}. Reverting to original query " +
+                                    "predicate."
+                        )
+                    }
                     resolvedQueryPredicate
                 }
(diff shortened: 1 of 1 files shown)


dnys1 (approved):

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -537,7 +537,14 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
                             }
                         )
                     }
-                    latch.await()
+                    try {
+                        latch.await()
+                    } catch (e: InterruptedException) {
```
dnys1, line 542: Is this the only exception we care about? Should this be `catch (e: Exception)`?

dnys1 (dismissed):

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
HuiSF: Yea `await` throws only `InterruptedException`.

Jordan-Nelson (dismissed):

dnys1 (approved):

Jordan-Nelson (approved):

PR #2554, opened 2023-01-09
chore(core): Pub Score Improvements

Equartey:
*Description of changes:*

- Removes Deprecated property `loginMechanism` & `loginMechanisms`
- Regenerates files with `deprecated_member_use_from_same_package` lint ignore


packages/amplify/amplify_flutter/example/integration_test/config_test.dart
```diff
@@ -71,7 +71,6 @@ const amplifyconfig = ''' {
                 "Auth": {
                     "Default": {
                         "authenticationFlowType": "USER_SRP_AUTH",
-                        "loginMechanisms": [],
```
Jordan-Nelson, line 74: *Not in scope for your PR*

@dnys1 - I see this was added in https://github.com/aws-amplify/amplify-flutter/pull/1011. I assume these values are either fake, or are from an app that no longer exists. We should make that clear (a comment would be fine) because this could set off some false alarms.

dnys1: Yeah, good point. I'll clean this all up in a separate PR.

packages/amplify_core/example/readme.md
```diff
@@ -0,0 +1,3 @@
+# amplify_core
+
+This is an internal package containing shared types and methods for use in amplify packages.
```
dnys1, line 3: Capitalize filename to `README.md` to match others.

I would change the text to be closer to what's in README as well:

```suggestion
A core package containing base types and utilities shared across Amplify packages. 
```

dnys1 (approved):

Issue #2548, opened 2023-01-06
Model helpers cannot covert the "query by model" into appropriate filter 


filippomenchini:
### Description

Given this schema:
```graphql
type Blog {
    id: ID!
    title: String!
    posts: [Post!] @hasMany(indexName: "postsByBlog", fields: ["id"])
}

type Post {
    id: ID!
    blogID: ID! @index(name: "postsByBlog")
    blog: Blog! @belongsTo(fields: ["userID"])
}
```

After launching `amplify codegen models` and trying to query `Post`'s related to a `Blog` an error is returned:
```
The variables input contains a field name 'blog' that is not defined for input object type 'ModelPostFilterInput'
```

Mentioned this bug [here](https://discord.com/channels/705853757799399426/1060978083538161774).

### Categories

- [ ] Analytics
- [ ] API (REST)
- [X] API (GraphQL)
- [ ] Auth
- [ ] Authenticator
- [ ] DataStore
- [ ] Storage

### Steps to Reproduce

1. Create the following schema:
```graphql
type Blog {
    id: ID!
    title: String!
    posts: [Post!] @hasMany(indexName: "postsByBlog", fields: ["id"])
}

type Post {
    id: ID!
    blogID: ID! @index(name: "postsByBlog")
    blog: Blog! @belongsTo(fields: ["userID"])
}
```
2. Create a `Blog` instance and some `Post` instances that reference the `Blog` instance.
3. Try to query posts linked to the previously created blog:
```dart
final request =
    ModelQueries.list(Post.classType, where: Post.BLOG.eq(blogId));
final response = await Amplify.API.query(request: request).response;
final data = response.data?.items ?? <Post?>[];
```
4. You should get the following error:
```
The variables input contains a field name 'blog' that is not defined for input object type 'ModelPostFilterInput'
```

### Platforms

- [X] iOS
- [X] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Environment

```bash
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.3.10, on macOS 13.0.1 22A400 darwin-arm, locale it-IT)
[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.0-rc2)
[✓] Xcode - develop for iOS and macOS (Xcode 14.2)
[✓] Android Studio (version 2021.1)
[✓] VS Code (version 1.74.2)
[✓] Connected device (2 available)
[...]

PR #2559, opened 2023-01-09
fix(api): model helpers use targetNames in schemas with CPK enabled

ragingsquirrel3:
Addresses https://github.com/aws-amplify/amplify-flutter/issues/2548

With the latest version of the CLI, new backends have CPK enabled by default. This changes the output of codegen in a way that breaks model helpers. Model helpers look for a field `.targetName` on schema fields in a few places when getting info from related models.

This PR does not add full support for CPK, just keeps model helpers working with new versions of CLI.

### DIFF
--- a/packages/amplify_test/lib/test_models/Address.dart
+++ b/packages/amplify_test/lib/test_models/Address.dart
@@ -22,7 +22,8 @@
 import 'package:amplify_core/amplify_core.dart';
 import 'package:flutter/foundation.dart';
 
-/// This is an auto generated class representing the Address type in your schema.
+
+/** This is an auto generated class representing the Address type in your schema. */
 @immutable
 class Address {
   final String? _line1;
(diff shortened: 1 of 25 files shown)


packages/amplify_test/lib/test_models/Address.dart
```diff
@@ -34,183 +35,159 @@ class Address {
...
       return _line1!;
-    } catch (e) {
-      throw AmplifyCodeGenModelException(
-          AmplifyExceptionMessages
-              .codeGenRequiredFieldForceCastExceptionMessage,
-          recoverySuggestion: AmplifyExceptionMessages
-              .codeGenRequiredFieldForceCastRecoverySuggestion,
-          underlyingException: e.toString());
+    } catch(e) {
+      throw new AmplifyCodeGenModelException(
```
dnys1, line 39: These changes don't seem correct

HuiSF: This is what codegen generated I think, we usually reformat these files.
Codegen code definitely needs some clean up...

packages/api/amplify_api/lib/src/graphql/graphql_request_factory.dart
```diff
@@ -207,8 +207,10 @@ class GraphQLRequestFactory {
 
     // e.g. { 'name': { 'eq': 'foo }}
     if (queryPredicate is QueryPredicateOperation) {
+      final association = schema.fields?[queryPredicate.field]?.association;
+      // TODO(ragingsquirrel3): Change key logic when supporting CPK.
       final associatedTargetName =
-          schema.fields?[queryPredicate.field]?.association?.targetName;
+          association?.targetName ?? association?.targetNames?.first;
```
HuiSF, line 213: mega nit: can be  `association?.targetNames?.first ?? association?.targetName` as the `targetNames` will be the default, `targetName` is the fallback for backwards compatibility.

HuiSF (approved):

packages/amplify_test/lib/test_models/Address.dart
ragingsquirrel3: I just ran the formatter on all these so hopefully that cleans up the diff.

HuiSF (approved):

dnys1 (approved):

dnys1 (approved):

HuiSF (approved):

PR #2561, opened 2023-01-10
ci(repo): Enable DDC/native workflows in PRs

dnys1:
- Missing GITHUB_TOKEN for Firefox tests
- Enables DDC/native workflows to run on PRs by default

### DIFF
--- a/.github/workflows/dart_ddc.yaml
+++ b/.github/workflows/dart_ddc.yaml
@@ -63,6 +63,8 @@ jobs:
       - name: Setup Firefox
         if: ${{ matrix.browser == 'firefox' }}
         uses: ./.github/composite_actions/setup_firefox
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Bootstrap
         id: bootstrap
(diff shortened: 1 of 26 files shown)


.github/workflows/dart_ddc.yaml
```diff
@@ -63,6 +64,8 @@ jobs:
       - name: Setup Firefox
         if: ${{ matrix.browser == 'firefox' }}
         uses: ./.github/composite_actions/setup_firefox
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
ragingsquirrel3, line 68: question: Does the token used here require some specific permission? Seems like consumer jobs set to `read-all` so I was just curious if there was some specific permission needed for this API? no change needed in PR

ragingsquirrel3 (approved):
approved w small question

.github/workflows/dart_ddc.yaml
dnys1: No special permissions required. It's just to let GitHub know who we are when making a request so we don't get pooled into the IP-based rate limiting bucket.

HuiSF (approved):

NikaHsn (approved):

PR #2564, opened 2023-01-10
ci(repo): Fix integration test runner

dnys1:
Broke due to changes in #2068.

### DIFF
--- a/.github/workflows/amplify_integration_tests.yaml
+++ b/.github/workflows/amplify_integration_tests.yaml
@@ -30,6 +30,7 @@ jobs:
       - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # 3.1.0
         with:
           persist-credentials: false
+          submodules: true
 
       # Flutter requires Java 11 to build android apps with Gradle.
       - uses: actions/setup-java@de1bb2b0c5634f0fc4438d7aa9944e68f9bf86cc # 3.6.0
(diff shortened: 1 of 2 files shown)


Jordan-Nelson (approved):

HuiSF (approved):

.github/workflows/amplify_integration_tests.yaml
```diff
@@ -30,6 +30,7 @@ jobs:
       - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # 3.1.0
         with:
           persist-credentials: false
+          submodules: true
```
ragingsquirrel3, line 33: question: Is this needed only for android or for other instances of checkout here for other platforms?

ragingsquirrel3:
Does this part need additional commands for submodule, libgit setup? https://github.com/aws-amplify/amplify-flutter/blob/next/.github/composite_actions/install_dependencies/action.yaml#L20

.github/workflows/amplify_integration_tests.yaml
dnys1: It's needed for `aft`

ragingsquirrel3: Makes sense, but this instance of checkout is only for android and all the other platforms in this file use checkout so was thinking they needed same change (as they also use aft in independent runner/job in matrix).

dnys1:
> Does this part need additional commands for submodule, libgit setup? https://github.com/aws-amplify/amplify-flutter/blob/next/.github/composite_actions/install_dependencies/action.yaml#L20

Ideally, no. Added a check to be safe.

ragingsquirrel3 (approved):

Issue #2493, opened 2022-12-15
[api] Multiple subscriptions cannot be established synchronously


dnys1:
In vNext API, multiple subscriptions cannot be established synchronously, e.g.

```dart
void setupSubscriptions() {
  final request1 = GraphQLRequest(...);
  final subscription1 = Amplify.API.subscribe(request1).listen(...);
  final request2 = GraphQLRequest(...);
  final subscription2 = Amplify.API.subscribe(request2).listen(...);
}
```

Results in the following error:

```
E/flutter (14866): [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: 'package:amplify_api/src/graphql/web_socket/blocs/web_socket_bloc.dart': Failed assertion: line 211 pos 7: '_currentState is ConnectingState': We should never receive an connection ack message while not connecting.
package:amplify_api/…/blocs/web_socket_bloc.dart:211
E/flutter (14866): #0      _AssertionError._doThrowNew (dart:core-patch/errors_patch.dart:51:61)
E/flutter (14866): #1      _AssertionError._throwNew (dart:core-patch/errors_patch.dart:40:5)
E/flutter (14866): #2      WebSocketBloc._connectionAck
package:amplify_api/…/blocs/web_socket_bloc.dart:211
E/flutter (14866): <asynchronous suspension>
E/flutter (14866):
```

dnys1:
This has been fixed in v1.0.0-next.5

PR #2567, opened 2023-01-11
fix(api): Web Socket Multiple Requests Synchronously

Equartey:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-flutter/issues/2493

*Description of changes:*
This fixes the respective error when trying to have multiple subscriptions established synchronously. 


packages/api/amplify_api/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart
```diff
@@ -206,9 +206,13 @@ class WebSocketBloc with AWSDebuggable, AmplifyLoggerMixin {
   Stream<WebSocketState> _connectionAck(
     ConnectionAckMessageEvent event,
   ) async* {
+    // Connection is already established
+    if (_currentState is ConnectedState) {
```
ragingsquirrel3, line 210: spoke offline, sounds like we are looking to a way to prevent this condition and only do 1 connection init in this scenario so we don't have to ignore a message like this (would only get 1 connection ack).

dnys1: Yeah this would appear to be creating multiple AppSync connections

dnys1 (commented):
The addition of a new state is fine, although it adds significant complexity IMO since now the initialization flow is broken down even further.

The crux of the issue is simply that two `RegisterEvent`s are enqueued before the first `InitEvent`. This is because in `subscribe`, we call `add(RegisterEvent(...))` which effectively schedules the `_registration` function to run in the next event loop. But this could all be avoided by running `_registration` directly instead of having an event for it.

Since `_registration` can never throw, and since it never `yield`s anything, it can safely be called in a synchronous context in this way. The function could be made into a `void _registration(RegisterEvent event)` to denote this further (and removing async/await) or you can use the `LazyStream` primitive from `package:async` so you can call `await _registration` in its constructor.

I think I would prefer this strategy since it will save us an event type and a state type which don't seem to add much structure and are potentially confusing.

dnys1:
A similar issue could happen when calling `add(InitEvent())` multiple times.

I think the simplest way to avoid this is by not adding InitEvent deep inside the logic of the class. Just call it when instantiating the bloc:

```dart
return _webSocketBlocPool[endpoint.name] ??= createWebSocketBloc(endpoint)
      ..add(const InitEvent())
      ..stream.listen((event) {
        _emitHubEvent(event);

        if (event is PendingDisconnect) {
          _webSocketBlocPool.remove(endpoint.name);
        }
      });
```

packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
```diff
@@ -105,6 +105,66 @@ void main({bool useExistingTestUser = false}) {
...
+
+            final stream1 = Amplify.API.subscribe(
+              subscriptionRequest1,
+              onEstablished: expectAsync0(readyCompleter.complete),
+            );
+            stream1.listen(
+              ((event) {
+                if (event.data?.name == name) {
+                  final blogFromEvent = event.data;
+                  expect(blogFromEvent?.name, equals(name));
```
ragingsquirrel3, line 132: nit: I think this is guaranteed to be true bc of the if statement it's inside. Is this added just so the test has an expect? If so, maybe this could be removed and the expect could be added to the final lines around the completer completing.

packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
```diff
@@ -105,6 +105,66 @@ void main({bool useExistingTestUser = false}) {
...
+                  dataCompleter.complete(event.data);
+                }
+              }),
+              onError: (Object e) => fail('Error in subscription stream: $e'),
+            );
+
+            final subscriptionRequest2 = ModelSubscriptions.onCreate(
```
ragingsquirrel3, line 139: nit: Could the 2nd request be `onUpdate` so the use case makes a little more sense? As written, both these streams have the same output so kinda strange use case to test (though it does effectively test the bug this PR is fixing).

packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
```diff
@@ -105,6 +105,66 @@ void main({bool useExistingTestUser = false}) {
...
+              onError: (Object e) => fail('Error in subscription stream: $e'),
+            );
+
+            await readyCompleter.future;
+            await readyCompleter2.future;
+
+            await addBlog(name);
+
+            await dataCompleter.future;
+            await dataCompleter2.future;
```
ragingsquirrel3, line 165:
```suggestion
            await expectLater(dataCompleter.future, completes);
            await expectLater(dataCompleter2.future, completes);
```

packages/api/amplify_api_dart/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart
```diff
@@ -264,9 +264,7 @@ class WebSocketBloc with AWSDebuggable, AmplifyLoggerMixin {
   // Init connection and add channel events to the event stream.
   Stream<WebSocketState> _init(InitEvent event) async* {
     assert(
-      _currentState is DisconnectedState ||
-          _currentState is ConnectingState ||
-          _currentState is ReconnectingState,
+      _currentState is InitializingState || _currentState is ReconnectingState,
```
ragingsquirrel3, line 267: question: Am I correct in understanding we no longer expect to get here while in `ConnectingState`? Seems like this is a branch we removed from here but seems like login in _registration prevents it.

ragingsquirrel3 (approved):

packages/api/amplify_api_dart/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart
Equartey: Discussed offline.

packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
Equartey: Swapped to `onDelete` for continuity.

packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
Equartey: Yah I think this copy-paste artifact is from another test. Removed.

packages/api/amplify_api_dart/test/util.dart
```diff
@@ -112,6 +114,12 @@ const mockAckMessage = {
   'payload': {'connectionTimeoutMs': 300000}
 };
 
+final isAIntegrationTestPost = isA<Post>().having(
```
ragingsquirrel3, line 117: nit: not an integration test post, maybe rename this `isATestPost`?

packages/api/amplify_api_dart/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart
```diff
@@ -121,7 +121,7 @@ class WebSocketBloc with AWSDebuggable, AmplifyLoggerMixin {
     final subBloc = _saveRequest<T>(event);
 
     // Add subscribe event to [WebSocketBloc]
-    add(RegistrationEvent(event.request));
+    _registration(RegistrationEvent(event.request));
```
ragingsquirrel3, line 124: nit: Could we change the signature of `_registration` (maybe rename `registerSubscriptionRequest`) to not take an event and just take the request itself? That way we could remove the `RegistrationEvent` class definition as it's no longer one of the events for the state machine, just a method used internally now.

packages/api/amplify_api_dart/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart
```diff
@@ -383,18 +381,11 @@ class WebSocketBloc with AWSDebuggable, AmplifyLoggerMixin {
 
   /// Sends registration message on ws channel when connected
   /// else init's connection
```
ragingsquirrel3, line 383: nit: Does it make sense to remove `else init's connection` from this comment as I don't think it does that anymore here?

dnys1 (approved):
LGTM!

Issue #2510, opened 2022-12-22
Enable transfer acceleration with S3


abdallahshaban557:
Enable transfer acceleration similar to the Amplify library for JavaScript

https://docs.amplify.aws/lib/storage/transfer-acceleration/q/platform/js/

HuiSF:
Implementation of transfer acceleration has merged in and released with the latest dev-preview release.
Closing as completed.

PR #2569, opened 2023-01-11
feat(storage): allow configuring transfer acceleration 

HuiSF:
*Issue #, if available:*

https://github.com/aws-amplify/amplify-flutter/issues/2510

*Description of changes:*

1. Update infra storage stack to enable transfer acceleration for the buckets
2. Add `useAcceleration` param to corresponding operation options interface
3. Update s3 service to create `S3Client` instance per each operation and not using a singleton (S3Client configuration may changes per operation, per use case)
4. Update `DateTime` comparison in unit tests

### DIFF
--- a/packages/aws_common/lib/src/io/aws_file_platform_html.dart
+++ b/packages/aws_common/lib/src/io/aws_file_platform_html.dart
@@ -125,18 +125,24 @@ class AWSFilePlatform extends AWSFile {
           return externalContentType;
         }
 
-        String blobType;
+        String? blobType;
 
         final file = _inputFile ?? _inputBlob;
+        final path = super.path;
+
         if (file != null) {
           blobType = file.type;
-        } else {
+        } else if (path != null) {
           blobType = (await _resolvedBlob).type;
         }
 
         // on Web blob.type may return an empty string
         // https://developer.mozilla.org/en-US/docs/Web/API/Blob/type#value
-        return blobType.isEmpty ? null : blobType;
+        if (blobType != null) {
+          return blobType.isEmpty ? null : blobType;
+        }
+
+        return blobType;
       });
 
   Future<Blob> get _resolvedBlob async {
--- a/packages/storage/amplify_storage_s3/example/integration_test/main_test.dart
+++ b/packages/storage/amplify_storage_s3/example/integration_test/main_test.dart
@@ -328,7 +328,7 @@ void main() async {
               getProperties: true,
               bytesRange: S3DataBytesRange(
                 start: start,
-                end: 5 * 1024 + 12,
+                end: end,
               ),
             ),
           ).result;
--- a/packages/storage/amplify_storage_s3/example/integration_test/main_test.dart
+++ b/packages/storage/amplify_storage_s3/example/integration_test/main_test.dart
@@ -17,6 +17,8 @@ import 'package:integration_test/integration_test.dart';
 import 'package:path/path.dart' as path;
 
 import 'content_type_infer/content_type_infer.dart';
+import 'transfer_acceleration/test_acceleration_config.dart';
+import 'transfer_acceleration/transfer_acceleration.dart';
 
 class CustomPrefixResolver implements S3PrefixResolver {
   const CustomPrefixResolver();
@@ -438,6 +440,27 @@ void main() async {
         testContentTypeInferTest(
           smallFileBytes: testBytes,
           largeFileBytes: testLargeFileBytes,
+        );
+
+        testTransferAcceleration(
+          dataPayloads: [
+            TestTransferAccelerationConfig(
+              targetKey: 'transfer-acceleration-datapayload-${uuid()}',
+              targetAccessLevel: StorageAccessLevel.guest,
+              uploadSource: S3DataPayload.bytes(
+                testBytes,
+              ),
+              referenceBytes: testBytes,
+            ),
+          ],
+          awsFiles: [
+            TestTransferAccelerationConfig(
+              targetKey: 'transfer-acceleration-awsfile-${uuid()}',
+              targetAccessLevel: StorageAccessLevel.private,
+              uploadSource: AWSFile.fromData(testLargeFileBytes),
+              referenceBytes: testLargeFileBytes,
+            )
+          ],
         );
       });
 
--- a/packages/storage/amplify_storage_s3_dart/example/bin/example.dart
+++ b/packages/storage/amplify_storage_s3_dart/example/bin/example.dart
@@ -183,6 +183,7 @@ Future<void> getUrlOperation() async {
   final accessLevel = promptStorageAccessLevel(
     'Choose the storage access level associated with the object: ',
   );
+  final useAccelerateEndpoint = promptUseAcceleration();
 
   final s3Plugin = Amplify.Storage.getPlugin(AmplifyStorageS3Dart.pluginKey);
   final getUrlOperation = s3Plugin.getUrl(
@@ -193,6 +194,7 @@ Future<void> getUrlOperation() async {
         minutes: 10,
       ),
       checkObjectExistence: true,
+      useAccelerateEndpoint: useAccelerateEndpoint,
     ),
   );
 
@@ -253,6 +255,8 @@ Future<void> downloadFileOperation() async {
   final destinationPath = prompt(
     'Enter the destination file path (ensure the file path is writable): ',
   );
+  final useAccelerateEndpoint = promptUseAcceleration();
+
   final localFile = AWSFile.fromPath(destinationPath);
 
   final s3Plugin = Amplify.Storage.getPlugin(AmplifyStorageS3Dart.pluginKey);
@@ -262,6 +266,7 @@ Future<void> downloadFileOperation() async {
     options: S3DownloadFileOptions(
       getProperties: true,
       accessLevel: accessLevel,
+      useAccelerateEndpoint: useAccelerateEndpoint,
     ),
     onProgress: onTransferProgress,
   );
@@ -330,6 +335,7 @@ Future<void> uploadFileOperation() async {
   final file = AWSFile.fromPath(filePath);
 
   final option = prompt('Upload size ${await file.size}, continue? (Y/n): ');
+  final useAccelerateEndpoint = promptUseAcceleration();
 
   if (option.toLowerCase() != 'y') {
     stdout.writeln('Upload canceled.');
@@ -347,6 +353,7 @@ Future<void> uploadFileOperation() async {
       metadata: {
         'nameTag': nameTag,
       },
+      useAccelerateEndpoint: useAccelerateEndpoint,
     ),
   );
 
@@ -506,6 +513,17 @@ StorageAccessLevel promptStorageAccessLevel(String message) {
   }
 
   return accessLevel;
+}
+
+bool promptUseAcceleration() {
+  String input;
+
+  do {
+    input = prompt('Use transfer acceleration for this operation? (y/n): ')
+        .toLowerCase();
+  } while (input != 'y' && input != 'n');
+
+  return input == 'y';
 }
 
 Never exitError(Object error, [StackTrace? stackTrace]) {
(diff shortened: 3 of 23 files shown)


packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
```diff
@@ -562,6 +563,20 @@ class StorageS3Service {
...
+  }) =>
+      // dependencyManager.get() => S3Client is used for unit tests
+      _dependencyManager.get() ??
+      s3.S3Client(
+        region: _defaultRegion,
+        credentialsProvider: _credentialsProvider,
+        s3ClientConfig: smithy_aws.S3ClientConfig(
+          useAcceleration: useAcceleration,
+        ),
+        client: _httpClient,
```
ragingsquirrel3, line 577: question: Will this create a new httpClient for each s3 client? If so, would it be possible to reuse the HTTP client? I'm not sure about our HTTP client, but the package:http HTTP client has a lot of performance gains from reusing the same client on the same host bc if you create new ones, it has to do a new handshake for each client. Even if our HTTP client doesn't have that issue, might be nice to reuse.

packages/storage/amplify_storage_s3/example/integration_test/main_test.dart
```diff
@@ -262,6 +265,7 @@ void main() async {
                   metadata: {
                     'filename': testObjectFileName3,
                   },
+                  useAcceleration: true,
```
ragingsquirrel3, line 268: question: Do we also need same test with accel set to `false` to test to keep coverage for use without it?

packages/storage/amplify_storage_s3_dart/lib/src/model/s3_download_data_options.dart
```diff
@@ -53,4 +58,10 @@ class S3DownloadDataOptions extends StorageDownloadDataOptions {
   /// Whether to retrieve properties for the downloaded object using the
   /// `getProperties` API.
   final bool getProperties;
+
+  /// {@template storage.amplify_storage_s3.transfer_acceleration}
```
ragingsquirrel3, line 62: nit/question: Nice comment, would it be possible to have same comment for download_file, upload_file/data, etc...?

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
```diff
@@ -218,8 +212,15 @@ class StorageS3Service {
...
 
-    final host =
-        '$_defaultBucket.${_getS3EndpointHost(region: _defaultRegion)}';
+    var host = '$_defaultBucket.${_getS3EndpointHost(region: _defaultRegion)}';
+
+    if (options.useAcceleration) {
+      // https: //docs.aws.amazon.com/AmazonS3/latest/userguide/transfer-acceleration-getting-started.html
+      host = host
+          .replaceFirst(RegExp('$_defaultRegion\\.'), '')
+          .replaceFirst(RegExp(r'\.s3\.'), '.s3-accelerate.');
```
HuiSF, line 221: The logic is from here:

https://github.com/aws-amplify/amplify-flutter/blob/fc6736512e7fbd595a3b376f3d7f2b3199e47d06/packages/storage/amplify_storage_s3_dart/lib/src/sdk/src/s3/operation/get_object_operation.dart#L393-L396

Could make this logic a util function in the sdk and export it to use here?

ragingsquirrel3 (commented):
A few small questions/nits, but I wanted to make sure I understand if HTTP client needs to be reused, though possible I'm missing something there.

packages/storage/amplify_storage_s3/example/integration_test/main_test.dart
HuiSF: The default value is false, the tests are not using `useAcceleration: true` should cover the case `useAcceleration: false`.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
HuiSF: Good to know, I think for now it's safe to use the same HTTP client instance, will update.

HuiSF: Updated to use a singleton of HTTP client with in the same instance of S3Service wrapper.

dnys1: This does seem pretty inefficient. Seems like we should have a max of 2 S3Clients floating around.

```dart
_dependencyManager.get() ?? (useAcceleration ? _accelerateClient : _normalClient);

late final _accelerateClient = s3.S3Client(...);
late final _normalClient = s3.S3Client(...);
```

dnys1: But actually it looks like this should be configurable on the client methods since it's configurable on the operation currently. So I can put up a PR for that at some point.

packages/storage/amplify_storage_s3/example/integration_test/main_test.dart
```diff
@@ -498,16 +506,20 @@ void main() async {
             );
           });
 
-          testWidgets('get url of object with access level guest',
-              (WidgetTester tester) async {
+          testWidgets(
```
dnys1, line 509: To Travis's earlier point, why are we replacing existing tests instead of adding new ones? It seems like you've flattened tests into guest+no acceleration / authorized+acceleration. Is there any harm making this 2 different dimensions so we get a full matrix of acceleration/authorization cases?

HuiSF: Yea, can just add a separate set upload-download-delete test case on accelerate endpoint for clarity, I changed existing tests as acceleration doesn't involve or is not involved with auth.

dnys1: That's fine I would just add it as a separate test, then, so:
- guest
- authorized
- accelerate

packages/storage/amplify_storage_s3_dart/lib/src/model/s3_download_data_options.dart
HuiSF: This is a good thought.

Developers are interacting with the Storage APIs via the Storage Category API interface, the documentation should be added here. The category interface, however, hides the S3 specifics, i.e. transfer acceleration is S3 unique. So I think the discoverability of this configuration would depend on Amplify library documentation and Flutter/Dart API docs on pub.dev. What do you think?

packages/storage/amplify_storage_s3/example/integration_test/main_test.dart
HuiSF: Added dedicated group to test.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
HuiSF: Thanks for updating the SDK, no it doesn't need to create multiple instances of `S3Client`.

packages/aws_common/lib/src/io/aws_file_platform_html.dart
```diff
@@ -125,18 +125,25 @@ class AWSFilePlatform extends AWSFile {
...
 
         final file = _inputFile ?? _inputBlob;
         if (file != null) {
           blobType = file.type;
-        } else {
+        }
+
+        final path = super.path;
+        if (path != null) {
           blobType = (await _resolvedBlob).type;
```
dnys1, line 137: Are you overriding blobType on purpose? Are these branches mutually exclusive? Can you add some notes here?

packages/storage/amplify_storage_s3/example/integration_test/main_test.dart
```diff
@@ -446,6 +448,27 @@ void main() async {
           smallFileBytes: testBytes,
           largeFileBytes: testLargeFileBytes,
         );
+
+        testTransferAcceleration(
+          dataPayloads: [
+            Tuple4(
```
dnys1, line 454: Can you create a class instead? Sorry 😅 In Dart 3, we'll get records, but until then I can only guess at what these params mean

packages/storage/amplify_storage_s3/example/integration_test/transfer_acceleration.dart
```diff
@@ -0,0 +1,154 @@
...
+import 'package:http/http.dart' as http;
+import 'package:tuple/tuple.dart';
+
+void testTransferAcceleration({
+  List<Tuple4<String, StorageAccessLevel, S3DataPayload, List<int>>>
+      dataPayloads = const [],
+  List<Tuple4<String, StorageAccessLevel, AWSFile, List<int>>> awsFiles =
+      const [],
+}) {
+  if (dataPayloads.isNotEmpty) {
```
dnys1, line 13: Why would these be empty? Can probably just remove this

packages/storage/amplify_storage_s3_dart/example/bin/example.dart
```diff
@@ -508,6 +515,17 @@ StorageAccessLevel promptStorageAccessLevel(String message) {
   return accessLevel;
 }
 
+bool promptUseAcceleration() {
+  String input;
+
+  do {
+    input = prompt('Use transfer acceleration for this operation? (Y/n): ')
+        .toLowerCase();
+  } while (input != 'y' && input != 'n');
```
dnys1, line 524: `(Y/n)` implies that an empty response == 'Y'.
`(y/n)` would imply no default value.

packages/storage/amplify_storage_s3_dart/lib/src/model/s3_client_config_copy.dart
```diff
@@ -0,0 +1,24 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import 'package:aws_signature_v4/aws_signature_v4.dart' as sigv4;
+import 'package:smithy_aws/smithy_aws.dart' as smithy_aws;
+
+/// Extension adds a copyWith method to [smithy_aws.S3ClientConfig] allowing
+/// to create a new instance of [smithy_aws.S3ClientConfig] with overrides.
+extension S3ClientConfigCopy on smithy_aws.S3ClientConfig {
+  /// Returns a new [smithy_aws.S3ClientConfig] with overrides.
+  smithy_aws.S3ClientConfig copyWith({
```
dnys1, line 11: Can you add this to `S3ClientConfig` instead?

dnys1 (approved):
Thanks!

packages/aws_common/lib/src/io/aws_file_platform_html.dart
HuiSF: Ugh didn't commit the final change on this file. Will update.

packages/storage/amplify_storage_s3/example/integration_test/transfer_acceleration.dart
HuiSF: Ah yea wanted to make it generic, but I guess it's unnecessary for this test cases.

packages/aws_common/lib/src/io/aws_file_platform_html.dart
HuiSF: Updated the logic resolve `blobType` from `File (Blob)` or from `Blob` resolved from path. The overriding was incorrect.

dnys1: Can you add another test for this since it wasn't caught before?

packages/storage/amplify_storage_s3/example/integration_test/main_test.dart
HuiSF: Updated

packages/storage/amplify_storage_s3_dart/lib/src/model/s3_client_config_copy.dart
HuiSF: Updated 👍

packages/aws_common/lib/src/io/aws_file_platform_html.dart
HuiSF: Added 👍

ragingsquirrel3 (approved):

PR #2579, opened 2023-01-13
fix(api): model helpers use targetNames in schemas with CPK enabled

ragingsquirrel3:
Cherry-picks https://github.com/aws-amplify/amplify-flutter/commit/809190ced5745ec270d2880cc736d5df265a30d5 off main to address https://github.com/aws-amplify/amplify-flutter/issues/2548 on next. I had to add another commit to clean up some conflicts in the codegen test models because they got moved to the Dart package on next and I apparently did not handle the move/renames perfectly.


packages/api/amplify_api_dart/test/test_models/Blog.dart
```diff
@@ -16,64 +16,6 @@ import 'ModelProvider.dart';
 /// This is an auto generated class representing the Blog type in your schema.
 @immutable
 class Blog extends Model {
```
ragingsquirrel3, line 18: callout: For codegen models in this dir I ended up copying the models from main like https://github.com/aws-amplify/amplify-flutter/blob/main/packages/amplify_test/lib/test_models/Blog.dart except for the the imports/header bc the codegen files have a Flutter dep so to use them in Dart tests I have manually changed the imports. I tried running `dart format` on them here to format them but there were no changes. Basically, maintaining these here is a pain and the diff is kinda big here, open to ideas to reduce/improve.

HuiSF (approved):

dnys1 (approved):

PR #2600, opened 2023-01-25
feat(analytics): Use indexedDB on Web

fjnoyp:
*Issue #, if available:*

*Description of changes:*

Use IndexDB to store cached Analytics Events on web. 


packages/aws_common/lib/src/js/common.dart
```diff
@@ -374,3 +374,14 @@ Object? dartify(Object? o) {
   }
   return o;
 }
+
+/// Convert a dart map to a JS object
+Object mapToJSObj(Map<Object, Object> a) {
```
fjnoyp, line 379: @dnys1 I tried using the _convertToJs but that didn't work so I created this method instead.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/string_database/dart_string_database.stub.dart
```diff
@@ -0,0 +1,41 @@
...
+
+import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_client/string_database/string_database.dart';
+import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_client/string_database/string_database_element.dart';
+import 'package:aws_common/aws_common.dart';
+import 'package:drift/drift.dart';
+
+/// {@template amplify_analytics_pinpoint_dart.dart_string_database}
+/// Stores strings using IndexDB for web and Drift for all other platforms
+/// {@endtemplate}
+class DartStringDatabase implements StringDatabase, Closeable {
```
dnys1, line 14: I don't get what a "string" database is... how about `EventDatabase` for the interface and `PlatformEventDatabase` for this wrapper class?

I feel like these methods should accept `Event` objects as well and be named appropriately, e.g. `saveEvent`, `deleteEvents`, `listEvents`

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/string_database/index_db/index_db_string_db.dart
```diff
@@ -0,0 +1,241 @@
...
+
+  @override
+  String get runtimeTypeName => 'IndexDbStringDb';
+
+  @override
+  FutureOr<void> addString(String string) async {
+    _inMemoryDb.addString(string);
+
+    try {
+      return _indexDb?.addString(string);
```
dnys1, line 88: `addString` needs to be awaited for the try/catch to work. Applies to all try/catches throughout where the body of `try` is a future.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/string_database/index_db/index_db_string_db.dart
```diff
@@ -0,0 +1,241 @@
...
+        store.deleteByKeyRange(
+          IDBKeyRange.bound(start, previous),
+        );
+        start = id;
+      }
+      previous = id;
+    }
+    await store
+        .deleteByKeyRange(IDBKeyRange.bound(start, idsToDelete.last))
+        .future;
```
dnys1, line 233: This logic makes no sense to me. If `idsToDelete = [1, 2, 3, 4, 5]`, this will result in the following calls:

```
deleteByKeyRange (1, 1)
deleteByKeyRange (1, 5)
```

I had to work this out on paper it's very confusing. Please clean up and add tests for this logic.

packages/analytics/amplify_analytics_pinpoint_dart/test/string_database_test.dart
```diff
@@ -0,0 +1,145 @@
+@TestOn('browser')
```
dnys1, line 1: Below the license header

packages/aws_common/lib/src/js/common.dart
dnys1: What didn't work? This is basically the same. Also, it should be marked `@internal` if it's not private.

packages/analytics/amplify_analytics_pinpoint_dart/test/string_database_test.dart
```diff
@@ -0,0 +1,145 @@
...
+        }),
+      );
+      await db.deleteStrings(idsToDelete);
+      validateElements(
+        await db.getStrings(10),
+        ['1', '3', '5', '7', '9'],
+      );
+    });
+
+    test('Concurrent writes, deletes, reads of strings', () async {
```
dnys1, line 122: I have no idea what this is testing - but awaiting writes/deletes in a for loop will make these happen serially, not concurrently.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/string_database/dart_string_database.web.dart
```diff
@@ -0,0 +1,42 @@
...
+import 'dart:async';
+
+import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_client/string_database/index_db/index_db_string_db.dart';
+import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_client/string_database/string_database.dart';
+import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_client/string_database/string_database_element.dart';
+import 'package:aws_common/aws_common.dart';
+import 'package:drift/drift.dart';
+
+/// {@macro amplify_analytics_pinpoint_dart.dart_string_database}
+class DartStringDatabase implements StringDatabase, Closeable {
```
dnys1, line 13: Move the logic from `IndexStringDb` to here. There's way too much indirection going on.

dnys1 (commented):
Please rename `IndexDB` to `IndexedDB` throughout

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/string_database/index_db/index_db_string_db.dart
```diff
@@ -0,0 +1,241 @@
...
+    }
+  }
+
+  @override
+  FutureOr<List<StringDatabaseElement>> getStrings(int maxToGet) async {
+    return _inMemoryDb.getStrings(maxToGet);
+  }
+
+  /// Clear indexDb data
+  /// The autoincrement of indexDb id is not reset!
```
Jordan-Nelson, line 111: What does this comment mean? Is the `!` intentional? If there is something that you are trying to alert the consumer to, I might do `/// Warning: Something that requires attention` on a new line.

packages/analytics/amplify_analytics_pinpoint_dart/test/string_database_test.dart
fjnoyp: Successive is a better way to describe.  It's meant to test that successive operations still result in a proper state.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/string_database/dart_string_database.stub.dart
fjnoyp: Hey this is how things are working now. 

The `EventDatabase` you suggest is the current `EventStorageAdapter` which also accepts `Event` objects and already has `saveEvent` `retrieveEvents` and `deleteEvents` methods. 

This is how storage works:

`EventStorageAdapter.saveEvent(Event event)` -> 
`DartStringDatabase.addString`( event encoded as json string) 

The `DartStringDatabase` is an internal class that just handles the storage of strings.  `EventStorageAdapter` deals with converting stored `Event` as strings and retrieving those strings and turning them back to `Event.

packages/aws_common/lib/src/js/common.dart
fjnoyp: Crashes code with exception: 

>   DataCloneError: Failed to execute 'add' on 'IDBObjectStore': _installSpecializedAsCheck(object) {
>         var t1, testRti = this,
>           asFn = A._generalAsCheckImplementati...<omitted>... } could not be cloned.
>   string_database_test.dart.browser_test.dart.js 7793:31   Object.callMethod
>   string_database_test.dart.browser_test.dart.js 22904:67  <fn>
> 

Reason: 

https://github.com/dart-lang/sdk/issues/34130

image

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/string_database/index_db/index_db_string_db.dart
```diff
@@ -0,0 +1,241 @@
...
+      final currentId = (await indexDb.getStrings(1)).first.id;
+      await indexDb.deleteStrings([currentId]);
+      return currentId + 1;
+    } else {
+      return readElements.last.id;
+    }
+  }
+
+  /// {@macro amplify_analytics_pinpoint_dart.index_db_string_db}
+  static Future<IndexDbStringDb> getInstance() async {
```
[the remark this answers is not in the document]
fjnoyp: Yeah good call, per your and Dillon's comments I have switched to using the pattern in secure storage using the async memoizer.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/string_database/index_db/index_db_string_db.dart
fjnoyp: Thanks added 'warning' and remove !

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/string_database/index_db/index_db_string_db.dart
fjnoyp: @dnys1 I think the main issue was I missed an edge case.  So updated code to: 

    var previous = idsToDelete.first - 1;

I've added comments explaining how the logic works. 

This is how it works basically: 

We want to delete sequential id ranges.  We keep track of the start index of the range.  We iterate until the next id is no longer sequential. We then delete that range and continue.

fjnoyp: I've added a lot of extra tests to make sure it's right.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/string_database/index_db/index_db_string_db.dart
```diff
@@ -0,0 +1,241 @@
...
+    final isIndexDbSupported = await _checkIsIndexDbSupported();
+
+    if (isIndexDbSupported) {
+      final indexDb = _IndexedDBStorage();
+      final readElements = await indexDb.getAllStrings();
```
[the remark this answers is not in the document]
fjnoyp: Yes O(1) I reran the tests on different data set sizes (200, 500, 1000 cached events) to see performance scale. 

Yeah there were new insights thanks to Dillon regarding performance.  

The reads are actually very very fast.  The O(n) performance comes from add/inserts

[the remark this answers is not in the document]
fjnoyp: And yes I removed the cursor impl and switched to GetAll

packages/analytics/amplify_analytics_pinpoint_dart/test/string_database_test.dart
```diff
@@ -0,0 +1,145 @@
...
+      db = await IndexDbStringDb.getInstance();
+    }
+
+    void validateElements(
+      Iterable<StringDatabaseElement> readElements,
+      List<String> targetValues,
+    ) {
+      expect(readElements.length, targetValues.length);
+
+      var curCount = 0;
```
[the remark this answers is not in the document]
Jordan-Nelson: You could use elementAt().

```dart
for (var i = 0; i < readElements.length; i++) {
  expect(readElements.elementAt(i).string, i.toString());
}
```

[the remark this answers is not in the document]
fjnoyp: Nice, thanks.  Updated

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/string_database/dart_string_database.stub.dart
Jordan-Nelson: I think Dillon might be noting that the name doesn't help give context as to what this is/does. A "String Database" is not term people would be familiar with, nor does it really describe what this does or what it is intended to be used for.

EventDatabase would indicate it is used to store events. It sounds like you intentionally wanted to keep a level of abstraction between this DB and Analytics events though, which I think is fine.

This does indeed store Strings, but so do other storage mechanisms we have, such as secure_storage. What makes this different? This data store has some unique characteristics. It stores objects (that happen to be strings) in a First in, First out (FIFO) manner. The delete method is an exception here, as it accepts specific IDs to delete, rather than deleting the first N items.

Naming is hard. We aren't going to find a perfect name to describe this unique data store, but I think we can do better than StringDatabase/StringDatabaseElement.

Below is one option. In addition to the name change, I updated the method names, and changed the API signature of delete to take items, instead of just Ids. If `StoredEvent` were to extend/implement FIFODataStoreItem, you could just pass in the StoredEvents to deleteEvents, and let this class be concerned with how to delete those items. It would help maintain the level of abstraction / separation of concerns that I think you wanted achieve. 

```dart
/// A Data store that inserts and retrieves data in a FIFO (First In, First Out)
/// manner.
abstract class FIFODataStore {
  /// Insert an item into the store.
  FutureOr<void> insert(String item);

  /// Get the first [count] items.
  FutureOr<Iterable<FIFODataStoreItem>> get(int count);

  /// Delete the [items] from the store.
  FutureOr<void> delete(Iterable<FIFODataStoreItem> items);

  /// Delete all items in database.
  FutureOr<void> clear();

  /// Closes this database and releases associated resources.
  FutureOr<void> close();
}

/// An item stored in the [FIFODataStore].
class FIFODataStoreItem {
  const FIFODataStoreItem({required this.id, required this.value});

  /// The ID in the database.
  final int id;

  /// The value of the item.
  final String value;
}
```

packages/analytics/amplify_analytics_pinpoint_dart/test/string_database_test.dart
```diff
@@ -0,0 +1,197 @@
+// Web only - database regeneration causes multiple Drift database open warning
+
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+@TestOn('browser')
```
Jordan-Nelson, line 6: Why just on web?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/string_database/index_db/in_memory_string_db.dart
```diff
@@ -0,0 +1,54 @@
...
+class InMemoryStringDb implements StringDatabase {
+  /// {@macro amplify_analytics_pinpoint_dart.in_memory_string_db}
+  InMemoryStringDb({
+    List<StringDatabaseElement> initValues = const [],
+    int lastId = 0,
+  })  : _lastId = lastId,
+        _database = initValues;
+
+  int _lastId = 0;
+  final List<StringDatabaseElement> _database;
```
Jordan-Nelson, line 22: I mentioned in another comment that this store is essentially a queue, with some custom delete behavior. You could consider using `Queue` from dart:collection to hold the data for the in mem impl

packages/analytics/amplify_analytics_pinpoint_dart/test/string_database_test.dart
```diff
@@ -0,0 +1,197 @@
...
+        idsToDelete: [0, 1, 2, 3],
+      );
+
+      await db.clear();
+
+      await deleteAndValidateValues(
+        writeCount: 10,
+        idsToDelete: [6, 7, 8, 9],
+      );
+    });
```
Jordan-Nelson, line 126: The code in these tests is _very_ DRY, which usually is a good thing, but sometimes in tests it is not very desirable as it reduces the readability. Take a look at https://stackoverflow.com/questions/6453235/what-does-damp-not-dry-mean-when-talking-about-unit-tests

Good tests should be easy to ready, and clearly explain the expected behavior of the code, even if that results in some duplication.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/string_database/dart_string_database.stub.dart
HuiSF: Agree with above, I think naming should indicate what's the database capable of, but not what format it stores "things" in. TBH I was very confused by StringDatabase when I just started looking at this PR 😅

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/string_database/dart_string_database.web.dart
```diff
@@ -0,0 +1,112 @@
...
+
+  @override
+  Future<Iterable<StringDatabaseElement>> getStrings(int maxToGet) async {
+    await _init();
+
+    try {
+      return _database.getStrings(maxToGet);
+    } on Exception catch (e) {
+      _handleException(e);
+      return [];
```
HuiSF, line 80: Shouldn't we throw this exception here instead of swallowing it? This seems like a critical piece that analytics functionality depends on; if there is an exception here, it seems that we did something wrong in the library in the first place.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/string_database/index_db/indexed_db_adapter.dart
```diff
@@ -0,0 +1,149 @@
...
+
+  /// The name of the database
+  ///
+  /// Reference: https://www.w3.org/TR/IndexedDB/#name
+  String get databaseName => 'amplify_analytics_pinpoint';
+
+  /// The name of the object store
+  ///
+  /// Reference: https://www.w3.org/TR/IndexedDB/#object-store-name
+  final storeName = 'analytics_cached_events';
```
HuiSF, line 29: Does this need to be prefixed with `amplify`?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/string_database/index_db/indexed_db_adapter.dart
```diff
@@ -0,0 +1,149 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:async';
+import 'dart:js_util';
+
+import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_client/string_database/string_database.dart';
+import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_client/string_database/string_database_element.dart';
+import 'package:amplify_core/amplify_core.dart';
+// ignore: implementation_imports
+import 'package:aws_common/src/js/indexed_db.dart';
```
HuiSF, line 11: Should `indexed_db` be exported from `aws_common/aws_common.dart`?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/string_database/index_db/indexed_db_adapter.dart
```diff
@@ -0,0 +1,149 @@
...
+  late final Future<void> _databaseOpenEvent = _openDatabase();
+
+  late final IDBDatabase _database;
+
+  /// Checks for IDB availability and attempts to open the database.
+  Future<void> _openDatabase() async {
+    if (indexedDB == null) {
+      throw const InvalidStateException('IndexedDB is not available');
+    }
+    final openRequest = indexedDB!.open(databaseName, 1)
```
HuiSF, line 40: Is `indexedDB` a nullable property? if so
```suggestion
    final db = indexedDB;
    if (db == null) {
      throw const InvalidStateException('IndexedDB is not available');
    }
    final openRequest = db.open(databaseName, 1)
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/string_database/dart_string_database.stub.dart
fjnoyp: Thanks guys that sounds good, I've updated the PR to have the new naming scheme / API that Jordan suggested.

packages/analytics/amplify_analytics_pinpoint_dart/test/string_database_test.dart
fjnoyp: Oh wow I didn't know that before but it makes a late of sense.  Rewrote the unit test to be DAMP instead of DRY.

packages/analytics/amplify_analytics_pinpoint_dart/test/string_database_test.dart
fjnoyp: Rewrote tests and refactored logic to allow for non web.  Before it was due to memory caching changes and later due to issue with db common executor.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/string_database/dart_string_database.web.dart
fjnoyp: Indeed, per Jordan's comments above rewrote this logic to just throw exception.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/string_database/index_db/indexed_db_adapter.dart
fjnoyp: Hmm, I don't think so because the database it's in is already called: 

amplify_analytics_pinpoint

Would every store within that database also need to be prefix with amplify after?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/string_database/index_db/indexed_db_adapter.dart
fjnoyp: I'm not sure ... in the aws_common package none of the js subfiles are exported.  

SecureStorage which uses `indexed_db` also does the same thing. 

I'm not 100% sure why but I can imagine it's because the files in `js` are more private/internal.  While the other types in `aws_common` could feasibly be used by outside developers in a more generic use case.  

@Jordan-Nelson does that reasoning seem correct to you?

dnys1: Shouldn't be exported since they are experimental and will likely be replaced by https://github.com/dart-lang/web/ in the not-too-distant future.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/dart_queued_item_store.stub.dart
```diff
@@ -0,0 +1,46 @@
...
+import 'package:drift/drift.dart';
+import 'package:meta/meta.dart';
+
+/// {@template amplify_analytics_pinpoint_dart.dart_queued_item_store}
+/// Stores strings using IndexedDB for web and Drift for all other platforms
+/// {@endtemplate}
+class DartQueuedItemStore implements QueuedItemStore, Closeable {
+  /// {@macro amplify_analytics_pinpoint_dart.dart_queued_item_store}
+  // ignore: avoid_unused_constructor_parameters
+  DartQueuedItemStore(QueryExecutor? driftQueryExecutor) {
```
Jordan-Nelson, line 17: Can we avoid passing in something specific to Drift? I think we can keep all the drift impl details within the the Drift impl of DartQueuedItemStore.

I think there are a few things we can simplify now that we are not use Drift on web.

If you update this constructor to pass in the storage path (which you are already retrieving in the configure fn, where DartQueuedItemStore is instantiated) instead of the QueryExecutor , you can import `connect` from amplify_db_common_dart from within the Drift impl of DartQueuedItemStore to create the QueryExecutor.

I think you can also drop the dependency on amplify_db_common and just depend on amplify_db_common_dart as long as you are retrieving the storage path in configure (which you currently are). You can then remove the connect arg from the `AmplifyAnalyticsPinpointDart` constructor to simplify it a bit.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/dart_queued_item_store.web.dart
```diff
@@ -0,0 +1,97 @@
...
+
+    try {
+      return db.clear();
+    } on Exception catch (e) {
+      _handleException(e);
+      rethrow;
+    }
+  }
+
+  void _handleException(Exception e) {
```
Jordan-Nelson, line 88: I'm not sure I see the value in catching these just to add a log that isn't particularly helpful.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/drift/drift_queued_item_store.dart
```diff
@@ -30,26 +34,37 @@ class DriftDatabaseJsonStrings extends _$DriftDatabaseJsonStrings {
...
     final statement = (select(driftJsonStrings)
       ..orderBy([(v) => OrderingTerm.asc(v.id)])
       ..limit(maxToGet));
 
-    return statement.get();
+    final retrievedJsonStrings = await statement.get();
+    return retrievedJsonStrings.map(
+      (jsonString) => QueuedItem(
+        id: jsonString.id,
+        value: jsonString.jsonString,
```
Jordan-Nelson, line 53: Anytime you have code like `foo.foo` that is a good indication that the naming could be clearer.

I would do the following:
- Rename DriftJsonStrings. I believe the serialization to and from json happens elsewhere in the code. The data does happen to JsonStrings, but this class doesn't know what. I would name it DriftString, or DriftQueuedItem.
-  Rename `jsonString` on the DriftJsonStrings class. Again, it happens to be a jsonString, but it could be any string. I would name it value or data.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/index_db/in_memory_queued_item_store.dart
```diff
@@ -0,0 +1,48 @@
...
+import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_client/queued_item_store/queued_item_store.dart';
+
+/// {@template amplify_analytics_pinpoint_dart.in_memory_queued_item_store}
+/// Stores string elements in device memory
+/// {@endtemplate}
+class InMemoryQueuedItemStore implements QueuedItemStore {
+  /// {@macro amplify_analytics_pinpoint_dart.in_memory_queued_item_store}
+  InMemoryQueuedItemStore();
+
+  int _lastId = 0;
```
Jordan-Nelson, line 15: "lastId" indicates this is the ID of the last item in the DB (equal to _database.last.id), which it isn't.

```suggestion
  /// The next ID that should be used when adding an item in the DB.
  int _nextId = 0;
```

packages/analytics/amplify_analytics_pinpoint_dart/test/queued_item_store_test.dart
```diff
@@ -0,0 +1,244 @@
...
+
+      await validateStoredValues(numToGet: 6, values);
+    });
+
+    test('Only requested values are returned', () async {
+      const values = ['0', '1', '2', '3', '4', '5'];
+      await writeValues(values);
+
+      await validateStoredValues(numToGet: 3, values.sublist(0, 3));
+    });
```
Jordan-Nelson, line 72: These tests are still very dry. I have to look at writeValues and validateStoredValues to understand the tests. I'm not sure what APIs are being tested without digging into those functions.

The test names aren't helping me understand the test either. "Only requested values are returned" doesn't really explain the expected behavior. The expectation is that this behaves like a queue. That is, the first n items are returned.

Below is an example of how I might re-write the first two tests.

```dart
test('writes values to storage', () async {
  const values = ['0', '1', '2', '3', '4', '5'];
  for (final value in values) {
    await db.add(value);
  }

  final storedItems = await db.get(6);
  final storedValues = storedItems.map((e) => e.value);

  expect(storedValues, values);
});

test('returns first n items in the queue', () async {
  const values = ['0', '1', '2', '3', '4', '5'];
  for (final value in values) {
    await db.add(value);
  }

  final storedItems = await db.get(3);
  final storedValues = storedItems.map((e) => e.value);

  expect(storedValues, values.take(3));
});
```

fjnoyp: Based on the DAMP vs. DRY link, the purpose of code duplication is to "promote readability".  The tradeoff seemed to be described as conciseness vs readability.  

I don't believe that the remaining method abstractions I have reduce readability, instead I think they actively increase readability.  

Beyond the three method abstractions (writeValues, validateStoredValues, deleteValues) there aren't any other helper methods.  Furthermore each method does a basic task that can be inferred from its name. 

When reading code isn't it helpful to quickly know the intent of what's happening?  When reading the above code example, it seems harder to do. 

There are more redundant basic details that distract attention from understanding what's going on.  

While the intent is to validate the stored values, one has to figure that out by reading this code: 

```
final storedItems = await db.get(6);
  final storedValues = storedItems.map((e) => e.value);
  expect(storedValues, values);
```

Which could be: 
`      await validateStoredValues(numToGet: 6, values);
`

Once one knows the intent and sees that code pattern throughout every test, one still needs to parse x3 as much code to understand which distracts from understanding the test's purpose. 

What benefits do you see for expanding the code out in this way? 

I agree that the naming of the tests can be better though and have updated that.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/dart_queued_item_store.stub.dart
fjnoyp: I've implemented your changes.  

A downside of this change is that the connect function is no longer injectable so it cannot be configured or controlled within tests as easily.  We'd need to have a lower level test to change that behavior.

packages/analytics/amplify_analytics_pinpoint_dart/test/queued_item_store_test.dart
```diff
@@ -0,0 +1,243 @@
...
+
+    test('deletes values in storage', () async {
+      const values = ['0', '1', '2', '3', '4', '5'];
+      await writeValues(values);
+
+      await db.deleteItems(await db.get(100));
+
+      final storedValues = await db.get(100);
+      expect(storedValues.length, 0);
```
HuiSF, line 87: nit:
```suggestion
      expect(storedValues, isEmpty);
```

packages/analytics/amplify_analytics_pinpoint_dart/test/queued_item_store_test.dart
```diff
@@ -0,0 +1,243 @@
...
+    }) async {
+      final readElements = await db.get(numToGet);
+
+      expect(readElements.length, targetValues.length);
+      for (var i = 0; i < readElements.length; i++) {
+        expect(
+          reason: 'Comparing element $i of readElements vs targetValues',
+          readElements.elementAt(i).value,
+          targetValues[i],
+        );
```
HuiSF, line 41: It looks like this assertion expects the orders of elements in these lists are the same. if so can just use 
```dart
expect(targetValues, equals(readElements));
```
?

packages/analytics/amplify_analytics_pinpoint_dart/test/queued_item_store_test.dart
```diff
@@ -0,0 +1,243 @@
...
+      db = DartQueuedItemStore('test');
+    });
+
+    /// Reload db instance
+    Future<void> refreshDb() async {
+      await db.clear();
+    }
+
+    setUp(() async {
+      await refreshDb();
```
HuiSF, line 22: nit: can just call `await db.clear()` as `refreshDb` is not used anywhere else :D

packages/analytics/amplify_analytics_pinpoint_dart/test/queued_item_store_test.dart
```diff
@@ -0,0 +1,243 @@
...
+      await writeValues(values);
+
+      var readValues = await db.get(100);
+
+      await db.deleteItems(readValues);
+      await db.deleteItems(readValues);
+
+      readValues = await db.get(100);
+
+      expect(readValues.length, 0);
```
HuiSF, line 240: nit: 
```suggestion
      expect(readValues, isEmpty);
```

HuiSF (approved):

packages/analytics/amplify_analytics_pinpoint_dart/test/queued_item_store_test.dart
Jordan-Nelson: The tests as they are do not directly call any of the APIs they are testing. I don't know what APIs you are trying to test without looking into the underlying functions, which have far more code to parse through. Unit tests should generally directly call the APIs they are testing so it is clear what API is under test.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -196,7 +187,9 @@ class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
       }),
     );
 
-    _eventStorageAdapter = EventStorageAdapter(driftQueryExecutor);
+    final driftStoragePath = await _pathProvider?.getApplicationSupportPath();
+    final itemStore = DartQueuedItemStore(driftStoragePath);
```
Jordan-Nelson, line 191: No need to expose the implementation details here. DartQueuedItemStore happens to use drift under the hood on VM, but if that were swapped out the underlying storage mechanism would likely still need to know what directory to put the data in.

```suggestion
    final eventStoragePath = await _pathProvider?.getApplicationSupportPath();
    final itemStore = DartQueuedItemStore(eventStoragePath);
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/dart_queued_item_store.web.dart
```diff
@@ -0,0 +1,65 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:async';
```
Jordan-Nelson, line 4: I believe this is unused.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/drift/drift_queued_item_store.dart
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  DriftQueuedItemStore._(super.driftQueryExecutor);
+
+  // Reminder: Bump this number whenever you change or add a table definition.
+  @override
+  int get schemaVersion => 1;
+
+  /// Retrieve all [DriftString] stored in the database
+  Future<List<DriftString>> get allJsonStrings => select(driftStrings).get();
```
Jordan-Nelson, line 43: I don't see this used. Am I missing it?

Similar to other comments, would rename this to `allStrings`. They happen to be json strings, this class is not serializing data to/from json. All this class knows is that it stores strings.

fjnoyp: Thanks it is legacy and can be safely deleted.

packages/analytics/amplify_analytics_pinpoint_dart/test/queued_item_store_test.dart
```diff
@@ -0,0 +1,299 @@
...
+import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_client/queued_item_store/dart_queued_item_store.dart';
+
+import 'package:test/test.dart';
+
+void main() {
+  late DartQueuedItemStore db;
+
+  group('DartQueuedItemStore ', () {
+    setUpAll(() {
+      db = DartQueuedItemStore('test');
```
Jordan-Nelson, line 13: *nit:* This creates a file in the current working dir, which is not cleaned up. In Amplify DB Common, I had used `/tmp` as the path. A better solution might be to use a temp dir and also clean it up in a tearDownAll step. I would at least use `/tmp`.

packages/analytics/amplify_analytics_pinpoint_dart/test/queued_item_store_test.dart
```diff
@@ -0,0 +1,299 @@
...
+      await db.clear();
+    });
+
+    test('writes values to storage', () async {
+      const values = ['0', '1', '2', '3', '4', '5'];
+      for (final value in values) {
+        await db.add(value);
+      }
+
+      final readItems = await db.get(6);
```
Jordan-Nelson, line 26: *nit:* It isn't immediately obvious what the significance of 6 is.
```suggestion
      final readItems = await db.get(values.length);
```

packages/analytics/amplify_analytics_pinpoint_dart/test/queued_item_store_test.dart
```diff
@@ -0,0 +1,299 @@
...
+        return valuesToDelete.contains(item.value);
+      });
+      await db.deleteItems(itemsToDelete);
+
+      final readItems = await db.get(100);
+      final readValues = readItems.map((e) => e.value);
+      expect(readValues, const ['0', '2', '4', '6', '8']);
+    });
+
+    test('deletes every item with an even index', () async {
```
Jordan-Nelson, line 187: *nit:* Some of the tests for delete seem redundant. I appreciate the desire to be thorough and capture all edge cases. However, I am not sure each of these is testing something unique.

These seem like the core test cases.
- Can delete a single item (first, last, middle)
- Can delete all items
- Can delete a range (first, last, middle)
- Can delete multiple ranges
- Does nothing if items were previously deleted - could test this in various ways (ranges, single deletes) but not sure that is really needed.

I think you covered those, but not sure that some of the tests are really different than others.

Jordan-Nelson (approved):

packages/analytics/amplify_analytics_pinpoint_dart/test/queued_item_store_test.dart
fjnoyp: switched to /tmp

packages/analytics/amplify_analytics_pinpoint_dart/test/queued_item_store_test.dart
fjnoyp: Okay thanks - I removed 4 tests:
even, odd, delete & save x2 tests

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/index_db/indexed_db_adapter.dart
```diff
@@ -0,0 +1,148 @@
...
+    await _databaseOpenEvent;
+    final store = _getObjectStore();
+    final request = store.getAll(null, maxToGet);
+
+    await request.future;
+
+    for (final elem in request.result) {
+      final value = elem as Object;
+      final id = getProperty<int>(value, 'id');
+      final string = getProperty<String>(value, 'string');
```
dnys1, line 85: Need a better name than "string"

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/index_db/indexed_db_adapter.dart
```diff
@@ -0,0 +1,148 @@
...
+        QueuedItem(id: id, value: string),
+      );
+    }
+    return readValues;
+  }
+
+  @override
+  Future<void> deleteItems(Iterable<QueuedItem> items) async {
+    final idsToDelete = items.map((item) => item.id);
+    if (idsToDelete.isEmpty) return;
```
dnys1, line 96: Can be the first line

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/index_db/indexed_db_adapter.dart
```diff
@@ -0,0 +1,148 @@
...
+        store.deleteByKeyRange(
+          IDBKeyRange.bound(start, previous),
+        );
+        start = id;
+      }
+      previous = id;
+    }
+    await store
+        .deleteByKeyRange(IDBKeyRange.bound(start, idsToDelete.last))
+        .future;
```
dnys1, line 120: Use `package:collection`. This is confusing.

```suggestion
    final ranges = idsToDelete.splitBetween((a, b) => b != a + 1).map(
          (range) => IDBKeyRange.bound(range.first, range.last),
        );
    await Future.wait<void>(
      ranges.map((range) => store.deleteByKeyRange(range).future),
    );
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/queued_item_store.dart
```diff
@@ -0,0 +1,39 @@
...
+  FutureOr<Iterable<QueuedItem>> get(int count);
+
+  /// Delete [items] from the store.
+  FutureOr<void> deleteItems(Iterable<QueuedItem> items);
+
+  /// Clear the queue of items
+  FutureOr<void> clear();
+
+  /// Closes this store and release resources
+  FutureOr<void> close();
```
dnys1, line 21: Do these need to be `FutureOr` instead of `Future`?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/queued_item_store.dart
```diff
@@ -0,0 +1,39 @@
...
+/// Database for storing strings
+abstract class QueuedItemStore {
+  /// Insert an item to the end of the queue.
+  FutureOr<void> add(String string);
+
+  /// Get the first [count] items from the queue.
+  FutureOr<Iterable<QueuedItem>> get(int count);
+
+  /// Delete [items] from the store.
+  FutureOr<void> deleteItems(Iterable<QueuedItem> items);
```
dnys1, line 15: Why does deleteItems take in the items and not the keys?

packages/analytics/amplify_analytics_pinpoint_dart/test/queued_item_store_test.dart
```diff
@@ -0,0 +1,190 @@
...
+      expect(readValues, values.sublist(0, 9));
+    });
+
+    test('throws no error when deleting all items twice', () async {
+      const values = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'];
+      for (final value in values) {
+        await db.add(value);
+      }
+
+      var readValues = await db.get(100);
```
dnys1, line 180: Extract `db.get(100)` into a helper function -> `getAll()`

packages/aws_common/lib/src/js/common.dart
dnys1: It sounds like you were passing the `Map` without calling `jsify`. Your helper function here is exactly what `jsify` does: https://github.com/dart-lang/sdk/blob/15b77bb03c073fb0613873b7ff395addfe0bb03c/sdk/lib/_internal/js_shared/lib/js_util_patch.dart#L28

Please double check this is needed.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_storage_adapter.dart
```diff
@@ -64,42 +56,43 @@ class EventStorageAdapter implements Closeable {
...
 
 /// Wrapper class around [Event]
 /// Includes DriftId to identify that event for deletion when calling deleteEvents()
 class StoredEvent {
-  /// Create StoredEvent from [DriftJsonString]
-  factory StoredEvent(DriftJsonString data, Serializers serializers) {
-    final event = serializers.deserialize(jsonDecode(data.jsonString)) as Event;
-    return StoredEvent._(data.id, event);
+  /// Create StoredEvent from [QueuedItem]
+  factory StoredEvent(QueuedItem data, Serializers serializers) {
```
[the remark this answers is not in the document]
fjnoyp: Because there is extra logic to convert `Event` to `String` and `String` to `Event`. 

Sure it's technically not needed, but I think by OOP it's better to keep that logic separate instead of adding extra complexity to the `EventClient`.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/dart_queued_item_store.vm.dart
```diff
@@ -0,0 +1,45 @@
...
+  final DriftQueuedItemStore _database;
+
+  @override
+  Future<void> add(String string) {
+    return _database.add(string);
+  }
+
+  @override
+  Future<void> deleteItems(Iterable<QueuedItem> items) {
+    return _database.deleteItems(items);
```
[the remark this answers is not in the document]
fjnoyp: Ok well we can't use delete as it's another method so we would have to switch to Items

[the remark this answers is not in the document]
dnys1: Where? I don't see `delete` on `QueuedItemStore`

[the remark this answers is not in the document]
fjnoyp: To clarify I switched to: 
addItems/getItems/deleteItems

We can't use delete because in ```
DriftQueuedItemStore extends _$DriftQueuedItemStore
    implements QueuedItemStore 
```
The delete method name is used in the superclass.

File: packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/index_db/in_memory_queued_item_store.dart [hunk 42]
```diff
@@ -0,0 +1,49 @@
...
+
+  @override
+  Iterable<QueuedItem> get(int maxToGet) {
+    return _database.take(maxToGet);
+  }
+
+  @override
+  void deleteItems(Iterable<QueuedItem> items) {
+    final idsToDelete = items.map((item) => item.id);
+    final hashSet = HashSet<int>.from(idsToDelete);
```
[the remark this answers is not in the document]
fjnoyp: I used a `HashedSet` here for performance reasons.  I assume an Iterable list's contain method would be O(n) time instead of O(1) for `HashedSet` when calling `_database.removeWhere(...`

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/index_db/in_memory_queued_item_store.dart
```diff
@@ -0,0 +1,49 @@
...
+/// {@template amplify_analytics_pinpoint_dart.in_memory_queued_item_store}
+/// Stores string elements in device memory
+/// {@endtemplate}
+class InMemoryQueuedItemStore implements QueuedItemStore {
+  /// {@macro amplify_analytics_pinpoint_dart.in_memory_queued_item_store}
+  InMemoryQueuedItemStore();
+
+  /// The next ID that should be used when adding an item in the DB
+  int _nextId = 0;
+  final Queue<QueuedItem> _database = Queue();
```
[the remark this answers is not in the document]
fjnoyp: Nice that's good to know it provides that. 

What is the benefit of using a map here though?  Isn't this implementing a Queue interface anyway?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/index_db/indexed_db_adapter.dart
```diff
@@ -0,0 +1,148 @@
...
+      mode: IDBTransactionMode.readwrite,
+    );
+    final store = transaction.objectStore(storeName);
+    return store;
+  }
+
+  @override
+  Future<void> add(String string) async {
+    await _databaseOpenEvent;
+    await _getObjectStore().push({'string': string}).future;
```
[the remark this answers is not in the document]
fjnoyp: There must be a key-value pair

packages/aws_common/lib/src/js/indexed_db.dart
```diff
@@ -184,19 +198,47 @@ extension PropsIDBObjectStore on IDBObjectStore {
...
   /// This is for retrieving specific records from an object store.
   IDBRequest<String?> getObject(String query) =>
       js_util.callMethod(this, 'get', [query]);
+
+  /// Returns an [IDBRequest] object, and, in a separate thread, returns
+  /// [count] records from the object store.
+  ///
+  /// This is for retrieving a specific [count] of records from the object store.
+  IDBRequest<List<dynamic>> getAll(String? query, int? count) =>
+      js_util.callMethod(this, 'getAll', [query, count]);
```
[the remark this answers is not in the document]
fjnoyp: Why the cast?  Also doesn't compile in that case.  just used: 

```
 IDBRequest<List<Object>> getAll([String? query, int? count]) =>
      js_util.callMethod(this, 'getAll', [query, count]);
}
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/index_db/indexed_db_adapter.dart
fjnoyp: switched to 'value'

packages/aws_common/lib/src/js/indexed_db.dart
```diff
@@ -184,19 +198,47 @@ extension PropsIDBObjectStore on IDBObjectStore {
   IDBRequest<void> add(String value, String key) =>
       js_util.callMethod(this, 'add', [value, key]);
 
+  /// Returns an [IDBRequest] object, and, in a separate thread, creates a
+  /// structured clone of the value, and stores the cloned value in the object
+  /// store.
+  ///
+  /// This is for adding new records to an object store created with keyPath set and autoincrement = true
+  IDBRequest<void> push(Map<Object, Object> item) =>
```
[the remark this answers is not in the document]
fjnoyp: This is for adding new records to an object store created with keyPath set and autoincrement = true

Otherwise if just calling add with 'value' you get: 
 ` DataError: Failed to execute 'add' on 'IDBObjectStore': A generated key could not be inserted into the value.
`

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/dart_queued_item_store.vm.dart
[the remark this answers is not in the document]
dnys1: Why does Drift DB need to implement `QueuedItemStore`?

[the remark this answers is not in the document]
dnys1: Would prefer `add`/`getCount`/`delete` as suggested

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/queued_item_store.dart
fjnoyp: Made it `FutureOr` because of `InMemoryQueuedItemStore`

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/queued_item_store.dart
fjnoyp: To improve abstraction - consuming code shouldn't need to worry about item ids and parse those ids to perform deletes.

fjnoyp:
@dnys1 you're right I need to use jsify first.  That seems a bit confusing to call this though: 

 _convertToJs(jsify(a))!;
 
 So I've kept the method name but replaced the implementation with above code.

packages/aws_common/lib/src/js/common.dart
```diff
@@ -374,3 +376,9 @@ Object? dartify(Object? o) {
   }
   return o;
 }
+
+/// Convert a dart map to a JS object
+@internal
+Object mapToJSObj(Map<Object, Object> a) {
+  return _convertToJs(jsify(a))!;
```
dnys1, line 383: This method is not needed - just call `jsify` directly instead for Maps.

packages/aws_common/lib/src/js/indexed_db.dart
```diff
@@ -151,8 +151,22 @@ extension PropsIDBDatabase on IDBDatabase {
...
+    final params = <Object, Object>{};
+    if (keyPath != null) {
+      params['keyPath'] = keyPath;
+    }
+    if (autoIncrement != null) {
+      params['autoIncrement'] = autoIncrement;
+    }
+
+    return js_util
+        .callMethod(this, 'createObjectStore', [name, mapToJSObj(params)]);
```
dnys1, line 168: Add a trailing comma

packages/analytics/amplify_analytics_pinpoint_dart/test/queued_item_store_test.dart
```diff
@@ -0,0 +1,195 @@
...
+import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_client/queued_item_store/queued_item_store.dart';
+
+import 'package:test/test.dart';
+
+void main() {
+  late DartQueuedItemStore db;
+
+  group('DartQueuedItemStore ', () {
+    setUpAll(() {
+      db = DartQueuedItemStore('/tmp');
```
dnys1, line 14:
```suggestion
      final tempDir = Directory.systemTemp.createTempSync('amplify_analytics');
      addTearDown(() => tempDir.deleteSync(recursive: true));
      db = DartQueuedItemStore(tempDir.path);
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/dart_queued_item_store.vm.dart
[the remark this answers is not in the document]
fjnoyp: the current api names are:

`addItem` `getCount` and `deleteItems`

Drift DB implements `QueuedItemStore` for uniformity as its purpose is to store queued items.   

I don't think we should make it no longer implement that interface just so that we can name an API method `delete` instead of `deleteItems`.  Furthermore while `delete` methods normally remove a single item the `deleteItems` is specifically meant to delete multiple items at once.

packages/analytics/amplify_analytics_pinpoint_dart/test/queued_item_store_test.dart
fjnoyp: I don't think we can do that as Directory is from `dart:io`

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/index_db/in_memory_queued_item_store.dart
```diff
@@ -0,0 +1,48 @@
...
+  Iterable<QueuedItem> getCount(int count) {
+    return _database.values.take(count);
+  }
+
+  @override
+  void deleteItems(Iterable<QueuedItem> items) {
+    for (final item in items.toList()) {
+      _database.remove(item.id);
+    }
+  }
```
dnys1, line 39:
```suggestion
  Iterable<QueuedItem> getCount(int count) {
    return _database.values.take(count).toList();
  }

  @override
  void deleteItems(Iterable<QueuedItem> items) {
    for (final item in items) {
      _database.remove(item.id);
    }
  }
```

packages/analytics/amplify_analytics_pinpoint_dart/test/queued_item_store_test.dart
dnys1: Sure you can.

```dart
String? path;
if (!zIsWeb) {
  final tempDir = Directory.systemTemp.createTempSync('amplify_analytics');
  addTearDown(() => tempDir.deleteSync(recursive: true));
  path = tempDir.path;
}
db = DartQueuedItemStore(path);
```

packages/aws_common/lib/src/js/indexed_db.dart
```diff
@@ -6,6 +6,7 @@ library amplify_secure_storage_dart.js.indexed_db;
 
 import 'dart:async';
 import 'dart:js_util' as js_util;
+import 'dart:js_util';
```
dnys1, line 9: Remove - use prefixed `js_util`

packages/aws_common/lib/src/js/common.dart
```diff
@@ -8,6 +8,8 @@ import 'dart:js_util' as js_util;
 
 import 'package:aws_common/src/util/recase.dart';
 import 'package:js/js.dart';
+import 'package:js/js_util.dart';
+import 'package:meta/meta.dart';
```
dnys1, line 12: Remove

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/drift/drift_queued_item_store.dart
```diff
@@ -0,0 +1,76 @@
...
+import 'dart:async';
+
+import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_client/queued_item_store/queued_item_store.dart';
+import 'package:amplify_db_common_dart/amplify_db_common_dart.dart';
+import 'package:drift/drift.dart';
+
+part 'drift_queued_item_store.g.dart';
+
+/// SQL schema of data stored in DriftDatabase
+class DriftStrings extends Table {
```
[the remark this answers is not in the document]
dnys1: Please address this

packages/analytics/amplify_analytics_pinpoint_dart/test/queued_item_store_test.dart
fjnoyp: Nice thanks!

packages/analytics/amplify_analytics_pinpoint_dart/test/queued_item_store_test.dart
```diff
@@ -0,0 +1,203 @@
...
+
+    test(
+        'returns all stored items when get request size exceeds stored item count',
+        () async {
+      const values = ['0', '1', '2', '3', '4', '5'];
+      for (final value in values) {
+        await db.addItem(value);
+      }
+
+      final readItems = await getAll();
```
dnys1, line 67: Here you should be explicit since it's what you're testing...

```suggestion
      final readItems = await db.getCount(100);
```

Once you test it, you can take it for granted and use `getAll`

dnys1 (approved):

PR #2606, opened 2023-01-27
feat(api)!: custom primary key support for GraphQL model helpers

ragingsquirrel3:
This PR enables CPK (custom primary key) in GraphQL model helpers. It is a breaking change from the call sites of model helpers although older codegen files are still supported here when possible given the user provides a `ModelIdentifier` somehow (for the APIs that require it). Example:

```dart
final req = ModelQueries.get(Blog.classType, blog.modelIdentifier);
```

Summary:
* remove model helper interface definitions in core to reduce the coupling between packages as there are no use cases outside the API plugin.
* changes interfaces of model helpers to accept `Modeldentifier`s instead of string IDs.
* implementation changes to support newer schemas with or without custom primary keys (mostly in `graphql_request_factory.dart`)
* lots of tests and test codegen files for these use cases although not as exhaustive as Datastore tests


packages/api/amplify_api_dart/lib/src/graphql/factories/graphql_request_factory.dart
```diff
@@ -98,8 +108,35 @@ class GraphQLRequestFactory {
...
+            indexes?.firstWhereOrNull((index) => index.name == null);
+        if (modelIndex != null) {
+          // custom index(es), e.g.
+          // upperOutput: ($customId: ID!, name: String!)
+          // lowerOutput: (customId: $customId, name: $name)
+          final bUpperOutput = StringBuffer(_openParen);
+          final bLowerOutput = StringBuffer(_openParen);
+          for (final field in modelIndex.fields) {
+            final isId = field == modelIndex.fields.first;
+            // Assume type is `String` if not specified as `ID`.
```
ragingsquirrel3, line 122: callout: want to check if other types can be used for index, then might have to map all these out

packages/api/amplify_api_dart/lib/src/graphql/factories/graphql_request_factory.dart
```diff
@@ -216,14 +253,14 @@ class GraphQLRequestFactory {
...
           association?.targetNames?.first ?? association?.targetName;
       var fieldName = queryPredicate.field;
       if (queryPredicate.field ==
-          '${_lowerCaseFirstCharacter(schema.name)}.$idFieldName') {
-        // check for the IDs where fieldName set to e.g. "blog.id" and convert to "id"
-        fieldName = idFieldName;
+          '${_lowerCaseFirstCharacter(schema.name)}.$_defaultIdFieldName') {
+        // Check for the IDs where fieldName set to e.g. "blog.id" and convert to "id".
+        // This is only needed in older codegen files (without CPK).
+        fieldName = _defaultIdFieldName;
```
Equartey, line 263: nit: Discussed offline. Could leave some comments as to why each branch exists or in what scenario.

packages/api/amplify_api/example/integration_test/graphql/iam_test.dart
```diff
@@ -226,7 +226,90 @@ void main({bool useExistingTestUser = false}) {
...
+        );
+        final res = await Amplify.API.query(request: req).response;
+        final data = res.data;
+        expect(res, hasNoGraphQLErrors);
+        expect(data, equals(cpkParent));
+      });
+
+      testWidgets(
+          'should GET a child and include parent with complex identifier and custom primary key',
+          (WidgetTester tester) async {
```
Equartey, line 249: nit: Following up on our offline discussion. This could be a test case for a relationship diagram-like comment.

Equartey (approved):
Nice job. Tested some happy cases locally and CPK worked well. Just left a couple nits.

packages/api/amplify_api/example/integration_test/graphql/iam_test.dart
ragingsquirrel3: thanks, will work on improving

packages/api/amplify_api_dart/lib/src/graphql/factories/graphql_request_factory.dart
```diff
@@ -98,8 +108,35 @@ class GraphQLRequestFactory {
...
+        if (modelIndex != null) {
+          // custom index(es), e.g.
+          // upperOutput: ($customId: ID!, name: String!)
+          // lowerOutput: (customId: $customId, name: $name)
+          final bUpperOutput = StringBuffer(_openParen);
+          final bLowerOutput = StringBuffer(_openParen);
+          for (final field in modelIndex.fields) {
+            final isId = field == modelIndex.fields.first;
+            final graphQLTypeString =
+                isId ? 'ID' : getGraphQLTypeForField(schema, field);
```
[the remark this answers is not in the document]
ragingsquirrel3: Per offline discussion, added comment to address this. Bc codegen doesn't capture scalar type in schema, will populate the document string with values instead of referencing variables (which requires scalar type declared) in the case of custom index in get() query.

packages/api/amplify_api/example/lib/models/Post.dart
```diff
@@ -248,16 +272,13 @@ class Post extends Model {
     modelSchemaDefinition.addField(ModelFieldDefinition.belongsTo(
         key: Post.BLOG,
         isRequired: false,
-        // TODO(Jordan-Nelson): Remove `targetName` when API category has been
-        // updated to support CPK changes. This was added manually.
-        // ignore: deprecated_member_use
-        targetName: "blogID",
-        ofModelName: (Blog).toString()));
+        targetNames: ['blogID'],
+        ofModelName: 'Blog'));
```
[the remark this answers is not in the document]
ragingsquirrel3: yes, sorry for slow response here

HuiSF (approved):

packages/api/amplify_api/example/integration_test/graphql/iam_test.dart
ragingsquirrel3: added small comment

packages/api/amplify_api_dart/lib/src/graphql/factories/graphql_request_factory.dart
ragingsquirrel3: tried to clarify comments, though not sure how much I added as there were already some comments

PR #2607, opened 2023-01-27
feat(auth): Send Pinpoint ID

fjnoyp:
*Issue #, if available:*

*Description of changes:*

Listen for Analytics hub events and send pinpoint id with auth requests related to cognito user pools: 

https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pools-pinpoint-integration.html

This is to achieve feature parity with Amplify iOS and Android.  Our new dart first implementation was lacking this. 


packages/amplify_core/lib/src/config/auth/cognito/user_pool.dart
```diff
@@ -15,13 +15,16 @@ class CognitoUserPoolConfig
   final CognitoOAuthConfig? hostedUI;
   final String? endpoint;
 
-  const CognitoUserPoolConfig({
+  String? pinpointEndpointId;
+
+  CognitoUserPoolConfig({
```
dnys1, line 20: These classes are a direct representation of the `amplifyconfiguration.dart` file, so this is not the right place for this.

Data classes should also be immutable. We override `==` and `hashCode` on all data classes with `AWSEquatable`, which is marked with `@immutable` to ensure these values never change over the lifetime of the object.

This is why the analyzer triggers an error for this line: 

```
 info - lib/src/config/auth/cognito/user_pool.dart:9:7 - This class (or a class that this class inherits from) is marked as '@immutable', but one or more of its instance fields aren't final: CognitoUserPoolConfig.pinpointEndpointId - must_be_immutable
```

The reasons for this are two-fold:

- Mutating data classes over time leads to difficult-to-reproduce errors and hard-to-follow data flow. This is why functional programming languages expressly prohibit this paradigm.
- Anytime `==` or `hashCode` are overridden for structural equality, they _must_ never be mutated. For example, imagine putting a CognitoUserPoolConfig in a HashSet, then setting the pinpointEndpointId value - the HashSet would tell you it's not there! This is pretty confusing without having created a new object.

fjnoyp:
Need to merge before: https://github.com/aws-amplify/amplify-flutter/pull/2518 
As this builds on that PR

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/auth_state_machine.dart
```diff
@@ -55,6 +58,20 @@ class AuthStateMachine extends AuthStateMachineBase {
           endpoint: userPoolConfig.endpoint,
         ),
       );
+
+      // Listen for Analytics Pinpoint ID
+      final analyticsConfig = event.config.analytics;
+      if (analyticsConfig != null) {
+        late StreamSubscription<AnalyticsHubEvent> subscription;
+        subscription = Amplify.Hub.listen(HubChannel.Analytics,
```
dnys1, line 66: What happens if configuration fails?

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_in_state_machine.dart
```diff
@@ -70,7 +70,7 @@ class SignInStateMachine extends StateMachine<SignInEvent, SignInState> {
   bool get hasIdentityPool => identityPoolConfig != null;
 
   /// The Cognito Identity Provider service client.
-  late final CognitoIdentityProviderClient cognitoIdentityProvider = expect();
+  late final CognitoIdentityProviderClient _cognito = expect();
```
dnys1, line 73: "Cognito" is not descriptive enough since we work with two Cognito APIs, one for user pools (called `CognitoIdentityProvider`) and one for identity pools (called `CognitoIdentity`)

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_in_state_machine.dart
```diff
@@ -414,6 +419,11 @@ class SignInStateMachine extends StateMachine<SignInEvent, SignInState> {
         })
         ..clientMetadata.addAll(event.clientMetadata);
 
+      if (config.pinpointEndpointId != null) {
```
dnys1, line 422: All `InitiateAuthRequest`s and `RespondToAuthChallengeRequest`s should have this, but some don't.

dnys1 (commented):
Can we add some tests for this change?

fjnoyp:
Hey this is draft

packages/amplify_core/lib/src/config/auth/cognito/user_pool.dart
fjnoyp: Thanks for the great insights here on that.  I've changed the code to add a analyticsMetadataProvider in the stateMachineManager's dependencyManager.

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/auth_state_machine.dart
fjnoyp: Which configuration?

If auth config fails - even if it ran the results are irrelevant as I assume remaining auth will not run. 

If analytics config fails - analytics hub event might not even get here but if it does, that should be okay because => 

Current iOS behavior is for a Pinpoint ID to be generated in iOS Auth even if Analytics is not configured so it appears there was previous consensus that sending non initialized Pinpoint Endpoint ids is okay.  

I'm clarifying this - waiting for Pinpoint team to respond.

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_in_state_machine.dart
fjnoyp: Ok, shouldn't it be private though?

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_in_state_machine.dart
fjnoyp: Thanks added

fjnoyp:
> Can we add some tests for this change?

Added tests in srp workers and in authStateMachine test to ensure analytics pinpoint id is added. 

I did not add tests to ensure the relevant auth requests included the pinpoint endpoint id.  I assumed this extra step wouldn't be necessary and also did not find any unit tests that tested the fields of a generated request.  

@dnys1 perhaps I'm missing something but I couldn't find unit tests checking the value of generated Auth Request objects.

dnys1:
> @dnys1 perhaps I'm missing something but I couldn't find unit tests verifying the fields of generated Auth Request objects. Is it because writing those tests would be difficult or am I missing something else?

Which fields?

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
```diff
@@ -183,6 +184,9 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface<
   DeviceMetadataRepository get _deviceRepo =>
       _stateMachine.getOrCreate(DeviceMetadataRepository.token);
 
+  /// Analytics Metadata Provider
+  AnalyticsMetadataTypeBuilder? get _analyticsMetadata => _stateMachine.get();
```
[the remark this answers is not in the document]
fjnoyp: Agreed but the requests it takes all require the builder object instead of the actual instance, so when feeding into the requests I'd have to say something like _analyticsMetadata.toBuilder()

fjnoyp:
Converting to draft again.  Based on further talks with Amplify and service teams it appears we might actually need to give Auth the capability to generate and store the Pinpoint id without the analytics plugin being added.

packages/amplify_core/lib/src/hub/hub_channel.dart
```diff
@@ -18,5 +18,8 @@ enum HubChannel<HubEventPayload, E extends HubEvent<HubEventPayload>> {
   DataStore<DataStoreHubEventPayload, DataStoreHubEvent>(),
 
   /// Events of the API category
-  Api<ApiHubEventPayload, ApiHubEvent>();
+  Api<ApiHubEventPayload, ApiHubEvent>(),
+
+  /// Events of the Analytics category
+  Analytics<AnalyticsHubEventPayload, AnalyticsHubEvent>();
```
Jordan-Nelson, line 24: I am not sure if adding a hub for this purpose makes sense. Let's discuss offline.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -63,10 +64,12 @@ class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
             'Please make sure to call: await Amplify.configure(amplifyconfig)',
       );
     }
+    if (_cachedConfigException != null) throw _cachedConfigException!;
   }
 
   var _isConfigured = false;
   var _analyticsEnabled = false;
+  AnalyticsException? _cachedConfigException;
```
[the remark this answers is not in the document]
fjnoyp: Removed, code was refactored.

packages/auth/amplify_auth_cognito_test/test/state/auth_state_machine_test.dart
```diff
@@ -77,5 +80,35 @@ void main() {
 
       await stateMachine.close();
     });
+
+    test('analytics metadata is configured', () async {
+      final authStateMachine = stateMachine.getOrCreate(AuthStateMachine.type);
+
+      stateMachine.dispatch(AuthEvent.configure(mockConfigWithPinpoint));
```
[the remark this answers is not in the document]
fjnoyp: If no config with pinpoint is present, the `AnalyticsMetadataTypeBuilder` is never registered and sent with the Auth requests.  

These assignments take nullable fields. 
`..analyticsMetadata = get<AnalyticsMetadataTypeBuilder>();`

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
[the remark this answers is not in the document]
dnys1: That's the best way to do it. A builder is only a temporary object for updating a value before concretizing it. It shouldn't be long-lasting since its mutable.

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
```diff
@@ -90,8 +93,10 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface<
   AmplifyAuthCognitoDart({
     SecureStorageInterface? credentialStorage,
     HostedUiPlatformFactory? hostedUiPlatformFactory,
+    EndpointInfoStoreManager? pinpointEndpointIdStore,
```
dnys1, line 96: Shouldn't be injected. We should use the dependency manager (state machine) for injecting in tests.

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
```diff
@@ -269,6 +280,19 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface<
...
+    if (analyticsConfig != null && _pinpointEndpointIdStore != null) {
+      final appId = analyticsConfig.appId;
+
+      await _pinpointEndpointIdStore!.init(pinpointAppId: appId);
+
+      stateMachine.addInstance(
+        AnalyticsMetadataTypeBuilder()
+          ..analyticsEndpointId = _pinpointEndpointIdStore!.endpointId,
+      );
+    }
```
dnys1, line 294: Move to ConfigurationStateMachine

packages/auth/amplify_auth_cognito_test/test/common/mock_config.dart
```diff
@@ -46,6 +49,14 @@ final mockConfig = AmplifyConfig(
   ),
 );
 
+final mockConfigWithPinpoint = mockConfig.copyWith(
+    auth: AuthConfig.cognito(
+  pinpointAnalyticsConfig: const CognitoPinpointAnalyticsConfig(
+    appId: testPinpointAppId,
+    region: testRegion,
+  ),
+));
```
dnys1, line 58: Trailing comma

packages/auth/amplify_auth_cognito_test/test/flows/srp/srp_device_password_verifier_worker_test.dart
```diff
@@ -29,6 +30,9 @@ void main() {
 
       worker.logs.listen(safePrint);
       await worker.spawn();
+      final analyticsMetadata = (AnalyticsMetadataTypeBuilder()
+            ..analyticsEndpointId = testPinpointAppId)
+          .build();
```
dnys1, line 35:
```suggestion
      final analyticsMetadata = AnalyticsMetadataType(
        analyticsEndpointId: testPinpointAppId,
      );
```

packages/auth/amplify_auth_cognito_test/test/state/configuration_state_machine_test.dart
```diff
@@ -83,5 +86,18 @@ void main() {
 
       await stateMachine.close();
     });
+
+    test('configures analytics metadata if configured', () async {
+      stateMachine
+        ..getOrCreate(ConfigurationStateMachine.type)
+        ..dispatch(ConfigurationEvent.configure(mockConfigWithPinpoint));
+
+      expect(
```
dnys1, line 95: Update to await configuration after moving logic to `ConfigurationStateMachine`. This should only be true after configuration completes.

dnys1: EndpointInfoManager probably should be mocked as well.

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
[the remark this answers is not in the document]
fjnoyp: Okay I used this form since it was shorter: 
`          ..analyticsMetadata = _analyticsMetadata?.toBuilder();
`

packages/auth/amplify_auth_cognito_dart/lib/src/util/secure_storage_factory.dart
```diff
@@ -0,0 +1,6 @@
+import 'package:amplify_secure_storage_dart/amplify_secure_storage_dart.dart';
+
+/// A factory constructor for a [SecureStorageInterface] instance.
+typedef SecureStorageFactory = SecureStorageInterface Function({
+  required String scope,
+});
```
dnys1, line 6: I think this is will be a bit cleaner

```dart
typedef SecureStorageFactory = SecureStorageInterface Function(String scope);
```

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
```diff
@@ -141,11 +140,10 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface<
...
   final OAuthParameters? _initialParameters = initialParameters;
 
-  /// The on-device credential storage for the Auth category.
+  /// Factory for creating secure storage.
   ///
-  /// Defaults to an instance of [AmplifySecureStorageDart] with a scope of
-  /// "auth".
-  final SecureStorageInterface? _credentialStorage;
+  /// Creates an instance of [AmplifySecureStorageDart] with scope input
+  final SecureStorageFactory? _secureStorageFactory;
```
dnys1, line 146: This can be non-null with a default in the constructor

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
```diff
@@ -269,6 +268,31 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface<
...
   }
 
+  @protected
+  Future<void> registerAnalyticsMetadata({
+    required AmplifyConfig config,
+    LegacyNativeDataProvider? dataProvider,
+  }) async {
+    // Setup AnalyticsMetadataType
+    final analyticsConfig = config.auth?.awsPlugin?.pinpointAnalytics?.default$;
+    if (analyticsConfig != null) {
```
dnys1, line 278:
```suggestion
    if (analyticsConfig == null) {
      return;
    }
```

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
```diff
@@ -269,6 +268,31 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface<
...
+      final appId = analyticsConfig.appId;
+
+      final storageFactory = _stateMachine.expect<SecureStorageFactory>();
+      final endpointStore = storageFactory(scope: 'analytics');
+      final endpointStoreManager = EndpointInfoStoreManager(
+          store: endpointStore, legacyNativeDataProvider: dataProvider);
+
+      await endpointStoreManager.init(pinpointAppId: appId);
```
dnys1, line 286: I think this whole method should ideally be done in the `ConfigurationStateMachine`. The only thing we want to do in `plugin_impl` is add everything to the dependency manager so that this method only gets called once.

packages/auth/amplify_auth_cognito_test/test/state/configuration_state_machine_test.dart
fjnoyp: I don't think EndpointInfoManager needs to be mocked, but we can inject a mock secure storage and test with that

packages/analytics/amplify_analytics_pinpoint/example/lib/main.dart
```diff
@@ -40,18 +40,19 @@ class _MyAppState extends State<MyApp> {
...
+            // FIXME: In your app, make sure to remove this line and set up
+            /// Keychain Sharing in Xcode as described in the docs:
+            /// https://docs.amplify.aws/lib/project-setup/platform-setup/q/platform/flutter/#enable-keychain
+            // ignore: invalid_use_of_visible_for_testing_member
+            macOSOptions: MacOSSecureStorageOptions(useDataProtection: false),
+          ),
+        );
+    final authPlugin = AmplifyAuthCognito(secureStorageFactory: storageFactory);
+    final analyticsPlugin =
+        AmplifyAnalyticsPinpoint(secureStorageFactory: storageFactory);
```
Jordan-Nelson, line 55: I'm wondering if we want to offer a new static method on AmplifySecureStrorage to make this API a little cleaner. We could also consider making the current constructor for AmplifySecureStorage internal since we don't really want to encourage use of in outside of our libs. Everyone should just be using the factory.

I need to put a little more thought into the naming and impl, but something like the below would work.

@dnys1, thoughts?

```dart
// main.dart
final macOSOptions = MacOSSecureStorageOptions(useDataProtection: false);
final authPlugin = AmplifyAuthCognito(
  secureStorageFactory: AmplifySecureStorage.fromFactory(
    macOSOptions: macOSOptions,
  ),
);
final analyticsPlugin = AmplifyAnalyticsPinpoint(
  secureStorageFactory: AmplifySecureStorage.fromFactory(
    macOSOptions: macOSOptions,
  ),
);
```

```dart
// amplify_secure_storage.vm.dart
class AmplifySecureStorage extends AmplifySecureStorageInterface {
  /// Generates a [AmplifySecureStorage] from a config.
  ///
  /// Should only be used within Amplify packages. See [fromFactory] for
  /// external use.
  @internal
  AmplifySecureStorage.fromConfig({
    required super.config,
  });

  /// Returns a factory for creating [AmplifySecureStorage] instances.
  static SecureStorageFactory fromFactory({
    WebSecureStorageOptions? webOptions,
    WindowsSecureStorageOptions? windowsOptions,
    LinuxSecureStorageOptions? linuxOptions,
    MacOSSecureStorageOptions? macOSOptions,
    IOSSecureStorageOptions? iOSOptions,
  }) {
    return (AmplifyStorageScope scope) {
      return AmplifySecureStorage.fromConfig(
        config: AmplifySecureStorageConfig(
          scope: scope.name,
          webOptions: webOptions,
          windowsOptions: windowsOptions,
          linuxOptions: linuxOptions,
          macOSOptions: macOSOptions,
          iOSOptions: iOSOptions,
        ),
      );
    };
  }
  ...
}

Comment (fjnoyp) [reply packages/analytics/amplify_analytics_pinpoint/example/lib/main.dart:55]: I like this, it hides internal implementation details that customers should not be seeing. I think we will still need to expose the general constructor though because we have several unit tests that need to seed data on specific secure storage instances.

**Comment by fjnoyp** [comment]
@Jordan-Nelson @dnys1 as this PR changes files across multiple plugin directories, I am thinking to do a merge commit with each commit related to a particular plugin (auth, analytics, storage, api) changes for future clarity.

File: packages/analytics/amplify_analytics_pinpoint/example/lib/main.dart (hunk 18 above)
Comment (dnys1) [reply packages/analytics/amplify_analytics_pinpoint/example/lib/main.dart:55]: Sounds good - should be called `factoryFrom` though or just `factory` so that it makes sense without arguments, e.g. `AmplifySecureStorage.factory()`

**Comment by dnys1** [comment]
> @Jordan-Nelson @dnys1 as this PR changes files across multiple plugin directories, I am thinking to do a merge commit with each commit related to a particular plugin (auth, analytics, storage, api) changes for future clarity.

Yep, make sense. Please make sure each commit is properly marked with an exclamation point when it's a breaking change to any public API.

File: packages/analytics/amplify_analytics_pinpoint/example/lib/main.dart (hunk 18 above)
Comment (Jordan-Nelson) [reply packages/analytics/amplify_analytics_pinpoint/example/lib/main.dart:55]: Agreed. `factoryFrom` or just `factory` makes more sense.

Comment (dnys1) [reply packages/analytics/amplify_analytics_pinpoint/example/lib/main.dart:55]: I think I prefer `factoryFrom` since you would never really call it without arguments - it's an optional parameter everywhere, so overriding it would only be to customize it in some way.

File: packages/auth/amplify_auth_cognito/lib/src/auth_plugin_impl.dart [hunk 19]
```diff
@@ -120,6 +120,20 @@ class AmplifyAuthCognito extends AmplifyAuthCognitoDart with AWSDebuggable {
     AmplifyConfig? config,
     required AmplifyAuthProviderRepository authProviderRepo,
   }) async {
+    // Dependencies for AnalyticsMetadataType
```
fjnoyp, line 123: I added the state machine dependencies here because the configuration state machine is called during the call to configure in dart. 

I don't think this code is ideal as I'm relying on the execution order of the code to get things to work. 

@dnys1 can I move this code to the constructor instead?  I'm a bit concerned these changes are going to make the code less maintainable in the future.

packages/auth/amplify_auth_cognito_dart/lib/src/state/state/configuration_state.dart
```diff
@@ -3,6 +3,33 @@
 
 import 'package:amplify_auth_cognito_dart/src/state/state.dart';
 import 'package:amplify_core/amplify_core.dart';
+import 'package:amplify_secure_storage_dart/amplify_secure_storage_dart.dart';
+import 'package:meta/meta.dart';
+
+/// Token for storing the [SecureStorageInterface] used to store Pinpoint Endpoint Id.
```
fjnoyp, line 9: I can't use the base Token class as secure storage is saved twice, once for Auth and another time for Analytics.

fjnoyp:
Hi guys thanks again for your help 

Overall feeling a bit hesitant with the code changes.  I feel like these changes are a bit harder to maintain than the original approach of passing a `FlutterEndpointInfoManager` that wraps all `EndpointInfoManager` sub dependencies like `SecureStorage` and `LegacyDataProvider`.  Since Analytics can directly export and use `FlutterEndpointInfoManager`, we also know the code will evolve together in one place as well. 

I know we agreed on the current approach already so I'm fine either way.  Wanted to mention this for you to keep it in mind when you review this PR again.

packages/secure_storage/amplify_secure_storage_dart/lib/src/types/secure_storage_factory.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:amplify_secure_storage_dart/amplify_secure_storage_dart.dart';
```
dnys1, line 1: License headers

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -30,11 +31,20 @@ class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
...
-  })  : _appLifecycleProvider = appLifecycleProvider,
-        _pathProvider = pathProvider,
-        _analyticsClient = analyticsClient ?? AnalyticsClient();
+    SecureStorageFactory? secureStorageFactory,
+  })  : _pathProvider = pathProvider,
+        _legacyNativeDataProvider = legacyNativeDataProvider,
+        _deviceContextInfoProvider = deviceContextInfoProvider,
+        _appLifecycleProvider = appLifecycleProvider,
+        _secureStorageFactory = secureStorageFactory ??
+            ((scope) => AmplifySecureStorageWorker(
```
dnys1, line 43: The secure storage Dart package should have an equivalent factory method and that should be used here

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
```diff
@@ -279,6 +283,16 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface<
       );
     }
 
+    // Dependencies for AnalyticsMetadataType
+    _stateMachine.addInstance(
+      AmplifySecureStorageDart(
```
dnys1, line 288: You're using `AmplifySecureStorageDart` and `AmplifySecureStorageWorker` interchangeably throughout. We should ideally just be using the worker variant

packages/auth/amplify_auth_cognito/lib/src/auth_plugin_impl.dart
```diff
@@ -126,6 +120,20 @@ class AmplifyAuthCognito extends AmplifyAuthCognitoDart with AWSDebuggable {
...
+    stateMachine
+      ..addInstance<LegacyNativeDataProvider>(
+        FlutterLegacyNativeDataProvider(),
+      )
+      ..addInstance(
```
dnys1, line 128: `addInstance` calls will override previous ones, so calling this here won't do anything since it will be overridden in `super.configure`.

packages/auth/amplify_auth_cognito_dart/lib/src/state/state/configuration_state.dart
dnys1: This doesn't belong here - let's put in its own file

packages/auth/amplify_auth_cognito_dart/lib/src/state/state/configuration_state.dart
```diff
@@ -3,6 +3,33 @@
...
+@immutable
+class AmplifySecureStorageToken extends Token<SecureStorageInterface> {
+  /// Create a token using [name].
+  const AmplifySecureStorageToken(this.name, [super.dependencies]);
+
+  /// The token name.  Used for comparisons.
+  final String name;
```
dnys1, line 16:
```suggestion
  /// The storage scope.
  final AmplifyStorageScope scope;
```

packages/auth/amplify_auth_cognito_dart/lib/src/state/state/configuration_state.dart
```diff
@@ -3,6 +3,33 @@
...
+  int get hashCode => name.hashCode;
+
+  @override
+  String toString() => name;
+}
+
+/// Secure Storage Token for Pinpoint Endpoint storage.
+AmplifySecureStorageToken analyticsStorageToken = AmplifySecureStorageToken(
+  AmplifyStorageScope.awsPinpointAnalyticsPlugin.name,
+);
```
dnys1, line 32: Not in love with this and doesn't seem to be used everywhere

packages/auth/amplify_auth_cognito_dart/lib/src/state/state/configuration_state.dart
fjnoyp: File removed due to switch to EndpointInfoStoreManager we discussed yesterday.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -90,15 +102,22 @@ class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
     final eventStoragePath = await _pathProvider?.getApplicationSupportPath();
     final eventStore = DartQueuedItemStore(eventStoragePath);
 
-    await _analyticsClient.init(
+    final analyticsClient = AnalyticsClient(
+      storageScope: AmplifyStorageScope.awsPinpointAnalyticsPlugin,
+      deviceContextInfoProvider: _deviceContextInfoProvider,
+      legacyNativeDataProvider: _legacyNativeDataProvider,
+      secureStorageFactory: _secureStorageFactory,
+    );
```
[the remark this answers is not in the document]
fjnoyp: `AnalyticsClient` will be used outside of Analytics in Push Notifications to interact with Pinpoint. 

While Push Notifications is internal so we have more control. I do prefer for external package usage to be strongly typed and controlled. 

Switching to taking in a storage instance directly could suggest or allow for storage instances that do not have the proper scope to be passed in.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/analytics_client.dart
```diff
@@ -17,18 +16,32 @@ import 'package:meta/meta.dart';
...
-    EndpointInfoStoreManager? endpointInfoStoreManager,
-  })  : _deviceContextInfoProvider = deviceContextInfoProvider,
-        _endpointInfoStoreManager = endpointInfoStoreManager ??
-            EndpointInfoStoreManager(
-              store: AmplifySecureStorageWorker(
-                config: AmplifySecureStorageConfig(
-                  scope: EndpointStorageScope.analyticsPinpoint.name,
-                ),
+    LegacyNativeDataProvider? legacyNativeDataProvider,
+    SecureStorageFactory? secureStorageFactory,
```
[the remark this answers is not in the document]
fjnoyp: Yes, AnalyticsClient will be a bit weird because it is used by Analytics and Push Notifications, and possibly more notification categories in the longer term future. 

Push Notifications will create AnalyticsClient with only storageScope and deviceContextInfoProvider defined.

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
```diff
@@ -86,9 +87,14 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface<
     implements Closeable {
   /// {@macro amplify_auth_cognito_dart.amplify_auth_cognito_dart}
   AmplifyAuthCognitoDart({
-    SecureStorageInterface? credentialStorage,
+    SecureStorageFactory? secureStorageFactory,
```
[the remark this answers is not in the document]
fjnoyp: Good point.  Depends on how we want to evolve the usage of this factory though.  

I assumed there might be other secure storage instances created, hence the reason we switched to factory.  If not, we can change back to credentialStorage, my only concern is if factory is used to create other instances and now the name becomes misleading.

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
fjnoyp: Outdated, removed for EndpointInfoStoreManager

packages/auth/amplify_auth_cognito/lib/src/auth_plugin_impl.dart
fjnoyp: Outdated, removed for EndpointInfoStoreManager

packages/auth/amplify_auth_cognito_dart/lib/src/state/state/configuration_state.dart
fjnoyp: Outdated, removed for EndpointInfoStoreManager

packages/analytics/amplify_analytics_pinpoint/example/lib/main.dart
```diff
@@ -39,21 +39,23 @@ class _MyAppState extends State<MyApp> {
...
-        /// Keychain Sharing in Xcode as described in the docs:
-        /// https://docs.amplify.aws/lib/project-setup/platform-setup/q/platform/flutter/#enable-keychain
-        // ignore: invalid_use_of_visible_for_testing_member
-        macOSOptions: MacOSSecureStorageOptions(useDataProtection: false),
-      ),
+    // FIXME: In your app, make sure to remove this line and set up
+    /// Keychain Sharing in Xcode as described in the docs:
+    /// https://docs.amplify.aws/lib/project-setup/platform-setup/q/platform/flutter/#enable-keychain
+    final storageFactory = AmplifySecureStorage.fromFactory(
+      // ignore: invalid_use_of_visible_for_testing_member
```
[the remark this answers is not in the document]
Jordan-Nelson: Ah okay.

packages/auth/amplify_auth_cognito/lib/src/auth_plugin_impl.dart
```diff
@@ -126,6 +120,11 @@ class AmplifyAuthCognito extends AmplifyAuthCognitoDart with AWSDebuggable {
     AmplifyConfig? config,
     required AmplifyAuthProviderRepository authProviderRepo,
   }) async {
+    // Dependencies for AnalyticsMetadataType
+    stateMachine.addInstance<FlutterEndpointInfoStoreManager>(
```
dnys1, line 124:
```suggestion
    stateMachine.addInstance<EndpointInfoStoreManager>(
```

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
```diff
@@ -279,6 +280,9 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface<
       );
     }
 
+    // Dependencies for AnalyticsMetadataType
+    _stateMachine.addInstance(EndpointInfoStoreManager());
```
dnys1, line 284:
```suggestion
    _stateMachine.addBuilder<EndpointInfoStoreManager>(EndpointInfoStoreManager.new);
```

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/configuration_state_machine.dart
```diff
@@ -124,4 +129,21 @@ class ConfigurationStateMachine
...
   Future<void> onConfigureSucceeded(ConfigureSucceeded event) async {}
+
+  Future<void> _registerAnalyticsMetadata(AuthConfiguration config) async {
+    final analyticsConfig = config.pinpointConfig;
+    if (analyticsConfig == null) {
+      return;
+    }
+    final appId = analyticsConfig.appId;
+
+    final endpointStoreManager = expect<EndpointInfoStoreManager>();
```
dnys1, line 140:
```suggestion
    final endpointStoreManager = getOrCreate<EndpointInfoStoreManager>();
```

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_in_state_machine.dart
```diff
@@ -238,6 +239,8 @@ class SignInStateMachine extends AuthStateMachine<SignInEvent, SignInState> {
         ..poolId = config.poolId
         ..deviceKey = user.deviceSecrets?.deviceKey
         ..challengeParameters = challengeParameters
+        // Passing in the Builder causes issues with serialization
+        ..analyticsMetadata = get<AnalyticsMetadataType>()
```
dnys1, line 243: Instead of passing it to the workers (since they don't really need it) and to each response individually, you could add it once here: https://github.com/aws-amplify/amplify-flutter/blob/cc8f3bdb049142a8171b8696d240e212196077fa/packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_in_state_machine.dart#L819

packages/auth/amplify_auth_cognito_test/test/state/configuration_state_machine_test.dart
```diff
@@ -82,5 +84,72 @@ void main() {
...
+      const testEndpointId = 'testEndpointId';
+
+      // Add state machine dependencies.
+      stateMachine.addInstance(
+        EndpointInfoStoreManager(
+          store: MockSecureStorage()
+            ..write(
+              key: testPinpointAppId + EndpointStoreKey.endpointId.name,
+              value: testEndpointId,
+            ),
```
dnys1, line 99: Remove this and call init in here instead

packages/secure_storage/amplify_secure_storage/lib/src/amplify_secure_storage.vm.dart
```diff
@@ -11,18 +11,43 @@ import 'package:amplify_secure_storage_dart/amplify_secure_storage_dart.dart';
...
 /// {@template amplify_secure_storage.amplify_secure_storage}
 /// The default Secure Storage implementation used in Amplify packages.
 /// {@endtemplate}
 class AmplifySecureStorage extends AmplifySecureStorageInterface {
-  /// {@macro amplify_secure_storage.amplify_secure_storage}
+  /// Generates a [AmplifySecureStorage] from a config.
+  ///
+  /// Should only be used within Amplify packages. See [factoryFrom] for
+  /// external use.
   AmplifySecureStorage({
```
dnys1, line 25: Mark with `@internal`

packages/secure_storage/amplify_secure_storage_dart/lib/src/amplify_secure_storage_dart.dart
```diff
@@ -42,11 +41,36 @@ class AmplifySecureStorageDart extends AmplifySecureStorageInterface
 /// {@endtemplate}
 class AmplifySecureStorageWorker extends AmplifySecureStorageInterface
     with AWSDebuggable, AWSLoggerMixin {
-  /// {@macro amplify_secure_storage_dart.amplify_secure_storage_worker}
+  /// Generates a [AmplifySecureStorageWorker] from a config.
+  ///
+  /// Should only be used within Amplify packages. See [factoryFrom] for
+  /// external use.
   AmplifySecureStorageWorker({
```
dnys1, line 48: Mark with `@internal`

packages/secure_storage/amplify_secure_storage_dart/lib/src/types/secure_storage_factory.dart
```diff
@@ -0,0 +1,21 @@
...
+
+/// A factory constructor for a [SecureStorageInterface] instance.
+typedef SecureStorageFactory = SecureStorageInterface Function(
+  AmplifyStorageScope amplifyScope,
+);
+
+/// Storage scope for config of [SecureStorageInterface] within Amplify.
+enum AmplifyStorageScope {
+  /// Storage scope for Auth plugin using Cognito.
+  awsCognitoAuthPlugin,
```
dnys1, line 14: Not in love with the naming of these. How about `auth`, `analytics`, and `pushNotifications`?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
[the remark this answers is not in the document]
Jordan-Nelson: If the goal is to prevent using AmplifySecureStorage with an improper scope, there are better ways to do that.

First, we should mark the `AmplifySecureStorage()` constructor as internal. This will discourage use of creating an instance with a scope other than AmplifyStorageScope because the only way to do so would require using an internal member. `AmplifySecureStorage.factoryFrom` only accepts `AmplifyStorageScope`.

```dart
final s = AmplifySecureStorage.factoryFrom()(
  AmplifyStorageScope.awsCognitoAuthPlugin, // a string would result in a compile time error
);
```

If we feel that this is not enough, we can update `AmplifySecureStorageConfig()` to accept type `AmplifyStorageScope`. 

`AnalyticsClient` is not the correct place to try to enforce this.

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
fjnoyp: Updated the dart and flutter level to use this.  

```
stateMachine.addBuilder<EndpointInfoStoreManager>(
      FlutterEndpointInfoStoreManager.new,
    );

_stateMachine
        .addBuilder<EndpointInfoStoreManager>(EndpointInfoStoreManager.new);
```

packages/secure_storage/amplify_secure_storage_dart/lib/src/types/secure_storage_factory.dart
Jordan-Nelson: I believe I had used `awsCognitoAuthPlugin` because that was what native was using for credential storage. We don't have to align with that, but including something related to Cognito in the name seems like a good idea to avoid collisions (even though that is not really possibly right now with only one auth plugin).

If/when we have a second auth plugin, I think we would want them to use different scopes, right? A customer probably wouldn't use two auth plugins at once, but they may migrate from one to the other.

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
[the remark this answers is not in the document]
Jordan-Nelson: Hm, I see your point. IMO the main benefit of using the factory is to "hide" to scope from the customer, but it could hypothetically be used to create other instances in the future. I think that is unlikely, but I guess the name change is fine.

packages/analytics/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
```diff
@@ -11,8 +11,7 @@ export 'package:amplify_analytics_pinpoint_dart/amplify_analytics_pinpoint_dart.
         AWSPinpointUserProfile,
 
         /// For Push Notifications.
-        // TODO(kylecheng): export via SecureStorage instead.
-        EndpointInfoStoreManager,
+        AnalyticsClient,
```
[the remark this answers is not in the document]
Jordan-Nelson: @dnys1 thoughts on how we expose classes for internal use?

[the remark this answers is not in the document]
Jordan-Nelson: This would apply to the export of flutter_device_context_info_provider below as well

packages/secure_storage/amplify_secure_storage_dart/lib/src/amplify_secure_storage_dart.dart
```diff
@@ -42,11 +41,36 @@ class AmplifySecureStorageDart extends AmplifySecureStorageInterface
 /// {@endtemplate}
 class AmplifySecureStorageWorker extends AmplifySecureStorageInterface
     with AWSDebuggable, AWSLoggerMixin {
-  /// {@macro amplify_secure_storage_dart.amplify_secure_storage_worker}
+  /// Generates a [AmplifySecureStorageWorker] from a config.
```
Jordan-Nelson, line 44: *nit:* Can use templates instead of copying the same comment in multiple places

packages/analytics/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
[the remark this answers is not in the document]
dnys1: Yeah, internal classes should not be exported- they should be imported from `/src/` and marked with `// ignore: implementation_imports`.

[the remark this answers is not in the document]
dnys1: Anything exported from the package is subject to our breaking change/deprecation policy.

packages/secure_storage/amplify_secure_storage_dart/lib/src/types/secure_storage_factory.dart
dnys1: Ah true, true.. okay, yeah maybe I'd suggest something a bit shorter then like `authCognito`. The scope string can be the full name, but maybe just for the enum so it appears cleaner where it shows up.

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
dnys1: If you update both, then we arrive at the same issue. What gets created in the code below?

```dart
stateMachine.addBuilder<EndpointInfoStoreManager>(FlutterEndpointInfoStoreManager.new);
stateMachine.addBuilder<EndpointInfoStoreManager>(EndpointInfoStoreManager.new);
final endpointManager = stateMachine.getOrCreate<EndpointInfoStoreManager>();
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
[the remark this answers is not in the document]
fjnoyp: Sounds good thanks.  Updated to accept SecureStorageInterface in AnalyticsClient.

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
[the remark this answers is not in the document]
fjnoyp: Okay, I will keep the name change as is then.  I'll keep comment open for others to consider.

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
```diff
@@ -136,11 +142,10 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface<
   /// {@macro amplify_auth_cognito.initial_parameters}
   final OAuthParameters? _initialParameters = initialParameters;
 
-  /// The on-device credential storage for the Auth category.
+  /// Factory for creating secure storage.
```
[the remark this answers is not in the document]
fjnoyp: Per above comments, going to keep the comment name change. 

Updated to include original information though: 

 ```
 /// Factory for creating secure storage.
  ///
  /// Creates an instance of [SecureStorageInterface] with scope input.
  /// 
  /// Used to create on-device credential storage. 
  final SecureStorageFactory _secureStorageFactory;

```

File: packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart (hunk 32 above)
Comment (fjnoyp) [reply packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart:284]: Of course.. thanks for pointing that out. 

I understand `getOrCreate` better now.

File: packages/secure_storage/amplify_secure_storage_dart/lib/src/types/secure_storage_factory.dart (hunk 38 above)
Comment (fjnoyp) [reply packages/secure_storage/amplify_secure_storage_dart/lib/src/types/secure_storage_factory.dart:14]: Thanks for the insights.  

We only use the enum in one place (auth_plugin_impl) so is it really that necessary to shorten? 

I prefer having the enum names mapping directly to their value when possible to reduce logic.

File: packages/analytics/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart (hunk 39 above)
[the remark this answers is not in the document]
Comment (fjnoyp) [reply packages/analytics/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart:14]: How might we handle the unique case of Auth internal usage of `EndpointInfoStoreManager`? 

Flutter Auth uses `FlutterEndpointInfoStoreManager` and `EndpointInfoStoreManager`. 

`EndpointInfoStoreManager` comes from Analytics Dart, but only Analytics Flutter is used by Flutter Auth... 

How can Flutter Auth get a Analytics Dart dependency without relying on Analytics Dart and without Analytics Flutter exporting it?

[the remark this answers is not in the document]
Comment (fjnoyp) [reply packages/analytics/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart]: Spoke with Dillon, will just import the Analytics dart component into Auth and PN to avoid unnecessary exports.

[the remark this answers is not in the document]
Comment (fjnoyp) [reply packages/analytics/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart]: Also I realized the exports of Analytics Flutter and Dart was too permissive so I redid all exports and subsequent imports.

File: packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_info_store_manager.dart [hunk 42]
```diff
@@ -9,25 +9,18 @@ import 'package:amplify_analytics_pinpoint_dart/src/impl/flutter_provider_interf
...
-}
-
 /// {@template amplify_analytics_pinpoint_dart.endpoint_info_store_manager}
 /// Manages and provides Pinpoint Endpoint id and global fields.
 /// {@endtemplate}
 class EndpointInfoStoreManager {
   /// {@macro amplify_analytics_pinpoint_dart.endpoint_info_store_manager}
   EndpointInfoStoreManager({
-    required SecureStorageInterface store,
+    SecureStorageInterface? store,
```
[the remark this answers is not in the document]
fjnoyp: It is nullable because it is an optional parameter.  If it is not set a SecureStorageWorker is created instead.  This is consistent with how Analytics worked beforehand except that the default value is now defined in `EndpointInfoStoreManager` instead of the top level Analytics class.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/flutter_provider_interfaces/device_platform.dart
```diff
@@ -0,0 +1,26 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// ignore: public_member_api_docs
```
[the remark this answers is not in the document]
fjnoyp: Okay, it is private as it's not set by developers.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_info_store_manager.dart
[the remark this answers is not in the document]
Jordan-Nelson: Why is it optional? I don't see a use of this constructor where a value is not provided. Is there a use case for that?

packages/auth/amplify_auth_cognito_test/test/state/configuration_state_machine_test.dart
dnys1: No, I mean the logic for where to store the endpoint should not be hardcoded here.

dnys1 (approved):

PR #2585, opened 2023-01-18
fix(auth)!: Fetch Auth Session offline behavior

Jordan-Nelson:
*Issue #, if available:* https://github.com/aws-amplify/amplify-flutter/issues/2398, https://github.com/aws-amplify/amplify-flutter/issues/760

*Description of changes:*
- Remove `getAWSCredentials` from Fetch Auth Session API
- Add new types, AWSResult and AuthResult, to hold the result of an operation that may have thrown.
- Update exception handling in fetch auth session state machine
- Update `CognitoAuthSession` members to type AuthResult
- Add Fetch Auth Session test cases 
- Update Authenticator to catch NetworkException.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2609, opened 2023-01-30
fix(auth): SessionExpired Auth Hub event

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*

Fixes a regression introduced by https://github.com/aws-amplify/amplify-flutter/pull/2585. When refreshing user pool tokens fail due to an expired refresh token, a hub event for Session Expired should be emitted.

### DIFF
--- a/packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
+++ b/packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
@@ -221,10 +221,13 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface<
         if (state is SignInSuccess) {
           hubEvent = AuthHubEvent.signedIn(state.user.authUser);
         }
-        if (state is FetchAuthSessionFailure &&
-            (state.exception is UnauthorizedException ||
-                state.exception is AuthNotAuthorizedException)) {
-          hubEvent = AuthHubEvent.sessionExpired();
+        if (state is FetchAuthSessionSuccess) {
+          final exception = state.session.userPoolTokensResult.exception;
+          if (exception is UnauthorizedException ||
+              exception is AuthNotAuthorizedException ||
+              exception is SessionExpiredException) {
+            hubEvent = AuthHubEvent.sessionExpired();
+          }
         }
 
         if (hubEvent != null) {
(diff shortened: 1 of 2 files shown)


packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
```diff
@@ -221,10 +221,16 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface<
         if (state is SignInSuccess) {
           hubEvent = AuthHubEvent.signedIn(state.user.authUser);
         }
-        if (state is FetchAuthSessionFailure &&
-            (state.exception is UnauthorizedException ||
-                state.exception is AuthNotAuthorizedException)) {
-          hubEvent = AuthHubEvent.sessionExpired();
+        if (state is FetchAuthSessionSuccess) {
+          final exception = state.session.userPoolTokensResult.exception;
+          // TODO(Jordan-Nelson): Update list of exceptions once FetchAuthSession
```
Jordan-Nelson, line 226: Rght now `AuthNotAuthorizedException` is thrown for an expired session, but it should be `SessionExpiredException`. I have a task to update this.

`UnauthorizedException` was in the list previously, so I left it for now. I'll look into what scenarios what is thrown in to see if this needs to be included when I take care of the task mentioned above.

packages/auth/amplify_auth_cognito_test/test/plugin/fetch_auth_session_test.dart
```diff
@@ -0,0 +1,68 @@
...
+                'Refresh Token has expired.',
+              ),
+            ),
+          ),
+        );
+      });
+
+      test('should add a sessionExpired event to Auth Hub', () async {
+        final authStream = Amplify.Hub.availableStreams[HubChannel.Auth];
+        plugin.fetchAuthSession().ignore();
```
dnys1, line 59: This might fail on DDC - it doesn't like `.ignore` in tests for some reason (maybe due to #50619 in the Dart repo). Alternative is the same but more verbose:

```dart
unawaited(
  plugin.fetchAuthSession().then((_) {}).onError((_, __) {}),
);
```

dnys1 (approved):

packages/auth/amplify_auth_cognito_test/test/plugin/fetch_auth_session_test.dart
Jordan-Nelson: They seem to be passing (at least on firefox, looks like they are not running on Chrome) so I will leave for now. If it causes an issue we can change it.

PR #2610, opened 2023-01-30
feat(storage): optimize part size for multipart upload

HuiSF:
*Issue #, if available:*

*Description of changes:*

Added `openRead` API to `AWSFile` in order to read chunks of a html File/Blob, or a io File in parallel.

As allowing part sizes to be larger than 5 MB makes caching bytes of parts in memory for uploading inefficient, this PR changes in parallel parts uploads strategy of multipart upload as following

1. If the `AWSFile` is back by a stream of bytes, and file size is known, the upload uses multipart upload, and the parts uploads will be performed one by one
2. If the `AWSFile` is backed by a stream the file size is unknown, the upload uses S3 `pubObject`
3. If the `AWSFile` is backed by html File/Blob or io File, the upload uses multipart upload, and the parts uploads can be initiated in parallel (max 4 at a time)


packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_upload_task.dart
```diff
@@ -535,11 +538,15 @@ class S3UploadTask {
     // create a new batching Completer to start the current batching
     _uploadPartBatchingCompleter = Completer();
 
-    final numToBatch = min(
-      _maxNumParallelTasks - _numOfOngoingSubtasks,
-      _expectedNumOfSubtasks -
-          (_numOfCompletedSubtasks + _numOfOngoingSubtasks),
-    );
+    // If the AWSFile is backed by a stream, then upload chunks one by one
+    // otherwise maximum 4 parts of the file can be read and uploaded in parallel
```
dnys1, line 542: Why can't a stream be uploaded in parallel?

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/part_size_util_test.dart
```diff
@@ -0,0 +1,34 @@
...
+        expect(result, part_size_util.minPartSize);
+      },
+    );
+
+    test(
+      'allow part size to be larger than ${part_size_util.minPartSize}',
+      () {
+        const testFileSize = 5 * 1024 * 1024 * part_size_util.maxNumberOfParts +
+            1000 * 1024 * 1024; // 51GiB
+        final expectedResult =
```
dnys1, line 25: Pick known quantities if you're gonna unit test simple functions like this. Otherwise, this doesn't really test anything.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_upload_task.dart
HuiSF: Stream cannot be read as chunks in parallel with random start and end positions, e.g. `ChunkedStreamReader` `readChunk` returns a Future of collected bytes that needs to be await'ed and doesn't allow calls in parallel. 

We surely can read a few chunks in advance caching the bytes in memory and then start uploads in parallel, but caching bytes in memory could lead to memory leak.

Although I could use additional optimization here, e.g. checking part size, if it's still 5MB, that we can cache 20MB bytes in memory to allow uploads in parallel.

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/part_size_util_test.dart
HuiSF: Makes sense, will update

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/part_size_util.dart
```diff
@@ -0,0 +1,23 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:math';
```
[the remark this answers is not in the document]
HuiSF: Yup meant to do so.

packages/aws_common/lib/src/io/aws_file_platform_html.dart
```diff
@@ -145,6 +142,38 @@ class AWSFilePlatform extends AWSFile {
...
       });
 
+  @override
+  ChunkedStreamReader<int> getChunkedStreamReader() {
+    return ChunkedStreamReader(stream);
+  }
+
+  @override
+  Stream<List<int>> openRead([int? start, int? end]) {
+    if (super.isStreamOnly) {
```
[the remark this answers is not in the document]
HuiSF: > Also technically you could implement this for any stream using the chunked stream reader

Yes, but it would a different behavior than `openRead` over a file, the latter allows in parallel random file access. 

If over a stream, e.g. `openRead(500, 600)`, it requires to listen to the stream, and ignore the first 499 events in order to get the 500th to start with, and after this,  `openRead(200, 300)` will be impossible correct?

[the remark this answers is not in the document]
dnys1: Gotcha- I didn't realize openRead was parallelized at the VM layer. You could use a `StreamSplitter` but it would buffer all bytes read to that point in order to allow random access.

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_upload_task_test.dart
```diff
@@ -1094,8 +1094,10 @@ void main() {
           'should throw exception if the file to be upload is too large to initiate a multipart upload',
```
ragingsquirrel3, line 1094: question: Are the existing tests in this file sufficient to ensure that certain use cases result in expected number of parts? e.g. "file size x as a Stream results in n parts, each of size z." It looks like there are already checks for that ([example](https://github.com/aws-amplify/amplify-flutter/blob/next/packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_upload_task_test.dart#L868)). Just checking if it's possible or makes sense to add checks for extra logic in this PR in `_startNextUploadPart`, in addition to the tests already here for the part size utils.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_upload_task.dart
```diff
@@ -721,22 +745,23 @@ class _CompletedSubtask {
...
     required this.request,
-  });
+  }) : _partBodyGetter = partBodyGetter;
 
-  final List<int> partBodyChunk;
+  final Stream<List<int>> Function() _partBodyGetter;
   final int partNumber;
   final Future<void> request;
 
+  Stream<List<int>> get partBody => _partBodyGetter();
```
ragingsquirrel3, line 757: Am I correct in understanding this would only get called once for this subtask? I just wasn't sure if there was some way a caller might generate 2 streams for the expected same result if called twice on same instance, in which case might make sense to memoize or change this somehow. Doesn't appear to be an issue currently, just making sure I'm not missing something.

ragingsquirrel3 (approved):
Approved w a few non-blocking questions.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_upload_task.dart
HuiSF: Thanks for the question. This is for making the part upload resumable. 

When pause upload operation, ongoing sub tasks will be canceled. The references of the sub tasks will be held in memory. When resume the upload tasks, it looks up in memory subtasks first and resume them if any, here we would need a new stream of the part content.

packages/aws_common/lib/src/io/aws_file_platform.dart
```diff
@@ -20,8 +20,10 @@ class AWSFilePlatform extends AWSFile {
     Stream<List<int>> stream, {
     super.name,
     super.contentType,
-    required int size,
-  }) : super.protected();
+    int? size = -1,
```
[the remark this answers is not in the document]
HuiSF: Updated :)

packages/aws_common/test/io/aws_file_html_test.dart
```diff
@@ -130,5 +131,74 @@ void main() {
...
+        expect(bytesBuffer.takeBytes(), equals(testBytes));
+      });
+
+      test(
+        'returns streams over the underlying file pointed by the path',
+        () async {
+          final bytesBuffer = BytesBuilder();
+          final awsFile = AWSFile.fromPath(testFilePath);
+
+          final collectedBytes1 = (await awsFile.openRead(0, 20).toList())
```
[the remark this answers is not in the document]
HuiSF: Updated to use `collectBytes` from the async package.

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_upload_task_test.dart
HuiSF: Good call, I've added two new tests, which verifies the new branching added to `_startNextUploadPart ` by asserting calling `AWSFile.getChunkedStreamReader` vs `AWSFile.openRead`.

packages/aws_common/lib/src/io/aws_file_platform_html.dart
dnys1, line 152: Is this property necessary? I don't like that it's part of the public API and am wondering if we can infer this information without it.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_upload_task.dart
```diff
@@ -206,10 +206,12 @@ class S3UploadTask {
...
-      if (_fileSize <= _minPartSize) {
-        // S3 multipart upload requires minimum part size to be 5MB. Multipart
-        // upload cannot be initiated for a file with size less than 5MB so use
-        // putObject
+      if (_fileSize > part_size_util.maxSingleObjectSize) {
+        _completeUploadWithError(S3Exception.uploadSourceIsTooLarge());
+        return;
+      }
+
+      if (_fileSize < part_size_util.minPartSize) {
```
dnys1, line 214: If _fileSize is unknown (`== -1`), wouldn't it make more sense to try a multipart upload and fallback to putobject if we reach the end before hitting the min part size?

packages/aws_common/lib/src/io/aws_file_platform_html.dart
dnys1: You still need to throw InvalidFileException at the end here, so wouldn't it make more sense to `try/catch` things in storage instead of check this flag?

HuiSF: Update now infer this info based on `InvalidFileException`.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_upload_task.dart
HuiSF: Change the `size` param back to required when instantiate with `AWSFile.fromStream`, which should eliminate this -1 implication.

We can add support of uploading from data stream with unknown size in the future if it's requested.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_upload_task.dart
```diff
@@ -573,15 +578,31 @@ class S3UploadTask {
...
     required int partNumber,
   }) async {
-    final chunk = await _fileReader.readChunk(_minPartSize);
+    late Stream<List<int>> Function() chunkGetter;
+
+    // If the AWSFile is backed by a stream of bytes, read chunks one by one
+    // using the chunked stream reader
+    if (_isAWSFileStream) {
+      final chunk = await _fileReader.readChunk(_optimalPartSize);
+      chunkGetter = () => Stream.value(chunk);
```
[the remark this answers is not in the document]
HuiSF: This is for the use case pausing -> resume, when a part upload is canceled and resumed, it needs to recreate the body stream for reuploading. The `chunkGetter` handles the recreation.

Issue #2526, opened 2023-01-03
There's no constant named 'GET' in ModelOperation constants


EricDunaway:
### Description

The documentation describes the `@auth` directive as supporting the following operations
```
  create
  update
  delete
  read # Short-hand to allow "get", "list", "sync", "listen", and "search"

  get # Retrieves an individual item
  list # Retrieves a list of items
  sync # Enables ability to sync offline/online changes (including via DataStore)
  listen # Subscribes to real-time changes
  search # Enables ability to search using @searchable directive
```
[Docs: Authorization rules](https://docs.amplify.aws/cli/graphql/authorization-rules/#how-it-works:~:text=%23%20Retrieves%20an%20individual%20item)

I get an error of `Error: Member not found: 'GET'.` when trying to build with models generated from a schema with a directive such as `@auth(rules: [{allow: owner, operations: [get]}])`

Looking at the source for the package the granular read operations are not available.
`enum ModelOperation { CREATE, UPDATE, DELETE, READ }` ([src: packages/amplify_core/lib/src/types/models/auth_rule.dart](https://github.com/aws-amplify/amplify-flutter/blob/33df283fe17f206aa7b0e22484e83b60d4d4f789/packages/amplify_core/lib/src/types/models/auth_rule.dart#L25))

### Categories

- [ ] Analytics
- [ ] API (REST)
- [X] API (GraphQL)
- [ ] Auth
- [ ] Authenticator
- [ ] DataStore
- [ ] Storage

### Steps to Reproduce

1. Create a schema using `@auth(rules: [{allow: owner, operations: [get]}])` or any of the granular read operations
2. Generate models `amplify codegen models`
3. Execute `flutter run`
4. See the build errors about the missing constants

### Screenshots

- VS Code Error
  - <img width="1445" alt="image" src="https://user-images.githubusercontent.com/4960630/210433024-291819d1-d046-46cd-b9a0-3ec55382b912.png">

- Error in terminal
  - <img width="441" alt="image" src="https://user-images.githubusercontent.com/4960630/210433223-9229de57-8747-4afe-b8f3-0b694263b268.png">

### Platforms

- [X] iOS
- [X] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Android Device/Emulator API Level

API 27

### Environment

```bash
Doctor summary (to see all details, run flutter doctor -v):
[...]

PR #2611, opened 2023-01-31
feat(core): add granular model read operations

AaronZyLee:
*Issue #, if available:*
Fix https://github.com/aws-amplify/amplify-flutter/issues/2526
The purpose of adding enum in library is to ensure the model files generated by modelgen will be compiled. However, this is not intended for Datastore usage but for those who use modelgen files along with API plugin. It is not recommended to use them in the datastore due to the prerequisite of `sync` and `listen`. This is documented in the [callout](https://github.com/aws-amplify/docs/pull/5084). 
*Description of changes:*
Add granular read operation enums in model auth operation rule
Refer https://docs.amplify.aws/cli/graphql/authorization-rules/#how-it-works

### DIFF
--- a/packages/amplify_core/lib/src/types/models/auth_rule.dart
+++ b/packages/amplify_core/lib/src/types/models/auth_rule.dart
@@ -10,7 +10,17 @@ import 'package:meta/meta.dart';
 
 enum AuthStrategy { OWNER, GROUPS, PRIVATE, PUBLIC }
 
-enum ModelOperation { CREATE, UPDATE, DELETE, READ }
+enum ModelOperation {
+  CREATE,
+  UPDATE,
+  DELETE,
+  READ,
+  GET,
+  LIST,
+  SYNC,
+  LISTEN,
+  SEARCH
+}
 
 enum AuthRuleProvider { APIKEY, OIDC, IAM, USERPOOLS, FUNCTION }
 


packages/amplify_core/lib/src/types/models/auth_rule.dart
```diff
@@ -22,7 +22,17 @@ import 'package:flutter/foundation.dart';
...
 
-enum ModelOperation { CREATE, UPDATE, DELETE, READ }
+enum ModelOperation {
+  CREATE,
+  UPDATE,
+  DELETE,
+  READ,
+  GET,
+  LIST,
+  SYNC,
```
HuiSF, line 32: Per my understanding these finer read operation controls work out of box for API plugin, are these finer controls supported in DataStore? If not, do we have a documentation calling this out - not to use these finer read control with DataStore?

AaronZyLee: These are not supported in Datastore especially for the local storage. The auth rules are only respected when it syncs with cloud. Adding these enums is to make the modelgen file compiled for those who uses these files only for API usage.

There is no document on this part but I will have a PR out later to call out that the finer read ops are not recommended for Datastore usage. If cx insists on using them, the `sync` and `listen` are required.

AaronZyLee: Update with the doc PR: https://github.com/aws-amplify/docs/pull/5084

HuiSF (dismissed):

HuiSF (dismissed):

HuiSF (approved):

ragingsquirrel3 (approved):

PR #2615, opened 2023-01-31
fix(aft): Topo sort transitive deps

dnys1:
Consider transitive dependencies (not just direct) when determining the publish precedence for packages.


Equartey (approved):
Nice! Don't see DFS in the wild too often 💯

NikaHsn (approved):

PR #2626, opened 2023-02-01
fix(auth): `federateToIdentityPool` discrepancies

dnys1:
Some discrepancies exist in our implementation of `federateToIdentityPool`:
- `fetchAuthSession` should return `isSignedIn=true` when federated to identity pool
- User Pool APIs should throw a specific `InvalidStateException`, not a generic `SignedOutException`

### DIFF
--- a/packages/auth/amplify_auth_cognito_dart/lib/amplify_auth_cognito_dart.dart
+++ b/packages/auth/amplify_auth_cognito_dart/lib/amplify_auth_cognito_dart.dart
@@ -28,7 +28,8 @@ export 'src/model/password/cognito_update_password_options.dart';
 export 'src/model/session/cognito_auth_session.dart';
 export 'src/model/session/cognito_auth_user.dart';
 export 'src/model/session/cognito_session_options.dart';
-export 'src/model/session/cognito_sign_in_details.dart';
+export 'src/model/session/cognito_sign_in_details.dart'
+    hide CognitoSignInDetailsFederated;
 export 'src/model/session/cognito_user_pool_tokens.dart';
 export 'src/model/session/federate_to_identity_pool_options.dart';
 export 'src/model/session/federate_to_identity_pool_request.dart';
--- a/packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
+++ b/packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
@@ -11,6 +11,7 @@ import 'package:amplify_auth_cognito_dart/src/flows/helpers.dart';
 import 'package:amplify_auth_cognito_dart/src/flows/hosted_ui/initial_parameters_stub.dart'
     if (dart.library.html) 'package:amplify_auth_cognito_dart/src/flows/hosted_ui/initial_parameters_html.dart';
 import 'package:amplify_auth_cognito_dart/src/model/auth_user_ext.dart';
+import 'package:amplify_auth_cognito_dart/src/model/session/cognito_sign_in_details.dart';
 import 'package:amplify_auth_cognito_dart/src/model/sign_in_parameters.dart';
 import 'package:amplify_auth_cognito_dart/src/model/sign_up_parameters.dart';
 import 'package:amplify_auth_cognito_dart/src/sdk/cognito_identity_provider.dart'
@@ -927,8 +928,14 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface<
     AuthUserOptions? options,
   }) async {
     final credentials = await stateMachine.loadCredentials();
-    final tokens = credentials.userPoolTokens;
     final signInDetails = credentials.signInDetails;
+    // Per the `federateToIdentityPool` design, users cannot access user pool
+    // methods while federated. They must first clear federation to the
+    // identity pool and then sign into the user pool normally.
+    if (signInDetails is CognitoSignInDetailsFederated) {
+      throw const InvalidStateException.federatedToIdentityPool();
+    }
+    final tokens = credentials.userPoolTokens;
     if (tokens == null || signInDetails == null) {
       throw const SignedOutException('No user is currently signed in');
     }
@@ -1050,7 +1057,7 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface<
     final CognitoUserPoolTokens tokens;
     try {
       tokens = await getUserPoolTokens();
-    } on AuthException {
+    } on SignedOutException {
       _hubEventController.add(AuthHubEvent.signedOut());
       return const CognitoSignOutResult.complete();
     }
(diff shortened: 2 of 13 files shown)


packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/credential_store_state_machine.dart
```diff
@@ -208,6 +209,14 @@ class CredentialStoreStateMachine extends StateMachine<CredentialStoreEvent,
           expiration,
         );
       }
+      final signInDetailsStr = await _secureStorage.read(
```
NikaHsn, line 212: apiBased and hostedUi do not store the signInDetails json string and instead store specified fields. would it be better to do the same for federated?

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/credential_store_state_machine.dart
```diff
@@ -292,6 +301,12 @@ class CredentialStoreStateMachine extends StateMachine<CredentialStoreEvent,
           deletions.add(keys[CognitoIdentityPoolKey.expiration]);
         }
       }
+      if (signInDetails is CognitoSignInDetailsFederated) {
```
NikaHsn, line 304: same as above

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/fetch_auth_session_state_machine.dart
```diff
@@ -263,7 +260,8 @@ class FetchAuthSessionStateMachine extends StateMachine<FetchAuthSessionEvent,
     emit(
       FetchAuthSessionState.success(
         CognitoAuthSession(
-          isSignedIn: userPoolTokens != null,
+          isSignedIn: userPoolTokens != null ||
+              result.signInDetails is CognitoSignInDetailsFederated,
```
NikaHsn, line 264: if `userPoolTokens == null && result.signInDetails is CognitoSignInDetailsFederated` then `isSignedIn=true`.
however line 284 checks for `if userPoolTokens == null` to set the results to signedOut. is this a wanted behaviour?

packages/auth/amplify_auth_cognito_dart/lib/amplify_auth_cognito_dart.dart
```diff
@@ -28,7 +28,8 @@ export 'src/model/password/cognito_update_password_options.dart';
 export 'src/model/session/cognito_auth_session.dart';
 export 'src/model/session/cognito_auth_user.dart';
 export 'src/model/session/cognito_session_options.dart';
-export 'src/model/session/cognito_sign_in_details.dart';
+export 'src/model/session/cognito_sign_in_details.dart'
+    hide CognitoSignInDetailsFederated;
```
NikaHsn, line 32: Q: why do we want to hide `CognitoSignInDetailsFederated` given it's peers `CognitoSignInDetailsApiBased` and   `CognitoSignInDetailsHostedUi` are not hidden

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/fetch_auth_session_state_machine.dart
dnys1: It is correct, unfortunately. This is a very confusing API, but basically the way to think about it is that we have a single API which wraps Cognito user pools and identity pools. User pools actually have a user directory and manage users, whereas identity pools just vend AWS credentials and allow authenticating with external OIDC providers.

When calling `federateToIdentityPool`, you're doing the latter. You're taking a Google OIDC token and saying hey give me some AWS credentials - I am who I say I am. In this case, all the user management APIs in our library are shutdown since we cannot change user info on the Google profile.

The choice to have `isSignedIn=true` under federation is not something I agree with, but is the design for this API. That's what this line checks - `isSignedIn` means authenticated via a user pool or authenticated via a third-party in the identity pool.

The check on line 284 is a different requirement - that you cannot authenticate with _both_ a user pool and an identity pool. If you're already logged into a user pool, line 284 stops any attempt to federate into an identity pool.

packages/auth/amplify_auth_cognito_dart/lib/amplify_auth_cognito_dart.dart
dnys1: I wanted to leverage the SignInDetails infrastructure for determining `isSignedIn`, but the Auth group wanted the `getUser` API to throw when federated (which is where this information is returned to users). So currently it's just an implementation detail, but I thought it the cleanest way to track this information.

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/credential_store_state_machine.dart
dnys1: Good question. I can change it to match the others. There is a tradeoff between time spent reading values and time spent decoding values - I think consistency may be a bigger win though.

Equartey (approved):

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
```diff
@@ -929,6 +930,9 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface<
     final credentials = await stateMachine.loadCredentials();
     final tokens = credentials.userPoolTokens;
     final signInDetails = credentials.signInDetails;
+    if (signInDetails is CognitoSignInDetailsFederated) {
+      throw const InvalidStateException.federatedToIdentityPool();
+    }
```
NikaHsn, line 935:
```suggestion
    final signInDetails = credentials.signInDetails;
    // can not access user pool data when using federation to a Cognito Identity Pool 
    if (signInDetails is CognitoSignInDetailsFederated) {
      throw const InvalidStateException.federatedToIdentityPool();
    }
    final tokens = credentials.userPoolTokens;
```

packages/auth/amplify_auth_cognito_dart/lib/amplify_auth_cognito_dart.dart
NikaHsn: understood. so we can also hide `CognitoSignInDetailsApiBased` and `CognitoSignInDetailsHostedUi`.

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/fetch_auth_session_state_machine.dart
NikaHsn: thanks for the detailed explanation. do we have test cases for cognito federated identity in `packages/auth/amplify_auth_cognito_test/test/state/fetch_auth_session_state_machine_test.dart`?

dnys1: Good catch! Meant to do this. Added 👍

PR #2629, opened 2023-02-02

dnys1:
- Runs `aft deps update`
- Runs `build_runner` in all packages
- Updates `db_common` to align with latest `sqlite3_flutter_libs` changes
- Updates `pigeon` version


HuiSF (approved):

packages/aft/lib/src/commands/deps_command.dart
```diff
@@ -235,6 +275,8 @@ class _DepsUpdateCommand extends _DepsSubcommand {
       exitCode = 1;
     }
 
-    await _run(_DepsAction.apply);
+    if (hasUpdates && failedUpdates.isEmpty) {
+      await _run(_DepsAction.apply);
```
fjnoyp, line 279: Do we need to warn here if failedUpdates was not empty?

fjnoyp (approved):
One minor non blocking comment/question.  Looks good otherwise.

packages/aft/lib/src/commands/deps_command.dart
dnys1: Yep, happens on the line above

PR #2630, opened 2023-02-02
feat(storage): enable pause and cancel APIs for download operations

HuiSF:
*Issue #, if available:*

*Description of changes:*

Functionality of canceling a download operation depends on https://github.com/aws-amplify/amplify-flutter/pull/2545.

### DIFF
--- a/packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_download_task_test.dart
+++ b/packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_download_task_test.dart
@@ -228,13 +228,18 @@ void main() {
 
     group('pause API()', () {
       test('it should pause the task', () async {
+        var bodyStreamHasBeenCanceled = false;
         const testOptions = S3DownloadDataOptions();
         final testGetObjectOutput = GetObjectOutput(
           contentLength: Int64(1024),
           body: Stream<List<int>>.periodic(
             const Duration(microseconds: 200),
             (_) => [101],
-          ).take(1024),
+          ).take(1024).asBroadcastStream(
+            onCancel: (StreamSubscription<List<int>> subscription) {
+              bodyStreamHasBeenCanceled = true;
+            },
+          ),
         );
         final smithyOperation = MockSmithyOperation<GetObjectOutput>();
         final receivedState = <S3TransferState>[];
@@ -262,10 +267,13 @@ void main() {
           },
         );
 
+        // pause only takes effect after `start` gets called.
+        // pause can be called without having to wait for start to complete.
         unawaited(downloadTask.start());
         await downloadTask.pause();
 
         expect(receivedState.last, S3TransferState.paused);
+        expect(bodyStreamHasBeenCanceled, isTrue);
       });
     });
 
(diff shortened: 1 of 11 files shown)


packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_download_task_test.dart
```diff
@@ -266,6 +271,7 @@ void main() {
         await downloadTask.pause();
```
dnys1, line 271: Why is `start` not awaited on the line above?

dnys1 (approved):

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_download_task_test.dart
HuiSF: `task.start` is an async function that creates async actions internally, e.g.

* prefix resolution
* Getting response from the `getObject` call


It's called when creating a download operation, and so is `unawaited`, as it should not block returning the created operation. The final state of `task.start` is determined by `await task.result.`


`task.pause` internally awaits the async action (if any) that is created by `task.start.` Using `unawaited` (task.start)  here replicates the actual flow in the implementation.

dnys1: Isn't it more common for a user to call `pause` sometime after `start` has completed?

HuiSF: A user is not aware of the existence of `task.start`, even the task itself is hidden and wrapped by the operation.

```dart
final operation = Amplify.Storage.downloadData(...); // creates a _task, and call _task.start()

// a developer can call operation.pause() any time after getting the operation reference 
operation.pause(); // calls _task.pause() and
                   // the pause logic starts executing only when the async
                   // actions created by _task.start() complete.
```

ragingsquirrel3 (approved):

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_download_task_test.dart
dnys1: Okay. Just wondering if waiting for the task to start makes a difference then? Ideally, it shouldn't right?

HuiSF: Correct, it doesn't. The behaviors of pausing on the following

```dart
unawaited(downloadTask.start());
await downloadTask.pause();
// OR
await downloadTask.start();
await downloadTask.pause();
```
are identical.

dnys1: Cool - can we change it to the latter then? 😅 `unawaited` makes it seem like you're specifically trying to pause before start finishes.

HuiSF: I'd like to keep it as is, as this is how it's being called when creating the operation :D will add a comment to clarify 👍

PR #2631, opened 2023-02-02
feat(storage): web implementation of transfer database using local storage

NikaHsn:
*Issue #, if available:*

*Description of changes:*
Add web implementation of TransferDataBase to stop using drift in web.
- add TransferDatabase stub
- add TransferData class
- update native implementation to implement the TransferDatabase stub
- implement TransferDatabase for web using local storage
- add unit tests 

### DIFF
--- a/packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/database_html.dart
+++ b/packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/database_html.dart
@@ -1,0 +1,58 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:html';
+
+import 'package:amplify_core/amplify_core.dart';
+import 'package:amplify_storage_s3_dart/src/storage_s3_service/transfer/database/database_stub.dart'
+    as stub;
+import 'package:amplify_storage_s3_dart/src/storage_s3_service/transfer/database/transfer_record.dart';
+
+/// a key prefix used by [TransferDatabase] when adding a key value pair into local storage.
+/// it is an identifier to differentiate the data stored by [TransferDatabase] into local storage from others.
+const _localStorageKeyPrefix = 'amplify_storage_transfer_data_';
+
+/// {@macro amplify_storage_s3_dart.transfer_database}
+class TransferDatabase implements stub.TransferDatabase {
+  /// {@macro amplify_storage_s3_dart.transfer_database_token}
+  static const token = Token<TransferDatabase>();
+
+  Iterable<MapEntry<String, String>> get _entries => window.localStorage.entries
+      .where((element) => element.key.startsWith(_localStorageKeyPrefix));
+
+  @override
+  Future<int> deleteTransferRecords(String uploadId) async {
+    var result = 0;
+    for (final element in _entries) {
+      final transferRecord = TransferRecord.fromJsonString(element.value);
+      if (transferRecord.uploadId == uploadId) {
+        window.localStorage.remove(element.key);
+        result++;
+      }
+    }
+
+    return result;
+  }
+
+  @override
+  Future<List<TransferRecord>> getMultipartUploadRecordsCreatedBefore(
+    DateTime dateTime,
+  ) async {
+    final result = <TransferRecord>[];
+    for (final element in _entries) {
+      final value = element.value;
+      final transferRecord = TransferRecord.fromJsonString(value);
+      if (transferRecord.createdAt.isBefore(dateTime)) {
+        result.add(transferRecord);
+      }
+    }
+    return result;
+  }
+
+  @override
+  Future<String> insertTransferRecord(TransferRecord data) async {
+    final key = _localStorageKeyPrefix + uuid();
+    window.localStorage[key] = data.toJsonString();
+    return key;
+  }
+}
--- a/packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/database_io.dart
+++ b/packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/database_io.dart
@@ -4,22 +4,24 @@
 import 'package:amplify_core/amplify_core.dart';
 import 'package:amplify_db_common_dart/amplify_db_common_dart.dart'
     as db_common;
+import 'package:amplify_storage_s3_dart/src/storage_s3_service/transfer/database/database_stub.dart'
+    as stub;
 import 'package:amplify_storage_s3_dart/src/storage_s3_service/transfer/database/tables.dart';
+import 'package:amplify_storage_s3_dart/src/storage_s3_service/transfer/database/transfer_record.dart'
+    as data;
 import 'package:drift/drift.dart';
 import 'package:meta/meta.dart';
 
-part 'database.g.dart';
+part 'database_io.g.dart';
 
 /// The name of database file created to persist [TransferRecords].
 @internal
 const dataBaseName = 'amplify_storage_transfer_records';
 
-/// {@template amplify_storage_s3_dart.transfer_database}
-/// Database to persist [TransferRecords] created by Storage S3 plugin upload
-/// operations.
-/// {@endtemplate}
+/// {@macro amplify_storage_s3_dart.transfer_database}
 @DriftDatabase(tables: [TransferRecords])
-class TransferDatabase extends _$TransferDatabase {
+class TransferDatabase extends _$TransferDatabase
+    implements stub.TransferDatabase {
   /// {@macro amplify_storage_s3_dart.transfer_database}
   TransferDatabase(
     db_common.Connect connect,
@@ -31,7 +33,7 @@ class TransferDatabase extends _$TransferDatabase {
           ),
         );
 
-  /// Token to identify [TransferDatabase] instance.
+  /// {@macro amplify_storage_s3_dart.transfer_database_token}
   static const token = Token<TransferDatabase>(
     [Token<db_common.Connect>(), Token<AppPathProvider>()],
   );
@@ -40,9 +42,8 @@ class TransferDatabase extends _$TransferDatabase {
   @override
   int get schemaVersion => 1;
 
-  /// Gets a list of multiple upload records that are created earlier than
-  /// [dateTime].
-  Future<List<TransferRecord>> getMultipartUploadRecordsCreatedBefore(
+  @override
+  Future<List<data.TransferRecord>> getMultipartUploadRecordsCreatedBefore(
     DateTime dateTime,
   ) {
     return (select(transferRecords, distinct: true)
@@ -51,16 +52,28 @@ class TransferDatabase extends _$TransferDatabase {
               dateTime.toIso8601String(),
             ),
           ))
+        .map(
+          (e) => data.TransferRecord(
+            objectKey: e.objectKey,
+            uploadId: e.uploadId,
+            createdAt: DateTime.parse(e.createdAt),
+          ),
+        )
         .get();
   }
 
-  /// Inserts [entry], a [TransferRecordsCompanion], to the database.
-  Future<int> insertTransferRecord(TransferRecordsCompanion entry) {
-    return into(transferRecords).insert(entry);
+  @override
+  Future<String> insertTransferRecord(data.TransferRecord record) async {
+    final entry = TransferRecordsCompanion.insert(
+      uploadId: record.uploadId,
+      objectKey: record.objectKey,
+      createdAt: record.createdAt.toIso8601String(),
+    );
+    final value = await into(transferRecords).insert(entry);
+    return value.toString();
   }
 
-  /// Deletes [TransferRecords] from the database where
-  /// [TransferRecords.uploadId] is equal to [uploadId].
+  @override
   Future<int> deleteTransferRecords(String uploadId) {
     return (delete(transferRecords)
           ..where((tbl) => tbl.uploadId.equals(uploadId)))
--- a/packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/database_stub.dart
+++ b/packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/database_stub.dart
@@ -1,0 +1,49 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'package:amplify_core/amplify_core.dart';
+import 'package:amplify_storage_s3_dart/src/storage_s3_service/transfer/database/transfer_record.dart';
+
+/// {@template amplify_storage_s3_dart.transfer_database}
+/// Database to persist [TransferRecord] created by Storage S3 plugin upload
+/// operations.
+/// {@endtemplate}
+class TransferDatabase {
+  /// {@template amplify_storage_s3_dart.transfer_database_token}
+  /// The dependency token for [TransferDatabase].
+  /// {@endtemplate}
+  static Token<TransferDatabase> token = const Token<TransferDatabase>();
+
+  /// {@template amplify_storage_s3_dart.transfer_database_get}
+  /// Gets a list of multipart upload records that are created earlier than
+  /// [dateTime].
+  /// {@endtemplate}
+  Future<List<TransferRecord>> getMultipartUploadRecordsCreatedBefore(
+    DateTime dateTime,
+  ) {
+    throw UnimplementedError(
+      'getMultipartUploadRecordsCreatedBefore method is not implemented in the current platform.',
+    );
+  }
+
+  /// {@template amplify_storage_s3_dart.transfer_database_insert}
+  /// Inserts [record], a [TransferRecord], to the database.
+  /// returns the key for the inserted record into the database.
+  /// {@endtemplate}
+  Future<String> insertTransferRecord(TransferRecord record) {
+    throw UnimplementedError(
+      'insertTransferRecord method is not implemented in the current platform.',
+    );
+  }
+
+  /// {@template amplify_storage_s3_dart.transfer_database_delete}
+  /// Deletes [TransferRecord]s from the database where
+  /// [TransferRecord.uploadId] is equal to [uploadId].
+  /// returns number of recordes deleted from the database.
+  /// {@endtemplate}
+  Future<int> deleteTransferRecords(String uploadId) {
+    throw UnimplementedError(
+      'deleteTransferRecords method is not implemented in the current platform.',
+    );
+  }
+}
--- a/packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/transfer_record.dart
+++ b/packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/transfer_record.dart
@@ -1,0 +1,48 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:convert';
+
+import 'package:json_annotation/json_annotation.dart';
+
+part 'transfer_record.g.dart';
+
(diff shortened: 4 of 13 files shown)


packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/database_html.dart
```diff
@@ -0,0 +1,72 @@
...
+      if (DateTime.parse(transferData.createdAt).isBefore(dateTime)) {
+        result.add(transferData);
+      }
+    });
+    return result;
+  }
+
+  @override
+  Future<String> insertTransferRecord(TransferData data) async {
+    final key = localStorageKeyPrefix + uuid();
```
dnys1, line 67: If you use the uploadId as the key you can make some performance improvements in `delete`.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/database_html.dart
```diff
@@ -0,0 +1,72 @@
...
+    final entry = {key: data.toJsonString()};
+    localStorage.addAll(entry);
```
dnys1, line 69:
```suggestion
    localStorage[key] = data.toJsonString();
```

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/database_io.dart
```diff
@@ -51,16 +52,31 @@ class TransferDatabase extends _$TransferDatabase {
...
+      createdAt: data.createdAt,
+    );
+    return into(transferRecords)
+        .insert(entry)
+        .then((value) => value.toString());
   }
 
-  /// Deletes [TransferRecords] from the database where
-  /// [TransferRecords.uploadId] is equal to [uploadId].
+  /// {@macro amplify_storage_s3_dart.transfer_database_delete}
```
dnys1, line 78: Overrides don't need doc comments since they'll inherit them from the super method.

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/transfer/database_html_test.dart
```diff
@@ -0,0 +1,104 @@
...
+      createdAt: testCreatedAt,
+    );
+
+    final testTransferDataJsonString = testTransferData.toJsonString();
+
+    late MockStorage mockStorage;
+    late MockTransferDatabase transferDatabase;
+
+    setUp(() {
+      mockStorage = MockStorage();
```
dnys1, line 26: Can you use a fake instead? Just a wrapper over a Map.

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/transfer/transfer_data_test.dart
```diff
@@ -0,0 +1,57 @@
+import 'package:amplify_storage_s3_dart/src/storage_s3_service/transfer/database/transfer_data.dart';
+import 'package:test/test.dart';
+
+void main() {
```
dnys1, line 4: Not sure I find this file useful - you're basically testing the generated JSON code and the `jsonEncode` function which I think we can take for granted.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/database_html.dart
```diff
@@ -0,0 +1,72 @@
...
+import 'package:amplify_storage_s3_dart/src/storage_s3_service/transfer/database/database_stub.dart'
+    as stub;
+import 'package:amplify_storage_s3_dart/src/storage_s3_service/transfer/database/transfer_data.dart';
+import 'package:meta/meta.dart';
+
+@visibleForTesting
+
+/// a key prefix used by [TransferDatabase] when adding a key value pair into local storage.
+/// it is an identifier to differentiate the data stored by [TransferDatabase] into local storage from others.
+const localStorageKeyPrefix = 'amplify_storage_transfer_data_';
```
HuiSF, line 13: `localStorageKeyPrefix` could be private as it's only used in the Web impl.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/transfer.dart
```diff
@@ -1,4 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
HuiSF: Used `transfer` here initially thinking we'd have `transfer_manager` aside `transfer_database` to support background operation functionalities.

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/transfer/database_html_test.dart
```diff
@@ -0,0 +1,104 @@
...
+    });
+
+    test('insertTransferRecord to local storage', () async {
+      when(
+        () => mockStorage.addAll(any()),
+      ).thenAnswer((_) => {});
+
+      final recordId =
+          await transferDatabase.insertTransferRecord(testTransferData);
+      expect(recordId.startsWith(transferDatabase.prefix), true);
```
HuiSF, line 37: Could use matchers instead of the raw values.

```suggestion
      expect(recordId.startsWith(transferDatabase.prefix), isTrue);
```

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/database_html.dart
NikaHsn: it is used in tests. I can make it private and add a const to the test with same value. between the two options I think using public const with @visibleForTesting is more readable and easier to maintain. wdyt?

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/database_html.dart
NikaHsn: uploadId may not be unique so using a prefix and uuid is to make sure the key is unique.

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/transfer/transfer_data_test.dart
NikaHsn: you are right, I can delete it if you think having it adds maintenance burden. Having it would be a safeguard for future development.

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/transfer/database_html_test.dart
```diff
@@ -0,0 +1,64 @@
...
+      transferDatabase = TransferDatabase();
+      window.localStorage.clear();
+      window.localStorage['${localStorageKeyPrefix}test-uuid'] =
+          testTransferRecordJsonString;
+    });
+
+    test('insertTransferRecord to local storage', () async {
+      final recordId =
+          await transferDatabase.insertTransferRecord(testTransferRecord);
+      expect(recordId.startsWith(localStorageKeyPrefix), isTrue);
```
HuiSF, line 37: Looking at the impl. of `insertTransferRecord` on Web, the returned `recordId` wasn't actually read from the storage, hence, I think this test can be improved a bit -

In addition to verify the `recordId`, after inserting, we can read it from the storage and compare the result.

HuiSF (approved):

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/database_html.dart
```diff
@@ -0,0 +1,61 @@
...
+
+import 'dart:html';
+
+import 'package:amplify_core/amplify_core.dart';
+import 'package:amplify_storage_s3_dart/src/storage_s3_service/transfer/database/database_stub.dart'
+    as stub;
+import 'package:amplify_storage_s3_dart/src/storage_s3_service/transfer/database/transfer_record.dart';
+import 'package:meta/meta.dart';
+
+@visibleForTesting
```
dnys1, line 12: Annotations should go below doc comments

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/transfer_record.dart
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  /// The multipart upload id.
+  final String uploadId;
+
+  /// The object key associated with the [uploadId].
+  final String objectKey;
+
+  /// Timestamp of [uploadId] creation.
+  final String createdAt;
```
dnys1, line 41: `json_serializable` can handle a `DateTime` here which will help clean up the DateTime handling throughout

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/database_stub.dart
```diff
@@ -0,0 +1,49 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import 'package:amplify_core/amplify_core.dart';
+import 'package:amplify_storage_s3_dart/src/storage_s3_service/transfer/database/transfer_record.dart';
+
+/// {@template amplify_storage_s3_dart.transfer_database}
+/// Database to persist [TransferRecord] created by Storage S3 plugin upload
+/// operations.
+/// {@endtemplate}
+class TransferDatabase {
```
dnys1, line 11: Suggest making this `abstract` and removing method bodies.

```suggestion
abstract class TransferDatabase {
```

The only reason we would choose to add method bodies w/ `UnimplementedError` would be if there was a risk of breaking customers by adding APIs. In contrast, `abstract` w/out bodies makes it breaking when adding/removing APIs and forces implementations to implement all methods which I think we want here.

packages/storage/amplify_storage_s3_dart/pubspec.yaml
```diff
@@ -30,5 +31,6 @@ dev_dependencies:
   build_verify: ^3.0.0
   built_value_generator: 8.4.2
   drift_dev: ^2.2.0+1
+  json_serializable: ^6.5.4
```
dnys1, line 34: Should be pinned since updating it potentially changes the generated code and we track that. Can run `aft deps apply` to get the latest.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/database_io.dart
```diff
@@ -51,16 +52,29 @@ class TransferDatabase extends _$TransferDatabase {
...
+  @override
+  Future<String> insertTransferRecord(data.TransferRecord record) {
+    final entry = TransferRecordsCompanion.insert(
+      uploadId: record.uploadId,
+      objectKey: record.objectKey,
+      createdAt: record.createdAt,
+    );
+    return into(transferRecords)
+        .insert(entry)
+        .then((value) => value.toString());
```
dnys1, line 74: Prefer async/await when in an async context. Control flow is more clear.

```suggestion
    final value = await into(transferRecords).insert(entry);
    return value.toString();
```

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/transfer/database_html_test.dart
```diff
@@ -0,0 +1,66 @@
...
+
+@TestOn('browser')
+import 'dart:html';
+
+import 'package:amplify_storage_s3_dart/src/storage_s3_service/transfer/database/database_html.dart';
+import 'package:amplify_storage_s3_dart/src/storage_s3_service/transfer/database/transfer_record.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('TransferDatabase for web', () {
```
dnys1, line 12: These tests make the code a bit brittle since they test implementation and cannot be run out of order (are inter-dependent). I would recommend some functional testing, i.e. tests which can run on Web or VM and only test the exposed API of `TransferDatabase`, and making tests self-contained so they are not dependent on any test running before/after it.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/database_io.dart
NikaHsn: [effective dart](https://dart.dev/guides/language/effective-dart/usage#avoid-using-completer-directly) considers both async/await and future.then as good style patterns. I might be missing other style guides that we should follow so please lmk.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/database_stub.dart
NikaHsn: I don't think we can use dependancy manager with abstract class

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/transfer_record.dart
NikaHsn: using String type because the drift table uses TextColumn for this field. 
I agree that the code would abe cleaner if using the DateTime her and hence I gave it a try.
Due to changing the drift table schema we need to bump the schema version and add MigrationStrategy. after making thses changes I feel like changing the field type from string to DateTime adds complexity to the native impl and am not convinced if we should do it. @HuiSF wdyt?

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/transfer/database_html_test.dart
NikaHsn: tests can be run out of order and have not dependency on each other. do they fail for you?

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/database_io.dart
dnys1: Both are fine, but async/await is preferred: https://dart.dev/guides/language/effective-dart/usage#prefer-asyncawait-over-using-raw-futures

dnys1: I would say only use `.then` in situations where you explicitly want to avoid async/await, like when you want to fire-and-forget a future. But 95% of the time, async/await is easier and will make the code more readable.

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/transfer/database_html_test.dart
dnys1: Gotcha - I was doing something wrong. The other point still holds, though. We are testing implementation which means the impl cannot easily change.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/transfer_record.dart
dnys1: I don't follow - in the Drift code you create a `TransferRecordsCompanion` from this anyway, so the table could remain a string.

HuiSF: SQLite DB doesn't have a DateTime type, the values of date time can be stored as TEXT, REAL or INTEGER. If we prefer using DateTime in the library types, we could call `createdAt.toIso8601String()` passing value to `TransferRecordsCompanion` while creating a DB entry.

If we make  `createdAt` `DateTime` type in `TransferRecord` to let json serializer handle the type conversion, I guess it will generate something like

```
createdAt: json['createdAt'] == null
          ? null
          : DateTime.parse(json['createdAt'] as String),
```
So it should still be compatible with the string value read from SQLite db?

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/database_stub.dart
dnys1: Hmm it's a bit confusing since this class acts as both the interface and the default implementation. That alone is not a pattern I would suggest, but I suppose is fine for this PR.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/database_io.dart
NikaHsn: thanks for sharing the ref and explanation. will update the code

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/transfer/database/transfer_record.dart
NikaHsn: I thought we want to cleanup the DateTime handling and use DateTime type instead of string throughout. means updating the `TextColumn createdAt` to `DateTimeColumn createdAt` in tables.dart and use the drift builder option to handle the date time values as text in build.yaml
```
builders:
      drift_dev:
        options:
          store_date_time_values_as_text: true
```
per latest comments we want to just use DateTime for the `TransferRecord` and keep the drift table as is and convert DateTime to String for drift impl.

dnys1: I'm most concerned about having conversion logic interspersed with the business logic, e.g. https://github.com/aws-amplify/amplify-flutter/pull/2631/files#diff-842ab00c5bcfb4a64fb7459e3815b98ad02db3442f22e2a0d01e2cf4dce20f1aR48

Moving the conversion logic to the Drift code and/or changing the table type make sense to me, but the latter is most important.

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/transfer/database_html_test.dart
NikaHsn: in terms of testing functionality and not implementation I beleive all tests here test for functionality exept the part at line 37 which checks `expect(recordId.startsWith(localStorageKeyPrefix), isTrue)`. I agree it is testing implementation and will clean up. I wanted to make sure this is the one that you referring to.

dnys1: `localStorage` is an implementation detail. Functional tests should allow swapping out implementation with IndexedDB or whatever without changes to the tests.

NikaHsn: we initially mocked the storage and then changed to use window.localStorage, https://github.com/aws-amplify/amplify-flutter/pull/2631#discussion_r1096107807

dnys1: Yeah, sorry for the confusion... this is fine for now. Overall, what I mean is that with a test like this:

```dart
test('insertTransferRecord to local storage', () async {
  final recordId =
      await transferDatabase.insertTransferRecord(testTransferRecord);
  final actual = window.localStorage[recordId];
  expect(actual, testTransferRecordJsonString);
});
```

It's tied to the implementation. So in order to swap out `localStorage` for something else inside the database, you would need to update this test, which introduces a maintenance burden.

Compare that to the rest of the tests which do not reference `localStorage` but work solely off the public interface. With just those tests, I think you get the same level of assurance in terms of behavior, but you are free to change the implementation at any point. And then also these tests could be reused for the IO database, again reducing the maintenance burden of two parallel sets of tests.

I will leave it up to you - these are just my two cents.

dnys1 (approved):
LGTM!

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/transfer/database_html_test.dart
NikaHsn: updated test to not use localStorage

PR #2482, opened 2022-12-13
refactor(auth)!: State machine facade

dnys1:
- Renames `AuthStateMachine` to `ConfigurationStateMachine`
- Refactors the Cognito state machine facade a bit to decrease coupling between the different sub state machines.
- Adds a distinct queue for the plugin so that public API calls are serially executed whereas internal state machine calls can be concurrently executed to avoid deadlock.
- Consolidates state machine API to encourage proper usage.

PR #2635, opened 2023-02-03
chore(auth): State machine cleanup

dnys1:
Clean up to address comments on #2482.

### DIFF
--- a/packages/auth/amplify_auth_cognito_test/test/state/credential_store_state_machine_test.dart
+++ b/packages/auth/amplify_auth_cognito_test/test/state/credential_store_state_machine_test.dart
@@ -38,7 +38,7 @@ void main() {
     // Load an empty credential store.
     test('loadCredentialStore (empty)', () async {
       stateMachine.dispatch(
-        const CredentialStoreEvent.migrateLegacyCredentialStore(),
+        const CredentialStoreEvent.loadCredentialStore(),
       );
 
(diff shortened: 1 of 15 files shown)


packages/auth/amplify_auth_cognito_test/test/state/credential_store_state_machine_test.dart
```diff
@@ -38,16 +38,16 @@ void main() {
     // Load an empty credential store.
     test('loadCredentialStore (empty)', () async {
       stateMachine.dispatch(
-        const CredentialStoreEvent.migrateLegacyCredentialStore(),
+        const CredentialStoreEvent.loadCredentialStore(),
```
ragingsquirrel3, line 41: I think I missed the comment(s) in first PR that is the sets the context for changes in this PR. Do you mind linking that or summarizing why this is changed?

dnys1: https://github.com/aws-amplify/amplify-flutter/pull/2482#discussion_r1095075656

The event is not really necessary and so this helps trim down the API and makes `manager.loadCredentials` single-purpose. Basically, it's easier to do the right thing always.

ragingsquirrel3: thanks

Jordan-Nelson (approved):

PR #2638, opened 2023-02-08
feat(analytics): refactor event/endpoint clients

fjnoyp:
Allow for modules to be exported for use in the Push Notification category

Cleaned the internal initialization logic of Event and Endpoint Clients to allow them to be created easily without needing to create their sub dependencies separately. 

Exported necessary classes from Analytics dart and flutter packages. 

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/analytics/amplify_analytics_pinpoint/example/macos/Runner/DebugProfile.entitlements
+++ b/packages/analytics/amplify_analytics_pinpoint/example/macos/Runner/DebugProfile.entitlements
@@ -10,9 +10,5 @@
 	<true/>
 	<key>com.apple.security.network.server</key>
 	<true/>
-	<key>keychain-access-groups</key>
-	<array>
-		<string>$(AppIdentifierPrefix)com.amazonaws.amplify.amplify-analytics-pinpoint-example</string>
-	</array>
 </dict>
 </plist>
--- a/packages/analytics/amplify_analytics_pinpoint/example/macos/Runner/DebugProfile.entitlements
+++ b/packages/analytics/amplify_analytics_pinpoint/example/macos/Runner/DebugProfile.entitlements
@@ -10,5 +10,9 @@
 	<true/>
 	<key>com.apple.security.network.server</key>
 	<true/>
+	<key>keychain-access-groups</key>
+	<array>
+		<string>$(AppIdentifierPrefix)com.amazonaws.amplify.amplify-analytics-pinpoint-example</string>
+	</array>
 </dict>
 </plist>
--- a/packages/analytics/amplify_analytics_pinpoint/example/macos/Runner/Release.entitlements
+++ b/packages/analytics/amplify_analytics_pinpoint/example/macos/Runner/Release.entitlements
@@ -8,5 +8,9 @@
 	<true/>
 	<key>com.apple.security.network.server</key>
 	<true/>
+	<key>keychain-access-groups</key>
+	<array>
+		<string>$(AppIdentifierPrefix)com.amazonaws.amplify.amplify-analytics-pinpoint-example</string>
+	</array>
 </dict>
 </plist>
--- a/packages/analytics/amplify_analytics_pinpoint/example/macos/Runner/Release.entitlements
+++ b/packages/analytics/amplify_analytics_pinpoint/example/macos/Runner/Release.entitlements
@@ -8,9 +8,5 @@
 	<true/>
 	<key>com.apple.security.network.server</key>
 	<true/>
-	<key>keychain-access-groups</key>
-	<array>
-		<string>$(AppIdentifierPrefix)com.amazonaws.amplify.amplify-analytics-pinpoint-example</string>
-	</array>
 </dict>
 </plist>
--- a/packages/analytics/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
+++ b/packages/analytics/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
@@ -4,9 +4,9 @@
 /// Amplify Analytics Pinpoint
 library amplify_analytics_pinpoint;
 
-/// Overridable Flutter injected dependencies
+/// Overridable Flutter injected dependencies.
 export 'package:amplify_analytics_pinpoint_dart/amplify_analytics_pinpoint_dart.dart'
     show AppLifecycleProvider, AWSPinpointUserProfile;
 
-/// Category Implementation
+/// Category Implementation.
 export 'src/analytics_plugin_impl.dart';
--- a/packages/analytics/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
+++ b/packages/analytics/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
@@ -10,6 +10,3 @@ export 'package:amplify_analytics_pinpoint_dart/amplify_analytics_pinpoint_dart.
 
 /// Category Implementation
 export 'src/analytics_plugin_impl.dart';
-
-/// For Push Notifications
-export 'src/device_context_info_provider/flutter_device_context_info_provider.dart';
--- a/packages/analytics/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
+++ b/packages/analytics/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
@@ -4,6 +4,9 @@
 /// Amplify Analytics Pinpoint
 library amplify_analytics_pinpoint;
 
+/// For Push Notifications
+export 'package:amplify_analytics_pinpoint/src/device_context_info_provider/flutter_device_context_info_provider.dart';
+
 /// Overridable Flutter injected dependencies
 export 'package:amplify_analytics_pinpoint_dart/amplify_analytics_pinpoint_dart.dart'
     show AppLifecycleProvider, AWSPinpointUserProfile;
--- a/packages/analytics/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
+++ b/packages/analytics/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
@@ -4,12 +4,12 @@
 /// Amplify Analytics Pinpoint
 library amplify_analytics_pinpoint;
 
-/// For Push Notifications
-export 'package:amplify_analytics_pinpoint/src/device_context_info_provider/flutter_device_context_info_provider.dart';
-
 /// Overridable Flutter injected dependencies
 export 'package:amplify_analytics_pinpoint_dart/amplify_analytics_pinpoint_dart.dart'
     show AppLifecycleProvider, AWSPinpointUserProfile;
 
 /// Category Implementation
 export 'src/analytics_plugin_impl.dart';
+
+/// For Push Notifications
+export 'src/device_context_info_provider/flutter_device_context_info_provider.dart';
--- a/packages/analytics/amplify_analytics_pinpoint_dart/lib/amplify_analytics_pinpoint_dart.dart
+++ b/packages/analytics/amplify_analytics_pinpoint_dart/lib/amplify_analytics_pinpoint_dart.dart
@@ -6,6 +6,7 @@ library amplify_analytics_pinpoint_dart;
 
 export 'src/analytics_plugin_impl.dart';
 export 'src/impl/analytics_client/endpoint_client/aws_pinpoint_user_profile.dart';
+export 'src/impl/analytics_client/endpoint_client/endpoint_store_keys.dart';
 export 'src/impl/flutter_provider_interfaces/app_lifecycle_provider.dart';
 export 'src/impl/flutter_provider_interfaces/cached_events_path_provider.dart';
 export 'src/impl/flutter_provider_interfaces/device_context_info_provider.dart';
--- a/packages/analytics/amplify_analytics_pinpoint_dart/lib/amplify_analytics_pinpoint_dart.dart
+++ b/packages/analytics/amplify_analytics_pinpoint_dart/lib/amplify_analytics_pinpoint_dart.dart
@@ -4,9 +4,10 @@
 /// Amplify Analytics Pinpoint for Dart
 library amplify_analytics_pinpoint_dart;
 
+export 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_client/event_client.dart';
 export 'src/analytics_plugin_impl.dart';
-
 export 'src/impl/analytics_client/endpoint_client/aws_pinpoint_user_profile.dart';
+export 'src/impl/analytics_client/endpoint_client/endpoint_client.dart';
 export 'src/impl/flutter_provider_interfaces/app_lifecycle_provider.dart';
 export 'src/impl/flutter_provider_interfaces/cached_events_path_provider.dart';
 export 'src/impl/flutter_provider_interfaces/device_context_info_provider.dart';
--- a/packages/analytics/amplify_analytics_pinpoint_dart/lib/amplify_analytics_pinpoint_dart.dart
+++ b/packages/analytics/amplify_analytics_pinpoint_dart/lib/amplify_analytics_pinpoint_dart.dart
@@ -6,8 +6,6 @@ library amplify_analytics_pinpoint_dart;
 
 export 'src/analytics_plugin_impl.dart';
 export 'src/impl/analytics_client/endpoint_client/aws_pinpoint_user_profile.dart';
-export 'src/impl/analytics_client/endpoint_client/endpoint_client.dart';
-export 'src/impl/analytics_client/event_client/event_client.dart';
 export 'src/impl/flutter_provider_interfaces/app_lifecycle_provider.dart';
 export 'src/impl/flutter_provider_interfaces/cached_events_path_provider.dart';
 export 'src/impl/flutter_provider_interfaces/device_context_info_provider.dart';
--- a/packages/analytics/amplify_analytics_pinpoint_dart/lib/amplify_analytics_pinpoint_dart.dart
+++ b/packages/analytics/amplify_analytics_pinpoint_dart/lib/amplify_analytics_pinpoint_dart.dart
@@ -4,10 +4,10 @@
 /// Amplify Analytics Pinpoint for Dart
 library amplify_analytics_pinpoint_dart;
 
-export 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_client/event_client.dart';
 export 'src/analytics_plugin_impl.dart';
 export 'src/impl/analytics_client/endpoint_client/aws_pinpoint_user_profile.dart';
 export 'src/impl/analytics_client/endpoint_client/endpoint_client.dart';
+export 'src/impl/analytics_client/event_client/event_client.dart';
 export 'src/impl/flutter_provider_interfaces/app_lifecycle_provider.dart';
 export 'src/impl/flutter_provider_interfaces/cached_events_path_provider.dart';
 export 'src/impl/flutter_provider_interfaces/device_context_info_provider.dart';
--- a/packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
+++ b/packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
@@ -2,17 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import 'dart:async';
-import 'dart:convert';
 
 import 'package:amplify_analytics_pinpoint_dart/amplify_analytics_pinpoint_dart.dart';
-import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/endpoint_client/endpoint_client.dart';
-import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/endpoint_client/endpoint_global_fields_manager.dart';
-import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/endpoint_client/endpoint_store_keys.dart';
-import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_client/event_client.dart';
 import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_client/event_storage_adapter.dart';
-import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_client/queued_item_store/dart_queued_item_store.dart';
-import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_creator/event_creator.dart';
-import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_creator/event_global_fields_manager.dart';
+import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_client/string_database/dart_string_database.dart';
 import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/session_manager.dart';
 import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/stoppable_timer.dart';
 import 'package:amplify_analytics_pinpoint_dart/src/sdk/pinpoint.dart';
@@ -69,16 +62,10 @@ class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
   /// Storage key for the static Pinpoint endpoint id
   @visibleForTesting
   static const String endpointIdStorageKey = 'UniqueId';
-  static const String _endpointGlobalAttrsKey = 'EndpointGlobalAttributesKey';
-  static const String _endpointGlobalMetricsKey = 'EndpointGlobalMetricsKey';
 
-  late final EventCreator _eventCreator;
   late final EndpointClient _endpointClient;
   late final EventClient _eventClient;
   late final SessionManager _sessionManager;
-  late final EndpointGlobalFieldsManager _endpointGlobalFieldsManager;
-  late final EventGlobalFieldsManager _eventGlobalFieldsManager =
-      EventGlobalFieldsManager();
   late final EventStorageAdapter _eventStorageAdapter;
   late final StoppableTimer _autoEventSubmitter;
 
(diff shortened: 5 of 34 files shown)


packages/analytics/amplify_analytics_pinpoint/example/macos/Runner/DebugProfile.entitlements
```diff
@@ -10,5 +10,9 @@
 	<true/>
 	<key>com.apple.security.network.server</key>
 	<true/>
+	<key>keychain-access-groups</key>
```
dnys1, line 13: Should not be committed

packages/analytics/amplify_analytics_pinpoint/example/macos/Runner/Release.entitlements
```diff
@@ -8,5 +8,9 @@
 	<true/>
 	<key>com.apple.security.network.server</key>
 	<true/>
+	<key>keychain-access-groups</key>
```
dnys1, line 11: Remove

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_client.dart
```diff
@@ -52,6 +109,18 @@ class EndpointClient {
   Future<void> removeMetric(String name) async =>
       _globalFieldsManager.removeMetric(name);
 
+  /// Set provided fields of the Endpoint
+  void setFields({
```
dnys1, line 113: So if I don't pass a value and it's already set, it gets overridden with `null`? That's a bit confusing. Although partly a limitation of the Dart type system, I wouldn't recommend that for an API.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_global_fields_manager.dart
```diff
@@ -17,15 +18,43 @@ import 'package:amplify_secure_storage_dart/amplify_secure_storage_dart.dart';
...
+    await fieldsManager.init();
+    return fieldsManager;
+  }
+
+  /// {@macro amplify_analytics_pinpoint_dart.endpoint_global_fields_manager}
+  @visibleForTesting
+  Future<void> init() async {
+    /// Retrieve stored GlobalAttributes
+    final cachedAttributes =
+        await _endpointInfoStore.read(key: _endpointGlobalAttrsKey);
```
dnys1, line 40: Can these keys be tracked via `EndpointKeys`?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_id_manager.dart
```diff
@@ -0,0 +1,79 @@
+import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/endpoint_client/endpoint_store_keys.dart';
```
dnys1, line 1: Missing license header

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_id_manager.dart
```diff
@@ -0,0 +1,79 @@
...
+    required SecureStorageInterface store,
+    LegacyNativeDataProvider? legacyNativeDataProvider,
+    String? pinpointAppId,
+  })  : _store = store,
+        _legacyNativeDataProvider = legacyNativeDataProvider,
+        _pinpointAppId = pinpointAppId;
+
+  final SecureStorageInterface _store;
+  final LegacyNativeDataProvider? _legacyNativeDataProvider;
+  final String? _pinpointAppId;
```
dnys1, line 21: Why nullable? How are endpoint IDs segregated if not by pinpoint app ID?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_client.dart
```diff
@@ -16,25 +21,77 @@ import 'package:built_collection/built_collection.dart';
...
+        ).toBuilder();
 
-  final String _fixedEndpointId;
+  /// {@macro amplify_analytics_pinpoint_dart.endpoint_client}
+  @visibleForTesting
+  Future<void> init({
+    required SecureStorageInterface endpointInfoStore,
+    LegacyNativeDataProvider? legacyNativeDataProvider,
+  }) async {
+    final endpointIdManager = EndpointIdManager(
```
dnys1, line 54: Should be passing `pinpointAppId`

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_client.dart
```diff
@@ -118,8 +184,6 @@ class EndpointClient {
     }
 
     _endpointBuilder.user = newUserBuilder;
-
-    await updateEndpoint();
```
dnys1, line 122: Why is this being removed?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
```diff
@@ -65,13 +73,36 @@ class EventClient implements Closeable {
...
+    final pinpointEvent = _eventCreator.createPinpointEvent(
+      eventType,
+      session,
+      properties,
+    );
+    await _storageAdapter.saveEvent(pinpointEvent);
+  }
+
+  /// Register [AnalyticsProperties] to be sent with all future events
+  Future<void> registerGlobalProperties(
```
dnys1, line 91: This is unnecessary abstraction. Construct the global properties manager at this level and inject into the event creator. The event creator should be single-purpose and just create events.

packages/analytics/amplify_analytics_pinpoint_dart/lib/amplify_analytics_pinpoint_dart.dart
```diff
@@ -4,9 +4,10 @@
 /// Amplify Analytics Pinpoint for Dart
 library amplify_analytics_pinpoint_dart;
 
+export 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_client/event_client.dart';
```
dnys1, line 7: Use relative path like the others

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_id_manager.dart
```diff
@@ -0,0 +1,79 @@
...
+
+  static const String _endpointIdStorageKey = 'UniqueId';
```
dnys1, line 23: This is duplicated from `analytics_plugin_impl`

packages/analytics/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
```diff
@@ -4,6 +4,9 @@
 /// Amplify Analytics Pinpoint
 library amplify_analytics_pinpoint;
 
+/// For Push Notifications
+export 'package:amplify_analytics_pinpoint/src/device_context_info_provider/flutter_device_context_info_provider.dart';
```
dnys1, line 8: Use relative path

packages/analytics/amplify_analytics_pinpoint/example/macos/Runner/DebugProfile.entitlements
fjnoyp: Hmm why not?  Without this change every time we want to run Analytics Example app we need to add this required permission in for SecureStorage..

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_client.dart
fjnoyp: Indeed, it would be better to have different setter fields then.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_global_fields_manager.dart
fjnoyp: As in an enum?  Sure updated

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_client.dart
fjnoyp: I don't think it's clear that calling setUser will also call updateEndpoint.  So to improve code clarity I removed it and updated the call here: 
```
 @override
  Future<void> identifyUser({
    required String userId,
    required AnalyticsUserProfile userProfile,
  }) async {
    _ensureConfigured();
    await _endpointClient.setUser(userId, userProfile);
    await _endpointClient.updateEndpoint();
  } 
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
fjnoyp: I don't see how this is unnecessary abstraction.  Managing the metadata that is sent with events seems related to EventCreator functionality.  

Furthermore, switching to this requires changing the EventClient constructor call to use a factory instead because the `EndpointGlobalFieldsManager` instance must be the same.  

This extra complexity doesn't seem worth it and seems more unnecessary.

packages/analytics/amplify_analytics_pinpoint/example/macos/Runner/DebugProfile.entitlements
Jordan-Nelson: What we have done in most all example apps is override the credential storage to not require this. We can do the same for analytics. I don't think you need to update the example app in this PR. I would probably just remove this change and in a follow up PR, add the override for endpointInfoStore.

Here is what that looks like for Auth:
```dart
final secureStorage = AmplifySecureStorage(
  config: AmplifySecureStorageConfig(
    scope: 'auth',
    // FIXME: In your app, make sure to remove this line and set up
    /// Keychain Sharing in Xcode as described in the docs:
    /// https://docs.amplify.aws/lib/project-setup/platform-setup/q/platform/flutter/#enable-keychain
    // ignore: invalid_use_of_visible_for_testing_member
    macOSOptions: MacOSSecureStorageOptions(useDataProtection: false),
  ),
);
final authPlugin = AmplifyAuthCognito(credentialStorage: secureStorage);
```

fjnoyp: Nice thanks!

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
fjnoyp: Also, the `EventCreator` uses the `EventGlobalFieldsManager` to manage its own behavior when sending events. 

It seems weird to expose an inner class (`EventGlobalFieldsManager`) that will have side effects on `EventCreator`.  

Especially when `EventCreator` is the only class that needs to use `EventGlobalFieldsManager`.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_client.dart
```diff
@@ -16,25 +21,78 @@ import 'package:built_collection/built_collection.dart';
...
-    this._appId,
-    this._fixedEndpointId,
-    this._pinpointClient,
-    this._globalFieldsManager,
-    this._endpointBuilder,
-  );
-
-  final String _appId;
-  final PinpointClient _pinpointClient;
+  @visibleForTesting
```
Jordan-Nelson, line 24: Is this used in a test? I don't see it used.

On a related note, there don't seem to be many unit tests in analytics. Is there a plan to add more?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_client.dart
```diff
@@ -16,25 +21,78 @@ import 'package:built_collection/built_collection.dart';
...
+            timezone: deviceContextInfo?.timezone,
+          ),
+          location: EndpointLocation(
+            country: deviceContextInfo?.countryCode,
+          ),
+        ).toBuilder();
 
-  final String _fixedEndpointId;
+  /// {@macro amplify_analytics_pinpoint_dart.endpoint_client}
+  @visibleForTesting
```
Jordan-Nelson, line 49: I don't see this used in a test.

Jordan-Nelson (commented):
I would like to see some more tests. Let's add tests for each public API - APIs exposed to customers, and APIs exposed to other Amplify categories (push).

packages/analytics/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
```diff
@@ -10,3 +10,6 @@ export 'package:amplify_analytics_pinpoint_dart/amplify_analytics_pinpoint_dart.
 
 /// Category Implementation
 export 'src/analytics_plugin_impl.dart';
+
+/// For Push Notifications
+export 'src/device_context_info_provider/flutter_device_context_info_provider.dart';
```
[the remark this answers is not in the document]
fjnoyp: This is outdated, the final way push notification will use this is subject to change in a future PR

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_client.dart
```diff
@@ -16,25 +21,78 @@ import 'package:built_collection/built_collection.dart';
...
+            platformVersion: deviceContextInfo?.platformVersion,
+            timezone: deviceContextInfo?.timezone,
+          ),
+          location: EndpointLocation(
+            country: deviceContextInfo?.countryCode,
+          ),
+        ).toBuilder();
 
-  final String _fixedEndpointId;
+  /// {@macro amplify_analytics_pinpoint_dart.endpoint_client}
```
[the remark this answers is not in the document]
fjnoyp: Added:

```
/// Initialize [EndpointClient] by retrieving the endpoint id and
  /// initializing the [EndpointGlobalFieldsManager]
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_store_keys.dart
```diff
@@ -5,6 +5,21 @@
 enum EndpointStoreKey {
   /// The current version of the endpoint version store
   version,
+
+  /// Global attributes key attached to endpoint
+  // Use upper case to support legacy naming
```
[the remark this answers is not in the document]
fjnoyp: Previous implementation in next.  Good point ... based on our previous discussions about legacy support in next I'm going to change the casing to lower camel case and remove these comments.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_client.dart
fjnoyp: I will make unit tests for the followup PR to refactor Endpoint Info Management

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_client.dart
fjnoyp: I will make unit tests for the followup PR to refactor Endpoint Info Management

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -310,51 +247,5 @@ class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
     _isConfigured = false;
     _autoEventSubmitter.stop();
     await _eventClient.close();
-    await _eventStorageAdapter.close();
-  }
-
-  /// Retrieve the stored pinpoint endpoint id
-  @visibleForTesting
-  Future<String> retrieveEndpointId({
-    required String pinpointAppId,
```
fjnoyp, line 319: Code moved to a new `EndpointIdManager` class.  All endpoint information retrieval will be refactored in this followup [PR](https://github.com/aws-amplify/amplify-flutter/pull/2639/files#diff-7415c6999da61d3f183ade45e81877137703db58cbab02ef3db08de3812eb9da) to consolidate data retrieval/management for Endpoints.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_global_fields_manager.dart
```diff
@@ -17,15 +19,45 @@ import 'package:amplify_secure_storage_dart/amplify_secure_storage_dart.dart';
...
+    /// Retrieve stored GlobalAttributes
+    final cachedAttributes = await _endpointInfoStore.read(
+      key: EndpointStoreKey.endpointGlobalAttributesKey.name,
+    );
+    _globalAttributes = cachedAttributes == null
+        ? <String, String>{}
+        : (jsonDecode(cachedAttributes) as Map<String, Object?>)
+            .cast<String, String>();
+
+    /// Retrieve stored GlobalMetrics
```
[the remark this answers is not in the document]
dnys1: You've marked this resolved but it's not.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_global_fields_manager.dart
```diff
@@ -17,15 +19,45 @@ import 'package:amplify_secure_storage_dart/amplify_secure_storage_dart.dart';
...
+  ) async {
+    final fieldsManager = EndpointGlobalFieldsManager(endpointInfoStore);
+    await fieldsManager.init();
+    return fieldsManager;
+  }
+
+  /// {@macro amplify_analytics_pinpoint_dart.endpoint_global_fields_manager}
+  @visibleForTesting
+  Future<void> init() async {
+    /// Retrieve stored GlobalAttributes
```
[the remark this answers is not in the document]
dnys1: You've marked this resolved but it's not.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_store_keys.dart
```diff
@@ -5,6 +5,15 @@
 enum EndpointStoreKey {
   /// The current version of the endpoint version store
   version,
+
+  /// Global attributes key attached to endpoint
+  endpointGlobalAttributesKey,
```
dnys1, line 10:
```suggestion
  globalAttributesKey,
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_store_keys.dart
```diff
@@ -5,6 +5,15 @@
...
+
+  /// Global metrics key attached to endpoint
+  endpointGlobalMetricsKey,
```
dnys1, line 13:
```suggestion
  globalMetricsKey,
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_id_manager.dart
dnys1: > Analytics and Auth will share the same endpoint by using the same SecureStorageInterface scope

How will that work? Can you add unit tests which showcase this? What if the Analytics category and Auth category point to different Pinpoint App IDs? And what if Push points to yet another? Why would Analytics/Auth share a scope in this case? And which one owns the `SecureStorageInterface`?

Also, still not sure why this is null here.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_store_keys.dart
fjnoyp: While it makes sense in code to drop the endpoint, I think it's important to actually keep an endpoint prefix in case we store more values into the secure storage and need a way to easily differentiate keys.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_id_manager.dart
fjnoyp: You have a good point here ... I was copying iOS behavior here but it is very strange.  I'm going to double check this behavior with Abdallah. 

In the meantime, here's more context: 

Analytics and Auth will use the same endpoint because they use the same SecureStorage database to retrieve and store endpoints.  That will be more obvious in the next PR to refactor endpoint information management. 

I will add unit tests - not sure about testing if Analytics/Auth share the same endpoint.  The idea is that Analytics and Auth will use the same `FlutterEndpointInformationManager` class so by using the same implementation it should be guaranteed the data used is the same. 


The expected behavior is NOT for Analytics and Auth to use different Pinpoint App Ids.  In general it's not expected that any of the Analytics, Auth, Notification category Pinpoint App Ids should be different at all per our recent Analytics group discussions.  Again, I'll double check that with Abdallah. 

Analytics/Auth share a scope because it was assumed they must use the same Pinpoint App Id.  

They both own the storage.  The idea is that even if Analytics is never added, Auth can create and manage endpoint itself and vice versa. 

The PinpointAppId is nullable here because it reflects the philosophy that the Endpoint Id is the same between Analytics and Auth.  It's only needed for retrieving legacy Endpoint Id from Android (and NOT iOS).

fjnoyp: In this PR I refactor Endpoint logic: https://github.com/aws-amplify/amplify-flutter/pull/2639

Since the behavior in this PR is not different than previous implementations, and the next PR is related to Endpoint information storage, I recommend we tackle the refactor of the logic in that followup PR. 

I also plan to add tests there as well.

fjnoyp:
You have a good point here ... I was copying iOS behavior here but it is very strange. I'm going to double check this behavior with Abdallah.

I think you're right, Endpoint retrieval and storage should be tied to Pinpoint App Id. We're not doing that right now ... I should probably rework how endpoint storage works.

In this PR I refactor Endpoint logic: https://github.com/aws-amplify/amplify-flutter/pull/2639

Since the behavior for Endpoint storage in this PR is not different than previous implementations, and the next PR is related to Endpoint information storage, I recommend we tackle the refactor of the logic in that followup PR.

I also plan to add tests there as well.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_store_keys.dart
fjnoyp: I switched the version to endpointStoreVersion as well to be consistent with the above logic

dnys1: This is already `EndpointStoreKey` - reiterating that these values are related to endpoints is not necessary. If you're concerned about the value in storage (and not in code), then you can override the `name` used for storage:

```dart
enum EndpointStoreKey {
  version('endpointStoreVersion'),

  ...;

  const EndpointStoreKey(this.name);

  final String name;
}
```

But even this should not be necessary since values should already be appropriately scoped to the endpoint store, making this, too, redundant. Overall, I'm not seeing what values you would be adding here which would need further designation as endpoint values.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_id_manager.dart
dnys1: > not sure about testing if Analytics/Auth share the same endpoint

That's not needed for this PR. But it should be possible to have some unit tests which shows how two different `EndpointIdManager`'s with different pinpoint app IDs would interact since it's not just endpoint IDs which are being stored but also global properties etc.

fjnoyp: After speaking with native teams and Chris, I've decided the best approach is to implement Android's strategy of using the pinpointAppId to store the endpoint information (iOS just uses a hardcoded key which is what we do now). 

I'll do that in the dedicated endpoint refactor PR and add unit tests there.  Made a note of this to make sure I do that.

dnys1 (approved):

PR #2639, opened 2023-02-08
chore(analytics): refactor analytics (endpoint info / analytics client / unit tests) 

fjnoyp:
**Final refactor of Analytics** 

### Explanation of Changes: 

1) Refactor all Endpoint related storage logic (endpoint id and endpoint global fields) into a dedicated 'EndpointInfoStoreManager' class

2) Prefix all endpoint information keys with the PinpointAppId.  This is closer to Android's approach and different than iOS.  While it does not actually change end behavior, it prevents an edge case where a different PinpointAppId in Auth vs. Analytics would cause the same endpoint id to be sent to 2 different Pinpoint Apps. 

3) Encapsulate `EndpointInfoStorageManager` as a `FlutterEndpointInfoStoreManager` that will be used in Auth to create/store endpoint id. 
I decided against having this logic in common/core because endpoint id logic relies on android pigeon code to retrieve legacy data.  Moving that logic would require moving some ~10 code files to common/core.  PM me if you want the full document explanation behind this choice.  

4) Added Dart unit tests for all major internal logic 

5) Created a new `AnalyticsClient` class that encapsulates the complex initialization logic for `EventClient` and `EndpointClient` for PushNotifications.  
The `AnalyticsClient` prevents PushNotifications from needing to import the Pinpoint SDK into the codebase. 

**I know there are a lot of changes in this one PR.  A lot of this work is interconnected so splitting it is time consuming and any change would require updating and rebasing the split PRs...** 

### Sample Use Cases of New Changes

1) Auth will use the `FlutterEndpointInfoStoreManager` to retrieve and manage the endpoint id independent of analytics.  TBD - Will need to clarify how a Flutter dependency will be moved into Auth's dart logic. 

```
final store = FlutterEndpointInfoStoreManager(); 
await store.init(); 
final endpointId = store.endpointId; 
```

2) PushNotifications will use `FlutterAnalyticsClient` to interact with Pinpoint endpoint and event logic.  

```
// Creation 
final client = FlutterAnalyticsClient(
      endpointInfoStoreManager: FlutterEndpointInfoStoreManager(
        storageScope: EndpointStorageScope.pushNotifications, 
      ),
);
client.init(
    pinpointAppId: 'pinpointAppId',
    region: 'region', 
    authProvider: AuthProvider(),
); 

// Usage 
final endpointClient = client.endpointClient;
final eventClient = client.eventClient; 

await endpointClient.updateEndpoint();
eventClient.recordEvent(eventType: 'eventType');
eventClient.flushEvents(); 
```


packages/analytics/amplify_analytics_pinpoint/lib/amplify_analytics_pinpoint.dart
```diff
@@ -7,9 +7,16 @@ library amplify_analytics_pinpoint;
 /// For Push Notifications
 export 'package:amplify_analytics_pinpoint/src/device_context_info_provider/flutter_device_context_info_provider.dart';
 
-/// Overridable Flutter injected dependencies
+/// Pinpoint Client Helpers
+export 'package:amplify_analytics_pinpoint/src/flutter_endpoint_info_store_manager.dart';
```
dnys1, line 11: Use relative exports when it's from the same package.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_client.dart
```diff
@@ -48,35 +46,29 @@ class EndpointClient {
...
-      store: endpointInfoStore,
-      legacyNativeDataProvider: legacyNativeDataProvider,
+    _fixedEndpointId = await endpointInfoStoreManager.retrieveEndpointId(
+      pinpointAppId: _pinpointAppId,
     );
-    _fixedEndpointId = await endpointIdManager.retrieveEndpointId();
 
-    _globalFieldsManager =
-        await EndpointGlobalFieldsManager.create(endpointInfoStore);
+    _globalFieldsManager = await endpointInfoStoreManager.endpointFields;
```
dnys1, line 55: Inject this

packages/analytics/amplify_analytics_pinpoint_dart/lib/amplify_analytics_pinpoint_dart.dart
```diff
@@ -5,9 +5,12 @@
 library amplify_analytics_pinpoint_dart;
 
 export 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_client/event_client.dart';
+export 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/event_client/string_database/index_db/in_memory_string_db.dart';
```
dnys1, line 8: Relative exports

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_info_store_manager.dart
```diff
@@ -1,34 +1,58 @@
...
+        _legacyNativeDataProvider = legacyNativeDataProvider;
 
   final SecureStorageInterface _store;
   final LegacyNativeDataProvider? _legacyNativeDataProvider;
-  final String? _pinpointAppId;
 
   static const String _endpointIdStorageKey = 'UniqueId';
 
+  /// Fields of the Pinpoint Endpoint
+  late final Future<EndpointGlobalFieldsManager> endpointFields = () async {
```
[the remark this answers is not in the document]
fjnoyp: Not sure I see the benefit of doing that. 

The behavior of EndpointGlobalFieldsManager is driven by the `SecureStorageInterface`.  The `SecureStorageInterface` should always be the same between EndpointGlobalFieldsManager and what is used to store the endpoint id.  

An important purpose of this class is to ensure that.  

Having EndpointGlobalFieldsManager as a DI suggests its functionality is separate from `SecureStorageInterface` which is not correct.  Consequently the user could give a EndpointGlobalFieldsManager initialized with another SecureStorageInterface than the one used in the EndpointInfoStoreManager which would cause endpoint info storage location to diverge. 

While I could refactor the EndpointGlobalFieldsManager to take a SecureStorageInterface in its init method, the user could still pass the EndpointGlobalFieldsManager DI initialized already.  And then if a second call to init changes that, that would revert what the user intended.

packages/analytics/amplify_analytics_pinpoint/lib/src/analytics_plugin_impl.dart
```diff
@@ -16,20 +15,15 @@ import 'package:meta/meta.dart';
 class AmplifyAnalyticsPinpoint extends AmplifyAnalyticsPinpointDart {
   /// {@macro amplify_analytics_pinpoint.analytics_plugin_impl}
   AmplifyAnalyticsPinpoint({
-    @visibleForTesting SecureStorageInterface? endpointInfoStore,
     @visibleForTesting AppLifecycleProvider? appLifecycleProvider,
+    @visibleForTesting EndpointInfoStoreManager? endpointInfoStoreManager,
```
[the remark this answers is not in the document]
fjnoyp: For testing.  In `setup_utils.dart` we pass a `MockEndpointInfoStoreManager`.

packages/analytics/amplify_analytics_pinpoint/example/macos/Flutter/GeneratedPluginRegistrant.swift
```diff
@@ -9,7 +9,7 @@ import amplify_secure_storage
 import connectivity_plus
 import device_info_plus
 import package_info_plus
-import path_provider_foundation
+import path_provider_macos
```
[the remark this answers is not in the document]
fjnoyp: Thanks that fixed it.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_client.dart
fjnoyp: Please see my thoughts here: https://github.com/aws-amplify/amplify-flutter/pull/2639#discussion_r1102854770

fjnoyp:
Note to self: Add Unit tests for EndpointGlobalFieldsManager and some test to ensure Endpoint info can be shared between Auth and Analytics.

packages/analytics/amplify_analytics_pinpoint/example/integration_test/utils/mock_secure_storage.dart
```diff
@@ -2,22 +2,35 @@
...
 /// between relaunches (e.g. between test groups).
-final SecureStorageInterface mockEndpointInfoStore = _MockEndpointInfoStore()
-  ..write(
-    key: EndpointStoreKey.endpointId.name,
-    value: mockEndpointId,
-  );
+final SecureStorageInterface mockPersistedSecuredStorage = () {
+  final json = jsonDecode(amplifyconfig) as Map;
+  final amplifyConfig = AmplifyConfig.fromJson(json.cast());
+  final pinpointAppId =
```
fjnoyp, line 19: Because we now use pinpointAppId to prefix endpoint information storage keys, this must be replicated in integration test setup.

packages/analytics/amplify_analytics_pinpoint/lib/src/flutter_analytics_client.dart
```diff
@@ -0,0 +1,41 @@
+import 'package:amplify_analytics_pinpoint/src/device_context_info_provider/flutter_device_context_info_provider.dart';
+import 'package:amplify_analytics_pinpoint/src/flutter_endpoint_info_store_manager.dart';
+import 'package:amplify_analytics_pinpoint_dart/amplify_analytics_pinpoint_dart.dart';
+
+/// {@template amplify_analytics_pinpoint_dart.flutter_endpoint_store_manager}
+/// Manages retrieval, storage, and management of Pinpoint Endpoint id and fields
+/// Auto injects Flutter DI
+/// {@endtemplate}
+class FlutterAnalyticsClient extends AnalyticsClient {
```
fjnoyp, line 9: This client class will be used by Analytics, Push Notifications and any further category that requires Endpoint/Events with Pinpoint. 

It is designed to encapsulate as much internal initialization logic as possible to minimize exports. 

As a result we no longer need to export: 
'FlutterDeviceContextInfoProvider' 
'AmplifySecureStorage' 

For users to be able to instance 'AnalyticsClient'.

packages/analytics/amplify_analytics_pinpoint/lib/src/flutter_endpoint_info_store_manager.dart
```diff
@@ -0,0 +1,20 @@
...
+
+/// {@template amplify_analytics_pinpoint_dart.flutter_endpoint_store_manager}
+/// Manages retrieval, storage, and management of Pinpoint Endpoint id and fields
+/// Auto injects Flutter DI
+/// {@endtemplate}
+class FlutterEndpointInfoStoreManager extends EndpointInfoStoreManager {
+  /// {@macro amplify_analytics_pinpoint_dart.flutter_endpoint_store_manager}
+  FlutterEndpointInfoStoreManager()
+      : super(
+          store: AmplifySecureStorage(
```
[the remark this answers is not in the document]
fjnoyp: This class will be exported to Cognito Auth to allow Auth to manage creation and retrieval of endpoint id.  

I reduce the DI here so that Auth can simply use this class without writing any additional config logic.  Doing so reduces risk of decoupled endpoint logic between Auth and Analytics.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -31,21 +29,12 @@ const zSessionStopEventType = '_session.stop';
...
     AppLifecycleProvider? appLifecycleProvider,
-    DeviceContextInfoProvider? deviceContextInfoProvider,
-    LegacyNativeDataProvider? legacyNativeDataProvider,
-  })  : _endpointInfoStore = endpointInfoStore ??
-            AmplifySecureStorageWorker(
-              config: AmplifySecureStorageConfig(
-                scope: 'analyticsPinpoint',
-              ),
-            ),
+    AnalyticsClient? analyticsClient,
```
fjnoyp, line 34: Since the AnalyticsClient will handle the endpointInfoStore, deviceContextInfo, legacyNativeData etc. parameters, they can now be removed from plugin class.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/endpoint_client/endpoint_client.dart
```diff
@@ -20,12 +18,15 @@ import 'package:built_collection/built_collection.dart';
 /// {@endtemplate}
 class EndpointClient {
   /// {@macro amplify_analytics_pinpoint_dart.endpoint_client}
-  EndpointClient._({
+  EndpointClient({
     required String pinpointAppId,
     required PinpointClient pinpointClient,
+    required EndpointInfoStoreManager endpointInfoStoreManager,
```
fjnoyp, line 24: We could technically required fixedEndpointId and globalFieldsManager as DI, but I decided to still required EndpointInfoStoreManager as that class is meant to enforce a consistent storage/retrieval behavior for those values.

packages/analytics/amplify_analytics_pinpoint_dart/test/analytics_client.dart
```diff
@@ -0,0 +1,114 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
fjnoyp, line 1: Unsure if this test file is really needed.  Since different storage instances are used there shouldn't really be any chance for data to intermingle.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/analytics_client.dart
```diff
@@ -0,0 +1,88 @@
...
+
+  /// Async initialize the [eventClient] and [endpointClient].
+  /// Requires missing values typically provided in plugin configure.
+  Future<void> init({
+    required String pinpointAppId,
+    required String region,
+    required AWSCredentialsProvider authProvider,
+    DartQueuedItemStore? eventStore,
+  }) async {
+    final pinpointClient = PinpointClient(
```
HuiSF, line 72: nit: If this could be injected, then we don't need to add `initWithClient`?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/queued_item_store/index_db/in_memory_queued_item_store.dart
```diff
@@ -45,4 +46,11 @@ class InMemoryQueuedItemStore implements QueuedItemStore {
 
   @override
   void close() {}
+
+  /// Reset all internal values for next unit test
+  @visibleForTesting
+  void reset() {
```
HuiSF, line 52: nit: possible to create new instance in the tests instead of adding a public method reseting properties for unit testing?

packages/analytics/amplify_analytics_pinpoint_dart/test/analytics_legacy_native_data_test.dart
```diff
@@ -29,20 +31,24 @@ void main() {
...
+      );
 
       final storeVersion = await store.read(
         key: EndpointStoreKey.version.name,
       );
       expect(storeVersion, EndpointStoreVersion.v1.name);
 
       final migratedEndpointId = await store.read(
-        key: EndpointStoreKey.endpointId.name,
+        key: pinpointAppId + EndpointStoreKey.endpointId.name,
```
HuiSF, line 51: nit: string template

packages/analytics/amplify_analytics_pinpoint_dart/test/analytics_legacy_native_data_test.dart
```diff
@@ -53,20 +59,23 @@ void main() {
...
```
HuiSF, line 78: nit: ditto

packages/analytics/amplify_analytics_pinpoint_dart/test/analytics_legacy_native_data_test.dart
```diff
@@ -80,45 +89,26 @@ void main() {
       final endpointId = uuid();
       store.seedData({
         EndpointStoreKey.version.name: EndpointStoreVersion.v1.name,
-        EndpointStoreKey.endpointId.name: endpointId
+        pinpointAppId + EndpointStoreKey.endpointId.name: endpointId
```
HuiSF, line 92: nit: ditto

packages/analytics/amplify_analytics_pinpoint_dart/test/analytics_legacy_native_data_test.dart
```diff
@@ -80,45 +89,26 @@ void main() {
...
+      await endpointIdManager.init(pinpointAppId: pinpointAppId);
+
+      expect(
+        endpointIdManager.endpointId,
+        endpointId,
+      );
 
       final migratedEndpointId = await store.read(
-        key: EndpointStoreKey.endpointId.name,
+        key: pinpointAppId + EndpointStoreKey.endpointId.name,
```
HuiSF, line 107: nit: ditto

packages/analytics/amplify_analytics_pinpoint_dart/test/common/endpoint_global_fields_storage.dart
```diff
@@ -0,0 +1,34 @@
...
+import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/endpoint_client/endpoint_store_keys.dart';
+import 'package:amplify_secure_storage_dart/amplify_secure_storage_dart.dart';
+
+/// Retrieve attributes map from [SecureStorageInterface]
+Future<Map<String, String>> getStoredAttributes(
+  SecureStorageInterface storage, {
+  String appId = '',
+}) async {
+  final cachedAttributes = await storage.read(
+    key: appId + EndpointStoreKey.globalAttributesKey.name,
```
HuiSF, line 14: nit: ditto

packages/analytics/amplify_analytics_pinpoint_dart/test/common/endpoint_global_fields_storage.dart
```diff
@@ -0,0 +1,34 @@
...
+          .cast<String, String>();
+}
+
+/// Retrieve metrics map from [SecureStorageInterface]
+Future<Map<String, double>> getStoredMetrics(
+  SecureStorageInterface storage, {
+  String appId = '',
+}) async {
+  final cachedMetrics = await storage.read(
+    key: appId + EndpointStoreKey.globalMetricsKey.name,
```
HuiSF, line 28: nit: ditto

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/analytics_client.dart
fjnoyp: `initWithClient` is specifically added for testing, see the comments in the method declaration: 

```
/// Allows mocking of [PinpointClient] for unit tests
  @visibleForTesting
  Future<void> initWithClient({
```

packages/analytics/amplify_analytics_pinpoint_dart/test/analytics_legacy_native_data_test.dart
fjnoyp: Is there a way to use string templates in Dart? 
https://stackoverflow.com/questions/57230201/generic-template-string-like-in-python-in-dart

Also, not sure I want to replace this with method indirection.  

Lastly, I don't think that using string interpolation is good here.  In fact it looks a bit more confusing: 

```
 key: '$pinpointAppId${EndpointStoreKey.endpointId.name}',
 
 key: pinpointAppId + EndpointStoreKey.endpointId.name,

```

packages/analytics/amplify_analytics_pinpoint/example/integration_test/utils/mock_secure_storage.dart
```diff
@@ -2,22 +2,35 @@
...
+      amplifyConfig.analytics!.awsPlugin!.pinpointAnalytics.appId;
+
+  final storage = _MockSecureStorage()
+    ..write(
+      key: pinpointAppId + EndpointStoreKey.endpointId.name,
```
dnys1, line 24: This logic should be consolidated somewhere - it's easy to mess up and could potentially change. Check out how it's handled in Auth: https://github.com/aws-amplify/amplify-flutter/blob/next/packages/auth/amplify_auth_cognito_dart/lib/src/credentials/cognito_keys.dart

packages/analytics/amplify_analytics_pinpoint/lib/src/flutter_analytics_client.dart
```diff
@@ -0,0 +1,16 @@
+import 'package:amplify_analytics_pinpoint/src/device_context_info_provider/flutter_device_context_info_provider.dart';
+import 'package:amplify_analytics_pinpoint/src/flutter_endpoint_info_store_manager.dart';
+import 'package:amplify_analytics_pinpoint_dart/amplify_analytics_pinpoint_dart.dart';
+
+/// {@macro amplify_analytics_pinpoint.analytics_client}
+/// Auto injects Flutter dependencies.
```
dnys1, line 6:
```suggestion
/// {@template amplify_analytics_pinpoint.analytics_client}
/// <more descriptive doc comment>
/// {@endtemplate
```

packages/analytics/amplify_analytics_pinpoint/lib/src/flutter_endpoint_info_store_manager.dart
```diff
@@ -0,0 +1,19 @@
+import 'package:amplify_analytics_pinpoint/src/legacy_native_data_provider/flutter_legacy_native_data_provider.dart';
+import 'package:amplify_analytics_pinpoint_dart/amplify_analytics_pinpoint_dart.dart';
+import 'package:amplify_secure_storage/amplify_secure_storage.dart';
+
+/// {@macro amplify_analytics_pinpoint_dart.endpoint_store_manager}
+/// Auto injects Flutter dependencies.
```
dnys1, line 6:
```suggestion
/// {@template amplify_analytics_pinpoint.flutter_endpoint_store_manager}
/// <more descriptive doc comment>
/// {@endtemplate
```

packages/analytics/amplify_analytics_pinpoint/lib/src/flutter_endpoint_info_store_manager.dart
```diff
@@ -0,0 +1,19 @@
...
+class FlutterEndpointInfoStoreManager extends EndpointInfoStoreManager {
+  /// {@macro amplify_analytics_pinpoint_dart.flutter_endpoint_store_manager}
```
dnys1, line 8:
```suggestion
  /// {@macro amplify_analytics_pinpoint.flutter_endpoint_store_manager}
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/analytics_client.dart
dnys1: Hui's asking why this can't be done in the constructor. And I agree - injecting dependencies in both the constructor and `init` is not a common pattern and is rather confusing.

packages/analytics/amplify_analytics_pinpoint/example/integration_test/utils/mock_secure_storage.dart
dnys1: Or expose `_EndpointStore` - just saw that

packages/analytics/amplify_analytics_pinpoint_dart/test/common/endpoint_global_fields_storage.dart
```diff
@@ -0,0 +1,34 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:convert';
+import 'package:amplify_analytics_pinpoint_dart/src/impl/analytics_client/endpoint_client/endpoint_store_keys.dart';
+import 'package:amplify_secure_storage_dart/amplify_secure_storage_dart.dart';
+
+/// Retrieve attributes map from [SecureStorageInterface]
+Future<Map<String, String>> getStoredAttributes(
```
dnys1, line 9: Why are you using these helpers instead of `EndpointGlobalFieldsManager`?

These make your tests very brittle because their tied to the implementation - if your implementation changes, these tests need to change. If you used `EndpoingGlobalFieldsManager`, the implementation can change and your tests then work to verify your changes are accurate.

packages/analytics/amplify_analytics_pinpoint_dart/test/endpoint_global_fields_manager.dart
```diff
@@ -0,0 +1,214 @@
...
+    test('attribute key/values longer than "maxLength" are truncated',
+        () async {
+      final stringBuffer = StringBuffer();
+      for (var i = 0;
+          i < EndpointGlobalFieldsManager.maxAttributeValueLength * 2;
+          i++) {
+        stringBuffer.write('a');
+      }
+
+      final longAttrKey = stringBuffer.toString();
```
dnys1, line 142: Shouldn't this use `maxKeyLength`?

packages/analytics/amplify_analytics_pinpoint_dart/test/endpoint_global_fields_manager.dart
```diff
@@ -0,0 +1,214 @@
...
+    });
+
+    test('attribute key/values longer than "maxLength" are truncated',
+        () async {
+      final stringBuffer = StringBuffer();
+      for (var i = 0;
+          i < EndpointGlobalFieldsManager.maxAttributeValueLength * 2;
+          i++) {
+        stringBuffer.write('a');
+      }
```
dnys1, line 140:
```suggestion
      'a' * EndpointGlobalFieldsManager.maxAttributeValueLength * 2;
```

packages/analytics/amplify_analytics_pinpoint_dart/test/event_client_test.dart
```diff
@@ -0,0 +1,406 @@
...
+        () async {
+      await eventClient.recordEvent(
+        eventType: successEventType,
+      );
+      await eventClient.recordEvent(
+        eventType: failEventType,
+      );
+
+      when(() => pinpointClient.putEvents(any<PutEventsRequest>()))
+          .thenThrow(BadRequestException());
```
dnys1, line 379: Should be the smithy operation completes with this error - `putEvents` never directly throws

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/analytics_client.dart
fjnoyp: Thanks for clarifying.  I agree, it would be ideal to pass PinpointClient in constructor. 

Issue is that AnalyticsClient is a DI which means it must be created in the Flutter top level. 

And we only receive appId, region, credentialsProvider in the configure method.  

Furthermore we shouldn't only have one `init` method with PinpointClient because `AnalyticsClient` is used by PushNotifications which must avoid a dependency on the Pinpoint SDK.

packages/analytics/amplify_analytics_pinpoint/lib/src/analytics_plugin_impl.dart
```diff
@@ -15,19 +13,12 @@ import 'package:meta/meta.dart';
 class AmplifyAnalyticsPinpoint extends AmplifyAnalyticsPinpointDart {
   /// {@macro amplify_analytics_pinpoint.analytics_plugin_impl}
   AmplifyAnalyticsPinpoint({
-    @visibleForTesting SecureStorageInterface? endpointInfoStore,
     @visibleForTesting AppLifecycleProvider? appLifecycleProvider,
+    AnalyticsClient? analyticsClient,
```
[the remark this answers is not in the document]
fjnoyp: Thanks for catching - yes it's only for mocking.  Added `visibleForTesting`

packages/analytics/amplify_analytics_pinpoint/example/integration_test/utils/mock_secure_storage.dart
fjnoyp: Switched to a new MockEndpointStore class.

fjnoyp: Switched to a new MockEndpointStore class.

packages/analytics/amplify_analytics_pinpoint_dart/test/common/endpoint_global_fields_storage.dart
fjnoyp: Good point.  Exposed internal getAttributes and getMetrics from `EndpointGlobalFieldsManager` to be used in tests.

packages/analytics/amplify_analytics_pinpoint_dart/test/event_client_test.dart
dnys1: I'm saying `putEvents` doesn't throw. `putEvents` returns a SmithyOperation which when you await its `result` property throws.

fjnoyp: Okay understood, updated the tests.

HuiSF (approved):

dnys1 (approved):

Issue #2535, opened 2023-01-05
[amplify_secure_storage] SecurityFrameworkError code: -25308


asaarnak:
I couldn't reproduce it myself.
Noticed this issue for 2 users on iOS crash logs.
Maybe related to this: https://developer.apple.com/forums/thread/114542
Seems that app is locked maybe and app is on background that`s why this error happens.
https://stackoverflow.com/questions/48068325/keychain-access-error-25308-errsecinteractionnotallowed 

```stacktrace
SecureStorageException(
  
  message: An unknown exception occurred.,
  
  recoverySuggestion: An unknown exception occurred. Please take a look at https://github.com/aws-amplify/amplify-flutter/issues to see if there are any existing issues that match your scenario, and file an issue with the details of the bug if there isn't.,
  
  underlyingException: SecurityFrameworkError: {code: -25308, message: No error string is available.}),
  
  stackTrace=
  
0  ???                            0x0 AmplifySecureStorageCupertino._add + 146 (amplify_secure_storage_cupertino.dart:146)
1  ???                            0x0 AmplifySecureStorageCupertino.write.<fn> + 65 (amplify_secure_storage_cupertino.dart:65)
2  ???                            0x0 (null).using + 124 (arena.dart:124)
3  ???                            0x0 AmplifySecureStorageCupertino.write + 63 (amplify_secure_storage_cupertino.dart:63)
4  ???                            0x0 AmplifySecureStorageDartMixin.write + 42 (amplify_secure_storage_mixin.vm.dart:42)
5  ???                            0x0 SecureStorageWorker.run + 73 (secure_storage_worker.dart:73)
6  ???                            0x0 (null)._run + 14 (secure_storage_worker.worker.vm.dart:14)
```

Seems to be related stacktrace:
```
0  ???                            0x0 CredentialStoreStateMachine.getCredentialsResult + 59 (credential_store_state_machine.dart:59)
1  ???                            0x0 FetchAuthSessionStateMachine.onFetchAuthSession + 179 (fetch_auth_session_state_machine.dart:179)
2  ???                            0x0 FetchAuthSessionStateMachineBase.resolve + 37 (fetch_auth_session_state_machine_base.dart:37)
3  ???                            0x0 StateMachine._listenForEvents + 171 (state_machine.dart:171)
```

Jordan-Nelson:
[...]

PR #2649, opened 2023-02-16
fix(secure_storage)!: use correct values for `kSecAttrAccessible`

Jordan-Nelson:
*Issue #, if available:* https://github.com/aws-amplify/amplify-flutter/issues/2535

*Description of changes:*
- Map KeychainAttributeAccessible value to correct values in keychain.
- Gracefully handle the scenario when KeychainAttributeAccessible changes
- Add tests for the updated behavior

### DIFF
--- a/packages/secure_storage/amplify_secure_storage/example/integration_test/cupertino_test.dart
+++ b/packages/secure_storage/amplify_secure_storage/example/integration_test/cupertino_test.dart
@@ -7,6 +7,7 @@ import 'dart:io';
 
 import 'package:amplify_secure_storage/amplify_secure_storage.dart';
 import 'package:amplify_secure_storage/src/messages.cupertino.g.dart';
+import 'package:amplify_secure_storage_dart/src/platforms/amplify_secure_storage_cupertino.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:integration_test/integration_test.dart';
 
@@ -16,18 +17,19 @@ const key2 = 'key_2';
 const value1 = 'value_1';
 const value2 = 'value_2';
 
-// Enabling useDataProtection requires that the app is
-// added to at least one access group.
-final macOSOptions = MacOSSecureStorageOptions(useDataProtection: false);
-
-/// MacOS & iOS app uninstall & re-install tests.
-///
-/// These tests are only relevant for amplify_secure_storage
-/// (not amplify_secure_storage_dart) as they use
-/// NSUserDefaultsAPI via pigeon.
 void main() {
   IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  /// MacOS & iOS app uninstall & re-install tests.
+  ///
+  /// These tests are only relevant for amplify_secure_storage
+  /// (not amplify_secure_storage_dart) as they use
+  /// NSUserDefaultsAPI via pigeon.
   group('${Platform.operatingSystem} app uninstall & re-install', () {
+    // Enabling useDataProtection requires that the app is added to at least one
+    // access group.
+    final macOSOptions = MacOSSecureStorageOptions(useDataProtection: false);
+
     final userDefaults = NSUserDefaultsAPI();
     final storage = AmplifySecureStorage(
       config: AmplifySecureStorageConfig(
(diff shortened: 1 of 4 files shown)


ragingsquirrel3 (approved):

packages/secure_storage/amplify_secure_storage_dart/lib/src/platforms/amplify_secure_storage_cupertino.dart
```diff
@@ -51,9 +51,9 @@ class AmplifySecureStorageCupertino extends AmplifySecureStorageInterface {
   void write({required String key, required String value}) {
     return using((Arena arena) {
       try {
-        _add(key: key, value: value, arena: arena);
```
Jordan-Nelson, line 54: Note: This change isn't related to the bug fix and has no real impact on behavior.

While reviewing Amplify-iOS code, I noticed they first attempt to update the key, and then add it if the key doesn't exist. We were doing the opposite. The update aligns closer with Amplify-swift and is slightly more efficient for updates, which we would expect to occur more frequently than insertions.

packages/secure_storage/amplify_secure_storage/example/integration_test/cupertino_test.dart
```diff
@@ -74,4 +76,92 @@ void main() {
...
+        storage1.removeAll();
+        storage2.removeAll();
+      });
+      test(
+        'read returns null when reading keys with a different kSecAttrAccessible value',
+        () async {
+          storage1.write(key: key1, value: value1);
+
+          expect(storage1.read(key: key1), value1);
+          expect(storage2.read(key: key2), isNull);
```
dnys1, line 121: Should this be `key1`?

packages/secure_storage/amplify_secure_storage/example/integration_test/cupertino_test.dart
```diff
@@ -74,4 +76,92 @@ void main() {
...
+          storage2.delete(key: key1);
+
+          expect(storage1.read(key: key1), isNull);
+        },
+      );
+
+      test(
+        'removeAll can delete keys regardless of the kSecAttrAccessible value',
+        () async {
+          storage1.write(key: key1, value: value1);
```
dnys1, line 141: Should the reverse also be tested? `storage2 write / storage1 removeAll`

packages/secure_storage/amplify_secure_storage_dart/lib/src/interfaces/amplify_secure_storage_interface.dart
```diff
@@ -21,7 +21,7 @@ abstract class AmplifySecureStorageInterface extends SecureStorageInterface
   /// {@template amplify_secure_storage_dart.amplify_secure_storage_interface.remove_all}
   /// Removes all key-value pairs for the current scope.
   ///
-  /// Only available on Linux. Will throw [UnimplementedError]
+  /// Only available on iOS, MacOs, and Linux. Will throw [UnimplementedError]
```
dnys1, line 24:
```suggestion
  /// Only available on iOS, macOS, and Linux. Will throw [UnimplementedError]
```

packages/secure_storage/amplify_secure_storage/example/integration_test/cupertino_test.dart
```diff
@@ -16,18 +17,19 @@ const key2 = 'key_2';
...
-
-/// MacOS & iOS app uninstall & re-install tests.
-///
-/// These tests are only relevant for amplify_secure_storage
-/// (not amplify_secure_storage_dart) as they use
-/// NSUserDefaultsAPI via pigeon.
 void main() {
   IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  /// MacOS & iOS app uninstall & re-install tests.
```
dnys1, line 23:
```suggestion
  /// macOS & iOS app uninstall & re-install tests.
```

packages/secure_storage/amplify_secure_storage/example/integration_test/cupertino_test.dart
```diff
@@ -16,18 +17,19 @@ const key2 = 'key_2';
...
+  ///
+  /// These tests are only relevant for amplify_secure_storage
+  /// (not amplify_secure_storage_dart) as they use
+  /// NSUserDefaultsAPI via pigeon.
```
dnys1, line 27: The new tests can be done in Dart too, though, no?

dnys1 (approved):

packages/secure_storage/amplify_secure_storage/example/integration_test/cupertino_test.dart
Jordan-Nelson: Oops, yes. Updated. Thanks.

packages/secure_storage/amplify_secure_storage/example/integration_test/cupertino_test.dart
Jordan-Nelson: They probably could with some extra setup. On macOS they require keychain entitlement, which makes them easier to run from a flutter app.

PR #2650, opened 2023-02-16
feat(api): GraphQL Subscription Where Filter

Equartey:
*Description of changes:*
Exposes a `where` parameter that accepts a `QueryPredicate` for server side filtering of GraphQL Subscription requests.


packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
```diff
@@ -164,6 +165,32 @@ void main({bool useExistingTestUser = false}) {
...
+            final subscriptionRequest = ModelSubscriptions.onCreate(
+              Post.classType,
+              authorizationMode: APIAuthorizationType.userPools,
+              where: Post.TITLE.eq(postTitle),
+            );
+
+            final eventResponse = await establishSubscriptionAndMutate<Post>(
+              subscriptionRequest,
+              () => addPost(postTitle, blog),
+              eventFilter: (response) => response.data?.title == postTitle,
```
dnys1, line 187: Shouldn't be necessary, right?

packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
```diff
@@ -164,6 +165,32 @@ void main({bool useExistingTestUser = false}) {
...
+            );
+            final blogFromEvent = eventResponse.data;
+
+            expect(blogFromEvent?.title, equals(postTitle));
```
dnys1, line 191:
```suggestion
            final postFromEvent = eventResponse.data;

            expect(postFromEvent?.title, equals(postTitle));
```

packages/api/amplify_api_dart/lib/src/graphql/factories/graphql_request_factory.dart
```diff
@@ -114,6 +114,7 @@ class GraphQLRequestFactory {
     ModelSchema schema,
     GraphQLRequestOperation operation,
     ModelIdentifier? modelIdentifier,
+    bool hasSubFilter,
```
dnys1, line 117: Is this needed? You should be able to add the variables without this. And it's kinda messy.

packages/api/amplify_api/example/integration_test/util.dart
```diff
@@ -138,6 +138,23 @@ Future<Blog> addBlog(String name) async {
   return blog;
 }
 
+// declare utility which creates post with title and blog as parameter
+Future<Post> addPost(String name, Blog blog) async {
```
ragingsquirrel3, line 142: Nice util! Request: Could you please remove similar code from end of `addPostAndBlog` and just call this method from there to DRY?

packages/api/amplify_api_dart/lib/src/graphql/factories/graphql_request_factory.dart
```diff
@@ -169,7 +169,8 @@ class GraphQLRequestFactory {
       case GraphQLRequestOperation.onCreate:
       case GraphQLRequestOperation.onUpdate:
       case GraphQLRequestOperation.onDelete:
-        // upperOutput and lowerOutput already '', done
+        upperOutput = '(\$filter: ModelSubscription${modelName}FilterInput)';
```
ragingsquirrel3, line 172: question: In an older backend w no server-side subscription support, would this throw an error from AppSync when caller provides no filter (would not expect `ModelSubscriptionBlogFilterInput` to be present in transformed schema)? If so, would it be possible to leave this blank when no filter provided?

packages/api/amplify_api_dart/lib/src/graphql/factories/graphql_request_factory.dart
```diff
@@ -251,6 +256,20 @@ class GraphQLRequestFactory {
...
   }
 
+  Map<String, dynamic> buildVariablesForSubscriptionRequest({
+    required ModelType modelType,
+    QueryPredicate? where,
+  }) {
+    if (where == null) {
+      return {};
+    }
+    final filter = GraphQLRequestFactory.instance
```
ragingsquirrel3, line 266: question: Will this work for a subscription w owner field? I wasn't sure if it would implicitly work from the filter or if the owner value had to be moved into another variable value as I thought those subscriptions then took 2 variables (filter and owner).

packages/api/amplify_api_dart/test/graphql_helpers_test.dart
```diff
@@ -774,7 +774,7 @@ void main() {
     group('ModelSubScriptions', () {
       test('ModelSubscriptions.onCreate() should build a valid request', () {
```
ragingsquirrel3, line 775: request: For the tests here with no `where` clause provided, could we add assertions for how the `req.variables` should look? If we want the provided value to be null, empty, `{}`  we should enforce that here with a test.

packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
```diff
@@ -164,6 +165,38 @@ void main({bool useExistingTestUser = false}) {
...
+
+            final blog = await addBlog(blogName);
+
+            // TODO(equartey): Replace this statement when Model object supports parent id field.
+            final pred =
+                const QueryField<String>(fieldName: 'blogID').eq(blog.id);
+
+            final subscriptionRequest = ModelSubscriptions.onCreate(
+              Post.classType,
+              authorizationMode: APIAuthorizationType.userPools,
```
ragingsquirrel3, line 184: nit/question: Why is this test in the api key suite if using userPools? Can it be moved to that suite?

packages/api/amplify_api_dart/lib/src/graphql/factories/graphql_request_factory.dart
```diff
@@ -194,6 +195,7 @@ class GraphQLRequestFactory {
     APIAuthorizationType? authorizationMode,
     Map<String, String>? headers,
     int depth = 0,
+    bool hasSubFilter = false,
```
ragingsquirrel3, line 198: Is this used?

packages/api/amplify_api_dart/lib/src/graphql/factories/graphql_request_factory.dart
ragingsquirrel3: discussed offline, sounds like no, which is ok IMO for this PR. Can be separate effort.

packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
```diff
@@ -164,6 +165,38 @@ void main({bool useExistingTestUser = false}) {
...
+            final blogName =
+                'Integration Test Blog - subscription filter ${uuid()}';
+
+            final postTitle =
+                'Integration Test Post - subscription filter ${uuid()}';
+
+            final blog = await addBlog(blogName);
+
+            // TODO(equartey): Replace this statement when Model object supports parent id field.
+            final pred =
```
ragingsquirrel3, line 179: question: Can this be `Post.BLOG.eq(blog.id)`?. We do that in a test for query and it converts correctly (some unit tests as well). If it does not convert correctly would be good to understand.

packages/api/amplify_api/example/integration_test/util.dart
```diff
@@ -184,24 +201,7 @@ Future<Post> addPostAndBlog(
...
-  final data = createPostRes.data;
-  if (data == null) {
-    throw Exception(
-      'Null response while creating post. Response errors: ${createPostRes.errors}',
-    );
-  }
-  postCache.add(data);
-
-  return data;
+  return addPost(title, rating, createdBlog);
```
ragingsquirrel3, line 204: request: Could you make a similar change here as well in `addPostAndBlog`? https://github.com/aws-amplify/amplify-flutter/blob/next/packages/api/amplify_api/example/integration_test/util.dart#L188

ragingsquirrel3: nm, my mistake

ragingsquirrel3 (approved):

packages/api/amplify_api/example/integration_test/graphql/iam_test.dart
```diff
@@ -443,6 +443,41 @@ void main({bool useExistingTestUser = false}) {
...
+              final postFromEvent = event.data;
+
+              expect(postFromEvent?.title, equals(postTitle));
+
+              dataCompleter.complete(postFromEvent);
+            },
+            onError: (Object e) => fail('Error in subscription stream: $e'),
+          );
+
+          await dataCompleter.future;
```
dnys1, line 478: I don't see this being used for anything except blocking test completion. Maybe wrap the `listen` callback in `expectAsync1` instead

packages/api/amplify_api/example/integration_test/graphql/iam_test.dart
```diff
@@ -443,6 +443,41 @@ void main({bool useExistingTestUser = false}) {
           expect(postFromEvent?.title, equals(title));
         });
 
+        testWidgets('should support where clause when using Model helpers',
```
dnys1, line 446: I think to really test this you should create two blogs and two posts:

```dart
final blog1 = Blog(name: 'blog 1');
final blog2 = Blog(name: 'blog 2');
final post1 = Post(title: 'post 1', blog: blog1);
final post2 = Post(title: 'post 2', blog: blog2);

final sub = subscribe(Post.classType, where: Post.BLOG.eq(blog2.id));

addBlog(blog1);
addBlog(blog2);
addPost(post1);
addPost(post2);
```

`where` should filter out the first post and only return the second.

Currently, this could just be returning the first post.

dnys1 (approved):

packages/api/amplify_api/example/integration_test/graphql/iam_test.dart
Equartey: That was the intent, updated.

PR #2651, opened 2023-02-16
test(auth): Hosted UI E2E tests

dnys1:
Adds E2E tests for Hosted UI on all platforms. The full range of support is described below. Unless otherwise documented in **Notes**, the full e2e workflow can be automatically tested in this PR.

| Platform | Dart | Flutter | Notes |
| -------- | ----- | ------ | ------ |
| iOS | - | ✅ | Everything but native UI |
| Android | - | ✅ | Everything but native UI |
| Web | ✅ | - | Chrome only. Flutter testing does not seem possible but also no longer varies from Dart Web. |
| macOS | ✅ | ✅ | |
| Windows | ✅ | ✅ | |
| Linux | ✅ | ✅ | |

Currently the tests are not running in CI. Changes will be needed to allow Dart package examples to host e2e tests, and to allow Flutter tests to run `flutter_driver` tests as well as `integration_test` tests.


packages/amplify_datastore/example/pubspec.yaml
```diff
@@ -12,6 +12,7 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
+  amplify_api: ">=1.0.0-next.5 <1.0.0-next.6"
   amplify_datastore: ">=1.0.0-next.5 <1.0.0-next.6"
   # Uncomment the below lines to enable online sync
   # amplify_api:
```
Jordan-Nelson, line 18: Can probably remove this commented out dep if you are adding API as a dep.

packages/amplify_test/lib/test_models/Address.dart
```diff
@@ -8,7 +8,7 @@
 // ignore_for_file: public_member_api_docs, annotate_overrides, dead_code, dead_codepublic_member_api_docs, depend_on_referenced_packages, file_names, library_private_types_in_public_api, no_leading_underscores_for_library_prefixes, no_leading_underscores_for_local_identifiers, non_constant_identifier_names, null_check_on_nullable_type_parameter, prefer_adjacent_string_concatenation, prefer_const_constructors, prefer_if_null_operators, prefer_interpolation_to_compose_strings, slash_for_doc_comments, sort_child_properties_last, unnecessary_const, unnecessary_constructor_name, unnecessary_late, unnecessary_new, unnecessary_null_aware_assignments, unnecessary_nullable_for_final_variable_declarations, unnecessary_string_interpolations, use_build_context_synchronously, implicit_dynamic_parameter, implicit_dynamic_map_literal, implicit_dynamic_type
 
 import 'package:amplify_core/amplify_core.dart';
-import 'package:flutter/foundation.dart';
+import 'package:meta/meta.dart';
```
Jordan-Nelson, line 11: What is the reason for this change? I think codegen uses includes `import 'package:flutter/foundation.dart';`

Jordan-Nelson: Oh I see. You are removing flutter as a dep from this package. I think the change is fine. Might be a little confusing if we regenerate these but not a big deal.

packages/auth/amplify_auth_cognito/example/integration_test/hosted_ui_webview_test.dart
```diff
@@ -0,0 +1,256 @@
...
+import 'package:flutter/services.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:integration_test/integration_test.dart';
+import 'package:webview_flutter/webview_flutter.dart';
+// ignore: implementation_imports
+import 'package:webview_flutter_wkwebview/src/foundation/foundation.dart';
+
+import 'utils/setup_utils.dart';
+
+late String username;
```
Jordan-Nelson, line 25: Are these top level so that HostedUiTestPlatform has access? You could probably pass them into the constructor for HostedUiTestPlatform

packages/auth/amplify_auth_cognito_dart/lib/src/state/state/credential_store_state.dart
```diff
@@ -181,7 +183,11 @@ class CredentialStoreFailure extends CredentialStoreState with ErrorState {
 /// {@template amplify_auth_cognito_dart.credential_store_state.credential_store_data}
 /// Data stored by the Credential Store.
 /// {@endtemplate}
-class CredentialStoreData with AWSEquatable<CredentialStoreData> {
+@zAmplifySerializable
+class CredentialStoreData
```
NikaHsn, line 187: is there a security risk on serializing credential data?

packages/auth/amplify_auth_cognito_dart/lib/src/flows/hosted_ui/hosted_ui_platform_html.dart
```diff
@@ -13,27 +13,31 @@ class HostedUiPlatformImpl extends HostedUiPlatform {
   HostedUiPlatformImpl(super.dependencyManager) : super.protected();
 
   /// The base URL
```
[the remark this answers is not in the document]
dnys1: It's not required any longer and I agree this wasn't very helpful to begin with. But in general, doc comments on private fields can still be helpful since they appear when you hover over them in code.

packages/auth/amplify_auth_cognito_dart/lib/src/state/state/credential_store_state.dart
dnys1: None that I can think of. The main risk around credentials is accidentally leaking them via logs - we are protected against that here by not overriding `toString`, so by default, printing an instance of this class will show (`Instance of 'CredentialStoreData'`)

packages/auth/amplify_auth_cognito_dart/lib/src/flows/hosted_ui/hosted_ui_platform.dart
```diff
@@ -95,27 +95,30 @@ abstract class HostedUiPlatform implements Closeable {
...
-      )
-      ..write(
-        key: _keys[HostedUiKey.codeVerifier],
-        value: codeVerifier,
-      )
-      ..write(
-        key: _keys[HostedUiKey.nonce],
-        value: nonce,
-      );
+    await Future.wait<void>(
```
Jordan-Nelson, line 106: Was this a bug that the tests uncovered?

Jordan-Nelson (approved):

packages/auth/amplify_auth_cognito/example/integration_test/hosted_ui_webview_test.dart
dnys1: True. I can change it

packages/amplify_test/lib/test_models/Address.dart
dnys1: Yeah, exactly. We can move test models to their own package if needed or add a `amplify_flutter_test` pkg as well if this becomes too cumbersome.

Hoping codegen gets fixed first though

packages/auth/amplify_auth_cognito_dart/lib/src/flows/hosted_ui/hosted_ui_platform.dart
dnys1: No but maybe I should verify my assumptions that it's needed

packages/auth/amplify_auth_cognito_dart/lib/src/state/state/credential_store_state.dart
NikaHsn: you are right, it does not add risk on accidental leak via prinitng an instance. do you see risk of leaking or malicious actions on instance.toJson() as we are making it serializable?

dnys1: No more risk than exposing the class to begin with, I think. Since you have to explicitly call `toJson` as the developer, just like you would explicitly have to retrieve instance variables if you wanted to print them or send them over the wire somewhere.

NikaHsn (approved):

Issue #2646, opened 2023-02-14
Custom Subscriptions not getting events


ignitum-ryan:
### Description

Hey Amplify Team,

I'm currently running into an issue where no custom subscriptions (both Android and iOS) are not triggering (no events are received by subscribed devices). [I used the following document to help in the setup of the custom subscription](https://docs.amplify.aws/guides/api-graphql/subscriptions-by-id/q/platform/flutter/). Still no luck. I will be pasting all relevant code below, please don't hesitate to ask for more details. Thanks all!

Additional notes on below screenshots:
- I originally did not have `decodePath` and `modelType` in the GraphQLRequest object. Having them there, or not having them in the instantiation results in the same outcome, no events.
- I originally had the sender and receiver objects in my `createdFriendshipGQL` query, however, that also did not result in any events. See below screenshot.

### Categories

- [ ] Analytics
- [ ] API (REST)
- [X] API (GraphQL)
- [ ] Auth
- [ ] Authenticator
- [ ] DataStore
- [ ] Storage

### Steps to Reproduce

1. Create the models.
2. Create the custom subscriptions.
3. Establish the subscriptions upon app start-up.
4. Create records (manually or programmatically, results in same outcome)
5. See that the events never are received by the listening devices.

### Friendship Model
<img width="461" alt="image" src="https://user-images.githubusercontent.com/97566760/218768812-c4d4f4dc-1871-4fec-8ec6-be21c133c3b0.png">

### User Model
<img width="715" alt="image" src="https://user-images.githubusercontent.com/97566760/218769121-6ef43c20-5a95-4a5f-9071-bc3e779f304e.png">

### Function setting up subscription:
![image](https://user-images.githubusercontent.com/97566760/218769697-bb63c526-ccea-4e9e-98b0-090382a87441.png)

### Subscription declaration in schema
<img width="857" alt="image" src="https://user-images.githubusercontent.com/97566760/218770158-ebfc0091-3eee-48e7-bb07-104bd8fd4245.png">

### Attempted with this query as well to no avail
<img width="491" alt="image" src="https://user-images.githubusercontent.com/97566760/218771723-8bf5307c-8085-40d4-857a-a6c0b75425c8.png">

### Platforms

- [X] iOS
- [X] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Flutter Version

3.7.1

### Amplify Flutter Version

^1.0.0-next.4

### Deployment Method

[...]

PR #2655, opened 2023-02-16
chore(api): model helpers, include parent IDs in selection set

ragingsquirrel3:
This small change will help reduce friction for a paper cut detail I noticed while debugging https://github.com/aws-amplify/amplify-flutter/issues/2646.

problem: If a user does a GQL subscription and filters by parent ID (e.g. get all comments created where post ID equals x) then that subscription won't get updates made creating comments from requests with `ModelMutations.create()`. This is because a subscription with this kind of filter requires the selection set of the mutation include the field itself (e.g. `"postID"`, not the ID of the post in the nested selection set).

This PR simply adds the ID fields to the selection sets of request documents when the schema has `belongsTo`. We already get the parent itself (which has the ID nested). However, here we essentially put the ID in 2 places. The new ID instance is the response is ignored while converted to model with `fromJson` while deserializing the server response. However, any subscription by parent ID will now pickup mutations made with these requests.

### DIFF
--- a/packages/api/amplify_api_dart/lib/src/graphql/factories/graphql_request_factory.dart
+++ b/packages/api/amplify_api_dart/lib/src/graphql/factories/graphql_request_factory.dart
@@ -70,6 +70,19 @@ class GraphQLRequestFactory {
           ignoreParents: true,
         ); // always format like a get, stop traversing parents
         fields.add('${belongsTo.name} { $parentSelectionSet }');
+
+        // Include the ID key(s) itself in selection set. This is ignored while
+        // deserializing response because model will use ID nested in `parentSelectionSet`.
+        // However, including the ID in selection set allows subscriptions that
+        // filter by these ID to be triggered by these mutations.
+        var belongsToKeys = belongsTo.association?.targetNames;
+        if (belongsToKeys == null &&
+            belongsTo.association?.targetName != null) {
+          belongsToKeys = [belongsTo.association?.targetName as String];
+        }
+        if (belongsToKeys != null) {
+          fields.addAll(belongsToKeys);
+        }
       }
     }
 
--- a/packages/api/amplify_api_dart/test/graphql_helpers_mtm_test.dart
+++ b/packages/api/amplify_api_dart/test/graphql_helpers_mtm_test.dart
@@ -28,9 +28,9 @@ void main() {
     });
 
     const manyToManyPrimarySelectionSet =
-        'manyToManyPrimary { id name createdAt updatedAt }';
+        'manyToManyPrimary { id name createdAt updatedAt } manyToManyPrimaryId';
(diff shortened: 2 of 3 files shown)


packages/api/amplify_api_dart/lib/src/graphql/factories/graphql_request_factory.dart
```diff
@@ -70,6 +70,19 @@ class GraphQLRequestFactory {
...
         fields.add('${belongsTo.name} { $parentSelectionSet }');
+
+        // Include the ID key(s) itself in selection set. This is ignored while
+        // deserializing response because model will use ID nested in `parentSelectionSet`.
+        // However, including the ID in selection set allows subscriptions that
+        // filter by these ID to be triggered by these mutations.
+        var belongsToKeys = belongsTo.association?.targetNames;
+        if (belongsToKeys == null &&
+            belongsTo.association?.targetName != null) {
+          belongsToKeys = [belongsTo.association?.targetName as String];
```
Equartey, line 81:
```suggestion
          belongsToKeys = [belongsTo.association!.targetName as String];
```
nit: you know there is value here from the check on the previous line.

packages/api/amplify_api_dart/lib/src/graphql/factories/graphql_request_factory.dart
```diff
@@ -70,6 +70,19 @@ class GraphQLRequestFactory {
...
+        }
+        if (belongsToKeys != null) {
+          fields.addAll(belongsToKeys);
```
Equartey, line 84: Question: Is this for when a Model has multiple belongsTo relationships?

HuiSF (approved):

Equartey (approved):
Nice catch! This would make a good integ test case the `where` filter too.

packages/api/amplify_api_dart/lib/src/graphql/factories/graphql_request_factory.dart
ragingsquirrel3: This is actually for if there is a custom primary key w complex identifier (multiple keys) on a single relationship. Even though the CPK part not merged to target branch yet, just wrote this in a way will work with CPK, or simpler, older codegens.

NikaHsn: Q: does this change interfear with `buildInputVariableForMutations`?

ragingsquirrel3: No, will not interfere with that. We are already including these IDs in the input object built by `buildInputVariableForMutations` as the input of the mutation, this PR just a small change to requested response from AppSync.

NikaHsn (approved):

PR #2679, opened 2023-02-17
fix(api): write null values in ModelMutations.create() unless owner field

ragingsquirrel3:
This PR has a correction for some logic recently changed in https://github.com/aws-amplify/amplify-flutter/pull/2504. While the previous PR changed `ModelMutations.create()` to no longer include any null values in the input object, this PR makes it so only owner fields with null values will be removed from input and other null values will remain.

While removing null owner values is essential to fix the issue mentioned in https://github.com/aws-amplify/amplify-flutter/issues/2492, removing all null fields from create has some consequences that are difficult for users to understand or predict. Such as:

* Data created with older versions of model helpers is different than newer versions. Newer versions have empty values in DynamoDB versus explicit nulls from older entries. This makes is so `.ne(null)` will match newer versions with undefined values (the issue that was reported). While breaking changes in dev preview are expected, I don't think this change has been given proper thought/communication.
* Because `ModelMutations.update()` includes null values (impossible to determine if undefined or explicit null there), any model updated for any reason will have all nulls, no undefined fields. For a user, it would be confusing why their data is in 2 states from create/update when the model is in the same state.

Ultimately, the issues is that the codegen models have no way to distinguish between explicit nulls and undefined. Without a solution at that level, this PR makes things more consistent while still preventing the originally reported issue of null values in owner fields.

### DIFF
--- a/packages/api/amplify_api_dart/lib/src/graphql/factories/graphql_request_factory.dart
+++ b/packages/api/amplify_api_dart/lib/src/graphql/factories/graphql_request_factory.dart
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 import 'package:amplify_api_dart/src/graphql/utils.dart';
 import 'package:amplify_core/amplify_core.dart';
+import 'package:collection/collection.dart';
 
 // ignore_for_file: public_member_api_docs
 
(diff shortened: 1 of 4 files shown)


packages/api/amplify_api_dart/lib/src/graphql/factories/graphql_request_factory.dart
```diff
@@ -322,6 +322,10 @@ class GraphQLRequestFactory {
       }
     }
 
+    final ownerFieldNames = (schema.authRules ?? [])
+        .where((authRule) => authRule.ownerField != null)
+        .map((authRule) => authRule.ownerField!)
```
dnys1, line 327:
```suggestion
        .map((authRule) => authRule.ownerField)
        .whereType<String>() // or `whereNotNull()` from `package:collection`
```

dnys1 (approved):

HuiSF (approved):

PR #2688, opened 2023-02-22
fix(auth): Transform session expired exceptions

dnys1:
In the fetch auth session state machine, any `NotAuthorizedException` thrown should be considered a `SessionExpiredException` since that is the only time we would get such an error.


packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/fetch_auth_session_state_machine.dart
```diff
@@ -563,6 +569,20 @@ extension on CredentialStoreData {
   }
 }
 
+extension on AuthNotAuthorizedException {
+  /// In the fetch auth session state machine, all [AuthNotAuthorizedException]s
+  /// are mapped to [SessionExpiredException]s, since the only time they would
+  /// be thrown is due to expired/revoked credentials being used.
+  SessionExpiredException toSessionExpired(String message) {
```
ragingsquirrel3, line 576: nice util!

ragingsquirrel3 (approved):

NikaHsn (approved):

PR #2690, opened 2023-02-22
chore(core): Mark event completer methods with `@useResult`

dnys1:
Event completers should be used or explicitly ignored since any errors are swallowed by them.

### DIFF
--- a/packages/auth/amplify_auth_cognito/lib/src/flows/hosted_ui/hosted_ui_platform_flutter.dart
+++ b/packages/auth/amplify_auth_cognito/lib/src/flows/hosted_ui/hosted_ui_platform_flutter.dart
@@ -74,13 +74,17 @@ class HostedUiPlatformImpl extends io.HostedUiPlatformImpl {
         options.isPreferPrivateSession,
         options.browserPackageName,
       );
-      dispatcher.dispatch(
-        HostedUiEvent.exchange(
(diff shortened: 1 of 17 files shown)


ragingsquirrel3 (approved):

packages/auth/amplify_auth_cognito/lib/src/flows/hosted_ui/hosted_ui_platform_flutter.dart
```diff
@@ -74,13 +74,17 @@ class HostedUiPlatformImpl extends io.HostedUiPlatformImpl {
         options.isPreferPrivateSession,
         options.browserPackageName,
       );
-      dispatcher.dispatch(
-        HostedUiEvent.exchange(
-          OAuthParameters.fromJson(queryParameters.cast()),
+      unawaited(
```
NikaHsn, line 77: Q: what is the difference between dispatch().ignore() and unwaited(dispatchAndComplete())

dnys1: `dispatchAndComplete` will throw an error if the event processing finishes with an `ErrorState`. `unawaited` allows this error to surface as an unhandled exception. It shouldn't happen but we also can't lose visibility on it.

`dispatch.ignore()` basically swallows any errors if the event finishes in an error state.

PR #2691, opened 2023-02-22
refactor(auth)!: Plugin options

dnys1:
Migrates all Auth APIs to new plugin options


Jordan-Nelson (approved):

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
```diff
@@ -98,39 +81,40 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface<
       CognitoUserAttributeKey,
       AuthUserAttribute<CognitoUserAttributeKey>,
       CognitoDevice,
-      CognitoSignUpOptions,
```
HuiSF, line 101: So when are not going to strongly type `pluginOptions` when using `Auth.getPlugin(AuthPluginKey)`?

dnys1: Nope

packages/auth/amplify_auth_cognito_dart/lib/src/model/attribute/cognito_resend_user_attribute_confirmation_code_options.dart
```diff
@@ -5,39 +5,87 @@ import 'package:amplify_core/amplify_core.dart';
...
+  @override
+  Map<String, Object?> toJson() =>
+      _$CognitoResendUserAttributeConfirmationCodeOptionsToJson(this);
+}
+
+/// {@template amplify_auth_cognito.model.cognito_resend_user_attribute_confirmation_code_plugin_options}
+/// Cognito options for `Amplify.Auth.resendUserAttributeConfirmationCode`.
+/// {@endtemplate}
+@zAmplifySerializable
+class CognitoResendUserAttributeConfirmationCodePluginOptions
```
NikaHsn, line 62: how about to move PluginOptions classes to their own file so that they follow the class -> file naming convention. I mean instead of having `CognitoResendUserAttributeConfirmationCodePluginOptions` defined in `cognito_resend_user_attribute_confirmation_code_options.dart` it will be defined in `cognito_resend_user_attribute_confirmation_code_plugin_options.dart`

NikaHsn (approved):

packages/auth/amplify_auth_cognito_dart/lib/src/model/attribute/cognito_resend_user_attribute_confirmation_code_options.dart
dnys1: Hm do you think we should do the same for the category plugin options as well?

NikaHsn: I don't think so bc they are abstract classes and used only by category level options and so it is okay to have them inside the options file.

PR #2694, opened 2023-02-23
chore(auth): Add `SignOutStateMachine`

dnys1:
Integrates the sign out flow into the state machine queueing system which helps prevent against some weird race conditions and ensure serializability of API calls.


packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_out_state_machine.dart
```diff
@@ -0,0 +1,189 @@
...
+      if (hostedUiResult != null) {
+        return emit(SignOutState.complete(hostedUiResult));
+      }
+    }
+
+    // Do not try to send Cognito requests for plugin configs without an
+    // Identity Pool, since the requests will fail.
+    if (_identityPoolConfig != null) {
+      if (options.globalSignOut) {
+        // Revokes the refresh token
```
NikaHsn, line 108:
```suggestion
        // Revokes the access token
```

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_out_state_machine.dart
```diff
@@ -0,0 +1,189 @@
...
+          await _cognitoIdp
+              .globalSignOut(
+                GlobalSignOutRequest(
+                  accessToken: tokens.accessToken.raw,
+                ),
+              )
+              .result;
+        } on Exception catch (e) {
+          globalSignOutException = GlobalSignOutException(
+            accessToken: tokens.accessToken.raw,
```
NikaHsn, line 119: what is the need to capture accessToken.raw in this exception?

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_out_state_machine.dart
```diff
@@ -0,0 +1,189 @@
...
+            underlyingException: e,
+          );
+        }
+      }
+      // Revokes the access token
```
NikaHsn, line 124:
```suggestion
      // Revokes the refresh token
```

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_out_state_machine.dart
```diff
@@ -0,0 +1,189 @@
...
+      if (globalSignOutException != null) {
+        revokeTokenException = RevokeTokenException(
+          refreshToken: tokens.refreshToken,
```
NikaHsn, line 127: what is the need to capture refreshToken in this exception?

dnys1: It gives users the opportunity to retry the revoke on their own. After this function completely, all the tokens are cleared locally.

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_out_state_machine.dart
dnys1: It's actually the reverse, confusingly. We revoke the access token with the refresh token and vice versa.

NikaHsn:
Q: in sign_out_state_machine.dart we do SignOutState.complete even if the CognitoSignOutResult is failed or partially failed. why not to add failure state to SignOutStateType and use it for these cases?

dnys1:
> Q: in sign_out_state_machine.dart we do SignOutState.complete even if the CognitoSignOutResult is failed or partially failed. why not to add failure state to SignOutStateType and use it for these cases?

Good point - I don't like that naming. I changed it to be more inline with other state machines and broke out the failure states.

ragingsquirrel3 (approved):

NikaHsn (approved):

PR #2695, opened 2023-02-24
chore: migrate Android unit tests to GH actions

ragingsquirrel3:
This PR migrates Android unit tests to GH action by doing the following:
* change `aft generate workflows` command to generate GH actions for Android unit tests
* the result of running that command
* deleted a test that was flaky in GH action runner (more info in inline comment)

Removing CircleCI equivalent should be separate change after similar change on main/stable and removing check from repo settings.


.github/workflows/flutter_android.yaml
```diff
@@ -0,0 +1,60 @@
+name: Flutter Android
```
ragingsquirrel3, line 1: will improve naming of this file I suppose

packages/amplify/amplify_flutter_android/android/src/test/kotlin/com/amazonaws/amplify/AtomicResultTest.kt
```diff
@@ -58,18 +54,4 @@ class AtomicResultTest {
         atomicResult.success(null)
         verify(mockResult, times(1)).success(any())
     }
-
-    @Test
```
ragingsquirrel3, line 62: This test was flaky on both stable/beta, example failure https://github.com/aws-amplify/amplify-flutter/actions/runs/4297577694/jobs/7490721332. Seems to fail half or 1/3 the time. Would love to keep the test if anyone has a suggestion/insight for why this may be that I could investigate further. I wasn't sure how much it was worth investigating.

packages/aft/lib/src/commands/generate/generate_workflows_command.dart
```diff
@@ -24,12 +24,27 @@ class GenerateWorkflowsCommand extends AmplifyCommand {
           !falsePositiveExamples.contains(package.name)) {
         continue;
       }
-      // Some packages contain only native code/dependencies and do not need
-      // to be tested by Dart/Flutter analyze or test.
+      // Some packages do not need to be tested.
```
dnys1, line 27: Instead of shortening it, maybe we need to add more context.

```suggestion
      // Packages without a `lib/` folder generally contain only native code,
      // e.g. `amplify_flutter_android`. These packages are tested via their
      // parent package, e.g. `amplify_flutter`, and do not require a workflow
      // of their own. 
```

packages/aft/lib/src/commands/generate/generate_workflows_command.dart
```diff
@@ -24,12 +24,27 @@ class GenerateWorkflowsCommand extends AmplifyCommand {
...
       final libDir = Directory(p.join(package.path, 'lib'));
       if (!libDir.existsSync()) {
         continue;
       }
+
+      final internalAndroidTestDir =
+          Directory(p.join(package.path, 'android', 'src', 'test'));
+      final externalAndroidPackageTestDir = Directory(
```
dnys1, line 35: `internal`/`external` is a bit confusing to me, since I would think of the `_android` package as being "internal".

Maybe we can borrow the language of [federated plugins](https://docs.flutter.dev/development/packages-and-plugins/developing-packages) and use `appFacing`/`platform` instead

packages/aft/lib/src/commands/generate/generate_workflows_command.dart
```diff
@@ -127,7 +145,64 @@ jobs:
...
+permissions: read-all
+
+jobs:
+  test:
+    uses: ./.github/workflows/$androidWorkflow
+    with:
+      working-directory: $repoRelativePath
+''';
+
+      androidWorkflowFile.writeAsStringSync(androidWorkflowContents);
```
dnys1, line 205: Can we move all the new logic to its own method since it seems completely independent of the rest?

.github/workflows/flutter_android.yaml
```diff
@@ -0,0 +1,60 @@
...
+        run: aft --config=aft.beta.yaml constraints apply
+
+      - name: Bootstrap
+        id: bootstrap
+        timeout-minutes: 10
+        run: aft bootstrap
+
+      - name: run Android unit tests
+        run: |
+          packageName=$(sed 's:.*/::' <<< ${{ inputs.working-directory }})
```
dnys1, line 58: Can you pass in the name as an input?

.github/workflows/flutter_android.yaml
```diff
@@ -0,0 +1,60 @@
...
+          - stable
+          - beta
+    steps:
+      - name: Git Checkout
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # 3.1.0
+
+      - name: Git Submodules
+        run: git submodule update --init
+
+      - name: Setup Gradle
```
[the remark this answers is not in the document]
ragingsquirrel3: good suggestion, thanks. With this change I found I had to run `flutter build apk` before running tests (like we currently do in CircleCI). LMK if this was not expected.

packages/aft/lib/src/commands/generate/generate_workflows_command.dart
ragingsquirrel3: good suggestion, thanks

.github/workflows/flutter_android.yaml
ragingsquirrel3: yep

packages/aft/lib/src/commands/generate/generate_workflows_command.dart
ragingsquirrel3: good suggestion, done. This will also help once iOS added here.

.github/workflows/flutter_android.yaml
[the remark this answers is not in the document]
dnys1: Sounds good - must be the wrapper is only generated when building the APK

.github/workflows/flutter_android.yaml
```diff
@@ -0,0 +1,57 @@
+name: Flutter Android
+on:
+  workflow_call:
+    inputs:
+      working-directory:
+        description: The working directory relative to the repo root
+        required: true
+        type: string
+      package-name:
+        description: The working directory relative to the repo root
```
dnys1, line 10:
```suggestion
        description: The name of the app-facing package being tested
```

.github/workflows/flutter_android.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+      - name: Git Submodules
+        run: git submodule update --init
+
+      - name: Setup Flutter
+        uses: subosito/flutter-action@dbf1fa04f4d2e52c33185153d06cdb5443aa189d # 2.8.0
+        with:
+          cache: true
+          channel: ${{ matrix.sdk }}
+          flutter-version: ${{ matrix.flutter-version }}
```
dnys1, line 38: Remove

.github/workflows/flutter_android.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Bootstrap
+        id: bootstrap
+        timeout-minutes: 10
+        run: aft bootstrap
+
+      - name: Run Android unit tests
+        run: |
+          flutter build apk --debug --verbose
+          ./gradlew :"${{ inputs.package-name}}":testDebugUnitTest
+        working-directory: ${{ inputs.working-directory }}/example/android
```
dnys1, line 57: Would almost rather pass the full path in since it's not being used anywhere else

packages/aft/lib/src/commands/generate/generate_workflows_command.dart
```diff
@@ -127,7 +148,79 @@ jobs:
...
+          packageName: package.name,
+          repoRelativePath: repoRelativePath,
+          platformPackageAndroidTestDirExists:
+              platformPackageAndroidTestDirExists,
+        );
+      }
     }
   }
+
+  Future<void> generateAndroidUnitTestWorkflow({
```
dnys1, line 166: I would move all the logic here, including the determination of paths.

packages/aft/lib/src/commands/generate/generate_workflows_command.dart
```diff
@@ -128,6 +130,93 @@ jobs:
...
+
+    if (package.flavor != PackageFlavor.flutter || !hasAndroidTests) {
+      return;
     }
+
+    final androidWorkflowFilepath = p.join(
+      rootDir.path,
+      '.github',
+      'workflows',
+      '${package.name}_android.yaml',
```
dnys1, line 173: Can we rename these to `{package.name}.android.yaml` to distinguish from a package named `{package.name}_android`?

dnys1 (commented):
LGTM thanks! 🚀

dnys1 (approved):

HuiSF (approved):

Issue #2647, opened 2023-02-15
(iOS) App crashes  when when trying to save/query with DataStore


jeemoney:
### Description

I'm seeing this exception consistently on iOS simulator when performing any DataStore query or save call:

> DataStoreException(message: The operation couldn’t be completed. (SQLite.Result error 0.), recoverySuggestion: The operation couldn’t be completed. (SQLite.Result error 0.), underlyingException: The operation couldn’t be completed. (SQLite.Result error 0.))

I created a new Flutter project and a new Amplify app with the default Blog/Post/Comment schema to test whether or not our schema was the issue. I see the same thing with this new project and boilerplate schema. I was able to test on my iPhone and I'm still seeing the same issue, but with a bit more detail in the error message suggesting that the table for the model doesn't exist.

Both our app and this test app are working fine on Android simulator and physical device.

### Categories

- [ ] Analytics
- [ ] API (REST)
- [X] API (GraphQL)
- [ ] Auth
- [ ] Authenticator
- [X] DataStore
- [ ] Storage

### Steps to Reproduce

```dart
class _MyHomePageState extends State<MyHomePage> {
  @override
  void initState() {
    super.initState();
    _configureAmplify();
  }

  List<String> posts = [];

  Future<void> addPost() async {
    final post = Post(title: 'New Post');
    final comment = Comment(content: 'New Comment', post: post);
    await Amplify.DataStore.save(post);
    await Amplify.DataStore.save(comment);
    final fetchedPosts = await Amplify.DataStore.query(Post.classType);

    setState(() {
      posts = fetchedPosts.map((post) => post.title).toList();
    });
  }

  Future<void> _configureAmplify() async {
    final authPlugin = AmplifyAuthCognito();
    final apiPlugin = AmplifyAPI();
    final datastorePlugin =
        AmplifyDataStore(modelProvider: ModelProvider.instance);
    await Amplify.addPlugins([authPlugin, apiPlugin, datastorePlugin]);

    try {
      await Amplify.configure(amplifyconfig);
    } on AmplifyAlreadyConfiguredException {
      safePrint(
          "Tried to reconfigure Amplify; this can occur when your app restarts on Android.");
    }
  }
```

This code throws an exception on the first DataStore.save call, regardless of which model. I also tried creating a blog in the web studio and allowing DataSource time to sync it down, then tried a fetch. Same result.

### Platforms

- [X] iOS
- [ ] Android
- [ ] Web
- [ ] macOS
[...]

PR #2698, opened 2023-02-24
fix(api): early call of Amplify creates wrong instance of AmplifyClass

HuiSF:
*Issue #, if available:*

https://github.com/aws-amplify/amplify-flutter/issues/2647

*Description of changes:*

The constructor of `AmplifyAPIDart` invokes `Amplify.hub` to add channel, as this happen during initiating the API plugin

```dart
final apiPlugin = AmplifyAPI();
```

https://github.com/aws-amplify/amplify-flutter/blob/d21259d326cd1ad2215d9cdba4cf6c04872f6e57/packages/api/amplify_api_dart/lib/src/api_plugin_impl.dart#L23-L33

It causes `Amplify` to be backed back wrong instance of `AmplifyClass` provided by  `amplify_core`, which should be the `AmplifyClass` provided by `amplify_flutter`. Therefore, when `Amplify.configure()` is called, the native plugins won't be configured via the `AmplifyHybridImpl.configurePlatform()` provided by `amplify_flutter`.

Moving the adding hub to `AmplifyAPIDart.configure()`.

### DIFF
--- a/packages/api/amplify_api_dart/lib/src/api_plugin_impl.dart
+++ b/packages/api/amplify_api_dart/lib/src/api_plugin_impl.dart
@@ -29,7 +29,6 @@ class AmplifyAPIDart extends APIPluginInterface with AWSDebuggable {
   })  : _baseHttpClient = baseHttpClient,
         _connectivity = connectivity {
     authProviders.forEach(registerAuthProvider);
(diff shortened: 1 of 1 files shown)


dnys1 (approved):
Good catch!

ragingsquirrel3 (approved):

Issue #2705, opened 2023-02-27
[cognito] Uri not awaited


passsy:
In `amplify_auth_cognito-1.0.0-next.5+1` is discovered that the uri is not awaited, thus trying to open url `Instance of 'Future<Uri>'` instead of the actual URL. (iOS)

This results in the login not working at all

https://github.com/aws-amplify/amplify-flutter/blob/84efa67b01a580150e28e1bc7c2e5bbb01395c62/packages/auth/amplify_auth_cognito/lib/src/flows/hosted_ui/hosted_ui_platform_flutter.dart#L78

Correct would be
```dart
final signInUri = await getSignInUri(provider: provider);
```

dnys1:
The fix for this has been released in v1.0.0-next.6 of the libraries. Apologies for the delay and thanks for your patience!

PR #2706, opened 2023-02-27
fix(auth): Await `signInUri` in Hosted UI platform

passsy:
Fixes https://github.com/aws-amplify/amplify-flutter/issues/2705

I don't think this ever worked. The URL always resolved to `"Instance of 'Future<Uri>'"`


dnys1:
Thanks, @passsy! The method signature was very recently updated to return a Future. I will look into adding tests which verify this behavior. In the meantime, yes this is the correct fix :)

NikaHsn (approved):

ragingsquirrel3 (approved):

PR #2707, opened 2023-02-27
feat(notifications): implement push notification Flutter iOS module

HuiSF:
*Issue #, if available:*

*Description of changes:*


packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
```diff
@@ -0,0 +1,89 @@
...
+        didFinishLaunchingWithOptions launchOptions: [AnyHashable : Any] = [:]
+    ) -> Bool {
+        if let remoteNotification = launchOptions[UIApplication.LaunchOptionsKey.remoteNotification] as? [AnyHashable: Any] {
+            // If the App is awaken by a remote notification when background mode is enabled
+            if application.applicationState == .background {
+                isBackgroundMode = true
+            }
+
+            // When the App is launched (from terminated state) by tapping on a notification
+            // This happens whne background mode is not enabled
```
dnys1, line 35:
```suggestion
            // This happens when background mode is not enabled
```

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin.swift
```diff
@@ -1,19 +1,183 @@
...
-    default:
-      result(FlutterMethodNotImplemented)
-    }
-  }
+    private let methodChannel: FlutterMethodChannel
+
+    internal var remoteNotificationCompletionHandlers: [String: (UIBackgroundFetchResult) -> Void] = [:]
+    internal var isBackgroundMode = false
+    internal var launchNotification: [AnyHashable: Any]?
+    internal var cachedDeviceToken: String?
```
dnys1, line 15: Needs doc comments

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin.swift
```diff
@@ -1,19 +1,183 @@
...
+                name: pushNotificationMethodChannelName,
+                binaryMessenger: registrar.messenger()
+            )
+        )
+
+        registrar.addMethodCallDelegate(pluginInstance, channel: pluginInstance.methodChannel)
+        registrar.addApplicationDelegate(pluginInstance)
+    }
+
+    public func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
```
dnys1, line 37: Why are we not using `pigeon`?

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin.swift
```diff
@@ -1,19 +1,183 @@
...
+    private func completeNotification(
+        args: Any?,
+        completionHandler: @escaping (PushNotificationOperationResult<Bool>) -> Void
+    ) {
+        guard let completionHandlerId = args as? String else {
+            completionHandler(.failure(invalidArgumentsError))
+            return
+        }
+
+        if let completionHanlder = remoteNotificationCompletionHandlers[completionHandlerId] {
```
dnys1, line 176:
```suggestion
        if let completionHandler = remoteNotificationCompletionHandlers[completionHandlerId] {
```

dnys1: Actually can you call it something else so it's different than the method's completion handler?

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
```diff
@@ -0,0 +1,89 @@
...
+
+            remoteNotificationCompletionHandlers[completionHandlerId] = completionHandler
+            userInfoCopy[completionHandlerIdKey] = completionHandlerId
+
+            EventsStreamHandlers.shared.backgroundMessageReceived.sendEvent(payload: userInfoCopy)
+        } else if UIApplication.shared.applicationState == .inactive {
+            // When the App entering foreground by tapping on a notification
+            // and it's previously awaken by a remote notification, we treat the tapped notification
+            // as the launchNotitication
+            if (isBackgroundMode) {
```
dnys1, line 72: Not following what this tracks and why it's different than `applicationState == .background`

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
dnys1: What is background mode?

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
```diff
@@ -0,0 +1,89 @@
...
+extension AmplifyPushNotificationsPlugin {
+    public func application(
+        _ application: UIApplication,
+        didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data
+    ) {
+        let token = deviceToken.map { String(format: "%02.2hhx", $0) }.joined()
+
+        // the cached token check is to reduce the frequence of emitting TOKEN_RECEIVED events
+        // the deviceToken that client gets is still from the system directly
+        if (cachedDeviceToken != token) {
```
dnys1, line 18: I wouldn't worry about this here - I would send all tokens and filter on the Dart side so the logic applies to both platforms

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        return true
+    }
+
+    public func application(
+        _ application: UIApplication,
+        didFailToRegisterForRemoteNotificationsWithError error: Error
+    ) {
+        print("registerForRemoteNotifications failed with error: \(error).")
```
dnys1, line 48: Print on Dart side using `safePrint` - not sure if this is safe to print in prod

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
```diff
@@ -0,0 +1,89 @@
...
+            }
+        } else {
+            EventsStreamHandlers.shared.foregroundMessageReceived.sendEvent(payload: userInfo)
+        }
+
+        return true
+    }
+
+    public func applicationDidBecomeActive(_ application: UIApplication) {
+        application.registerForRemoteNotifications()
```
Samaritan1011001, line 87: Should this happen in `didFinishLaunchingWithOptions`?

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
```diff
@@ -0,0 +1,89 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import Foundation
+import Flutter
+
+private let completionHandlerIdKey = "completionHandlerId"
+
+extension AmplifyPushNotificationsPlugin {
```
Samaritan1011001, line 9: Is it a valid filename and way of doing it? `AmplifyPushNotificationsPlugin+AppDelegate.swift` 
Would it still allow apps to have their own AppDelegate?

packages/notifications/push/amplify_push_notifications/ios/Classes/PushNotificationEventsStreamHandler.swift
```diff
@@ -0,0 +1,141 @@
...
+        if let eventSink = eventSink {
+            eventSink(event.toMap())
+        } else {
+            eventQueue.append(event)
+        }
+    }
+
+    func sendError(error: FlutterError) {
+        if let eventSink = eventSink {
+            eventSink(error)
```
Samaritan1011001, line 97: This should go to the `eventQueue` too. I ran into this problem where there may be an error before the eventsink is ready, that time the error gets lost. Going through the queue will ensure error gets emitted when the stream is listened to and sink is ready.

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
HuiSF: This naming indicate this file contains additional `AppDelegate` behaviors added to `AmplifyPushNotificationsPlugin`. 

Here we implement `FlutterPlugin` protocol defined App life cycle methods, it doesn't block developers to do anything with their own `AppDelegate`.

packages/notifications/push/amplify_push_notifications/ios/Classes/PushNotificationEventsStreamHandler.swift
HuiSF: updated to queue the error as well.

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin.swift
HuiSF: Added

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin.swift
HuiSF: Using pigeon now.

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
HuiSF: We are sending the error over to the Dart impl. now this print is unnecessary, removed.

packages/notifications/push/amplify_push_notifications/pigeons/native_push_notification_plugin.dart
```diff
@@ -0,0 +1,56 @@
...
+  String getPermissionStatus();
+
+  @async
+  bool requestPermissions(PermissionsOptions withPermissionOptions);
+
+  @async
+  Map<Object?, Object?>? getLaunchNotification();
+
+  @async
+  int getBadgeCount();
```
dnys1, line 51:
```suggestion
  int getBadgeCount();
```

packages/notifications/push/amplify_push_notifications/pigeons/native_push_notification_plugin.dart
```diff
@@ -0,0 +1,56 @@
...
+@HostApi()
+abstract class NativePushNotificationsPlugin {
+  @async
+  String getPermissionStatus();
+
+  @async
+  bool requestPermissions(PermissionsOptions withPermissionOptions);
+
+  @async
+  Map<Object?, Object?>? getLaunchNotification();
```
dnys1, line 48:
```suggestion
  Map<Object?, Object?>? getLaunchNotification();
```

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin.swift
```diff
@@ -1,19 +1,114 @@
...
+        if withPermissionOptions.badge.boolValue {
+            options.insert(.badge)
+        }
+
+        if withPermissionOptions.sound.boolValue {
+            options.insert(.sound)
+        }
+
+        UNUserNotificationCenter.current().requestAuthorization(options: options) { granted, error in
+            if (error != nil) {
```
dnys1, line 77:
```suggestion
            guard error == nil else {
```

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin.swift
```diff
@@ -1,19 +1,114 @@
...
+        )
+
+        NativePushNotificationsPluginSetup(messenger, pluginInstance)
+
+        registrar.addApplicationDelegate(pluginInstance)
+    }
+
+    public func getPermissionStatus(completion: @escaping (String?, FlutterError?) -> Void) {
+        UNUserNotificationCenter.current().getNotificationSettings { settings in
+            var status: String
```
dnys1, line 39: Pigeon supports enums - would prefer that over strings

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
```diff
@@ -0,0 +1,82 @@
...
+    ) -> Bool {
+        if let remoteNotification = launchOptions[UIApplication.LaunchOptionsKey.remoteNotification] as? [AnyHashable: Any] {
+            // If the App is awaken by a remote notification when remote notifications background mode is enabled
+            if application.applicationState == .background {
+                awakenInBackground = true
+            }
+
+            // When the App is launched (from terminated state) by tapping on a notification
+            // This happens when remote notifications background mode is NOT enabled
+            if application.applicationState != .background {
```
dnys1, line 30: This should be an `else` clause since it's orthogonal to the `if` above and is the reason we don't set `launchNotification`

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
```diff
@@ -0,0 +1,82 @@
...
+            remoteNotificationCompletionHandlers[completionHandlerId] = completionHandler
+            userInfoCopy[completionHandlerIdKey] = completionHandlerId
+
+            sharedEventsStreamHandlers.backgroundMessageReceived.sendEvent(payload: userInfoCopy)
+        } else if UIApplication.shared.applicationState == .inactive {
+            // If the App has been awaken in the background by remote notification from terminated state,
+            // when the App is entering foreground from background the first time by tapping on a notification,
+            // we treat this tapped notification as the launch notitification, this only happens when
+            // remote notification background mode is enabled.
+            if (awakenInBackground) {
```
dnys1, line 65:
```suggestion
            if awakenInBackground {
```

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
```diff
@@ -0,0 +1,82 @@
...
+            error: FlutterError(code: "DeviceTokenError", message: error.localizedDescription, details: error)
+        )
+    }
+
+    public func application(
+        _ application: UIApplication,
+        didReceiveRemoteNotification userInfo: [AnyHashable : Any],
+        fetchCompletionHandler completionHandler: @escaping (UIBackgroundFetchResult) -> Void
+    ) -> Bool {
+        if UIApplication.shared.applicationState == .background {
```
dnys1, line 52: Can you switch on `applicationState` instead of if/else?

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
dnys1: Not sure I understand this still. I think some diagrams will be helpful since it's not clear why we need this flag if we end up setting `launchNotification` anyway to (I'm guessing) the same payload.

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
dnys1: Same Q

HuiSF: No we discussed during the development of react-native iOS module, that we invoke `registerForRemoteNotifications`  when App becomes active, according to Apple documentation `registerForRemoteNotifications` returns fast after the very first call there shouldn't be a performance concern here.

packages/notifications/push/amplify_push_notifications/pubspec.yaml
```diff
@@ -20,6 +20,7 @@ dev_dependencies:
     path: ../../../amplify_test
   flutter_test:
     sdk: flutter
+  pigeon: ^3.1.6
```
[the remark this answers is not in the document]
HuiSF: The base branch is outdated I presume. Will update when rebase the main feature branch.

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin.swift
HuiSF: Yea, was thinking Android and iOS returns different values implementing this API, could aggregate and remap the result on Dart side, could make the enum a super set of all the value that may return on Android and iOS, but still need to remap somehow.

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
HuiSF: Yea good to use `@unknown default` to avoid handling of unexpected case that falls into the `else`.

packages/notifications/push/amplify_push_notifications/ios/Classes/PushNotificationEventsStreamHandler.swift
```diff
@@ -0,0 +1,151 @@
...
+                case .error(let error):
+                    eventSink(error)
+                }
+            }
+        }
+    }
+
+    private func createEventMap(payload: [AnyHashable: Any]) -> [AnyHashable: Any] {
+        return [
+            "eventType": streamEventType.eventName,
```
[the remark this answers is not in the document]
HuiSF: It's redundant as the event is scoped by the event channels already, was going to remove but haven't done so somehow 😅

packages/notifications/push/amplify_push_notifications/ios/Classes/PushNotificationEventsStreamHandler.swift
```diff
@@ -0,0 +1,151 @@
...
+    static var binaryMessenger: FlutterBinaryMessenger?
+
+    let tokenReceived: PushNotificationEventsStreamHandler
+    let notificationOpened: PushNotificationEventsStreamHandler
+    let launchNotificationOpened: PushNotificationEventsStreamHandler
+    let foregroundMessageReceived: PushNotificationEventsStreamHandler
+    let backgroundMessageReceived:PushNotificationEventsStreamHandler
+
+    private init() {
+        guard let binaryMessenger = EventsStreamHandlers.binaryMessenger else {
```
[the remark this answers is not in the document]
HuiSF: Agreed, no shared instance needed here, will update

packages/notifications/push/amplify_push_notifications/ios/amplify_push_notifications.podspec
```diff
@@ -5,17 +5,19 @@
...
+  s.source           = { :git => 'https://github.com/aws-amplify/amplify-flutter.git' }
+  s.source_files     = 'Classes/**/*'
+  s.platform         = :ios, '13.0'
+
   s.dependency 'Flutter'
-  s.platform = :ios, '9.0'
+  s.dependency 'amplify_flutter_ios'
 
   # Flutter.framework does not contain a i386 slice.
   s.pod_target_xcconfig = { 'DEFINES_MODULE' => 'YES', 'EXCLUDED_ARCHS[sdk=iphonesimulator*]' => 'i386' }
```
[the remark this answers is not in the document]
HuiSF: Interesting, I haven't seen any issue around this while testing builds (both debug and release and signed bundles). 

`pigeon` v9 still says generating Swift is experimental, I'd avoid it and add the custom module map if necessary... 

Do you recall how to surface this linked issue?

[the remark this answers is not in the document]
HuiSF: Did more reading on the linked cocoapods issue, it seems the circular deps issue happens only when importing swift into objc, which is not applicable to this implementation (yet). I think we can add the custom module map when we need this?

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
```diff
@@ -0,0 +1,103 @@
...
+
+    public func application(
+        _ application: UIApplication,
+        didFinishLaunchingWithOptions launchOptions: [AnyHashable : Any] = [:]
+    ) -> Bool {
+        // when the App launch is caused by a notification
+        if let remoteNotification = launchOptions[UIApplication.LaunchOptionsKey.remoteNotification] as? [AnyHashable: Any] {
+            // This happens when:
+            //   1. remote notifications background mode is enabled on the host App AND
+            //   2. the arrived remote notification has content-available: 1 AND
```
cshfang, line 35: Nit for clarity
```suggestion
            //   2. the arrived remote notification has `content-available: 1` AND
```

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
```diff
@@ -0,0 +1,103 @@
...
+            else {
+                launchNotification = remoteNotification
+                sharedEventsStreamHandlers.launchNotificationOpened.sendEvent(payload: remoteNotification)
+            }
+        }
+
+        return true
+    }
+
+    // Method is being called differently at differnet states of App life cycle
```
cshfang, line 54:
```suggestion
    // Method is being called differently at different states of App life cycle
```

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
```diff
@@ -0,0 +1,103 @@
...
+    // https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623013-application#discussion
+    // Specifically:
+    // 1. .background case triggers when:
+    //   a. remote notifications background mode is enabled on the host App AND -
+    //   b. remote notifications has content-avilable: 1 AND -
```
cshfang, line 59:
```suggestion
    //   b. remote notifications has `content-available: 1` AND -
```

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
```diff
@@ -0,0 +1,103 @@
...
+    //   c. the host App is in the background
+    // 2. .inactive case triggers when:
+    //   a. end user tapped on a notification to open the host App regardless if the App was terminated
```
cshfang, line 62: Nit
```suggestion
    //   a. end user taps on a notification to open the host App regardless of whether the App was terminated
```

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
```diff
@@ -0,0 +1,103 @@
...
+
+            remoteNotificationCompletionHandlers[completionHandlerId] = completionHandler
+            userInfoCopy[completionHandlerIdKey] = completionHandlerId
+
+            sharedEventsStreamHandlers.backgroundMessageReceived.sendEvent(payload: userInfoCopy)
+        case .inactive:
+            // If the App was awaken but has not been launched, we record the tapped notification
+            // as the launch notification
+            if awakenInBackgroundAndHasNotLaunched {
+                awakenInBackgroundAndHasNotLaunched = false
```
cshfang, line 84: Might've missed this in JS review but would suggest simple past tense` or `awoke` since this is a boolean.

Would it be a more accurate descriptor of this state as well if we said awokeInBackground**But**HasNotLaunched?

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin.swift
```diff
@@ -1,19 +1,111 @@
...
+            sharedEventsStreamHandlers: EventsStreamHandlers(binaryMessenger: messenger)
+        )
+
+        NativePushNotificationsPluginSetup(messenger, pluginInstance)
+
+        registrar.addApplicationDelegate(pluginInstance)
+    }
+
+    public func getPermissionStatus(completion: @escaping (String?, FlutterError?) -> Void) {
+        UNUserNotificationCenter.current().getNotificationSettings { settings in
```
cshfang, line 36: We might want to update this to use the Swift utils

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -0,0 +1,248 @@
...
+  Future<bool> requestPermissions({
+    bool alert = true,
+    bool badge = true,
+    bool sound = true,
+  }) {
+    return _nativePlugin.requestPermissions(
+      PermissionsOptions(
+        alert: true,
+        sound: true,
+        badge: true,
```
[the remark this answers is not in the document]
HuiSF: oops

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -0,0 +1,248 @@
...
+  }
+
+  @override
+  PushNotificationMessage? get launchNotification => PushNotificationMessage();
+
+  Future<PushNotificationMessage?> _getLaunchNotification() async {
+    await _nativePlugin.getLaunchNotification();
+
+    return PushNotificationMessage();
+  }
```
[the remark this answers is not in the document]
HuiSF: Yes. I just moved the file into `lib/src` with necessary pigeon relevant changes, this PR is focusing on the iOS implementation, I believe there will be another PR to complete the Dart implementation.

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -0,0 +1,248 @@
...
+
+  Future<void> _registerDevice(String address) async {
+    try {
+      await _serviceProviderClient.registerDevice(address);
+      _logger.info('Successfully registered device with the servvice provider');
+      // return true;
+    } on Exception {
+      throw const PushNotificationException(
+        'Error when registering device with the service provider: ',
```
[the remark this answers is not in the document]
HuiSF: Same as above, the Dart implementation will be completed in a separate PR.

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin.swift
```diff
@@ -1,19 +1,111 @@
...
+
+    public func getBadgeCountWithError(_ error: AutoreleasingUnsafeMutablePointer<FlutterError?>) -> NSNumber? {
+        return UIApplication.shared.applicationIconBadgeNumber as NSNumber
+    }
+
+    public func setBadgeCountWithBadgeCount(_ withBadgeCount: NSNumber, error: AutoreleasingUnsafeMutablePointer<FlutterError?>) {
+        UIApplication.shared.applicationIconBadgeNumber = withBadgeCount.intValue
+    }
+
+    public func completeNotification(withCompletionHandlerId: String, error: AutoreleasingUnsafeMutablePointer<FlutterError?>) {
```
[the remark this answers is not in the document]
HuiSF: Was thinking about doing the way you described. I didn't do so because this setup is pretty much iOS specific - it runs the callback function after the App wakes up, where Android creates a "headless" Dart isolate to run the callback, so Android unnecessarily need this platform channel call into Flutter. Will update.

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin.swift
HuiSF: Now using `AmplifyUtilsNotifications` pod to handle `requestPermissions` and `getPermissionStatus`

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin.swift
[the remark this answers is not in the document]
HuiSF: Updated 👍

packages/notifications/push/amplify_push_notifications/pubspec.yaml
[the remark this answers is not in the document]
HuiSF: Right now in the base feature branch, aft.yaml doesn't specify pigeon package version. I updated to use `^7.1.5` following the `next` branch.

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin.swift
HuiSF: Updated 👍

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -205,44 +170,89 @@ class AmplifyPushNotifications extends PushNotificationsPluginInterface {
...
+        },
+      ),
+    );
+  }
+
+  @override
+  void onLaunchNotificationOpened(Map<Object?, Object?> payload) {
+    for (final callback in _onLaunchNotificationOpenedCallbacks) {
+      callback(PushNotificationMessage.fromJson(payload));
+    }
```
cshfang, line 256: Since it's intended to be used internally, there should be just one callback and then in theory you can just drop the reference to it too after calling it

HuiSF: I'm thinking that even it's internal, we may need to add more than 1 callback to handle different use cases, e.g.
1. register a callback to set the launch notification for the `get launchNotification` getter
2. and one to execute the logic to record analytics event.

I've pushed an update to clear out the registered callbacks after execute them once (as onLaunchNotification should happen only once in one App life cycle)

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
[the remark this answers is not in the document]
HuiSF: Pushed new commits to update some of the place holder and todos, to make the basic functionality of APIs, event streams working in dev example App using `AmplifyPushNotificationsPinpoint()` directly without any analytics integration.

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -0,0 +1,248 @@
...
+
+    // Block configure if registering device is not complete
+    // TODO: Catch and throw exceptions originating from token or device registration
+    final deviceToken = await onTokenReceived.first;
+    await _registerDevice(deviceToken);
+
+    // Initialize listeners
+    onTokenReceived.listen(_tokenReceivedListener);
+    onNotificationReceivedInForeground.listen(_foregroundNotificationListener);
+    onNotificationOpened.listen(_notificationOpenedListener);
```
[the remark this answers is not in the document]
HuiSF: Added logging on the errors.

In the native we only actively emit `FlutterError` to the onTokenReceived channel, which should happen on the very first event (registering device for PN) when the App is incorrectly configured with PN capability (certain behavior on iOS, need to verify on Android when the implementation gets ready). We are throwing this error within `Amplify.configure()`.

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -0,0 +1,290 @@
...
+  final PushNotificationsHostApi _hostApi;
+  final _PushNotificationsFlutterApi _flutterApi;
+
+  late final Stream<String> _onTokenReceived;
+  late final Stream<PushNotificationMessage> _onForegroundNotificationReceived;
+  late final Stream<PushNotificationMessage> _onNotificationOpened;
+
+  var _isConfigured = false;
+  PushNotificationMessage? _launchNotification;
+  PushNotificationMessage? _launchNotificationForAnalytics;
```
HuiSF, line 71: This copy will be used to send an analytic event record after `Amplify.configure` is completed, used internally.

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin.swift
```diff
@@ -1,19 +1,92 @@
...
+        )
+
+        PushNotificationsHostApiSetup(messenger, pluginInstance)
+
+        registrar.addApplicationDelegate(pluginInstance)
+    }
+
+    public func getPermissionStatus(completion: @escaping (GetPermissionStatusResult?, FlutterError?) -> Void) {
+        Task {
+            switch await AUNotificationPermissions.status {
```
dnys1, line 38: If I call `getPermissionStatus` before calling `requestPermissionStatus` it should return `.notRequested`, no?

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin.swift
```diff
@@ -1,19 +1,92 @@
...
+            }
+
+            do {
+                let granted = try await AUNotificationPermissions.request(options)
+                completion(granted as NSNumber, nil)
+            } catch {
+                completion(
+                    nil,
+                    FlutterError(
+                        code: "RequsetPermissionsError",
```
dnys1, line 76:
```suggestion
                        code: "RequestPermissionsError",
```

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
dnys1: If a launch notification is defined as a notification which opens the app from killed, then I'm not I understand why the platform calls into the Dart code instead of, for example:

```dart
Future<void> configure(...) async {
  // ...
  _launchNotification = await _hostApi.getLaunchNotification(); 
  for (final callback in callbacks) {
    // ...
  }
}
```

Having this separate from `configure` doesn't seem to make sense.

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
dnys1: This should be retrieved once from the platform - if we need to use it for analytics purposes it should be used then and there. It does not need to be stored on the class.

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
HuiSF: On iOS when a remote notification wakes the App in the background, it's the entire App wakes up, including developers Flutter App. `Amplify.configure()` would be executed at this point, `_launchNotification = await _hostApi.getLaunchNotification();` won't get the correct launch notification (the App is still running in the background, and there is no notification being tapped)

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin.swift
```diff
@@ -1,19 +1,100 @@
...
+        let launchNotification = launchNotification
+        self.launchNotification = nil
+        return launchNotification
+    }
+
+    public func getBadgeCountWithError(_ error: AutoreleasingUnsafeMutablePointer<FlutterError?>) -> NSNumber? {
+        return UIApplication.shared.applicationIconBadgeNumber as NSNumber
+    }
+
+    public func setBadgeCountWithBadgeCount(_ withBadgeCount: NSNumber, error: AutoreleasingUnsafeMutablePointer<FlutterError?>) {
```
Samaritan1011001, line 97: Should be just `setBadgeCount` ?

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin.swift
```diff
@@ -1,19 +1,100 @@
...
+                        code: "RequestPermissionsError",
+                        message: "Error occurred requesting notitication center authorization.",
+                        details: error.localizedDescription
+                    )
+                )
+            }
+        }
+    }
+
+    public func getLaunchNotificationWithError(_ error: AutoreleasingUnsafeMutablePointer<FlutterError?>) -> [AnyHashable : Any]? {
```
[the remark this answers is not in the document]
HuiSF: This is how pigeon generates the function names, if tried my best adding `with` with function parameters which generates cleaner function names, but not sure what to do about this. However, these are internal APIs so the names don't matter much.

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin.swift
HuiSF: Ditto

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
```diff
@@ -0,0 +1,120 @@
...
+                completionHandler(.noData)
+            }
+        case .inactive:
+
+            if let launchNotification = launchNotification {
+                if !NSDictionary(dictionary: launchNotification).isEqual(to: userInfo) {
+                    // When a launch notification is recorded in didFinishLaunchingWithOptions above,
+                    // but the last tapped notification is not the recorded launch notification, the last
+                    // tapped notification will be sent to Flutter via notificationOpened.
+                    // This may happen when an end user rapidly tapped on multiple notifications.
```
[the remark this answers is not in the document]
HuiSF: Yes it would an edge case.

Samaritan1011001 (approved):

packages/amplify_core/lib/src/types/notifications/push/push_notification_permission_status.dart
```diff
@@ -5,8 +5,8 @@
 /// Enum to indicate the state of the current push notificaiton permissions
 /// {@endtemplate}
 enum PushNotificationPermissionRequestStatus {
```
dnys1, line 7:
```suggestion
enum PushNotificationPermissionStatus {
```

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin.swift
```diff
@@ -1,19 +1,100 @@
...
+        PushNotificationsHostApiSetup(messenger, pluginInstance)
+
+        registrar.addApplicationDelegate(pluginInstance)
+    }
+
+    public func getPermissionStatus(completion: @escaping (GetPermissionStatusResult?, FlutterError?) -> Void) {
+        Task {
+            switch await AUNotificationPermissions.status {
+            case .notDetermined:
+                completion(.make(with: .shouldRequestWithRationale), nil)
```
dnys1, line 42: I think pigeon needs to be re-run

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin.swift
```diff
@@ -1,19 +1,100 @@
...
-    switch call.method {
-    case "getPlatformVersion":
-      result("iOS " + UIDevice.current.systemVersion)
-    default:
-      result(FlutterMethodNotImplemented)
+import amplify_flutter_ios
+import AmplifyUtilsNotifications
+
+public class AmplifyPushNotificationsPlugin: NSObject, FlutterPlugin, PushNotificationsHostApi {
+    var remoteNotificationCompletionHandlers: [String: (UIBackgroundFetchResult) -> Void] = [:]
```
dnys1, line 9: Still needed?

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
```diff
@@ -0,0 +1,120 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import Foundation
+import Flutter
+
+private let completionHandlerIdKey = "completionHandlerId"
```
dnys1, line 7: Still needed?

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
```diff
@@ -0,0 +1,120 @@
...
+        case .background:
+            flutterApi.onNotificationReceivedInBackground(withPayload: userInfo) { _ in
+                // Wait for Flutter side to complete calling registered callbacks
+                // then notifiy the system that the background task is completed
+                completionHandler(.noData)
+            }
+        case .inactive:
+
+            if let launchNotification = launchNotification {
+                if !NSDictionary(dictionary: launchNotification).isEqual(to: userInfo) {
```
dnys1, line 91: I think this just compares references, though, right? Does this work?

packages/notifications/push/amplify_push_notifications/ios/Classes/PushNotificationEventsStreamHandler.swift
```diff
@@ -0,0 +1,113 @@
...
+}
+
+class PushNotificationEventsStreamHandler: NSObject, FlutterStreamHandler {
+    private let eventChannel: FlutterEventChannel
+    private var eventQueue: [AmplifyPushNotificationsEvent] = []
+    private var eventSink: FlutterEventSink?
+
+    init(eventType: NativeEvent, binaryMessenger: FlutterBinaryMessenger) {
+        eventChannel = FlutterEventChannel(name: eventType.eventChannelName, binaryMessenger: binaryMessenger)
+        super.init()
```
dnys1, line 42: Calling `super` in the middle seems weird

```suggestion
        super.init()
        eventChannel = FlutterEventChannel(name: eventType.eventChannelName, binaryMessenger: binaryMessenger)
```

packages/notifications/push/amplify_push_notifications/ios/Classes/PushNotificationEventsStreamHandler.swift
```diff
@@ -0,0 +1,113 @@
...
+        } else {
+            eventQueue.append(
+                AmplifyPushNotificationsEvent.error(error: error)
+            )
+        }
+    }
+
+    private func flushEvents() {
+        if let eventSink = eventSink {
+            while(!eventQueue.isEmpty) {
```
dnys1, line 85: Format

packages/notifications/push/amplify_push_notifications/ios/amplify_push_notifications.podspec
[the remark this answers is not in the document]
dnys1: Can you confirm this is still true given that we've added the `@FlutterApi`?

dnys1 (approved):

packages/notifications/push/amplify_push_notifications/ios/Classes/AmplifyPushNotificationsPlugin+AppDelegate.swift
HuiSF: `isEqual` does a deep comparison

>Two dictionaries have equal contents if they each hold the same number of entries and, for a given key, the corresponding value objects in each dictionary satisfy the [isEqual(_:)](https://developer.apple.com/documentation/objectivec/nsobjectprotocol/1418795-isequal) test.

packages/notifications/push/amplify_push_notifications/ios/amplify_push_notifications.podspec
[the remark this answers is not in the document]
HuiSF: Yup, still haven't seen any issues building with both debug and release models and signed bundles...

packages/notifications/push/amplify_push_notifications/ios/Classes/PushNotificationEventsStreamHandler.swift
HuiSF: This is on purpose, and the compiler enforces this -
1. `eventChannel` needs to be initialized before `super.init()` as it's not a property that exists in the `super`
2. `eventChannel.setStreamHandler(self)` the `self` can only be called after `super.init()`

PR #2709, opened 2023-02-28
feat(push): Permissions, background & notification opened APIs on Android 

Samaritan1011001:
*Issue #, if available:*

*Description of changes:*
Adds implementation for `getPermissionStatus`, `requestPermissions`, `onNotificationOpened` & `launchNotification` APIs. 


packages/notifications/push/amplify_push_notifications/lib/callback_dispatcher.dart
```diff
@@ -0,0 +1,50 @@
...
+        CallbackHandle.fromRawHandle(element['handle'] as int),
+      );
+      assert(callback != null, 'Callback not found');
+
+      // ignore: avoid_dynamic_calls
+      await callback!(
+        // ignore: avoid_dynamic_calls
+        PushNotificationMessage.fromJson(element['notification'] as Map),
+      );
+    }
```
dnys1, line 43:
```suggestion
    for (final element in args.cast<Map>()) {
      final callback = PluginUtilities.getCallbackFromHandle(
        CallbackHandle.fromRawHandle(element['handle'] as int),
      );
      if (callback == null) {
        throw StateError('Callback not found');
      }
      if (callback is! OnRemoteMessageCallback) {
        throw StateError('Invalid callback type: ${callback.runtimeType}');
      }

      await callback(
        PushNotificationMessage.fromJson(element['notification'] as Map),
      );
    }
```

packages/notifications/push/amplify_push_notifications/lib/amplify_push_notifications.dart
```diff
@@ -110,51 +117,81 @@ class AmplifyPushNotifications extends PushNotificationsPluginInterface {
...
-    await _serviceProviderClient.init(
-      config: config,
-      authProviderRepo: authProviderRepo,
-    );
-
-    // Block configure if registering device is not complete
-    // TODO: Catch and throw exceptions originating from token or device registration
-    final deviceToken = await onTokenReceived.first;
-    await _registerDevice(deviceToken);
+    await _initiliazeServiceClientAndRegisterDevice(config, authProviderRepo);
```
dnys1, line 138: sp

packages/notifications/push/amplify_push_notifications/lib/amplify_push_notifications.dart
```diff
@@ -110,51 +117,81 @@ class AmplifyPushNotifications extends PushNotificationsPluginInterface {
...
+  @override
+  Future<void> reset() async {
+    if (!_isConfigured) {
+      return;
+    }
+    _isConfigured = false;
+  }
+
+  Future<void> _initiliazeServiceClientAndRegisterDevice(
+    AmplifyConfig? config,
```
dnys1, line 177: `config` is never null here

```suggestion
    AmplifyConfig config,
```

packages/notifications/push/amplify_push_notifications/lib/amplify_push_notifications.dart
```diff
@@ -110,51 +117,81 @@ class AmplifyPushNotifications extends PushNotificationsPluginInterface {
...
+      await _serviceProviderClient.init(
+        config: config,
+        authProviderRepo: authProviderRepo,
+      );
+
+      final deviceToken = await onTokenReceived.first;
+      await _registerDevice(deviceToken);
+    } on Exception catch (e) {
+      throw PushNotificationException(
+        e.toString(),
```
dnys1, line 190: Use underlying exception instead

packages/notifications/push/amplify_push_notifications/lib/amplify_push_notifications.dart
```diff
@@ -110,51 +117,81 @@ class AmplifyPushNotifications extends PushNotificationsPluginInterface {
...
 
     _isConfigured = true;
   }
 
+  Future<void> _registerCallbackDispatcher() async {
+    try {
+      final callback = PluginUtilities.getCallbackHandle(callbackDispatcher);
+      await _methodChannel.invokeMethod<void>(
+        'initializeService',
+        <dynamic>[callback?.toRawHandle()],
```
dnys1, line 158: Callback will never be `null`. The only time it is is when you pass a non-static/top-level method.

```suggestion
        <dynamic>[callback!.toRawHandle()],
```

packages/notifications/push/amplify_push_notifications/lib/amplify_push_notifications.dart
```diff
@@ -110,51 +117,81 @@ class AmplifyPushNotifications extends PushNotificationsPluginInterface {
...
+    await _initiliazeServiceClientAndRegisterDevice(config, authProviderRepo);
 
     // Initialize listeners
     onTokenReceived.listen(_tokenReceivedListener);
     onNotificationReceivedInForeground.listen(_foregroundNotificationListener);
     onNotificationOpened.listen(_notificationOpenedListener);
-
-    // TODO: Register the callback dispatcher
-
+    await _registerBgCallback(
```
dnys1, line 144: I don't follow the need for all these callbacks - why not just one?

packages/notifications/push/amplify_push_notifications/lib/amplify_push_notifications.dart
Samaritan1011001: What do you mean? 😅

packages/notifications/push/amplify_push_notifications/lib/amplify_push_notifications.dart
Samaritan1011001: So we can record Analytics but that can be done in the callback dispatcher too before calling the external callback function. Do we want to change to that?

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
```diff
@@ -3,110 +3,203 @@
...
+        val BG_EXTERNAL_CALLBACK_HANDLE_KEY = "bg_user_callback_handle"
+        val BG_INTERNAL_CALLBACK_HANDLE_KEY = "bg_internal_callback_handle"
+
         private val TAG = "AmplifyPushNotificationsPlugin"
     }
 
     override fun onAttachedToActivity(binding: ActivityPluginBinding) {
         activityBinding = binding
+        this.mainActivity = binding.activity
         onNewIntent(binding.activity.intent)
```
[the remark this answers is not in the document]
Samaritan1011001: It is necessary for the onNewIntent to be called when the activity is being created (from killed state) [Auth](https://github.com/aws-amplify/amplify-flutter/blob/next/packages/auth/amplify_auth_cognito_android/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt#L140) does the same

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
```diff
@@ -3,110 +3,203 @@
...
+            Handler(context.mainLooper).post {
+                if (!task.isSuccessful) {
+                    task.exception?.let { StreamHandlers.tokenReceived.sendError(it) }
+                    return@post
+                }
+                StreamHandlers.tokenReceived.send(
+                    mapOf(
+                        "token" to task.result
+                    )
+                )
```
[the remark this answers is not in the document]
Samaritan1011001: send is a generic method that needs to accept even notificationMap

packages/notifications/push/amplify_push_notifications/lib/amplify_push_notifications.dart
dnys1: Sorry, means "spelling"

packages/notifications/push/amplify_push_notifications/lib/amplify_push_notifications.dart
dnys1: I think that would make more sense, yeah

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
```diff
@@ -3,110 +3,250 @@
...
+
+    override fun onDetachedFromEngine(@NonNull binding: FlutterPlugin.FlutterPluginBinding) {
+        channel.setMethodCallHandler(null)
     }
 
     override fun onAttachedToActivity(binding: ActivityPluginBinding) {
+        mainActivity = binding.activity
         activityBinding = binding
-        onNewIntent(binding.activity.intent)
+        mainBinaryMessenger?.let {
```
dnys1, line 91: Since the binary messenger is only available in `onAttachedToEngine`, these should probably only be initialized there

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
```diff
@@ -3,110 +3,250 @@
...
+            StreamHandlers.initStreamHandlers()
+            StreamHandlers.initEventChannels(it)
+        }
         binding.addOnNewIntentListener(this)
-
-        // TODO: also fetchToken on app resume
+        onNewIntent(binding.activity.intent)
```
dnys1, line 96: Shouldn't be needed

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
dnys1: Same for deinit - move to `onDetachedFromEngine`

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
```diff
@@ -0,0 +1,155 @@
...
+
+    override fun onHandleWork(intent: Intent) {
+        Log.i(TAG, "Handling work @ PushNotificationBackgroundService...")
+        if (!intent.isPushNotificationIntent) {
+            return
+        }
+        val remoteMessage = RemoteMessage(intent.extras)
+        val notificationMap = remoteMessage.asPayload().asChannelMap()
+
+        val bgExternalCallbackHandle = baseContext.getSharedPreferences(
```
dnys1, line 124: Let's get better names for these - and reduce to a single callback handler if possible

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
Samaritan1011001: Yes, that was the initial approach but when we create the Background Isolate upon receiving a notification in the background/killed state, `onAttachedToEngine` gets called again with the Background Isolate's binaryMessenger. We don't want to initialize the stream handlers with that messenger but rather keep it attached to the main Flutter Engine that displays the app. Which is why I store the messenger and only initialize when the plugin is attached to the activity.

packages/notifications/push/amplify_push_notifications/android/src/main/AndroidManifest.xml
```diff
@@ -8,14 +8,9 @@
...
-        <receiver
-            android:name=".PushNotificationReceiver"
-            android:exported="true"
-            android:permission="com.google.android.c2dm.permission.SEND">
-
-            <intent-filter>
-                <action android:name="com.google.android.c2dm.intent.RECEIVE" />
-            </intent-filter>
-        </receiver>
+        <service
```
cshfang, line 11: Is this permission not needed?
```xml
<uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
```

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
```diff
@@ -3,110 +3,250 @@
...
+open class AmplifyPushNotificationsPlugin : FlutterPlugin, MethodCallHandler, ActivityAware,
     PluginRegistry.NewIntentListener {
+    private companion object {
+        const val TAG = "AmplifyPushNotificationsPlugin"
+
+        /**
+         * The scope in which to spawn tasks which should not be awaited from the main thread.
+         */
+        val scope =
+            CoroutineScope(Dispatchers.IO) + CoroutineName("amplify_flutter.PushNotifications")
```
cshfang, line 41: Why does this need to be in the companion object as opposed to just a member var in the class?

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
```diff
@@ -3,110 +3,250 @@
...
+
+
+private const val TAG = "AmplifyPushNotificationsPlugin"
 
 /** AmplifyPushNotificationsPlugin */
-class AmplifyPushNotificationsPlugin : FlutterPlugin, MethodCallHandler, ActivityAware,
+open class AmplifyPushNotificationsPlugin : FlutterPlugin, MethodCallHandler, ActivityAware,
     PluginRegistry.NewIntentListener {
+    private companion object {
+        const val TAG = "AmplifyPushNotificationsPlugin"
```
cshfang, line 35: I don't think you need this since you already have the TAG declared on this file and you don't access it statically anywhere else (I think?)

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
```diff
@@ -3,110 +3,250 @@
...
+     * The plugin binding for [mainActivity], used to manage lifecycle callback registration.
+     */
     private var activityBinding: ActivityPluginBinding? = null
-    companion object {
-        private val TAG = "AmplifyPushNotificationsPlugin"
+
+    /**
+     * The application context.
+     */
+    private var applicationContext: Context? = null
```
cshfang, line 57: If this were `lateinit` would you still need the `!!` assertions?

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
```diff
@@ -3,110 +3,250 @@
...
+
+    /**
+     * Shared Preference used to persist callback handles.
+     */
+    private lateinit var sharedPreferences: SharedPreferences
+    private lateinit var channel: MethodChannel
+
+    /**
+     * Main engine's binary messenger holder.
```
cshfang, line 66: Nit:
```suggestion
     * Main engine's binary messenger.
```

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationConstants.kt
```diff
@@ -0,0 +1,19 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.amazonaws.amplify.amplify_push_notifications
+
+class PushNotificationConstants {
```
cshfang, line 6:
```suggestion
class PushNotificationPluginConstants {
```
It's too confusing that the constants from the android utils is almost identically named

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
```diff
@@ -0,0 +1,155 @@
...
+        private const val TAG = "PushBackgroundService"
+
+        @JvmStatic
+        private val JOB_ID = UUID.randomUUID().mostSignificantBits.toInt()
+
+        @JvmStatic
+        private var backgroundFlutterEngine: FlutterEngine? = null
+
+        @JvmStatic
+        private val serviceStarted = AtomicBoolean(false)
```
cshfang, line 46: Do these need to be in the companion object?

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
```diff
@@ -0,0 +1,155 @@
...
+        backgroundChannel = MethodChannel(
+            backgroundFlutterEngine!!.dartExecutor.binaryMessenger,
+            "plugins.flutter.io/amplify_push_notification_plugin_background"
+        )
+        backgroundChannel.setMethodCallHandler(this)
+    }
+
+    override fun onMethodCall(call: MethodCall, result: MethodChannel.Result) {
+        when (call.method) {
+            "PushNotificationBackgroundService.initialized" -> {
```
cshfang, line 100: I think this string should be treated as a method name. So exactly what is this method doing? Does this dot notation do anything special? Maybe this should just be `startBackgroundService` or something?

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
```diff
@@ -0,0 +1,155 @@
...
+        synchronized(serviceStarted) {
+            if (backgroundFlutterEngine == null) {
+
+                val callbackHandle = context.getSharedPreferences(
+                    PushNotificationConstants.SHARED_PREFERENCES_KEY, Context.MODE_PRIVATE
+                ).getLong(
+                    PushNotificationConstants.CALLBACK_DISPATCHER_HANDLE_KEY, 0
+                )
+                if (callbackHandle == 0L) {
+                    Log.w(TAG, "Warning: no callback registered")
```
cshfang, line 69: When and why would this happen?

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
```diff
@@ -0,0 +1,155 @@
...
+                    context.assets, FlutterMain.findAppBundlePath(), callbackInfo
+                )
+
+                // DartCallback must only be executed after the engine has been created
+                backgroundFlutterEngine!!.dartExecutor.executeDartCallback(args)
+            }
+        }
+        backgroundChannel = MethodChannel(
+            backgroundFlutterEngine!!.dartExecutor.binaryMessenger,
+            "plugins.flutter.io/amplify_push_notification_plugin_background"
```
cshfang, line 93: Hmm isn't there already an established convention for this within Amplify Flutter?

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
```diff
@@ -0,0 +1,155 @@
...
+                    serviceStarted.set(true)
+                }
+            }
+            else -> result.notImplemented()
+        }
+        result.success(null)
+    }
+
+    override fun onHandleWork(intent: Intent) {
+        Log.i(TAG, "Handling work @ PushNotificationBackgroundService...")
```
cshfang, line 117: Should clean up all the logs so they're not using trailing ellipses or @ symbol

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationUtils.kt
```diff
@@ -63,24 +60,35 @@ fun NotificationPayload.asChannelMap(): Map<String, Any?> {
...
+            // TODO(Samaritan1011001): add channelId if needed
         ),
         "data" to this.rawData
     )
 }
 
-fun Bundle.asPayloadFromExtras(): NotificationPayload? {
-    return this.getBundle(PAYLOAD_KEY)?.toNotificationsPayload()
+val Intent.isPushNotificationIntent: Boolean
+    get() = extras?.containsKey("google.message_id") ?: false
```
cshfang, line 71: I think this is misleading since it wouldn't handle the intents from our created notifications correctly. So universally extending Intent like this is probably going to cause issues for future contributors to this project

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
Samaritan1011001: There is only one callback function that the customer provides to the background API. 
The handle to that is in `bgExternalCallbackHandle` in the above code, this is sent to the [Callback Dispatcher](https://github.com/aws-amplify/amplify-flutter/blob/base-project-setup-bg/packages/notifications/push/amplify_push_notifications/lib/callback_dispatcher.dart) maintained by the plugin which then calls the customer's callback function.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
Samaritan1011001: Following the same convention as [AuthCognito](https://github.com/aws-amplify/amplify-flutter/blob/next/packages/auth/amplify_auth_cognito_android/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt#L140)

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
Samaritan1011001: right, I missed to remove the one defined for the file. It's in the object because I was following the convention from [AuthCognito](https://github.com/aws-amplify/amplify-flutter/blob/next/packages/auth/amplify_auth_cognito_android/android/src/main/kotlin/com/amazonaws/amplify/amplify_auth_cognito/AuthCognito.kt#L140) but you are right on not wanting to use it statically.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
Samaritan1011001: That's true. I initially had it that way I believe, it was changed to match AuthCognito like I mentioned above. 
Should I make them `lateinit` and non-nullable?

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
```diff
@@ -0,0 +1,155 @@
...
+     * The Background Channel that is used to send messages to the callback dispatcher
+     */
+    private lateinit var backgroundChannel: MethodChannel
+
+    companion object {
+
+        private const val TAG = "PushBackgroundService"
+
+        @JvmStatic
+        private val JOB_ID = UUID.randomUUID().mostSignificantBits.toInt()
```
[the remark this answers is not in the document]
Samaritan1011001: JOB_ID is needed by the enqueueWork but moved the others out of the companion object

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
Samaritan1011001: dot doesn't do anything special. I renamed it to `callbackDispatcherInitialized` since that is what it signifies. It's called by the [Callback Dispatcher](https://github.com/aws-amplify/amplify-flutter/blob/base-project-setup-bg/packages/notifications/push/amplify_push_notifications/lib/callback_dispatcher.dart#L50) after initializing the background channel.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
Samaritan1011001: If the Callback Dispatcher's handle was not persisted in Shared Preference. Updated the warning message.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
Samaritan1011001: Moved it to the constants file and referenced that.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationUtils.kt
Samaritan1011001: Yeah, that check is not actually needed. So removing it since it is only called from `onMessageReceived` .

packages/notifications/push/amplify_push_notifications/lib/amplify_push_notifications.dart
```diff
@@ -110,97 +116,129 @@ class AmplifyPushNotifications extends PushNotificationsPluginInterface {
...
+      );
+    }
+  }
+
   void _foregroundNotificationListener(
     PushNotificationMessage pushNotificationMessage,
   ) {
     _logger.info(
-      'Successfully listrening to foreground events: $pushNotificationMessage',
+      'Successfully listrening to FOREGROUND events in the plugin',
```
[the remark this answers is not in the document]
Samaritan1011001: Actually, these are not needed to be logged so removing it.

packages/notifications/push/amplify_push_notifications/lib/amplify_push_notifications.dart
```diff
@@ -110,97 +116,129 @@ class AmplifyPushNotifications extends PushNotificationsPluginInterface {
...
-
   void _notificationOpenedListener(
     PushNotificationMessage pushNotificationMessage,
   ) {
     _logger.info(
-      'Successfully listrening to notification opened events: $pushNotificationMessage',
+      'Successfully listrening to NOTIFICATIONOPENED events: $pushNotificationMessage',
     );
     // TODO(Samaritan1011001): Record Analytics
+    _launchNotification = pushNotificationMessage;
```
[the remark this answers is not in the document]
Samaritan1011001: Yes, that's the idea. Is that not desirable?

packages/notifications/push/amplify_push_notifications/lib/amplify_push_notifications.dart
```diff
@@ -110,97 +116,129 @@ class AmplifyPushNotifications extends PushNotificationsPluginInterface {
...
-  ) {
-    // TODO(Samaritan1011001): Record Analytics
-  }
-
   void _notificationOpenedListener(
     PushNotificationMessage pushNotificationMessage,
   ) {
     _logger.info(
-      'Successfully listrening to notification opened events: $pushNotificationMessage',
+      'Successfully listrening to NOTIFICATIONOPENED events: $pushNotificationMessage',
```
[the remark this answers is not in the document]
Samaritan1011001: Log not needed, removed

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
Samaritan1011001: Both the [situations](https://api.flutter.dev/javadoc/io/flutter/embedding/engine/plugins/activity/ActivityAware.html#:~:text=1%20of%202-,situations,-%3A) mentioned in the API doc for `onAttachedToActivity` indicates that this order of connection `Plugin -> FlutterEngine -> Activity` 
So, no it can't be called before a FlutterEngine is present.

packages/notifications/push/amplify_push_notifications/android/src/main/AndroidManifest.xml
Samaritan1011001: It is according to the docs. Added it.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationUtils.kt
```diff
@@ -63,24 +60,35 @@ fun NotificationPayload.asChannelMap(): Map<String, Any?> {
...
+    return null
 }
 
-fun RemoteMessage.isRemoteMessageSupported(): Boolean {
-    return !this.data[PushNotificationsConstants.PINPOINT_CAMPAIGN_CAMPAIGN_ID].isNullOrEmpty() or
-            !this.data[PushNotificationsConstants.JOURNEY_ID].isNullOrEmpty()
-}
\ No newline at end of file
+val RemoteMessage.isSupported: Boolean
+    get() = !this.data[PushNotificationsConstants.PINPOINT_CAMPAIGN_CAMPAIGN_ID].isNullOrEmpty() or !this.data[PushNotificationsConstants.JOURNEY_ID].isNullOrEmpty()
```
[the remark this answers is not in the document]
Samaritan1011001: The `PINPOINT_PREFIX` is in the data payload I believe. Do you mean to say we should check that? 

Updated it to `get() = data.any { it.key.contains(PushNotificationsConstants.PINPOINT_PREFIX) }`

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationUtils.kt
```diff
@@ -3,34 +3,32 @@
...
-        ?: data[PushNotificationsConstants.PINPOINT_NOTIFICATION_BODY]
-    val title = this.notification?.title
-        ?: data[PushNotificationsConstants.TITLE_ATTRIBUTE_KEY]
-        ?: data[PushNotificationsConstants.PINPOINT_NOTIFICATION_TITLE]
-    val imageUrl = this.notification?.imageUrl?.toString()
-        ?: data[PushNotificationsConstants.IMAGEURL_ATTRIBUTE_KEY]
-        ?: data[PushNotificationsConstants.PINPOINT_NOTIFICATION_IMAGEURL]
-
-
+    val body = this.notification?.body ?: data[PushNotificationsConstants.MESSAGE_ATTRIBUTE_KEY]
```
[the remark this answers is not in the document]
Samaritan1011001: Yeah, it's `processRemoteMessage` so changed to that.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationFirebaseMessagingService.kt
```diff
@@ -3,16 +3,92 @@
...
+        }
+        // Otherwise, try to handle the message
+        onMessageReceived(remoteMessage)
+    }
+
+    /**
+     * Method to handle and forward messages received in foreground & background using isolates and event channels
+     */
+    override fun onMessageReceived(remoteMessage: RemoteMessage) {
+        baseContext?.let {
```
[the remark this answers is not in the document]
Samaritan1011001: wouldn't be, removing it.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
cshfang: I would do what makes the most sense. Just because we've done it one way before doesn't always mean it was the right way. If it's possible, we can also ask the original author if we're missing something by not adopting the same convention/pattern - it's always a possibility that AuthCognito should be updated instead.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
cshfang: I would do what makes the most sense. Just because we've done it one way before doesn't always mean it was the right way. If it's possible, we can also ask the original author if we're missing something by not adopting the same convention/pattern - it's always a possibility that AuthCognito should be updated instead.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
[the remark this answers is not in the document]
cshfang: Can't it be moved to the file? Outside of the class altogether I mean?

packages/notifications/push/amplify_push_notifications/lib/amplify_push_notifications.dart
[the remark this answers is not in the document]
cshfang: Wouldn't every notification that was opened become a launch notification in that case?

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationUtils.kt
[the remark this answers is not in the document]
Samaritan1011001: Synced offline and changed this to the suggested code

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
```diff
@@ -3,110 +3,263 @@
...
-    // TODO: update this function to be more robust
-    override fun onNewIntent(intent: Intent): Boolean {
-        Log.d(TAG, "onNewIntent in push plugin $intent")
+    private fun registerCallbackToCache(
+        context: Context,
+        callbackHandle: Long,
+        callbackKey: String,
+    ) {
+            Log.i(TAG, "Registering callback function with key $callbackKey")
+            context.getSharedPreferences(
```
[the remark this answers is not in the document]
Samaritan1011001: Missed to use that reference after I created it. changed.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationEventsStreamHandler.kt
```diff
@@ -63,54 +64,97 @@ class PushNotificationEventsStreamHandler constructor(
...
             }
+        } catch (e: Exception) {
+            Log.e(TAG, "error when flushing event queue: $e")
         }
     }
 }
 
 class StreamHandlers {
     companion object {
+        @JvmStatic
```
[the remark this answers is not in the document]
Samaritan1011001: I thought that was convention at first but missed to remove it after I learnt it's use. Removed.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
```diff
@@ -3,110 +3,263 @@
...
+        flutterApi = null
     }
 
     override fun onAttachedToActivity(binding: ActivityPluginBinding) {
+        mainActivity = binding.activity
         activityBinding = binding
-        onNewIntent(binding.activity.intent)
+        mainBinaryMessenger?.let {
+            StreamHandlers.initStreamHandlers()
+            StreamHandlers.initEventChannels(it)
```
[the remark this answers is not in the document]
Samaritan1011001: Added a `deInit` method and nulled all event channels and stream handlers. 
Called the `StreamHandlers.deInit` when detaching from activity.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
```diff
@@ -3,110 +3,263 @@
...
+        flutterApi = PushNotificationsFlutterApi(mainBinaryMessenger)
+        applicationContext = flutterPluginBinding.applicationContext
+        sharedPreferences =
+            applicationContext.getSharedPreferences(
+                PushNotificationPluginConstants.SHARED_PREFERENCES_KEY,
+                Context.MODE_PRIVATE
+            )
+    }
+
+    override fun onDetachedFromEngine(@NonNull binding: FlutterPlugin.FlutterPluginBinding) {
```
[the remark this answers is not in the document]
Samaritan1011001: Removed all such instances.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
```diff
@@ -3,110 +3,263 @@
...
-        StreamHandlers.initialize(flutterPluginBinding.binaryMessenger)
+    // TODO(Samaritan1011001): 1. This gets called with intent only when a specific notification is tapped and not when the group is tapped
+    //      2. The intent here is the last notification device got rather than the one that was tapped
+    override fun onNewIntent(intent: Intent): Boolean {
+        intent.extras?.let {
+            val payload = it.asPayload()
+            if (payload != null) {
+                val notificationHashMap = payload.asChannelMap()
+                flutterApi?.onLaunchNotificationOpened(
+                    notificationHashMap as Map<Any, Any?>
```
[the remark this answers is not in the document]
Samaritan1011001: Added the cast since this was the exception case of sending the notification via `flutterApi` and conforming to it's required type as opposed to `eventChannel` but I guess it does make sense to change type for `asChannelMap`. So made that change.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
```diff
@@ -3,110 +3,263 @@
...
-//            PushNotificationEventsStreamHandler.sendEvent(
-//                PushNotificationsEvent(
-//                    NativeEvent.NOTIFICATION_OPENED,
-//                    notificationHashMap
-//                )
-//            )
+    private fun refreshToken() {
+        FirebaseMessaging.getInstance().token.addOnCompleteListener { task ->
+            if (!task.isSuccessful) {
+                task.exception?.let { StreamHandlers.tokenReceived.sendError(it) }
```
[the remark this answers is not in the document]
Samaritan1011001: It cannot be null at this particular call site since we are checking for `!task.isSuccessful`. `task.exception` is null only when the task was not completed fully or has completed successfully according to the [docs](https://developers.google.com/android/reference/com/google/android/gms/tasks/Task#public-abstract-exception-getexception). 

In any case, I changed it to log an error if it is null. 

```dart
if (!task.isSuccessful) {
    if (task.exception == null) {
        Log.e(TAG, "UnknownError: fetching device token.")
    }else{
        StreamHandlers.tokenReceived?.sendError(task.exception!!)
    }
    return@addOnCompleteListener
}
```

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationUtils.kt
```diff
@@ -3,84 +3,55 @@
...
 }
 
-fun RemoteMessage.isRemoteMessageSupported(): Boolean {
-    return !this.data[PushNotificationsConstants.PINPOINT_CAMPAIGN_CAMPAIGN_ID].isNullOrEmpty() or
-            !this.data[PushNotificationsConstants.JOURNEY_ID].isNullOrEmpty()
-}
\ No newline at end of file
+// Check if the notification's data payload is supported by us
+val Bundle.isSupported: Boolean
+    get() = keySet().any { it.contains(PushNotificationsConstants.PINPOINT_PREFIX) }
```
dnys1, line 56:
```suggestion
    get() = keySet().any { it.startsWith(PushNotificationsConstants.PINPOINT_PREFIX) }
```

packages/notifications/push/amplify_push_notifications/lib/callback_dispatcher.dart
```diff
@@ -0,0 +1,51 @@
...
+              'Invalid callback type: ${externalCallback.runtimeType}',
+            );
+          }
+          await externalCallback(
+            PushNotificationMessage.fromJson(
+              callbackInfo['notification'] as Map,
+            ),
+          );
+        }
+      }
```
dnys1, line 48: Use early exits to reduce nesting and improve readability:
```suggestion
      final externalHandle = callbackInfo['externalHandle'] as int?;
      if (externalHandle == null) {
        return;
      }
      final externalCallback = PluginUtilities.getCallbackFromHandle(
        CallbackHandle.fromRawHandle(externalHandle),
      );
      if (externalCallback == null) {
        _logger.debug('Could not locate callback for handle: $externalHandle');
        return;
      }
      if (externalCallback is! OnRemoteMessageCallback) {
        throw StateError(
          'Invalid callback type: ${externalCallback.runtimeType}',
        );
      }
      await externalCallback(
        PushNotificationMessage.fromJson(
          callbackInfo['notification'] as Map,
        ),
      );
```

packages/notifications/push/amplify_push_notifications_pinpoint/lib/src/pinpoint_provider.dart
```diff
@@ -1,25 +1,221 @@
...
+
+  // TODO(Samaritan1011001): Enable use of FlutterAnalyticsClient after merging to feature branch
+  late dynamic _analyticsClient;
+  final _campaginIdKey = 'campaign_id';
+  final _campaignActivityIdKey = 'campaign_activity_id';
+  final _treatmentIdKey = 'treatment_id';
+  String _androidCampaignIdKey = 'pinpoint.campaign.';
+  String _androidCampaignActivityIdKey =
+      'pinpoint.campaign.campaign_activity_id';
+  String _androidCampaignTreatmentIdKey = 'pinpoint.campaign.treatment_id';
```
dnys1, line 34: Not sure what's going on here - everything but `_analyticsClient` should be `static const` it seems and should not change ever.

```suggestion
  PinpointProvider() {
    _androidCampaignIdKey += _campaginIdKey;
    _androidCampaignActivityIdKey += _campaignActivityIdKey;
    _androidCampaignTreatmentIdKey += _treatmentIdKey;
  }

  // TODO(Samaritan1011001): Enable use of FlutterAnalyticsClient after merging to feature branch
  late dynamic _analyticsClient;
  final _campaginIdKey = 'campaign_id';
  final _campaignActivityIdKey = 'campaign_activity_id';
  final _treatmentIdKey = 'treatment_id';
  String _androidCampaignIdKey = 'pinpoint.campaign.';
  String _androidCampaignActivityIdKey =
      'pinpoint.campaign.campaign_activity_id';
  String _androidCampaignTreatmentIdKey = 'pinpoint.campaign.treatment_id';
```

packages/amplify_core/lib/src/types/notifications/push/push_notification_event_type.dart
```diff
@@ -0,0 +1,41 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+enum PushNotificationEventType {
+  notificaitonReceived,
+  notificationOpened,
+  tokenReceived
+}
+
+enum AWSPinpointMessageEvent {
```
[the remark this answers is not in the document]
Samaritan1011001: Actually only one is needed - `PinpointEventType`. Updated that and also separated each to their own file.

packages/notifications/push/amplify_push_notifications_pinpoint/lib/src/pinpoint_provider.dart
Samaritan1011001: Cleaned it up.

packages/amplify_core/lib/src/types/notifications/push/pinpoint_event_type.dart
```diff
@@ -0,0 +1,24 @@
...
+/// {@template amplify_core.push.pinpoint_event_type}
+/// Enum that indicates the different types recorded with Pinpoint
+/// {@endtemplate}
+enum PinpointEventType {
+  notificationOpened,
+  backgroundMessageReceived,
+  foregroundMessageReceived,
+}
+
+extension PinpointEventTypeExtension on PinpointEventType {
```
HuiSF, line 13: nit: could use enhanced enum instead of extension as we are enforcing 2.18+ Dart SDK, same to other enum.

```dart
enum PinpointEventType {
  notificationOpened('opened_notification'),
  backgroundMessageReceived('received_background'),
  foregroundMessageReceived('received_foreground');
  
  const PinpointEventType(this.name);
  
  final String name;
}
```

dnys1 (approved):
LGTM thanks! 🚀

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationEventsStreamHandler.kt
```diff
@@ -61,56 +61,110 @@ class PushNotificationEventsStreamHandler constructor(
...
-                NativeEvent.BACKGROUND_MESSAGE_RECEIVED,
-            )
+
+        var tokenReceived: PushNotificationEventsStreamHandler? = null
+
+        var notificationOpened: PushNotificationEventsStreamHandler? = null
+
+        var foregroundMessageReceived: PushNotificationEventsStreamHandler? = null
+
+        var isInitStreamHandlers: Boolean = false
```
HuiSF, line 119: nit
```suggestion
        private var isInitStreamHandlers: Boolean = false
```

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -184,19 +207,37 @@ class AmplifyPushNotifications extends PushNotificationsPluginInterface {
...
 
+  Future<void> _registerCallback(
+    Function callback,
+    CallbackType callbackType,
+  ) async {
+    final callbackHandle = PluginUtilities.getCallbackHandle(callback);
+    if (callbackHandle == null) {
+      _logger.error(
+        'Callback is not a global or static function',
+      );
```
HuiSF, line 225: Should we thrown exception instead of/in addition to logging an error?

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -184,19 +207,37 @@ class AmplifyPushNotifications extends PushNotificationsPluginInterface {
...
+    );
+    _logger.info('Successfully registered callback');
+  }
+
   Future<void> _registerDeviceWhenConfigure() async {
     late String deviceToken;
 
     try {
       deviceToken = await onTokenReceived.first;
+      _logger.info('Device Token: $deviceToken');
```
HuiSF, line 240: Needed?

HuiSF (approved):

.vscode/settings.json
```diff
@@ -1,8 +1,6 @@
 {
   "dart.runPubGetOnPubspecChanges": "always",
-  "dart.analysisExcludedFolders": [
-    "templates"
-  ],
+  "dart.analysisExcludedFolders": ["templates"],
```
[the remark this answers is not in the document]
Samaritan1011001: I think because I added a line in the beginning of this PR to limit the char length and because of that, this file kept getting formatted. Removed it now and should be good.

PR #2713, opened 2023-02-28
fix(analytics): event retry logic & max fail tries

fjnoyp:
*Issue #, if available:*

*Description of changes:*

Add logic to:
1) Treat AWSHttpException as retryable to fix no internet case
2) Retryable statusCode is 500-600
3) If a specific event fails 3 times, delete it.

### DIFF
--- a/packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
+++ b/packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
@@ -107,7 +107,7 @@ class EventClient implements Closeable {
     return completer.future;
   }
 
-  final _numFailuresByEvent = <String, int>{};
+  final _failureCountByEvent = <String, int>{};
 
   Future<void> _flushEvents() async {
     final storedEvents = await _eventStorage.retrieveEvents();
@@ -200,21 +200,19 @@ class EventClient implements Closeable {
       if (e.statusCode != null && _isRetryable(e.statusCode!)) {
         eventsToDelete.removeWhere((eventId, _) => _shouldRetryEvent(eventId));
         _logRecoverableException(e);
+      } else {
+        // TODO(kylecheng): convert to AnalyticsException and rethrow
+        // Convert status codes based on pinpoint api reference:
+        // https://docs.aws.amazon.com/pinpoint/latest/apireference/apps-application-id-events.html
+        // Follow logic structure of:
+        // https://github.com/aws-amplify/amplify-flutter/blob/next/packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart#L166
+        rethrow;
       }
-      // TODO(kylecheng): convert to AnalyticsException and log warning
-      // Convert status codes based on pinpoint api reference:
-      // https://docs.aws.amazon.com/pinpoint/latest/apireference/apps-application-id-events.html
-      // Follow logic structure of:
-      // https://github.com/aws-amplify/amplify-flutter/blob/next/packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart#L166
-      else {
-        _logger.warn('Unrecoverable HttpException Encountered', e);
-      }
-    } on AWSHttpException catch (e) {
+    } on AWSHttpException {
       // AWSHttpException indicates request did not complete
       // Due to no internet or unable to reach server.
       // These exceptions are always retryable.
       eventsToDelete.clear();
-      _logRecoverableException(e);
     } on Exception catch (e) {
       _logger
         ..warn('putEvents - exception encountered: ', e)
@@ -224,6 +222,11 @@ class EventClient implements Closeable {
       // Unless a retryable exception has been received (see above)
       if (eventsToDelete.isNotEmpty) {
         await _eventStorage.deleteEvents(eventsToDelete.values);
+
+        // Update _numFailuresByEvent to avoid memory leak
+        for (final eventId in eventsToDelete.keys) {
+          _failureCountByEvent.remove(eventId);
+        }
       }
     }
   }
@@ -241,13 +244,13 @@ class EventClient implements Closeable {
   }
 
   bool _shouldRetryEvent(String eventId) {
-    final timesFailed = _numFailuresByEvent[eventId] ?? 0;
+    final timesFailed = _failureCountByEvent[eventId] ?? 1;
     if (timesFailed >= 3) {
       _logger.warn('Retry limit exceeded, deleting event: $eventId');
-      _numFailuresByEvent.remove(eventId);
+      _failureCountByEvent.remove(eventId);
       return false;
     }
-    _numFailuresByEvent[eventId] = timesFailed + 1;
+    _failureCountByEvent[eventId] = timesFailed + 1;
     return true;
   }
 
--- a/packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
+++ b/packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
@@ -202,9 +202,10 @@ class EventClient implements Closeable {
       if (e.statusCode != null && _isRetryable(e.statusCode!)) {
         _handleRecoverableException(e, eventsToDelete);
       }
-    } on AWSHttpException catch (e) {
+    } on AWSHttpException {
       // AWSHttpException indicates request did not complete
-      _handleRecoverableException(e, eventsToDelete);
+      // Due to no internet.  These exceptions are always retryable.
+      eventsToDelete.clear();
     } on Exception catch (e) {
       _logger
         ..warn('putEvents - exception encountered: ', e)
@@ -235,7 +236,7 @@ class EventClient implements Closeable {
   }
 
   void _removeRetryableEvents(Map<String, StoredEvent> eventsToDelete) {
-    for (final eventId in eventsToDelete.keys) {
+    for (final eventId in eventsToDelete.keys.toList()) {
       if (_shouldEventRetry(eventId)) {
         eventsToDelete.remove(eventId);
       }
--- a/packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
+++ b/packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
@@ -107,7 +107,7 @@ class EventClient implements Closeable {
     return completer.future;
   }
 
-  final _failedEventIds = <String, int>{};
+  final _numFailuresByEvent = <String, int>{};
 
   Future<void> _flushEvents() async {
     final storedEvents = await _eventStorage.retrieveEvents();
@@ -180,14 +180,12 @@ class EventClient implements Closeable {
             'putEvents - issue with eventId: $eventId \n $eventItemResponse',
           );
 
-          if (_isRetryable(eventItemResponse.statusCode)) {
+          if (_isRetryable(eventItemResponse.statusCode) &&
+              _shouldEventRetry(eventId)) {
             _logger.warn(
               'putEvents - recoverable issue, will attempt to resend: $eventId in next FlushEvents',
             );
-
-            if (_shouldEventRetry(eventId)) {
-              eventsToDelete.remove(eventId);
-            }
+            eventsToDelete.remove(eventId);
           }
         }
       });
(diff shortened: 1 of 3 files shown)


packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
```diff
@@ -107,6 +107,8 @@ class EventClient implements Closeable {
     return completer.future;
   }
 
+  final _failedEventIds = <String, int>{};
```
HuiSF, line 110: nit: the naming seems a bit confusing when I initially read this line, `_numOfFailuresByEvent`?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
```diff
@@ -172,35 +174,38 @@ class EventClient implements Closeable {
...
-            'putEvents - issue with eventId: $eventID \n $eventItemResponse',
+            'putEvents - issue with eventId: $eventId \n $eventItemResponse',
           );
 
-          if (_isRetryable(eventItemResponse.message ?? '')) {
+          if (_isRetryable(eventItemResponse.statusCode)) {
             _logger.warn(
-              'putEvents - recoverable issue, will attempt to resend: $eventID in next FlushEvents',
+              'putEvents - recoverable issue, will attempt to resend: $eventId in next FlushEvents',
             );
```
HuiSF, line 186: nit: reading this log message, it should be inside `if (_shouldEventRetry(eventId))`

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
```diff
@@ -172,35 +174,38 @@ class EventClient implements Closeable {
...
       _logger
         ..warn('putEvents - exception encountered: ', e)
         ..warn(
             'Pinpoint event batch limits exceeded: 100 events / 4 mb total size / 1 mb max size per event \n '
             'Reduce your event size or change number of events in a batch')
         ..warn('Unrecoverable issue, deleting cache for local event batch');
+    } on SmithyHttpException catch (e) {
+      if (e.statusCode != null && _isRetryable(e.statusCode!)) {
+        _handleRecoverableException(e, eventsToDelete);
+      }
```
HuiSF, line 204: Smithy `HttpOperation` handles retry on `SmithyHttpException` with a condition, not sure if you need to handle it by yourself here.

https://github.com/aws-amplify/amplify-flutter/blob/8a7ce85f292f92403078a049aa72a073a391e23c/packages/smithy/smithy/lib/src/behavior/retryer.dart#L17-L19

fjnoyp: Thanks, took a look at that code.  

Unsure the conditions for when retryConfig is set or not.  

What is the general HTTPException returned?  Because all non 200 status codes aren't retryable right so in this case what Http exception would be returned

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
fjnoyp: Good call thanks

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
fjnoyp: Good point here, I've created another task and will handle general error handling for Analytics in that separate PR to come.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
```diff
@@ -172,35 +174,36 @@ class EventClient implements Closeable {
...
```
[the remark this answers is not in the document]
fjnoyp: Yes, this behavior should change.  

In a separate PR: 

1) AWSHttpExceptions should be thrown as NetworkException outside of event_client. 

2) For non recoverable SmithyHttpException, handle and throw according to similar S3 logic [here](https://github.com/aws-amplify/amplify-flutter/blob/19d8738831b7bcdb494cbaea9d77d70845364909/packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart#L166).

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
```diff
@@ -215,21 +229,26 @@ class EventClient implements Closeable {
...
-        !_equalsIgnoreCase(message, 'SerializationException') &&
-        !_equalsIgnoreCase(message, 'BadRequestException');
+  bool _isRetryable(int statusCode) {
+    // Pinpoint service policy is for 500-599 status codes to be retryable
+    return statusCode >= 500 && statusCode < 600;
+  }
+
+  bool _shouldRetryEvent(String eventId) {
+    final timesFailed = _numFailuresByEvent[eventId] ?? 0;
+    if (timesFailed >= 3) {
```
dnys1, line 245: If we're in this method, we've failed at least once. If I understand this correctly, you're retrying 3 times, yeah? This is the logic needed for 3 retries:

```suggestion
    final timesFailed = _numFailuresByEvent[eventId] ?? 1;
    if (timesFailed >= 3) {
```

packages/analytics/amplify_analytics_pinpoint_dart/test/event_client_test.dart
```diff
@@ -379,23 +377,42 @@ void main() {
...
+      when(
+        () => mockOperation.result,
+      ).thenThrow(PayloadTooLargeException());
+
+      when(() => pinpointClient.putEvents(any<PutEventsRequest>()))
+          .thenReturn(mockOperation);
+
+      await eventClient.flushEvents();
+
+      final items = eventStore.getCount(100);
```
dnys1, line 409: You should be checking the count before and after `flushEvents` to verify this

dnys1 (approved):

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
```diff
@@ -215,21 +229,26 @@ class EventClient implements Closeable {
...
+      _logger.warn('Retry limit exceeded, deleting event: $eventId');
+      _numFailuresByEvent.remove(eventId);
+      return false;
+    }
+    _numFailuresByEvent[eventId] = timesFailed + 1;
```
ragingsquirrel3, line 250: question: If a retry is successful, will the IDs for that event stay in `_numFailuresByEvent` creating a memory leak? I didn't see where this is cleared when successful, but maybe I'm missing something.

As a larger question, why is it necessary to track which events need retry? If the failure is by HTTP status code, can't you assume all the events in a request need a retry and just retry the request itself? something like https://pub.dev/packages/retry?

Also, will the retry here be immediate (seemed so from testing), would we want some kind of backoff for the retry?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
```diff
@@ -107,6 +107,8 @@ class EventClient implements Closeable {
     return completer.future;
   }
 
+  final _numFailuresByEvent = <String, int>{};
```
Jordan-Nelson, line 110: *nit:* Easier to understand without the abbreviation
```suggestion
  final _failureCountByEvent = <String, int>{};
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
```diff
@@ -215,21 +229,26 @@ class EventClient implements Closeable {
   }
 
   // If exception is recoverable, do not delete eventIds from local cache
-  void _handleRecoverableException(
-    SmithyHttpException e,
-    Map<String, StoredEvent> eventIdsToDelete,
-  ) {
+  void _logRecoverableException(Exception e) {
```
Jordan-Nelson, line 232: Why log a recoverable exception as a warning? This would be expected if a client goes offline.

Imagine a customer is streaming all warning / error logs to a remote server. If a client goes offline for a while, their logs will get filled with these warnings.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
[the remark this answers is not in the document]
Jordan-Nelson: Is there a good reason not to re throw it?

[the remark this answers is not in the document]
Jordan-Nelson: I am imagining a customer calling flush events manually and not knowing it failed because we just caught and logged it.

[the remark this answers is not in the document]
fjnoyp: Yeah we should throw ... updated

packages/analytics/amplify_analytics_pinpoint_dart/test/event_client_test.dart
```diff
@@ -404,15 +421,68 @@ void main() {
...
       when(() => pinpointClient.putEvents(any<PutEventsRequest>()))
           .thenReturn(mockOperation);
 
       await eventClient.flushEvents();
 
       final items = eventStore.getCount(100);
+      expect(items.length, 1);
+    });
+
+    test('flushEvents deletes events that fail >3 times', () async {
```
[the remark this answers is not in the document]
fjnoyp: Yes you're right it should be > 3

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
fjnoyp: Good call yes it would have, added code to fix that. 

Events can fail individually.  The underlying Pinpoint service sends separate responses for each event.  

Events send every 10 seconds.  There is no exponential backoff here, not sure if that's necessary since we always send events and can just retry on the next round.

ragingsquirrel3: Makes sense, thanks for explanation.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
Jordan-Nelson: I don't think any changes were made around this. Is there a good reason to have these as a warning? Recoverable issues that will be retried don't seem like warnings to me. They are expected.

Jordan-Nelson: Maybe log it as a warning after it hits the retry limit?

Jordan-Nelson (approved):
Approved. I left some comments around logging warnings. Please take a look. If recoverable errors occur, that doesn't seem like it should be a warning.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
fjnoyp: Hi Jordan I updated the code to only log recoverable exceptions for non network offline exceptions.  So the logs should not be filled with these warnings. 

I don't think they are expected, though that doesn't mean we have to log it though.  

The underlying error received is a Pinpoint `InternalServerErrorException` due to an 'unknown server error'.  

I guess we could argue since the user probably can't fix an Pinpoint internal error and that events will be retried anyway that this logging isn't necessary. 

Do you know if we have an Amplify specific concept of what a warning should be?  Maybe it should be logged still but at the info level as this is not expected, but still recoverable. 

And yes we already log if retry limit is hit.

Jordan-Nelson: If the network failures are not being logged, then this is fine. My main concern was what happened when a user went offline.

PR #2714, opened 2023-03-01
fix(storage): GetUrl signing

### LINKED ISSUE #2711: [Storage] GetURL incorrect signing for special characters
### Description

GetURL creates an invalid URL when using `DateTime.now().toString()` as the image key

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [ ] Auth
- [ ] Authenticator
- [ ] DataStore
- [X] Storage

### Steps to Reproduce

_No response_

### Screenshots

_No response_

### Platforms

- [ ] iOS
- [ ] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Flutter Version

3.7

### Amplify Flutter Version

next

### Deployment Method

Amplify CLI

### Schema

_No response_

dnys1:
Fixes #2711.

Presigned URLs should use `AWSHttpRequest.raw` since S3 requests require special handling of paths which doesn't align with `Uri` constructors.


packages/storage/amplify_storage_s3/example/integration_test/issue_2711_test.dart
```diff
@@ -0,0 +1,62 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
fjnoyp, line 1: Should we use a more descriptive key like getUrl_test / getUrl_special_characters / getUrl_signing ?

fjnoyp (approved):
looks good, just one none blocking comment

packages/storage/amplify_storage_s3/example/integration_test/issue_2711_test.dart
HuiSF: Should add this test to `main_test.dart` to run as a part of the Storage integ test suite?

HuiSF (approved):

Issue #2718, opened 2023-03-01
Getting SmithyException when user is not confirmed or does not exist


primico:
### Description

I am upgrading my flutter project to 1.0.0-next.5 and when I call Amplify.Auth.signIn and the user does not exist or is not confirmed, I get a SmithyException exception thrown in http_operation.dart:

```
UserNotConfirmedException {
message=User is not confirmed.,
}
UserNotFoundException {
  message=User does not exist.,
}
```

According to this, I don't think I should be getting an exception:
https://github.com/aws-amplify/amplify-flutter/issues/2390

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [X] Auth
- [ ] Authenticator
- [ ] DataStore
- [ ] Storage

### Steps to Reproduce

1. Call Amplify.Auth.signIn
2. SmithyException exception thrown in http_operation.dart when user does not exist or is not confirmed.

### Platforms

- [ ] iOS
- [X] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Flutter Version

3.7.5

### Amplify Flutter Version

1.0.0-next.5

### Deployment Method

Amplify CLI

dnys1:
Hi @primico - can you confirm that `Amplify.Auth.signIn` is throwing the exception and not returning? This should still be correct:

```dart
final signInRes = await Amplify.Auth.signIn(
  username: username,
  password: password,
);
final userConfirmationNeeded =
    signInRes.nextStep.signInStep == AuthSignInStep.confirmSignUp;
```

If you run your app in Debug mode with breakpoints for exceptions, it will show the exceptions being thrown, but the above code should still work.

dnys1:
I do agree that it shouldn't be tripping the "Uncaught Exceptions" breakpoint. I've opened a PR to hopefully resolve this.

primico:
Yes, my code is exactly what you just posted, and it works, and I use a try/catch to catch AuthExceptions which work, but this other SmithyException is something new that I have only seen after upgrading.

brunozorima:
Hi, is this resolved?

primico:
Along with the others above, I also get this SmithyException prior to the AuthException being caught:

NotAuthorizedException {
  message=Incorrect username or password.,
}

PR #2720, opened 2023-03-01
fix(smithy): Handle uncaught exception

dnys1:
Related to #2718.

When running a Flutter app in Debug mode w/ breakpoints on "Uncaught Exceptions", there are false negatives being tripped inside Smithy.

I'm not sure why this happens, but it's likely related to the interplay of Completer's and Zone's within CancelableOperation. Since these exceptions are already being handled properly, simply catching and rethrowing them seems to at least suppress the breakpoint trigger.


brunozorima:
Hi, is this resolved?

Jordan-Nelson (approved):

HuiSF (approved):

PR #2738, opened 2023-03-08
chore: migrate iOS unit tests to GH actions

ragingsquirrel3:
This PR does the following:
* changes `aft generate workflows` command to generate GH actions for iOS unit tests of Flutter packages that have iOS unit tests (just 4 packages).
* adds the generated workflows from running that command
* adds the base action to run iOS unit tests (consumed by generated workflows).

Removing Circle CI equivalent should be separate PR after similar change on main/stable and removing checks from repo settings.


.github/workflows/flutter_ios.yaml
```diff
@@ -0,0 +1,68 @@
...
+      working-directory:
+        description: The working directory relative to the repo root
+        required: true
+        type: string
+      package-name:
+        description: The name of the app-facing (or federated) iOS package being tested
+        required: true
+        type: string
+
+# Cancels in-progress job when there is another push to same ref.
```
ragingsquirrel3, line 14: callout: We do this in integration tests but I noticed we don't do it many unit test flows. Getting the result of a previous commit is nice to know when iterating on something, but IMO macos targets can get scarce so better to add this to conserve. I'm open to removing this here or if ppl think we should add more widely to unit test jobs.

.github/workflows/flutter_ios.yaml
```diff
@@ -0,0 +1,68 @@
...
+    name: Test
+    runs-on: macos-latest
+    timeout-minutes: 60
+    strategy:
+      # Allows other matrix items to run if one fails
+      fail-fast: false
+      matrix:
+        sdk:
+          - stable
+          - beta
```
dnys1, line 30: Instead of testing different Flutter versions (or maybe in addition to) can we test different Xcode or iOS versions? 

It looks like `macos-latest` (which is `macos-12`) uses Xcode 14.2 and `macos-11` uses Xcode 13.2. But it might be worth to compile for iOS 13 and iOS 16 - our min + current stable

Doesn't have to be this PR, just food for thought

.github/workflows/flutter_ios.yaml
```diff
@@ -0,0 +1,68 @@
...
+        run: |
+          set -o pipefail
+          flutter build ios --simulator
+          XCODEBUILD_DESTINATION="platform=iOS Simulator,name=iPhone 14 Pro Max,OS=latest"
+          xcodebuild test \
+              -workspace Runner.xcworkspace \
+              -scheme Runner \
+              -destination "$XCODEBUILD_DESTINATION" | xcpretty \
+              -r "junit" \
+              -o "test-results/${{ inputs.package-name }}-xcodebuild-test.xml"
```
dnys1, line 67: Don't think we need this without CircleCI

packages/aft/lib/src/commands/generate/generate_workflows_command.dart
```diff
@@ -219,4 +223,90 @@ jobs:
...
+  push:
+    branches:
+      - main
+      - stable
+      - next
+  pull_request:
+    paths:
+      - '$repoRelativePath/**/*.yaml'
+      - '$repoRelativePath/ios/**/*'
+      - '$repoRelativePath/example/ios/**/*'
```
dnys1, line 293: I would only include the unit tests folder in the example. This seems more appropriate for the integ tests/canaries.

packages/aft/lib/src/commands/generate/generate_workflows_command.dart
```diff
@@ -219,4 +223,90 @@ jobs:
...
+  Future<void> generateIosUnitTestWorkflow({
+    required PackageInfo package,
+    required String repoRelativePath,
+  }) async {
+    const iosWorkflow = 'flutter_ios.yaml';
+    final platformPackageName = '${package.name}_ios'; // federated _ios package
+    final platformPackagePath = '${package.path}_ios';
+    final appFacingIosTestDir =
+        Directory(p.join(package.path, 'example', 'ios', 'unit_tests'));
+    final platformIosPackageDir = Directory(platformPackagePath);
```
dnys1, line 237: Naming changed from without `ios` to with `ios`. I would prefer without I think, since we're already in an iOS context. But either way - just better to be consistent.

packages/aft/lib/src/commands/generate/generate_workflows_command.dart
```diff
@@ -219,4 +223,90 @@ jobs:
...
+    final platformIosPackageTestDir = Directory(
+      p.join(platformPackagePath, 'example', 'ios', 'unit_tests'),
+    ); // federated _ios package
+
+    final appFacingPackageIosTestsDirExists = appFacingIosTestDir.existsSync();
+    final platformPackageDirExists = platformIosPackageDir.existsSync();
+    final platformPackageIosTestDirExists =
+        platformIosPackageTestDir.existsSync();
+    final hasIosTests =
+        (appFacingPackageIosTestsDirExists || platformPackageIosTestDirExists);
```
dnys1, line 247:
```suggestion
        appFacingPackageIosTestsDirExists || platformPackageIosTestDirExists;
```

dnys1 (approved):

.github/workflows/flutter_ios.yaml
ragingsquirrel3: I agree, we should add iOS version matrix. I've been thinking about doing this in integration tests, too, maybe only running them on a smaller schedule load than other tests and adding extra matrices, as well. Will think about more for a future PR.

HuiSF: 👍 I think the regular tests should always run on the minimum version that we support. Others can run with longer interval. Beta version of Xcode is not offered correct?

HuiSF (approved):

.github/workflows/flutter_ios.yaml
ragingsquirrel3: I don't think beta is guaranteed to be available, the versions, sims and CLI tools they have are on https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md

PR #2742, opened 2023-03-09
chore(storage)!: remove unnecessary HeadObject call from download operations

HuiSF:
*Issue #, if available:*

*Description of changes:*

The `GetObject` call has already included object metadata in the API response, there is no need to make a separate `HeadObject` call to get properties, hence, removing the `getProperties` field from download operation options, and remove the `HeadObject` call from download task handling. 

Add metadata fields verification in the existing integration test suite. 

**NOTE:**
On Web, the CORS policy of Amplify managed S3 bucket doesn't allow exposing `x-amz-meta-*` headers, hence, user defined metadata may not be retrievable out-of-box. A [manual CORS policy update](https://github.com/aws-amplify/amplify-js/issues/2903#issuecomment-475012164) is required to make it work on Web. 

### DIFF
--- a/packages/storage/amplify_storage_s3_dart/lib/src/platform_impl/download_file/download_file_html.dart
+++ b/packages/storage/amplify_storage_s3_dart/lib/src/platform_impl/download_file/download_file_html.dart
@@ -69,18 +69,19 @@ Future<S3DownloadFileResult> _downloadFromUrl({
     name: request.localFile.path,
   );
 
+  // download is based on the presigned url on Web, so we make a separate
+  // HeadObject call to include object metadata in the operation result to
+  // maintain the same behavior of download on VM.
   return S3DownloadFileResult(
-    downloadedItem: s3Options.getProperties
-        ? (await storageS3Service.getProperties(
-            key: request.key,
-            options: targetIdentityId == null
-                ? S3GetPropertiesOptions(
-                    accessLevel: s3Options.accessLevel,
-                  )
-                : S3GetPropertiesOptions.forIdentity(targetIdentityId),
-          ))
-            .storageItem
-        : S3Item(key: request.key),
+    downloadedItem: (await storageS3Service.getProperties(
+      key: request.key,
+      options: targetIdentityId == null
+          ? S3GetPropertiesOptions(
+              accessLevel: s3Options.accessLevel,
+            )
+          : S3GetPropertiesOptions.forIdentity(targetIdentityId),
+    ))
+        .storageItem,
     localFile: request.localFile,
   );
 }
--- a/infra/lib/storage/stack.ts
+++ b/infra/lib/storage/stack.ts
@@ -110,6 +110,10 @@ class StorageIntegrationTestEnvironment extends IntegrationTestStackEnvironment<
             "x-amz-request-id",
             "x-amz-id-2",
             "ETag",
+            // This is required workaround on Web to retrieve metadata fields
+            // via GetObject and HeadObject.
+            // https://github.com/aws-amplify/amplify-js/issues/2903#issuecomment-475012164
+            "x-amz-meta-filename",
           ],
           maxAge: 3000,
         },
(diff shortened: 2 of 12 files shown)


packages/storage/amplify_storage_s3_dart/lib/src/platform_impl/download_file/download_file_html.dart
```diff
@@ -69,18 +69,19 @@ Future<S3DownloadFileResult> _downloadFromUrl({
...
-            key: request.key,
-            options: targetIdentityId == null
-                ? S3GetPropertiesOptions(
-                    accessLevel: s3Options.accessLevel,
-                  )
-                : S3GetPropertiesOptions.forIdentity(targetIdentityId),
-          ))
-            .storageItem
-        : S3Item(key: request.key),
+    downloadedItem: (await storageS3Service.getProperties(
```
dnys1, line 76: I don't follow this change? `getProperties` exists on other APIs still - should it be removed from those too?

dnys1 (commented):
Might be worth testing multiple metadata values maybe with some non-ASCII chars (which may need special handling according to [this](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#UserMetadata))

Also, should `Amplify.Storage.getProperties` be tested as well?

packages/storage/amplify_storage_s3_dart/lib/src/platform_impl/download_file/download_file_html.dart
HuiSF: Um no -

`getProperties` allows devs to configure retrieving metadata after an operation completes and return it along with the operation result, e.g. uploading an object with metadata, I want the upload result to include the metadata (not only the metadata I define, but also etag etc.) so I can use the info somewhere else.

For operation backed by `GetObject`, the metadata is directly available in the response. So it's available by default, setting `getProperties` to retrieve metadata is redundant. For other operations, setting `getProperties` to retrieve metadata is still required to support above use case.

NikaHsn (approved):

packages/storage/amplify_storage_s3_dart/lib/src/platform_impl/download_file/download_file_html.dart
dnys1: Hmm okay - I guess I'm just worried about the lopsidedness this creates in the APIs. If all of them take in the `getProperties` flag except this, it could be confusing. And if the flag actually does something on Web, might it be worth it to keep it?

HuiSF: That's a valid concern.

The opposite alternative would be keeping `getProperties` for download operations and manually suppressing the wrapping of `metadata` in the operation result on the VM if `getProperties` is set to false (so Web and VM will have the same behavior), even though it's already available via the underlying `GetObject` call.

Considering the cost of API calls, keeping `getProperties` would be more efficient on the Web, as calling `HeadObject` is not a default behavior. So after all this may be a better option. 

What do you think @NikaHsn ?

NikaHsn: at line 48 we call `storageS3Service.getUrl` with `checkObjectExistence: true`. this makes a call to`storageS3Service.getProperties` under the hood. so we can eliminate the second api call to `getProperties` (some code refactoring is needed for sure). with this change the `getProperties` flag won't be needed for web either.
in terms of keeping the flag for APIs consistancy, given that not all APIs have `getProperties` flag and it is not needed for `downloadFile` and `downloadData` APIs we should be good to remove it.

NikaHsn: we should be able to do it for download file html as `checkObjectExistence: true` always. so this won't be waiting till we make `checkObjectExistence: true` a default behavior, right?

PR #2744, opened 2023-03-10
feat(secure_storage)!: Use Local App Data Directory on Windows

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Use Local App Data Directory instead of Roaming dir on Windows

### DIFF
--- a/packages/secure_storage/amplify_secure_storage/lib/src/amplify_secure_storage.vm.dart
+++ b/packages/secure_storage/amplify_secure_storage/lib/src/amplify_secure_storage.vm.dart
@@ -6,6 +6,7 @@ import 'dart:io';
 
 import 'package:amplify_secure_storage/src/amplify_secure_storage.android.dart';
 import 'package:amplify_secure_storage/src/messages.cupertino.g.dart';
+import 'package:amplify_secure_storage/src/path_provider_local.dart';
 import 'package:amplify_secure_storage_dart/amplify_secure_storage_dart.dart';
 // ignore: implementation_imports
 import 'package:amplify_secure_storage_dart/src/utils/file_key_value_store.dart';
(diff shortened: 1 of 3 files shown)


dnys1 (approved):
Nice!

I feel like this introduces a bit of tech debt since we're extending `PathProviderWindows`. If `path_provider`/`path_provider_windows` ever change their implementation, we might be hosed.

Might be worth it at some point to fork `path_provider` and have our own `aws_path_provider` library, esp. since we could then use this in Dart-only apps.

Jordan-Nelson:
> Nice!
> 
> I feel like this introduces a bit of tech debt since we're extending `PathProviderWindows`. If `path_provider`/`path_provider_windows` ever change their implementation, we might be hosed.
> 
> Might be worth it at some point to fork `path_provider` and have our own `aws_path_provider` library, esp. since we could then use this in Dart-only apps.

@dnys1 - This was essentially my plan initially. However, the logic to get the app specific directory ([this code](https://github.com/flutter/packages/blob/5d4086e4656c670236fac2f6f0535c64e199a224/packages/path_provider/path_provider_windows/lib/src/path_provider_windows_real.dart#L191-L258)) was more complicated that I had thought it would be. If we fork/copy it it would introduce some operational overhead of maintaining that code (and tests for it). I felt that wasn't worth the tradeoff here. I think it is very unlikely that the impl changes in a way that is problematic for us without it being a major version bump.

I'm not totally opposed to maintaining our own version in order to better support non-flutter use cases in the future.

dnys1:
Ah, makes sense. Yeah, I think this is a lower maintenance burden than the original code. And I agree, this could benefit many other people if Flutter would open support to Dart apps natively.

ragingsquirrel3 (approved):

PR #2747, opened 2023-03-13
feat(notifications): killed state notification handling and Analytics integration

Samaritan1011001:
*Issue #, if available:*

*Description of changes:*
Previously, we created a Background Flutter Engine and communicated to the Dart side via a callback dispatcher. This has a limitation of not been able to access the required config file in a killed state. 

This change creates an Engine when not already created and calls the app's main.dart which initializes our plugin that then is used to record the BG event. Hence, the callback dispatcher is not needed as the whole app is booted up upon receiving a notification when the app is in killed state. This also means the callback function provided to BG API need not be a global or static function anymore. 


packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -274,40 +260,51 @@ class AmplifyPushNotifications extends PushNotificationsPluginInterface {
...
+    for (final element in [..._eventQueue, withItem]) {
+      if (element != null) {
+        await Future.wait(
+          _onNotificationReceivedInBackgroundCallbacks.map(
+            (callback) async {
+              await callback(PushNotificationMessage.fromJson(element));
+            },
+          ),
+        );
+      }
```
HuiSF, line 307:
```suggestion
      if (element != null) {
        final notification = PushNotificationMessage.fromJson(element);
        await Future.wait(
          _onNotificationReceivedInBackgroundCallbacks.map(
            (callback) async {
              await callback(notification);
            },
          ),
        );
      }
```

packages/notifications/push/amplify_push_notifications_pinpoint/example/lib/main.dart
```diff
@@ -86,6 +68,30 @@ class _MyAppState extends State<MyApp> {
...
+    //       globalBgCallbackCount != null ? (globalBgCallbackCount + 1) : 1;
+    //   await prefs.setInt(
+    //     globalBgCallbackKey,
+    //     globalBgCallbackCount,
+    //   );
+    //   print('globalBgCallbackCount in handler -> $globalBgCallbackCount');
+    // } on Exception catch (e) {
+    //   print(' error in handler: $e');
+    // }
+    // return;
```
HuiSF, line 92: Should remove?

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationFirebaseMessagingService.kt
```diff
@@ -67,15 +76,17 @@ class PushNotificationFirebaseMessagingService : FirebaseMessagingService() {
...
                     )
-                    FlutterMain.startInitialization(baseContext)
-                    FlutterMain.ensureInitializationComplete(baseContext, null)
-                    PushNotificationBackgroundService.enqueueWork(
-                        baseContext, remoteMessage.toIntent()
-                    )
+                    val mainEngine = FlutterEngineCache.getInstance()
+                        .get(PushNotificationPluginConstants.FLUTTER_ENGINE_ID)
+                    if (mainEngine == null) {
+                        engineGroup.createAndRunDefaultEngine(baseContext)
```
HuiSF, line 85: I would add comments as doc here to describe that this runs from `lib/main.dart` if the App is at the killed state at this moment and why, or wrap this in a method `runAppFromKilledState` or something similar to make it explicit.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
```diff
@@ -60,53 +66,54 @@ open class AmplifyPushNotificationsPlugin : FlutterPlugin, ActivityAware,
...
      * when the app is in killed state and null otherwise.
      */
     private var launchNotification: MutableMap<Any, Any?>? = null
 
+
     override fun onAttachedToEngine(flutterPluginBinding: FlutterPlugin.FlutterPluginBinding) {
         mainBinaryMessenger = flutterPluginBinding.binaryMessenger
+        FlutterEngineCache.getInstance().put(
+            PushNotificationPluginConstants.FLUTTER_ENGINE_ID,
+            flutterPluginBinding.flutterEngine
```
HuiSF, line 81: The underlying Java method [`getFlutterEngine()` is deprecated](https://api.flutter.dev/javadoc/io/flutter/embedding/engine/plugins/FlutterPlugin.FlutterPluginBinding.html#getFlutterEngine()), how would this work when Flutter finally drops it?

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
```diff
@@ -60,53 +66,54 @@ open class AmplifyPushNotificationsPlugin : FlutterPlugin, ActivityAware,
...
 
     /**
      * Launch notification has the notification map when app was launched by tapping on the notification
      * when the app is in killed state and null otherwise.
      */
     private var launchNotification: MutableMap<Any, Any?>? = null
 
+
     override fun onAttachedToEngine(flutterPluginBinding: FlutterPlugin.FlutterPluginBinding) {
         mainBinaryMessenger = flutterPluginBinding.binaryMessenger
```
HuiSF, line 78: Looking at this file, this assignment used only in this method implementation, can make it local, so you don't need `!!` and reset it as null below.

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -58,10 +55,11 @@ class AmplifyPushNotifications extends PushNotificationsPluginInterface {
           .createChild('AmplifyPushNotification');
   final ServiceProviderClient _serviceProviderClient;
   final PushNotificationsHostApi _hostApi;
-  final _PushNotificationsFlutterApi _flutterApi;
+  late final _PushNotificationsFlutterApi _flutterApi;
```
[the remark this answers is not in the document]
Samaritan1011001: Artifact of debugging. removed.

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -131,11 +129,10 @@ class AmplifyPushNotifications extends PushNotificationsPluginInterface {
       _recordAnalyticsForLaunchNotification(launchNotification);
     }
 
-    // Register the callback dispatcher
-    await _registerCallback(
-      callbackDispatcher,
-      CallbackType.dispatcher,
-    );
+    // TODO(Samaritan1011001): This gets called twice in killed state and suspends when token.first is not available
+    await _registerDeviceWhenConfigure();
```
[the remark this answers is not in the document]
Samaritan1011001: No, this was before I added the Flutter cache logic. Removed it now.

packages/notifications/push/amplify_push_notifications_pinpoint/example/lib/main.dart
Samaritan1011001: needed to demonstrate working of this API in killed state. So, uncommented it.

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -257,7 +273,7 @@ class _PushNotificationsFlutterApi implements PushNotificationsFlutterApi {
   final _eventQueue = <Map<Object?, Object?>>[];
 
   final _onNotificationReceivedInBackgroundCallbacks =
-      <OnRemoteMessageCallback>[];
+      <OnRemoteMessageCallback>{};
```
HuiSF, line 276: Is changing to a `Map` intended here?

Samaritan1011001: No, changed back to list.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
Samaritan1011001: Good question. Thinking about it, this is pretty much only to indicate that a Fluter Engine exists in the group. Perhaps the same can be done with a bool in Shared Prefs.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationFirebaseMessagingService.kt
```diff
@@ -84,4 +92,16 @@ class PushNotificationFirebaseMessagingService : FirebaseMessagingService() {
             }
         }
     }
+
+    private fun runAppFromKilledState(){
+        // Check if there is already a main Flutter Engine running
+        val mainEngine = FlutterEngineCache.getInstance()
+            .get(PushNotificationPluginConstants.FLUTTER_ENGINE_ID)
+        if (mainEngine == null) {
+            StreamHandlers.deInit()
```
HuiSF, line 101: Is it possible that `StreamHandlers` could have been `init`'ed when the App is at the killed state?

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -181,62 +172,56 @@ class AmplifyPushNotifications extends PushNotificationsPluginInterface {
...
+  ) =>
+      _serviceProviderClient.recordNotificationEvent(
+        eventType: PinpointEventType.notificationOpened,
+        notification: pushNotificationMessage,
+      );
 
-  void _tokenReceivedListener(String deviceToken) {
-    unawaited(_registerDevice(deviceToken));
+  Future<void> _tokenReceivedListener(String deviceToken) async {
+    await _registerDevice(deviceToken);
```
HuiSF, line 218: `_registerDevice` may throw below, when that happens, the exception may become unhandled exception at this point.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
Samaritan1011001: `launchNotification` needs to be nullable so we can assign it null after it's consumed.

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -274,40 +259,53 @@ class AmplifyPushNotifications extends PushNotificationsPluginInterface {
...
   final _onNotificationReceivedInBackgroundCallbacks =
       <OnRemoteMessageCallback>[];
 
-  void registerOnReceivedInBackgroundCallback(
+  Future<void> registerOnReceivedInBackgroundCallback(
     OnRemoteMessageCallback callback,
-  ) {
+  ) async {
     _onNotificationReceivedInBackgroundCallbacks.add(callback);
+    await _flushEvents();
```
HuiSF, line 282: nit: Does `_flushEvents()` need to be `await`'ed for? the call of `registerOnReceivedInBackgroundCallback` at line 88 seems not need to wait for it to complete.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
HuiSF: OK if there is another way to do check, can add a TODO comment and update later.

HuiSF (approved):
LGTM with some comments 👍

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationFirebaseMessagingService.kt
Samaritan1011001: Since I added the refresh flag to `StreamHandlers` init method and use that in `onAttachedToEngine`, this is not required. Removing it. 
The `StreamHandlers` are `init'ed` when the app receives a notification in the killed state and user opens the app. Hence we refresh the handlers so they are attached to the main engine.

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
Samaritan1011001: right, changed to unawaited

dnys1 (approved):

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -181,23 +172,47 @@ class AmplifyPushNotifications extends PushNotificationsPluginInterface {
     await _hostApi.setBadgeCount(badgeCount);
   }
 
+  Future<void> _registerDeviceWhenConfigure() async {
+    late String deviceToken;
+
+    try {
+      deviceToken = await onTokenReceived.first;
```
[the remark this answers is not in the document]
Samaritan1011001: `deviceToken` needs to declared earlier and initialized later because I moved the `_registerDevice` call after the catch block because it throws its own Exception if any.

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -274,40 +259,53 @@ class AmplifyPushNotifications extends PushNotificationsPluginInterface {
...
 
-    await Future.wait(
-      _onNotificationReceivedInBackgroundCallbacks.map(
-        (callback) async {
-          await callback(notification);
-        },
-      ),
-    );
+  Future<void> _flushEvents({Map<Object?, Object?>? withItem}) async {
+    for (final element in [..._eventQueue, withItem]) {
```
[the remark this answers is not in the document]
Samaritan1011001: Using whereNotNull requires the collection pacakge so replaced it with whereTyp

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationEventsStreamHandler.kt
```diff
@@ -125,7 +125,13 @@ class StreamHandlers {
          * for queuing work when binary messenger is not ready e.g FirebaseService's onNewToken
          */
         @JvmStatic
-        fun initStreamHandlers() {
+        fun initStreamHandlers(refresh: Boolean) {
+            // Refresh the Stream handlers when needed.
+            // When app opens up from a killed state, the stream handlers associated with the background engine
+            // needs to deInit and new ones to be initialized. Used in onAttachedToEngine.
+            if (refresh) {
+                deInit()
```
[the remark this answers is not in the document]
Samaritan1011001: When the app is in killed state and notification arrives, we create and start the default engine using `createAndRunDefaultEngine` method that basically calls the flutter app's `lib/main.dart `as the entrypoint which wires up the event channels in our plugin to the default engine. 
When the app is opened by the user, a new Flutter Engine gets created which does not seem to call the `onDetached` methods for the previous default engine. Hence I make sure to `deInit` and create new stream handlers.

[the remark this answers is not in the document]
dnys1: Is this parameter needed, then? Couldn't you always call `deInit` first?

packages/notifications/push/amplify_push_notifications_pinpoint/example/lib/main.dart
```diff
@@ -86,6 +65,28 @@ class _MyAppState extends State<MyApp> {
     });
   }
 
+  void bgHandler(PushNotificationMessage pushNotificationMessage) async {
```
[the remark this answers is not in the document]
Samaritan1011001: Now that we are using `engineGroup.createAndRunDefaultEngine(baseContext)` to create and call the default engine, Flutter app's `lib/main.dart` gets called. Which means the Customer's app gets woken up in the background that calls the `Amplify.configure` which in turn configures our event channels and listeners. 

If they want to listen to updates when the app is killed, calling the `Amplify.Notifications.Push.onNotificationReceivedInBackground` in the initState will be good enough. Since the app gets woken up in killed state, this API will get wired up and get fired.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationEventsStreamHandler.kt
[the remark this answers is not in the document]
Samaritan1011001: That's true. `deInit`ing it always now.

packages/smithy/goldens/bin/main/ModelConverter.kt
```diff
@@ -0,0 +1,44 @@
+import software.amazon.smithy.model.Model
```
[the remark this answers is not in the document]
Samaritan1011001: Unintended, removed

HuiSF (approved):

PR #2748, opened 2023-03-14
chore: migrate Android unit tests to GH actions

ragingsquirrel3:
This PR migrates Android unit tests to GH actions. It largely copies the actions from next with some manual modifications (like only running against Flutter stable, not beta). I looked at the CircleCI config and made sure to test the same packages as that aspect is a little different from next.

Removing CircleCI equivalents should be separate PR. 

### DIFF
--- a/.github/workflows/amplify_analytics_pinpoint.android.yaml
+++ b/.github/workflows/amplify_analytics_pinpoint.android.yaml
@@ -4,7 +4,6 @@ on:
     branches:
       - main
       - stable
-      - next
   pull_request:
     paths:
       - 'packages/analytics/amplify_analytics_pinpoint/**/*.yaml'


packages/amplify_core/android/src/test/kotlin/com/amazonaws/amplify/amplify_core/AtomicResultTest.kt
```diff
@@ -66,21 +63,4 @@ class AtomicResultTest {
         atomicResult.success(null)
         verify(mockResult, times(1)).success(any())
     }
-
-    @Test
-    fun multipleConcurrentRepliesAreNotSent() = coroutinesTestRule.testDispatcher.runBlockingTest {
```
ragingsquirrel3, line 71: this test was flaky on gh actions, not sure why (same issue on next). I just removed it here (and on next) but open to keeping if anyone has suggestions. I didn't think it was worth investigation.

Jordan-Nelson (dismissed):

.github/workflows/amplify_analytics_pinpoint.android.yaml
```diff
@@ -0,0 +1,28 @@
+name: amplify_analytics_pinpoint Android
+on:
+  push:
+    branches:
+      - main
+      - stable
+      - next
```
NikaHsn, line 7: do we want this for next branch?

NikaHsn, line 7: do we want it for next branch?

ragingsquirrel3: nope! Good catch, will remove.

NikaHsn (approved):

Jordan-Nelson (approved):

PR #2752, opened 2023-03-14
chore(analytics): throw AnalyticsExceptions

fjnoyp:
Throw exceptions typed as AnalyticsExceptions.

Update UnknownException to be category agnostic.

*Issue #, if available:*

*Description of changes:*


packages/amplify_core/lib/src/types/analytics/analytics_types.dart
```diff
@@ -1,8 +1,13 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+/// Exceptions.
+export '../exception/analytics/analytics_exception.dart';
+export '../exception/network_exception.dart';
+export '../exception/unknown_exception.dart';
```
HuiSF, line 7: As categories now share the same `unknown_exception.dart` should it be exported from `amplift_core.dart` instead of from each category type entry?

fjnoyp: Thanks updated the exports.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/exception/pinpoint_analytics_exception.dart
```diff
@@ -0,0 +1,63 @@
...
+      underlyingException: exception,
+    );
+  }
+
+  /// Creates a [PinpointException] from a [NotFoundException].
+  factory PinpointException.fromNotFoundException(NotFoundException exception) {
+    return PinpointException(
+      'The Pinpoint resource was not found.',
+      recoverySuggestion:
+          'Please ensure Analytics is properly configured in your amplifyconfiguration.dart file.',
```
Jordan-Nelson, line 34: Is this message helpful? When would this be expected to happen?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/exception/pinpoint_analytics_exception.dart
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  /// Creates a [PinpointException] from a [ForbiddenException].
+  factory PinpointException.fromForbiddenException(
+    ForbiddenException exception,
+  ) {
+    return PinpointException(
+      'Forbidden from accessing the Pinpoint resource.',
+      recoverySuggestion:
+          'Please ensure Analytics is properly configured in your amplifyconfiguration.dart file and that your auth rules allow for access.',
```
Jordan-Nelson, line 46: > auth rules allow for access

Is this referring to unauthenticated guest access?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/exception/pinpoint_analytics_exception.dart
```diff
@@ -0,0 +1,63 @@
...
+class PinpointException extends AnalyticsException {
+  /// {@macro amplify_analytics_pinpoint_dart.pinpoint_exception}
+  const PinpointException(
+    super.message, {
+    required String super.recoverySuggestion,
+    super.underlyingException,
+  });
+
+  /// Creates a [PinpointException] from a [BadRequestException].
+  factory PinpointException.fromBadRequestException(
```
Jordan-Nelson, line 19: Should be private?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/exception/pinpoint_analytics_exception.dart
```diff
@@ -0,0 +1,63 @@
...
+  ) {
+    return PinpointException(
+      'Error in request data sent to Pinpoint.',
+      recoverySuggestion: _fileIssueMessage,
+      underlyingException: exception,
+    );
+  }
+
+  /// Creates a [PinpointException] from a [NotFoundException].
+  factory PinpointException.fromNotFoundException(NotFoundException exception) {
```
Jordan-Nelson, line 30: Should be private?

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/exception/pinpoint_analytics_exception.dart
```diff
@@ -0,0 +1,63 @@
+import 'package:amplify_analytics_pinpoint_dart/src/sdk/pinpoint.dart';
+import 'package:amplify_core/amplify_core.dart';
+
+const _fileIssueMessage =
+    'This exception is not expected. Please try again. If the exception persists, please file an issue at https://github.com/aws-amplify/amplify-flutter/issues';
```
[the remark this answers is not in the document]
fjnoyp: Yes it was.  Thanks I switched to `AmplifyExceptionMessages.missingExceptionMessage` 

Storage has that wording too but it's used throughout in about 8 different exception messages.  

@HuiSF what are your thoughts about removing your `_fileIssueMessage` and replacing it with `AmplifyExceptionMessages.missingExceptionMessage` ?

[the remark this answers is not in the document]
HuiSF: Yes that makes sense to me to have a single source to get the messages and maintain consistency cross plugins.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/exception/pinpoint_analytics_exception.dart
fjnoyp: The idea was to direct users to inspect their setup in case of this exception instead of throwing it within the generic AnalyticsException.  

In my previous tests I thought it was triggered on a amplifyconfig mistep but now only getting ForbiddenExceptions so we can remove this as I'm unsure when it's expected to happen if at all now.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/exception/pinpoint_analytics_exception.dart
fjnoyp: Yes.  I can make this better: 

and that unauthenticated guest access is allowed or an authorized user is signed in.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/exception/pinpoint_analytics_exception.dart
```diff
@@ -0,0 +1,63 @@
...
+
+/// {@template amplify_analytics_pinpoint_dart.pinpoint_exception}
+/// Represents exceptions that may be thrown calling Pinpoint Analytics plugin APIs.
+/// {@endtemplate}
+class PinpointException extends AnalyticsException {
```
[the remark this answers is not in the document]
fjnoyp: That makes sense.  

Looking at the code, it seems Storage is the only category that's throwing plugin specific exceptions (S3Exceptions) now.  The logic of your comment might apply there too.  

Though for storage, the user might want to catch the individual exception types while in the case for Analytics here there really isn't much unique logic that I'd anticipate customers would write for a BadRequest/NotFound/Forbidden case beyond seeing it in their debugging. 

I'll update Analytics. 


However, perhaps Storage should also change and now have Storage vs. S3 exceptions as well. 

In fact, shouldn't we throw the specific error types for Storage? 

For example: 

S3IncompleteDownloadException instead of a S3Exception with an exception message of incomplete download string? 

I'd anticipate customers might actually want to have custom logic based on that specific case. 

cc @HuiSF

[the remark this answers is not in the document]
HuiSF: Good point from you both.
The `S3Exception` serves as a collection of the exception parser function but ultimately they are `StorageException`. Throwing specific exception variants here is unnecessary, only the specific storage exception types defined in core should be thrown. I can find some time to update the `S3Exception`.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/exception/pinpoint_analytics_exception.dart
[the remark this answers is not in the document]
fjnoyp: Great thank you for confirming Hui.  I'll update your code to use the common source as well.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/exception/pinpoint_analytics_exception.dart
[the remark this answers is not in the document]
fjnoyp: Sounds good.  If you are busy I can also find time to redo the S3Exception and make a PR for your review @HuiSF.  I'll see how my other work goes and then followup with you in DM.

packages/analytics/amplify_analytics_pinpoint_dart/test/endpoint_client_test.dart
```diff
@@ -260,5 +260,18 @@ void main() {
...
+
+    test(
+        'updateEndpoint throws AnalyticsException from unrecognized exceptions',
+        () async {
+      when(() => pinpointClient.updateEndpoint(any())).thenThrow(Exception());
+
+      expect(
+        endpointClient.updateEndpoint(),
+        throwsA(
+          const TypeMatcher<AnalyticsException>(),
```
dnys1, line 272:
```suggestion
          isA<AnalyticsException>(),
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/impl/analytics_client/event_client/event_client.dart
```diff
@@ -175,48 +176,52 @@ class EventClient implements Closeable {
         );
       }
       eventsItemResponse?.forEach((eventId, eventItemResponse) {
+        final eventType = eventsToDelete[eventId]?.event.eventType ?? eventId;
```
dnys1, line 179: Wouldn't `eventsToSend` make more sense since it's never modified?

dnys1 (approved):

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/exception/pinpoint_analytics_exception.dart
```diff
@@ -0,0 +1,38 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'package:amplify_analytics_pinpoint_dart/src/sdk/pinpoint.dart';
+import 'package:amplify_core/amplify_core.dart';
+
+AnalyticsException _fromBadRequestException(
+  BadRequestException exception,
+) {
+  return AnalyticsException(
```
Jordan-Nelson, line 10: Following the pattern in Auth/Storage, I think we should make `AnalyticsException` abstract to force sub types to be thrown. This would then become UnknownException. That can be a follow up PR though if you would prefer.

```suggestion
  return UnknownException(
```

Jordan-Nelson (approved):

PR #2757, opened 2023-03-15
fix(db_common): Windows build

dnys1:
- Fixes windows build by including the `sqlite3` lib in the library's outputs
- Adds `amplify_db_common` integration tests to CI


packages/common/amplify_db_common/example/pubspec.yaml
```diff
@@ -1,16 +1,21 @@
...
-  sdk: '>=2.18.0 <4.0.0'
+  sdk: ">=2.18.0 <4.0.0"
 
 dependencies:
   amplify_db_common: ">=0.1.2+3 <0.2.0"
   drift: ">=2.4.0 <2.6.0"
   flutter:
     sdk: flutter
 
+dependency_overrides:
```
fjnoyp, line 14: Could we add comments here explaining why we have to use overrides instead of regular dependencies for these imports?

fjnoyp (approved):

packages/common/amplify_db_common/example/pubspec.yaml
dnys1: Added 👍

HuiSF (approved):

PR #2765, opened 2023-03-20
test(api): Expand custom authorizer tests

dnys1:
- Adds more custom authorizer tests for query parameters, headers, and HTTP/2.
- Fixes header casing in HTTP/2 requests
- Fixes query parameter encoding issue
- Fixes API using access token instead of ID token

### DIFF
--- a/packages/api/amplify_api_dart/lib/src/decorators/authorize_http_request.dart
+++ b/packages/api/amplify_api_dart/lib/src/decorators/authorize_http_request.dart
@@ -87,7 +87,7 @@ T _validateAuthProvider<T extends AmplifyAuthProvider>(
   if (authProvider == null) {
     throw ApiException(
       'No auth provider found for auth mode ${authType.name}.',
-      recoverySuggestion: 'Ensure auth plugin correctly configured.',
+      recoverySuggestion: 'Ensure API plugin correctly configured.',
     );
   }
   return authProvider;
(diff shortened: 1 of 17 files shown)


packages/api/amplify_api_dart/lib/src/decorators/authorize_http_request.dart
```diff
@@ -87,7 +87,7 @@ T _validateAuthProvider<T extends AmplifyAuthProvider>(
   if (authProvider == null) {
     throw ApiException(
       'No auth provider found for auth mode ${authType.name}.',
-      recoverySuggestion: 'Ensure auth plugin correctly configured.',
+      recoverySuggestion: 'Ensure API plugin correctly configured.',
```
ragingsquirrel3, line 90: :face-palm, good catch!

ragingsquirrel3 (approved):

packages/aws_signature_v4/lib/src/request/canonical_request/canonical_request_util.dart
```diff
@@ -3,11 +3,16 @@
 
 part of 'canonical_request.dart';
 
+final RegExp _encodedComponent = RegExp('%[A-F0-9]{2}');
+
 /// Encodes a query parameter while preventing double-encoding.
 String _safeEncode(String queryComponent) {
```
Jordan-Nelson, line 9: Not fully in scope of these changes, but I wonder if we are handling all edge cases appropriately with this. Two edge cases come to mind:
1. Strings that have a %XX pattern in them, but are not actually encoded. For example `foo bar%20`. This function will leave the string unchanged. Should this be encoded to `foo%20bar%2520`?
2. Strings that are double/triple/etc. encoded. For example,`foo%2520bar` (which is `foo bar` double encoded). Should these be fully decoded, and then single encoded? Or should we leave them encoded multiple times?

Jordan-Nelson (approved):

PR #2774, opened 2023-03-23
fix(storage): add handling of AWSHttpException

HuiSF:
*Issue #, if available:*

*Description of changes:*

* No longer export `S3Exception` where `StorageException` from amplify_core should be used to handle exceptions at runtime
* Added handling of `AWSHttpException` calling S3 SDK APIs

### DIFF
--- a/packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
+++ b/packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
@@ -224,6 +224,16 @@ class S3Exception extends StorageException {
         recoverySuggestion: AmplifyExceptionMessages.missingExceptionMessage,
       );
 
+  /// Creates a [NetworkException] over a [AWSHttpException] thrown from
+  /// [s3.S3Client] APIs.
+  static NetworkException fromAWSHttpException(AWSHttpException exception) {
+    return NetworkException(
+      'The request failed due to a network error.',
+      recoverySuggestion: 'Ensure that you have an active network connection',
+      underlyingException: exception,
+    );
+  }
+
   @override
   String get runtimeTypeName => 'S3Exception';
 }
--- a/packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
+++ b/packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
@@ -1,8 +1,10 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+@internal
 import 'package:amplify_core/amplify_core.dart';
 import 'package:amplify_storage_s3_dart/src/sdk/s3.dart' as s3;
+import 'package:meta/meta.dart';
 import 'package:smithy/smithy.dart';
 
 const _keyNotFoundRecoveryMessage =
--- a/packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
+++ b/packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
@@ -6,6 +6,7 @@ import 'dart:math';
 
 import 'package:amplify_core/amplify_core.dart' hide PaginatedResult;
 import 'package:amplify_storage_s3_dart/amplify_storage_s3_dart.dart';
+import 'package:amplify_storage_s3_dart/src/exception/s3_storage_exception.dart';
 import 'package:amplify_storage_s3_dart/src/sdk/s3.dart' as s3;
 import 'package:amplify_storage_s3_dart/src/sdk/src/s3/common/endpoint_resolver.dart'
     as endpoint_resolver;
--- a/packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
+++ b/packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
@@ -124,6 +124,8 @@ class StorageS3Service {
       } on smithy.UnknownSmithyHttpException catch (error) {
         // S3Client.headObject may return 403 error
         throw S3Exception.fromUnknownSmithyHttpException(error);
+      } on AWSHttpException catch (error) {
+        throw S3Exception.fromAWSHttpException(error);
       }
     }
 
@@ -158,6 +160,8 @@ class StorageS3Service {
     } on smithy.UnknownSmithyHttpException catch (error) {
       // S3Client.headObject may return 403 error
       throw S3Exception.fromUnknownSmithyHttpException(error);
+    } on AWSHttpException catch (error) {
+      throw S3Exception.fromAWSHttpException(error);
     }
   }
 
@@ -425,6 +429,8 @@ class StorageS3Service {
     } on smithy.UnknownSmithyHttpException catch (error) {
       // S3Client.copyObject may return 403 or 404 error
       throw S3Exception.fromUnknownSmithyHttpException(error);
+    } on AWSHttpException catch (error) {
+      throw S3Exception.fromAWSHttpException(error);
     }
 
     return S3CopyResult(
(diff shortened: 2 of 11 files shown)


packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
```diff
@@ -211,6 +213,16 @@ class S3Exception extends StorageException {
         recoverySuggestion: _fileIssueMessage,
       );
 
+  /// Creates a [NetworkException] over a [AWSHttpException] thrown from
+  /// [s3.S3Client] APIs.
+  static NetworkException fromAWSHttpException(AWSHttpException exception) {
```
NikaHsn, line 218: other categories do similar, `throw NetworkException on AWSHttpException`. How about to define this at amplify_core (types/exception/network_exception.dart) instead?

HuiSF: Agreed, will refactor (actually remove S3Exception) in the following PRs, will move this parser around accordingly as part of the following PR.

packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
```diff
@@ -222,6 +224,16 @@ class S3Exception extends StorageException {
         recoverySuggestion: AmplifyExceptionMessages.missingExceptionMessage,
       );
 
+  /// Creates a [NetworkException] over a [AWSHttpException] thrown from
+  /// [s3.S3Client] APIs.
+  static NetworkException fromAWSHttpException(AWSHttpException exception) {
+    return NetworkException(
+      'The request failed due to a network error.',
```
fjnoyp, line 231: Should we define a common NetworkException preset as well?  Since the text here is the same as Analytics?

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
```diff
@@ -424,6 +429,8 @@ class StorageS3Service {
     } on smithy.UnknownSmithyHttpException catch (error) {
       // S3Client.copyObject may return 403 or 404 error
       throw S3Exception.fromUnknownSmithyHttpException(error);
+    } on AWSHttpException catch (error) {
```
fjnoyp, line 432: Instead of duplicating the same pattern 'on smithy throw, on awsHttpexception, throw' can you wrap this in common code.  For Analytics I did that in a `pinpointAnalyticsException.fromPinpointException` method.

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_download_task_test.dart
```diff
@@ -589,11 +599,16 @@ void main() {
...
-            } else {
+            } else if (exception is NoSuchKey) {
               expect(
                 downloadTask.result,
                 throwsA(isA<StorageKeyNotFoundException>()),
               );
+            } else {
+              expect(
+                downloadTask.result,
+                throwsA(isA<NetworkException>()),
```
fjnoyp, line 610: So in this and similar tests are you expecting all 4 exceptions to be thrown?

StorageAccessDeniedException
StorageHttpStatusException
StorageKeyNotFoundException
NetworkException

HuiSF: This test is driven by a for loop, each loop is an exception type to be tested.
They are not expected to be thrown at the same time.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
HuiSF: I personally prefer to keep the exception catching inline and close to the API call so I'd know immediately what kind of exception I would expect from the API call.

packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
HuiSF: Same as my comment above, will make the NetworkException conversion universal in a [separate PR](https://github.com/aws-amplify/amplify-flutter/pull/2781/commits/8925c5f93455136eaadb0c60077e2e46b9933e23).

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_download_task_test.dart
fjnoyp: Okay that makes sense.  I've had similar approaches to unit tests in the past, but Jordan had a good point about keeping tests DAMP. 

Personally as an outside observer of your code, the use of for loops makes it hard for me to understand what you're testing.  I know it saves a lot of redundant code, but I don't think that's necessarily the point of writing unit tests.  It seems the file is already quite large, so perhaps it can be broken in to sub files if you're worried about file size. 

[HERE](https://enterprisecraftsmanship.com/posts/dry-damp-unit-tests/#:~:text=DRY%20stands%20for%20%22Don't,representation%20in%20your%20code%20base.&text=DAMP%20stands%20for%20%22Descriptive%20and,the%20readability%20of%20the%20code.)

fjnoyp (approved):
Okay I left a comment about unit testing.  I had a similar approach before but changed it based on discussions with Jordan.

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_download_task_test.dart
HuiSF: OK thanks for providing the reference, I think it makes sense.

I would argue however, this forEach driven test is DAMP as it's "for each exception type I'm doing this test"

I think the confusing point is the assertion part that is not somehow clear as the forEach itself. I will improve this part a bit to make it more clear.

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_download_task_test.dart
```diff
@@ -467,136 +521,87 @@ void main() {
...
+        test('on resume ${item.description}', () async {
+          final testGetObjectOutput1 = GetObjectOutput(
+            contentLength: Int64(1024),
+            body: Stream<List<int>>.periodic(
+              const Duration(microseconds: 200),
+              (_) => [101],
+            ).take(1024),
+          );
+          final smithyOperation1 = MockSmithyOperation<GetObjectOutput>();
+          final receivedState = <S3TransferState>[];
```
Jordan-Nelson, line 562: Do you need this? Can you get rid of the `onProgress` callback below?

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_download_task_test.dart
```diff
@@ -467,136 +521,87 @@ void main() {
...
-            );
+      for (final item in exceptionTestItems) {
+        test('on resume ${item.description}', () async {
+          final testGetObjectOutput1 = GetObjectOutput(
+            contentLength: Int64(1024),
+            body: Stream<List<int>>.periodic(
+              const Duration(microseconds: 200),
+              (_) => [101],
+            ).take(1024),
+          );
```
Jordan-Nelson, line 560: Can this be moved to test setup?

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_download_task_test.dart
```diff
@@ -467,136 +521,87 @@ void main() {
...
+            defaultS3ClientConfig: defaultS3ClientConfig,
+            prefixResolver: testPrefixResolver,
+            bucket: testBucket,
+            key: testKey,
+            options: defaultTestOptions,
+            logger: logger,
+            onProgress: (progress) {
+              receivedState.add(progress.state);
+            },
+          );
```
Jordan-Nelson, line 586: Can this be moved to test setup?

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_download_task_test.dart
```diff
@@ -467,136 +521,87 @@ void main() {
...
+
+          final downloadTask = S3DownloadTask(
+            s3Client: s3Client,
+            defaultS3ClientConfig: defaultS3ClientConfig,
+            prefixResolver: testPrefixResolver,
+            bucket: testBucket,
+            key: testKey,
+            options: defaultTestOptions,
+            logger: logger,
+          );
```
Jordan-Nelson, line 544: Can this be moved to test setup?

Jordan-Nelson (commented):
Left a few questions/comments on the tests. I do not consider any of them blocking. I haven't looked at the rest of the PR, so will let other approve.

PR #2777, opened 2023-03-24
chore(notifications): unit tests and refactor to include Android utils changes 

Samaritan1011001:
*Issue #, if available:*

*Description of changes:*
Contains the following major changes,
- Updated NotificationPayload structure based on the changes from released shared Android utils 
- Unit tests for the Dart code
- Unit tests for Android native code


packages/amplify_core/lib/src/types/notifications/push/push_notification_message.dart
```diff
@@ -30,42 +30,50 @@ class PushNotificationMessage
...
-            : DateTime.fromMillisecondsSinceEpoch(sentTimeInt);
-        fcmOptions = FcmPlatformOptions(
-          channelId: fcmOptionsMap['channelId'] as String?,
-          messageId: fcmOptionsMap['messageId'] as String?,
-          senderId: fcmOptionsMap['sender'] as String?,
-          sentTime: sentTime,
-        );
       }
+    } on Error catch (e) {
+      safePrint('Error when converting Json to PushNotificationMessage: $e');
```
HuiSF, line 76: What's the source of error here if the code has already handled fields nullability correctly?

packages/notifications/push/amplify_push_notifications_pinpoint/lib/src/pinpoint_provider.dart
```diff
@@ -72,18 +80,21 @@ class PinpointProvider implements ServiceProviderClient {
   }) async {
     try {
       if (!_isInitialized) {
-        _logger.error(
-          'Pinpoint provider not configured, re-run Amplify.configure',
+        throw Exception(
+          'Pinpoint provider is not initialized. Re-run Amplify.configure.',
```
HuiSF, line 84: This can't be generic `Exception` IMO
We need proper error message, recovery suggestion and underlying exception (if any).

In addition "re-run" is not accurate, Amplify doesn't support re-configure, so when the exception happens, the only case is that configure has never been called.

packages/notifications/push/amplify_push_notifications_pinpoint/lib/src/push_notifications_background_processing.dart
```diff
@@ -19,14 +19,17 @@ const _backgroundChannel = MethodChannel(
...
+  AmplifySecureStorage? amplifySecureStorage,
+}) async {
   WidgetsFlutterBinding.ensureInitialized();
-  final amplifySecureStorage = AmplifySecureStorage(
-    config: AmplifySecureStorageConfig(
-      scope: 'amplifyPushNotifications',
-    ),
-  );
-  final amplifyconfigStr = await amplifySecureStorage.read(
+  final amplifySecureStorage0 = amplifySecureStorage ??
```
HuiSF, line 26: Setting the default value in the parameter list, so `amplifySecureStorage` can be non-nullable.

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -75,21 +106,29 @@ class AmplifyPushNotifications extends PushNotificationsPluginInterface {
...
-
   late final Stream<PushNotificationMessage> _onNotificationOpened;
-
   var _isConfigured = false;
   PushNotificationMessage? _launchNotification;
   final Future<void> Function() _backgroundProcessor;
+
+  final bool _androidCheck =
+      Zone.current[androidCheckOverride] as bool? ?? Platform.isAndroid;
+  final bool _iosCheck =
```
dnys1, line 124:
```suggestion
  final bool _isIOS =
```

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -29,12 +31,29 @@ const _foregroundNotificationEventChannel = EventChannel(
...
 
-const _externalHandleKey = 'externalHandleKey';
+/// {@template amplify_push_notifications.external_handle_key}
+/// Key used to store the external callback handle function.
+///
+/// Used in this file to store the callback function and used in Flutter API class to access it.
+/// {@endtemplate}
+const externalHandleKey = 'externalHandleKey';
 
 const _needsConfigurationException = PushNotificationException(
```
dnys1, line 41: Make this a `ConfigurationError`

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -29,12 +31,29 @@ const _foregroundNotificationEventChannel = EventChannel(
...
   'Configure Amplify with Notifications Plugin before using this method.',
 );
 
+/// {@template amplify_push_notifications.android_check_override}
+/// Internal Platform check exposed for testing purposes only.
+/// {@endtemplate}
+@visibleForTesting
+const androidCheckOverride = #_androidCheckOverride;
```
dnys1, line 49: This basically reimplements the functionality of `package:os_detect`. I would probably recommend moving to that. You can see example usage in `aws_common`.

packages/notifications/push/amplify_push_notifications/lib/src/push_notifications_flutter_api.dart
```diff
@@ -0,0 +1,121 @@
...
+final AmplifyLogger _logger = AmplifyLogger.category(Category.pushNotifications)
+    .createChild('AmplifyPushNotification');
+
+/// {@template amplify_push_notifications.amplify_push_notifications_flutter_api}
+/// Internal Platform check exposed for testing purposes only.
+/// {@endtemplate}
+class AmplifyPushNotificationsFlutterApi
+    implements PushNotificationsFlutterApi {
+  /// {@macro amplify_push_notifications.amplify_push_notifications_flutter_api}
+  AmplifyPushNotificationsFlutterApi() {
```
dnys1, line 19: Make this a singleton

packages/notifications/push/amplify_push_notifications/lib/src/push_notifications_flutter_api.dart
```diff
@@ -0,0 +1,121 @@
...
+    PushNotificationsFlutterApi.setup(this);
+  }
+
+  final _eventQueue = <PushNotificationMessage>[];
+
+  /// {@template amplify_push_notifications.event_queue}
+  /// Internal eventQueue getter exposed for testing purposes only.
+  /// {@endtemplate}
+  @visibleForTesting
+  List<PushNotificationMessage> get eventQueue => _eventQueue;
```
dnys1, line 29:
```suggestion
  List<PushNotificationMessage> get eventQueue => UnmodifiableListView(_eventQueue);
```

packages/notifications/push/amplify_push_notifications/lib/src/push_notifications_flutter_api.dart
```diff
@@ -0,0 +1,121 @@
...
+  final _onNotificationReceivedInBackgroundCallbacks =
+      <OnRemoteMessageCallback>[];
+
+  /// {@template amplify_push_notifications.on_notification_received_in_background_callbacks}
+  /// Internal callbacks list getter exposed for testing purposes only.
+  /// {@endtemplate}
+  @visibleForTesting
+  List<OnRemoteMessageCallback>
+      get onNotificationReceivedInBackgroundCallbacks =>
+          _onNotificationReceivedInBackgroundCallbacks;
```
dnys1, line 40:
```suggestion
          UnmodifiableListView(_onNotificationReceivedInBackgroundCallbacks);
```

packages/notifications/push/amplify_push_notifications/lib/src/push_notifications_flutter_api.dart
```diff
@@ -0,0 +1,121 @@
...
+        'Could not locate external callback for the stored external handle',
+      );
+      return;
+    }
+    if (externalCallback is! OnRemoteMessageCallback) {
+      throw PushNotificationException(
+        'Invalid callback type: ${externalCallback.runtimeType}',
+      );
+    }
+    externalCallback(pushNotificationMessage);
```
dnys1, line 85:
```suggestion
    await externalCallback(pushNotificationMessage);
```

packages/notifications/push/amplify_push_notifications/lib/src/push_notifications_flutter_api.dart
```diff
@@ -0,0 +1,121 @@
...
+    } else {
+      _eventQueue.add(notification);
+    }
+
+    await _dispatchToExternalHandle(notification);
+    await Future.wait(
+      _onNotificationReceivedInBackgroundCallbacks.map(
+        (callback) async {
+          await callback(notification);
+        },
```
dnys1, line 106:
```suggestion
        (callback) => callback(notification),
```

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -0,0 +1,444 @@
...
+  [
+    PushNotificationsHostApi,
+    ServiceProviderClient,
+    AmplifySecureStorage,
+    PluginUtilities,
+    AmplifyPushNotificationsFlutterApi,
+  ],
+)
+void main() {
+  const tokenReceivedEventChannel = EventChannel(
```
dnys1, line 34: This should be shared with the impl

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -49,15 +68,27 @@ class AmplifyPushNotifications extends PushNotificationsPluginInterface {
   AmplifyPushNotifications({
     required ServiceProviderClient serviceProviderClient,
     required Future<void> Function() backgroundProcessor,
```
dnys1, line 70: Why is this required?

dnys1: This class should probably be `abstract`. I don't see when it would be instantiated directly ever.

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -0,0 +1,444 @@
...
+
+      testWidgetsFlutterBinding.defaultBinaryMessenger.setMockMethodCallHandler(
+        MethodChannel(tokenReceivedEventChannel.name),
+        handler,
+      );
+    });
+    test('should configure correctly', () async {
+      log.clear();
+      when(mockPushNotificationsHostApi.getLaunchNotification()).thenAnswer(
+        (_) => Future(() => standardAndroidPushMessage),
```
dnys1, line 127: Update throughout

```suggestion
        (_) async => standardAndroidPushMessage,
```

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -0,0 +1,444 @@
...
+        mockServiceProviderClient.recordNotificationEvent(
+          eventType: anyNamed('eventType'),
+          notification: anyNamed('notification'),
+        ),
+      ).called(1);
+    });
+
+    test('should throw exception when configuring if there is no appId present',
+        () async {
+      final config0 = AmplifyConfig.fromJson(
```
dnys1, line 148: It's okay to shadow variable names

```suggestion
      final config = AmplifyConfig.fromJson(
```

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -0,0 +1,444 @@
...
+          authProviderRepo: authProviderRepo,
+          config: config0,
+        ),
+        throwsA(const TypeMatcher<PushNotificationException>()),
+      );
+    });
+
+    test('should fail configure when registering device is unsuccessfull',
+        () async {
+      log.clear();
```
dnys1, line 162: This should be in `setUp`

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -0,0 +1,444 @@
...
+      testWidgetsFlutterBinding.defaultBinaryMessenger.setMockMethodCallHandler(
+        MethodChannel(tokenReceivedEventChannel.name),
+        throwErrorHandler,
+      );
+      expect(
+        () async => plugin.configure(
+          authProviderRepo: authProviderRepo,
+          config: config,
+        ),
+        throwsA(const TypeMatcher<PushNotificationException>()),
```
dnys1, line 172: Should be checking more than just that it throws, e.g.

```suggestion
        throwsA(
          isA<PushNotificationException>()
            .having((e) => e.message, 'message', contains('token'),
        ),
```

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -0,0 +1,444 @@
...
+    AmplifyPushNotificationsFlutterApi,
+  ],
+)
+void main() {
+  const tokenReceivedEventChannel = EventChannel(
+    'com.amazonaws.amplify/push_notification/event/TOKEN_RECEIVED',
+  );
+
+  final log = <MethodCall>[];
+  late TestWidgetsFlutterBinding testWidgetsFlutterBinding;
```
dnys1, line 39: Should only be called once per file.

```suggestion
  final testWidgetsFlutterBinding = TestWidgetsFlutterBinding.ensureInitialized();
```

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -0,0 +1,444 @@
...
+      log.clear();
+      testWidgetsFlutterBinding.defaultBinaryMessenger.setMockMethodCallHandler(
+        MethodChannel(tokenReceivedEventChannel.name),
+        throwErrorHandler,
+      );
+      expect(
+        () async => plugin.configure(
+          authProviderRepo: authProviderRepo,
+          config: config,
+        ),
```
HuiSF, line 171:
```suggestion
        plugin.configure(
          authProviderRepo: authProviderRepo,
          config: config,
        ),
```

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -0,0 +1,444 @@
...
+      );
+      expect(
+        () async => plugin.launchNotification,
+        throwsA(const TypeMatcher<PushNotificationException>()),
+      );
+      expect(
+        () async => plugin.identifyUser(
+          userId: 'userId',
+          userProfile: AnalyticsUserProfile(),
+        ),
```
HuiSF, line 197:
```suggestion
        plugin.identifyUser(
          userId: 'userId',
          userProfile: AnalyticsUserProfile(),
        ),
```

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -0,0 +1,444 @@
...
+        throwsA(const TypeMatcher<PushNotificationException>()),
+      );
+      expect(
+        () async => plugin.requestPermissions(),
```
HuiSF, line 201:
```suggestion
        plugin.requestPermissions,
```

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -0,0 +1,444 @@
...
+        throwsA(const TypeMatcher<PushNotificationException>()),
+      );
+      expect(
+        () async => plugin.getPermissionStatus(),
```
HuiSF, line 205:
```suggestion
        () async => plugin.getPermissionStatus,
```

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -0,0 +1,444 @@
...
+      );
+
+      expect(
+        () async => plugin.setBadgeCount(42),
+        throwsA(const TypeMatcher<PushNotificationException>()),
+      );
+      expect(
+        () async => plugin.getBadgeCount(),
+        throwsA(const TypeMatcher<PushNotificationException>()),
+      );
```
HuiSF, line 216: These two assertion actually caught the exception thrown for `!_iosCheck` I believe, which made the test inaccurate...

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -0,0 +1,444 @@
...
+      expect(
+        res,
+        42,
+      );
+    });
+
+    test('setBadgeCount calls the native layer to set', () async {
+      await plugin.setBadgeCount(42);
+      verify(
+        mockPushNotificationsHostApi.setBadgeCount(any),
```
HuiSF, line 301:
```suggestion
        mockPushNotificationsHostApi.setBadgeCount(42),
```

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -0,0 +1,444 @@
...
+          final pref = await SharedPreferences.getInstance();
+          plugin = AmplifyPushNotifications(
+            serviceProviderClient: mockServiceProviderClient,
+            backgroundProcessor: () async => {},
+          )..onNotificationReceivedInBackground(
+              testGlobalCallbackFunction,
+            );
+          await Future.delayed(const Duration(seconds: 3), () {});
+
+          expect(pref.containsKey(externalHandleKey), true);
```
HuiSF, line 386: nit: use a matcher when possible (same to other similar spots)
```suggestion
          expect(pref.containsKey(externalHandleKey), isTrue);
```

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -0,0 +1,444 @@
...
+          SharedPreferences.setMockInitialValues({});
+          final pref = await SharedPreferences.getInstance();
+          plugin = AmplifyPushNotifications(
+            serviceProviderClient: mockServiceProviderClient,
+            backgroundProcessor: () async => {},
+            dependencyManager: dependencyManager,
+          )..onNotificationReceivedInBackground(
+              testGlobalCallbackFunction,
+            );
+          await Future.delayed(const Duration(seconds: 3), () {});
```
HuiSF, line 407: nit: Does it really need 3 secs?

```suggestion
          await Future<void>.delayed(const Duration(seconds: 3));
```

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -0,0 +1,444 @@
...
+            serviceProviderClient: mockServiceProviderClient,
+            backgroundProcessor: () async => {},
+            dependencyManager: dependencyManager,
+          )..onNotificationReceivedInBackground(
+              (testGlobalCallbackFunction) {},
+            );
+          verify(
+            mockPushNotificationsNativeToFlutterApi
+                .registerOnReceivedInBackgroundCallback(
+              any,
```
HuiSF, line 434: verify the `callback` is the one.

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
Samaritan1011001: No, it's just some arbitrary time I put in to check. Can be lesser.

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
Samaritan1011001: The ios check only logs an error and doesn't throw

packages/notifications/push/amplify_push_notifications_pinpoint/lib/src/pinpoint_provider.dart
```diff
@@ -157,7 +168,7 @@ class PinpointProvider implements ServiceProviderClient {
       }
       if (data.containsKey(_androidCampaignTreatmentIdKey)) {
         campaign['treatment_id'] =
-            data[_androidCampaignTreatmentIdKey] as String;
+            data[_androidCampaignTreatmentIdKey].toString();
```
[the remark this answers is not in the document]
Samaritan1011001: Unintentional. Reverting.

packages/amplify_core/lib/src/types/notifications/push/push_notification_message.dart
Samaritan1011001: I added it to gracefully fail if the type of any field changes. Since we pass the rawData and other fields as we get in the native code, the type of them may be changed by cx when composing their messages I think. So, catching it and printing it rather than crashing. 

What would be the alternative here?

packages/notifications/push/amplify_push_notifications/test/push_notification_message_test.dart
```diff
@@ -0,0 +1,53 @@
...
+      expect(parsedAndroidMessage.imageUrl, 'TEST_URL');
+
+      // ios shares the same key to pass along deeplink and url
+      final parseiOSMessage =
+          PushNotificationMessage.fromJson(imageUrliOSMessage);
+      expect(parseiOSMessage.imageUrl, 'TEST_URL');
+    });
+
+    test('should not crash when there is a type mismatch', () {
+      PushNotificationMessage.fromJson(defectivePushMessage);
```
HuiSF, line 50:
```suggestion
      expect(
        () {
          PushNotificationMessage.fromJson(defectivePushMessage);
        },
        returnsNormally,
      )
```

packages/notifications/push/amplify_push_notifications/test/push_notification_flutter_api_test.dart
```diff
@@ -0,0 +1,92 @@
...
+      const OperatingSystem('android', ''),
+      () async {
+        SharedPreferences.setMockInitialValues({});
+        final pref = await SharedPreferences.getInstance();
+        AmplifyPushNotifications(
+          serviceProviderClient: MockServiceProviderClient(),
+          backgroundProcessor: () async => {},
+          // dependencyManager: dependencyManager,
+        ).onNotificationReceivedInBackground(
+          testGlobalCallbackFunction,
```
dnys1, line 78:
```suggestion
          expectAsync1(testGlobalCallbackFunction),
```

You need to set an expectation that this will be called. As it stands, this test would complete successfully whether the function is called or not.

packages/notifications/push/amplify_push_notifications/test/push_notification_flutter_api_test.dart
```diff
@@ -0,0 +1,92 @@
...
+  test('should call the external callback function', () async {
+    SharedPreferences.setMockInitialValues({});
+    void externalCallback(PushNotificationMessage pushMessage) {
+      expect(
+        pushMessage.title,
+        PushNotificationMessage.fromJson(standardAndroidPushMessage).title,
+      );
+    }
+
+    flutterApi.registerOnReceivedInBackgroundCallback(externalCallback);
```
dnys1, line 61:
```suggestion
    flutterApi.registerOnReceivedInBackgroundCallback(
      expectAsync1(externalCallback),
    );
```

packages/notifications/push/amplify_push_notifications/test/push_notification_flutter_api_test.dart
```diff
@@ -0,0 +1,92 @@
...
+void main() {
+  late AmplifyPushNotificationsFlutterApi flutterApi;
+
+  setUp(() {
+    TestWidgetsFlutterBinding.ensureInitialized();
+    flutterApi = AmplifyPushNotificationsFlutterApi.instance;
+  });
+
+  test('should queue events when the serviceClient is not ready', () {
+    SharedPreferences.setMockInitialValues({});
```
dnys1, line 28: Put this in `setUp`

packages/notifications/push/amplify_push_notifications/test/push_notification_flutter_api_test.dart
```diff
@@ -0,0 +1,92 @@
...
+        final pref = await SharedPreferences.getInstance();
+        AmplifyPushNotifications(
+          serviceProviderClient: MockServiceProviderClient(),
+          backgroundProcessor: () async => {},
+          // dependencyManager: dependencyManager,
+        ).onNotificationReceivedInBackground(
+          testGlobalCallbackFunction,
+        );
+        flutterApi
+            .registerOnReceivedInBackgroundCallback(testGlobalCallbackFunction);
```
dnys1, line 81: Why does this need to be called if `onNotificationReceived` is called above?

packages/notifications/push/amplify_push_notifications/test/push_notification_flutter_api_test.dart
Samaritan1011001: Tried `expectAsync1` but won't work for this case because `onNotificationReceivedInBackground` expects the callback function provided to be global or static but when I use `expectAsync1`, I believe it wraps it and makes it a local function?

packages/notifications/push/amplify_push_notifications/test/push_notification_flutter_api_test.dart
```diff
@@ -0,0 +1,92 @@
...
+
+        await Future.delayed(const Duration(seconds: 3), () {});
```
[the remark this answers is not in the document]
Samaritan1011001: It's just an arbitrary time I set to wait for the callback function to be registered. Since we want to keep the `onNotificationReceivedInBackground` async we are not waiting for the registration process to finish, hence I wait before triggering the function that is responsible to handle the notification coming from the native side.

packages/notifications/push/amplify_push_notifications/test/push_notification_flutter_api_test.dart
Samaritan1011001: Since I changed `flutterApi` to be a `singleton`, any registered callback here will be registered in the next tests as well. Making this `expectAsync1` errors out since the next test also invokes `externalCallback`.

packages/notifications/push/amplify_push_notifications/test/push_notification_message_test.dart
Samaritan1011001: Thanks, but removing this test based on the decision to remove the try catch block in `fromJson`.

packages/notifications/push/amplify_push_notifications/test/push_notification_flutter_api_test.dart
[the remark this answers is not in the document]
HuiSF: If it's arbitrary please use a smaller delay to make the tests run faster. The tests may be roughly 6 seconds slower waiting for this arbitrary time :D

HuiSF (approved):

packages/notifications/push/amplify_push_notifications/test/push_notification_flutter_api_test.dart
dnys1: There should be no state left over between tests - this will greatly increase the risk of false positives in tests (like here potentially). I would recommend adding a `reset` method to the static class and call that in between tests.

packages/notifications/push/amplify_push_notifications/test/push_notification_flutter_api_test.dart
dnys1: There should be no state left over between tests - this will greatly increase the risk of false positives in tests (like here potentially). I would recommend adding a reset method to the static class and call that in between tests.

packages/notifications/push/amplify_push_notifications/test/push_notification_flutter_api_test.dart
dnys1: You can implement it yourself with completers

packages/notifications/push/amplify_push_notifications/android/build.gradle
```diff
@@ -71,6 +66,16 @@ def amplifyFlutter = project.rootProject.findProject(':amplify_flutter_android')
...
-    implementation 'com.amplifyframework:aws-push-notifications-pinpoint-utils:2.1.0'
+    // Import support library for Amplify push utils
+    implementation 'com.amplifyframework:aws-push-notifications-pinpoint-common:2.5.0'
+    implementation "org.jetbrains.kotlinx:kotlinx-serialization-json:1.4.1"
     implementation project(path: ':flutter_plugin_android_lifecycle')
+    implementation 'androidx.test:core-ktx:1.5.0'
+
+    testImplementation 'junit:junit:4.13.2'
+    testImplementation 'io.mockk:mockk:1.13.4'
+    testImplementation 'org.slf4j:slf4j-nop:2.0.7'
```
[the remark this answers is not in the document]
dnys1: I'm guessing it's used by `mockk`, then? None of the packages we typically use require this. Can you add a note that it's needed for `mockk`?

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
dnys1: Can you update to make this abstract?

packages/notifications/push/amplify_push_notifications/lib/src/push_notifications_flutter_api.dart
```diff
@@ -0,0 +1,129 @@
+import 'dart:async';
```
dnys1, line 1: Still missing several license headers throughout. To add them all at once:

```
cd packages/notifications
../../tool/license.sh
```

packages/notifications/push/amplify_push_notifications/test/push_notification_flutter_api_test.dart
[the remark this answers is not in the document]
dnys1: If you implement the `expectAsync` pattern with a completer, you can avoid this wait

dnys1 (approved):

packages/notifications/push/amplify_push_notifications/test/push_notification_flutter_api_test.dart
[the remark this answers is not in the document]
Samaritan1011001: I am having to wait in-between callback registration and notification arrival because when registering we are using `unawaited` within the `onNotificationReceivedInBackground` [function](https://github.com/aws-amplify/amplify-flutter/blob/feat/push/android-utils/packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart#L162-L167). This choice was made to allow for using the API without awaiting it but as we can see it can introduce problems if the notification arrives too early as mimicked in our test by the use of `flutterApi.onNotificationReceivedInBackground`. 

I don't completely understand how completer is useful here but one proposal I have is to make the `onNotificationReceivedInBackground` API function async which will ensure the async operation to store the callback handle in shared preference is successful before it's ready to receive notifications.

PR #2781, opened 2023-03-27
refactor(storage)!: simplify storage s3 exceptions throwing

HuiSF:
*Issue #, if available:*

*Description of changes:*

This PR depends on https://github.com/aws-amplify/amplify-flutter/pull/2774

* Promoted `StorageTransferState` to amplify_core package
* No longer throwing exception on resuming already canceled task (the call of `resume()` doesn't do anything)
* Added generic `StorageOperationCanceled` exception to amplify_core
* Simplified `S3Exception` handling and remove unnecessary logics around it

### DIFF
--- a/packages/amplify_core/lib/src/types/exception/storage/operation_canceled_exception.dart
+++ b/packages/amplify_core/lib/src/types/exception/storage/operation_canceled_exception.dart
@@ -1,0 +1,19 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'package:amplify_core/amplify_core.dart';
+
+/// {@template amplify_core.storage.operation_canceled_exception}
+/// Exception thrown when a storage operation is canceled.
+/// {@endtemplate}
+class StorageOperationCanceledException extends StorageException {
+  /// {@macro amplify_core.storage.operation_canceled_exception}
+  const StorageOperationCanceledException(
+    super.message, {
+    super.recoverySuggestion,
+    super.underlyingException,
+  });
+
+  @override
+  String get runtimeTypeName => 'StorageOperationCanceledException';
+}
--- a/packages/amplify_core/lib/src/types/storage/transfer_progress.dart
+++ b/packages/amplify_core/lib/src/types/storage/transfer_progress.dart
@@ -1,30 +1,51 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-/// {@template amplify_storage_s3.transfer_progress}
+/// {@template amplify_core.storage.transfer_state}
+/// State of a download task.
+/// {@endtemplate}
+enum StorageTransferState {
+  /// The download task is in progress.
+  inProgress,
+
+  /// The download task is paused.
+  paused,
+
+  /// The download task is canceled.
+  canceled,
+
+  /// The download task completed successfully.
+  success,
+
+  /// The download task failed.
+  failure,
+}
+
+/// {@template amplify_core.storage.transfer_progress}
 /// The progress of a storage transfer operation.
 /// {@endtemplate}
 class StorageTransferProgress {
-  /// {@macro amplify_storage_s3.transfer_progress}
+  /// {@macro amplify_core.transfer_progress}
   const StorageTransferProgress({
     required this.transferredBytes,
     required this.totalBytes,
+    required this.state,
   });
 
   /// The current progress, in bytes, for the storage transfer operation.
   final int transferredBytes;
 
   /// The total number of bytes for the storage transfer operation.
+  ///
+  /// In an upload operation, the value may be `-1` if the upload source size
+  /// cannot be determined.
   final int totalBytes;
 
-  @Deprecated('Use transferredBytes instead')
-  int get currentBytes => transferredBytes;
+  /// The [StorageTransferState] of the [StorageTransferProgress].
+  final StorageTransferState state;
 
   /// The fractional progress of the storage transfer operation.
   ///
   /// 0 <= `fractionCompleted` <= 1
   double get fractionCompleted => transferredBytes / totalBytes;
-
-  @Deprecated('Use fractionCompleted instead')
-  double getFractionCompleted() => fractionCompleted;
 }
--- a/packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
+++ b/packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
@@ -77,15 +77,6 @@ class S3Exception extends StorageException {
     );
   }
 
-  /// An exception created when attempt to resume a canceled operation.
-  factory S3Exception.resumeCanceledOperation() {
-    return const S3Exception(
-      'The operation has been canceled and can\'t be resumed.',
-      recoverySuggestion:
-          'You can only resume an operation that is paused by calling `pause()`.',
-    );
-  }
-
   /// An exception thrown when the service doesn't return a valid uploadId on a
   /// successful `CreateMultipartUploadRequest`.
   factory S3Exception.unexpectedMultipartUploadId() {
--- a/packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
+++ b/packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
@@ -59,16 +59,6 @@ class S3Exception extends StorageException {
     );
   }
 
-  /// Creates a [S3Exception] that represents an expected exception
-  /// thrown when invoke cancel() on an eligible storage operation object.
-  factory S3Exception.controllableOperationCanceled() {
-    return const S3Exception(
-      'The requested operation has been canceled.',
-      recoverySuggestion:
-          'This is an expected exception when you call cancel() API on the storage operation object. You need to handle this exception to take further action on operation cancellation.',
-    );
-  }
-
   /// An exception thrown when the `body` returned from GetObject is unexpectedly null.
   factory S3Exception.unexpectedGetObjectBody() {
     return const S3Exception(
@@ -228,3 +218,11 @@ class S3Exception extends StorageException {
   @override
   String get runtimeTypeName => 'S3Exception';
 }
+
+/// The exception thrown when cancel a controllable operation.
+const s3ControllableOperationCanceledException =
+    StorageOperationCanceledException(
+  'The operation has been canceled.',
+  recoverySuggestion:
+      'This is an expected exception when you call cancel() API on the storage operation object. You need to handle this exception to take further action on operation cancellation.',
+);
--- a/packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
+++ b/packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
@@ -5,185 +5,70 @@
 import 'package:amplify_core/amplify_core.dart';
 import 'package:amplify_storage_s3_dart/src/sdk/s3.dart' as s3;
 import 'package:meta/meta.dart';
-import 'package:smithy/smithy.dart';
+import 'package:smithy/smithy.dart' as smithy;
 
 const _keyNotFoundRecoveryMessage =
-    'Please ensure that correct object key is provided, and/or correct `StorageAccessLevel` and `targetIdentityId` are included in the options.';
+    'Ensure that correct object key is provided, and/or correct `StorageAccessLevel`'
+    ' and `targetIdentityId` are included in the options.';
 const _httpErrorRecoveryMessage =
-    'HTTP error returned from service, please review the `underlyingException` for details.';
+    'HTTP error returned from service, review the `underlyingException` for details.';
 
-/// {@template amplify_storage_s3_dart.s3_storage_exception}
-/// Represents exceptions that may be thrown calling Storage S3 plugin APIs.
-/// {@endtemplate}
-class S3Exception extends StorageException {
-  /// {@macro amplify_storage_s3_dart.s3_storage_exception}
-  const S3Exception(
-    super.message, {
-    required String super.recoverySuggestion,
-    super.underlyingException,
-  });
+/// The exception thrown when cancel a controllable operation.
+const s3ControllableOperationCanceledException =
+    StorageOperationCanceledException(
+  'The operation has been canceled.',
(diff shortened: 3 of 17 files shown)


packages/amplify_core/lib/src/types/exception/storage/operation_canceled_exception.dart
```diff
@@ -0,0 +1,19 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'package:amplify_core/amplify_core.dart';
+
+/// {@template amplify_core.storage.operation_canceled_exception}
+/// Exception thrown when a storage operation is canceled.
+/// {@endtemplate}
+class StorageOperationCanceled extends StorageException {
```
fjnoyp, line 9: I think we should come together and formalize our policy for category exception naming.  

In Analytics I didn't prefix exceptions with 'Analytics'. 

For Auth, it seems we do and we don't sometimes.  For example: 

ie. InvalidStateException, UserCancelledException, etc.
but not AuthValidationException, etc. 

Perhaps I'm missing why we do/don't prefix with Auth? 

Regardless isn't it more clear to always prefix with Auth/Analytics/Storage? 

Thoughts? 
 @dnys1 @Jordan-Nelson

packages/amplify_core/lib/src/types/storage/transfer_progress.dart
```diff
@@ -1,32 +1,51 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-/// {@template amplify_storage_s3.transfer_progress}
+/// {@template amplify_core.transfer_state}
```
fjnoyp, line 4: nit: 
amplify_core.storage.transfer_state

Same for other comments.

packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
```diff
@@ -12,228 +11,69 @@ const _keyNotFoundRecoveryMessage =
...
+  recoverySuggestion:
+      'This is an expected exception when you call cancel() API on the storage operation object. You need to handle this exception to take further action on operation cancellation.',
+);
+
+/// Creates a [StorageKeyNotFoundException] with the [underlyingException].
+StorageKeyNotFoundException createKeyNotFoundException(
+  Exception underlyingException,
+) =>
+    StorageKeyNotFoundException(
+      'Key is not found.',
```
fjnoyp, line 26: Should we specify what specific key this is? 

Such as specifying this is the key for the storage item, like:

Cannot find key for the storage item. 

The storage item specified with the provided key doesn't exist.

packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
```diff
@@ -12,228 +11,69 @@ const _keyNotFoundRecoveryMessage =
...
-    return const S3Exception(
-      'Create multipart upload request succeeded, but it did not return a valid uploadId.',
-      recoverySuggestion: AmplifyExceptionMessages.missingExceptionMessage,
-    );
-  }
+/// Creates a [NetworkException] with the [AWSHttpException].
+NetworkException createNetworkExceptionFromAWSHttpException(
+  AWSHttpException exception,
+) {
+  return NetworkException(
```
fjnoyp, line 35: Since NetworkException will be exact same between Analytics, Storage and perhaps other categories, can we specify the exact NetworkException text in core?

packages/storage/amplify_storage_s3_dart/lib/src/model/s3_item.dart
```diff
@@ -44,7 +43,10 @@ class S3Item extends StorageItem {
 
     // In S3 plugin, key is required property presenting an object
     if (key == null) {
-      throw S3Exception.getS3ObjectMissingKeyException(object);
+      throw const UnknownException(
```
fjnoyp, line 46: For comment above, did you mean: 

    // In S3 plugin, key is required property presenting an object

    // In S3 plugin, key is required property present in object

packages/storage/amplify_storage_s3_dart/lib/src/model/s3_item.dart
```diff
@@ -44,7 +43,10 @@ class S3Item extends StorageItem {
...
+        'Object key is unexpectedly missing in a S3Object.',
```
fjnoyp, line 47: Nit: 

S3Object does not contain a key.

packages/storage/amplify_storage_s3_dart/lib/src/platform_impl/download_file/download_file_io.dart
```diff
@@ -101,12 +100,20 @@ Future<String> _ensureDestinationWritable(AWSFile file) async {
   final destinationPath = file.path;
 
   if (destinationPath == null) {
-    throw S3Exception.downloadDestinationFilePathIsNull;
+    throw const UnknownException(
+      'Download destination file path is null.',
+      recoverySuggestion:
+          'Ensure a valid file path with `AWSFile` was provided.',
```
fjnoyp, line 106: nit: Ensure your `AWSFile` contains a valid file path.

packages/storage/amplify_storage_s3_dart/lib/src/platform_impl/download_file/download_file_io.dart
```diff
@@ -116,7 +123,11 @@ Future<String> _ensureDestinationWritable(AWSFile file) async {
       await destination.delete();
     }
   } on FileSystemException catch (error) {
-    throw S3Exception.fromFileSystemException(error);
+    throw UnknownException(
+      'Expected file system error',
```
fjnoyp, line 127: Should we say: Encountered a file system error?

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
```diff
@@ -482,11 +486,13 @@ class StorageS3Service {
               : const S3CopyPluginOptions(),
         ),
       );
-    } on S3Exception catch (error) {
-      throw S3Exception(
+    } on StorageException catch (error) {
+      // Normally copy should not fail during moving, if it fails it's a
+      // "unknown" failure wrapping the underlying exception.
+      throw UnknownException(
         'Copy the source object failed while moving it due to: ${error.message}',
```
fjnoyp, line 493:
```suggestion
        'Moving object during Copy operation failed due to: ${error.message}',
```

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
```diff
@@ -505,11 +511,13 @@ class StorageS3Service {
         bucket: _defaultBucket,
         key: keyToRemove,
       );
-    } on S3Exception catch (error) {
-      throw S3Exception(
+    } on StorageException catch (error) {
+      // Normally delete should not fail during moving, if it fails it's a
+      // "unknown" failure wrapping the underlying exception.
+      throw UnknownException(
         'Delete the source object failed while moving it due to: ${error.message}',
```
fjnoyp, line 518:
```suggestion
        'Moving object during Delete operation failed due to: ${error.message}',
```

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
```diff
@@ -665,7 +675,7 @@ class StorageS3Service {
       );
     } on Exception catch (error, st) {
       logger.error('Error happened while resolving prefix', error, st);
-      throw S3Exception(
+      throw UnknownException(
         'Error happened while resolving prefix.',
```
fjnoyp, line 679:
```suggestion
        'Error while resolving prefix.',
```

packages/storage/amplify_storage_s3_dart/lib/src/platform_impl/download_file/download_file_io.dart
fjnoyp, line 106:
```suggestion
          'Ensure your `AWSFile` contains a valid file path.',
```

packages/storage/amplify_storage_s3_dart/lib/src/platform_impl/download_file/download_file_io.dart
fjnoyp, line 127:
```suggestion
      'Encountered file system error',
```

packages/storage/amplify_storage_s3_dart/lib/src/model/s3_item.dart
fjnoyp, line 47:
```suggestion
        'S3Object does not contain a key.',
```

packages/amplify_core/lib/src/types/exception/storage/operation_canceled_exception.dart
HuiSF: We added the prefix mainly to align with amplify-swift, although Swift can do an enum over an error type, e.g.

```
StorageError.unknown
StorageError.keyNotFound
```

We don't have this support on the same pattern (yet) for caching the specific exception types, so we have to keep them separate, but in the meantime, we still want to maintain the grouping.

packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
HuiSF: [Added an extension to `AWSHttpException`](https://github.com/aws-amplify/amplify-flutter/pull/2781/commits/8925c5f93455136eaadb0c60077e2e46b9933e23) to create a `NetworkException` with consistent messaging. Please let me know if other alternatives are preferred.

packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
```diff
@@ -5,235 +5,82 @@
...
-  factory S3Exception.unexpectedMultipartUploadId() {
-    return const S3Exception(
-      'Create multipart upload request succeeded, but it did not return a valid uploadId.',
-      recoverySuggestion: AmplifyExceptionMessages.missingExceptionMessage,
+/// The exception thrown when cancel a controllable operation.
+const s3ControllableOperationCanceledException = StorageOperationCanceled(
+  'The operation has been canceled.',
+  recoverySuggestion:
+      'This is an expected exception when you call cancel() API on the storage operation object. You need to handle this exception to take further action on operation cancellation.',
+);
```
NikaHsn, line 20: why return an exception if the cancel API was called and completed successfully?
in code we map `CancellationException` to `s3ControllableOperationCanceledException`. `CancellationException` is thrown when an operation is canceled before completion, which I think is due to some issue and not because of expected behavior of calling cancel API.

HuiSF: Looking at the `operation.result` type `Future<Result>`, where `Result` is non-nullable.

If we don't throw exception, but complete the operation successfully on `cancel()` then `Result` has to be nullable, and return `null`, where developer would need a null-check even on a really successful operation.

packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
```diff
@@ -5,235 +5,82 @@
...
-
-  /// Creates [StorageException] variants from [UnknownSmithyHttpException].
-  static StorageException fromUnknownSmithyHttpException(
-    UnknownSmithyHttpException exception,
-  ) {
-    final statusCode = exception.statusCode;
-
-    if (statusCode ~/ 100 == 3 || statusCode ~/ 100 == 5) {
+    } else if (statusCode == 404) {
+      return exception.toStorageKeyNotFoundException();
```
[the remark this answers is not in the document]
HuiSF: Agreed, I can't remember why I used the extension here...

packages/amplify_core/lib/src/types/exception/storage/operation_canceled_exception.dart
Jordan-Nelson: If I understand correctly, this exception is thrown when a `StorageOperation` is cancelled. If it was thrown when an `UploadOperation` was canceled it might be called an `UploadOperationCanceledException`. So I am not sure that the `Storage` prefix is really coming from the category in this case. I think it is describing the the operation.

It would be nice to follow a common pattern. I don't think we need to hold this PR up for it, but we can continue discussion.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_upload_task.dart
```diff
@@ -661,11 +676,14 @@ class S3UploadTask {
...
-      throw S3Exception.fromS3NoSuchUpload(error);
+      throw s3_exception
+          .createStorageExceptionFromUnknownSmithyHttpException(error);
     } on AWSHttpException catch (error) {
-      throw S3Exception.fromAWSHttpException(error);
+      throw error.toNetworkException();
+    } on s3.NoSuchUpload {
+      // this exception will be handled within the `on Exception` block
+      // in the function _handleUploadPart below
+      rethrow;
```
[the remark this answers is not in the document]
dnys1: If the exception is already being handled above, there's no need to catch and rethrow it. I think it only confuses things.

dnys1 (approved):

fjnoyp (approved):
Looks good, per Jordan's comments we'll deal with Exception naming prefixes in another PR.

packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
```diff
@@ -5,235 +5,73 @@
...
-  );
-
-  /// Creates a [StorageKeyNotFoundException] from an [Exception]
-  static StorageKeyNotFoundException getKeyNotFoundException(
-    Exception underlyingException,
-  ) =>
-      StorageKeyNotFoundException(
-        'Key is not found.',
+  if (statusCode ~/ 100 == 4) {
+    if ([401, 403].contains(statusCode)) {
```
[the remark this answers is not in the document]
HuiSF: Yea was thinking the possibility to expand the collection of status code, but for this use case it probably never going to expand 😅 updated.

PR #2791, opened 2023-03-28
fix(auth): User pool-only support

dnys1:
Fixes up support for user pool-only/identity pool-only modes and adds e2e tests for each.


packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_in_state_machine.dart
```diff
@@ -735,9 +732,7 @@ class SignInStateMachine extends AuthStateMachine<SignInEvent, SignInState> {
     if (authenticationResult != null) {
       final accessToken = await _saveAuthResult(authenticationResult);
       final newDeviceMetadata = authenticationResult.newDeviceMetadata;
-      if (newDeviceMetadata != null &&
-          // ConfirmDevice API requires an identity pool
-          hasIdentityPool) {
```
NikaHsn, line 740: Q: no need to check for `identityPoolConfig != null` anymore?

NikaHsn (approved):

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_in_state_machine.dart
dnys1: Nope not needed. If there's no identity pool, it will throw an exception later but that's okay. It's handled by the new changes to the Cognito SDK allowing optional auth.

Jordan-Nelson (approved):

PR #2793, opened 2023-03-28
feat(storage): make access level optional for storage APIs

NikaHsn:
*Issue #, if available:*

*Description of changes:*
- make access level optional for storage APIs
- `defaultAccessLevel` can be set by `Amplify.Amplify.configure(...)` with fallback to `guest`. 

### DIFF
--- a/packages/amplify_core/lib/src/types/storage/base/storage_operation_options.dart
+++ b/packages/amplify_core/lib/src/types/storage/base/storage_operation_options.dart
@@ -5,8 +5,8 @@ import 'package:amplify_core/src/types/storage/access_level.dart';
 
 abstract class StorageOperationOptions {
   const StorageOperationOptions({
-    required this.accessLevel,
+    this.accessLevel,
   });
 
-  final StorageAccessLevel accessLevel;
+  final StorageAccessLevel? accessLevel;
 }
--- a/packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
+++ b/packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
@@ -29,20 +29,18 @@ const testDateTimeNowOverride = #_testDateTimeNowOverride;
 class StorageS3Service {
   /// {@macro amplify_storage_s3.storage_s3_service}
   StorageS3Service({
-    required String defaultBucket,
-    required String defaultRegion,
+    required S3PluginConfig s3PluginConfig,
     required S3PrefixResolver prefixResolver,
     required AWSIamAmplifyAuthProvider credentialsProvider,
     required AWSLogger logger,
     String? delimiter,
     required DependencyManager dependencyManager,
-  })  : _defaultBucket = defaultBucket,
-        _defaultRegion = defaultRegion,
+  })  : _s3PluginConfig = s3PluginConfig,
         _delimiter = delimiter ?? '/',
         // dependencyManager.get() => S3Client is used for unit tests
         _defaultS3Client = dependencyManager.get() ??
             s3.S3Client(
-              region: defaultRegion,
+              region: s3PluginConfig.region,
               credentialsProvider: credentialsProvider,
               s3ClientConfig: _defaultS3ClientConfig,
               client: AmplifyHttpClient()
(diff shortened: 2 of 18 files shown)


packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
```diff
@@ -99,7 +96,7 @@ class StorageS3Service {
     final resolvedPrefix = await getResolvedPrefix(
       prefixResolver: _prefixResolver,
       logger: _logger,
-      accessLevel: options.accessLevel,
+      accessLevel: options.accessLevel ?? _s3PluginConfig.defaultAccessLevel,
```
HuiSF, line 99: I understand this `??` `if null` check is due to `Storage<Operation>Options.accessLevel` is nullable now, and its default value should be set by the plugin, but this default value has been set [here](https://github.com/aws-amplify/amplify-flutter/pull/2793/files#diff-3b6ec49d2366c0c7278efaa1cfc1c8f3d9b845c1f4fcca468113e3bb15cf9b08R172-R175) in the StorageS3Dart implementation (where I think it's a proper location). 

Hum we can't find a better way than `accessLevel: options.accessLevel!` we probably need to remove the default value setting from the StorageS3Dart implementation.

NikaHsn: using the default value here instead of `accessLevel: options.accessLevel!` makes the code more robust as we do not rely on callers to set value for nullable field.

dnys1: I agree. The storage options interface is not appropriate since it has a nullable access level. Either passing a non-null access level as an additional parameter to the service (e.g. `resolvedAccessLevel`) or handling it inline like this seems fine.

Since other S3 plugin options logic lives here, I think the current handling is appropriate.

packages/amplify_core/lib/src/types/storage/base/storage_operation_options.dart
```diff
@@ -5,8 +5,8 @@ import '../access_level.dart';
 
 abstract class StorageOperationOptions {
   const StorageOperationOptions({
-    required this.accessLevel,
+    this.accessLevel,
   });
 
-  final StorageAccessLevel accessLevel;
+  final StorageAccessLevel? accessLevel;
```
dnys1, line 11: I might recommend adding the following and using throughout.

```dart
StorageAccessLevel resolveAccessLevel(StorageAccessLevel defaultAccessLevel) =>
  accessLevel ?? defaultAccessLevel;
```

dnys1 (approved):

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
NikaHsn: @dnys1 thanks for suggesting the `resolvedAccessLevel`. I'm more toward handling it inline as there is no logic than checking for null.

HuiSF (approved):

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/storage_s3_service_test.dart
```diff
@@ -390,6 +392,47 @@ void main() {
         );
       });
 
+      test('should invoke S3Client.headObject with default access level',
```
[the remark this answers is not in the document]
NikaHsn: I think the key passed to headObject is prefixed with access level, something like `public#object-key`

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/storage_s3_service_test.dart
```diff
@@ -261,7 +263,7 @@ void main() {
       });
 
       test(
-          'should invoke S3Object.listObjectV2 with expected parameters and return expected results with listAll is set to true in the options',
+          'should invoke S3Object.listObjectV2 with expected parameters and return expected results with listAll and acess level is set to true in the options',
```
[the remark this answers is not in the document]
NikaHsn: nope, will update

PR #2795, opened 2023-03-29
fix(api): Include owner field in selection set

Equartey:
*Description of changes:*
An issue was observed when using the following auth rule `@auth(rules: [{allow: owner}])` and model helper methods. Subscriptions could not receive data events without the triggering mutation including the `owner` field in its selection set. 

This PR adds the owner field to the selection set when the auth rule is present in the model schema.  


ragingsquirrel3 (approved):

dnys1 (approved):
Please add an e2e test as well

ragingsquirrel3 (approved):

PR #2796, opened 2023-03-29
fix(auth): Process sign-in events once

### LINKED ISSUE #2794: [Flutter] Amplify 1.0.0-next.7 custom auth flow and set new password (confirm new password after temporary) causes errors
### Description

Framework: Flutter

Upon upgrading the development version from 1.0.0-next.6 to 1.0.0-next.7 there were a couple of issues that started appearing.

1. Passwordless custom authentication flow 

I implemented a custom authentication flow with the CreateAuthChallenge, DefineAuthChallenge and VerifyAuthChallenge lambda triggers. Upon upgrading the version of the dart libraries, it seems like upon failing a challenge, there would be a repeated sending of the same incorrect challenge code afterwards to Cognito until either a limit was hit (in our case we set 3 attempts maximum) or infinitely (if there was no maximum limit set). This would cause the challenge to end and prevent the user from submitting any more attempts.

2. Setting a new password after using the `admin_create_user` API

If a new user was created using the `admin_create_user` API, we use `Amplify.Auth.signIn(username: username, password: temporaryPassword` to confirm the temporary password and then `Amplify.Auth.confirmSignIn(confirmationValue: newPassword)` to set their new password. However, upon upgrading the libraries to 1.0.0-next.7, the second call would still work (set the new password and send an MFA code if necessary) but then raise an error of type `CodeMismatchException`.

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [X] Auth
- [ ] Authenticator
- [ ] DataStore
- [ ] Storage

### Steps to Reproduce

_No response_

### Screenshots

_No response_

### Platforms

- [ ] iOS
- [ ] Android
- [X] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Flutter Version

3.7.8

### Amplify Flutter Version

1.0.0-next.7

### Deployment Method

Amplify CLI

### Schema

_No response_

dnys1:
Fixes #2794.

- Fixes a regression caused by a recent change ([8d87ea8#diff-9d1c23e771518da5d10ac1f888bd38c5f865868332b126a16ab6d19fd3ef028fR825](https://github.com/aws-amplify/amplify-flutter/commit/8d87ea871b2d8a716f0f7f087e2660b7a8e61994#diff-9d1c23e771518da5d10ac1f888bd38c5f865868332b126a16ab6d19fd3ef028fR653))
- Improves sign-in exception handling logic (now aligned with iOS/Android)
- Adds more custom auth tests

### DIFF
--- a/packages/auth/amplify_auth_cognito_dart/lib/src/state/event/sign_in_event.dart
+++ b/packages/auth/amplify_auth_cognito_dart/lib/src/state/event/sign_in_event.dart
@@ -136,8 +136,29 @@ class SignInRespondToChallenge extends SignInEvent {
 
   @override
   PreconditionException? checkPrecondition(SignInState currentState) {
-    if (currentState is SignInFailure &&
-        currentState.exception is CodeMismatchException) {
+    if (currentState is SignInFailure) {
+      // If the state machine did not reach a challenge state before the
+      // exception was thrown, then any attempts to confirm sign in should
+      // not be allowed and the entire flow must be restarted by the user.
+      if (currentState.previousState.type != SignInStateType.challenge) {
+        return AuthPreconditionException(
+          'Sign-in previously failed',
+          recoverySuggestion:
+              'Restart the sign-in flow by calling Amplify.Auth.signIn',
+          underlyingException: currentState.exception,
+        );
+      }
+      // However, if the state machine was in a respond to challenge state when
+      // the exception was raised, further attempts to confirm sign in should
+      // be allowed.
+      //
+      // Recoverable exceptions are expected at this stage, such as
+      // CodeMismatchException and InvalidPasswordException. These exceptions
+      // prompt the user to retry if they've provided an invalid code or weak
+      // password, for example. It is not feasible to determine which exception
+      // types are non-recoverable, though. For all others, it is better to
+      // allow confirm sign in attempts and have Cognito continue to throw the
+      // exception.
       return null;
     }
     if (currentState.type != SignInStateType.challenge) {
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,6 +1,11 @@
 {
   "configurations": [
     {
+      "name": "Run",
+      "type": "dart",
+      "request": "launch",
+    },
+    {
       "name": "Dart Test in Chrome",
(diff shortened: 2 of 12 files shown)


packages/auth/amplify_auth_cognito_dart/lib/src/state/event/sign_in_event.dart
```diff
@@ -136,8 +136,29 @@ class SignInRespondToChallenge extends SignInEvent {
...
+      if (currentState.previousState.type != SignInStateType.challenge) {
+        return AuthPreconditionException(
+          'Sign-in previously failed',
+          recoverySuggestion:
+              'Restart the sign-in flow by calling Amplify.Auth.signIn',
+          underlyingException: currentState.exception,
+        );
+      }
+      // However, if the state machine was in a respond to challenge state when
+      // the exception wass raised, further attempts to confirm sign in should
```
Jordan-Nelson, line 152:
```suggestion
      // the exception was raised, further attempts to confirm sign in should
```

packages/auth/amplify_auth_cognito_dart/lib/src/state/event/sign_in_event.dart
```diff
@@ -136,8 +136,29 @@ class SignInRespondToChallenge extends SignInEvent {
...
+      // be allowed.
+      //
+      // Recoverable exceptions are expected at this stage, such as
```
Jordan-Nelson, line 155: Would this be more straightforward if confirmSignIn were split into multiple methods (confirmSignInMFA, confirmSignInCustomAuth, and confirmSignInNewPassword)?

Jordan-Nelson (approved):

packages/auth/amplify_auth_cognito_dart/lib/src/state/event/sign_in_event.dart
dnys1: On the surface, yes. But it's best that we don't prescribe retryability to exception types and instead let Cognito either allow the retry or rethrow the same exception.

ragingsquirrel3 (approved):

PR #2804, opened 2023-03-31
refactor(auth)!: Remove generic types

dnys1:
Removes generic types from the category/plugin interface and from all data classes. This fixes JSON serialization issues and cleans up the API, preventing accidental misuse of types and making the source code more accessible for first and third-party developers.

In the past, generic types were necessary since plugin keys returned category wrappers over plugins. The category had to remain generic over any plugin and the plugin key carried the generic types to make the category wrapper statically resemble the plugin. Now that the category and plugin have the exact same interface and the plugin key directly returns the plugin, there is no need for generics.


packages/amplify_core/lib/src/types/auth/attribute/cognito_user_attribute_key.dart
```diff
@@ -16,17 +16,20 @@ import 'package:meta/meta.dart';
 class CognitoUserAttributeKey extends AuthUserAttributeKey
```
NikaHsn, line 16: why do we have Cognito types in core? I recommend moving the category level behaviors to `AuthUserAttributeKey` and those that are specific to Cognito out of core and into the Auth package.

packages/amplify_core/lib/src/types/auth/session/auth_user.dart
```diff
@@ -3,36 +3,14 @@
 
 import 'package:amplify_core/amplify_core.dart';
 
-part 'auth_user.g.dart';
-
-@zAmplifyGenericSerializable
-class AuthUser<Details extends SignInDetails>
+abstract class AuthUser
```
NikaHsn, line 6: why to change to abstract?

dnys1: This is to match iOS/Android

packages/amplify_core/lib/src/types/auth/attribute/cognito_user_attribute_key.dart
dnys1: The Cognito configuration (and all plugin configurations) are in `core` and this is needed as part of it.

> category level behaviors to AuthUserAttributeKey

This has already been done unless I missed something?

packages/amplify_core/lib/src/types/auth/session/auth_user.dart
NikaHsn: `AuthUser` is an interface in iOS and a final class in Android. I think it make sense to not change it to abstract because it is a data class with no behaviour.

packages/amplify_core/lib/src/types/auth/attribute/cognito_user_attribute_key.dart
NikaHsn: this is right, I pulled the PR and looked at the code in details. now I understand the structure and how config is done in core package.

packages/amplify_core/lib/src/types/auth/session/auth_user.dart
dnys1: It is a template for a data class given that plugins would extend it and override the `SignInDetails` type. I'm okay making it instantiable in this case.

packages/amplify_core/lib/src/types/auth/attribute/auth_user_attribute_key.dart
```diff
@@ -15,7 +16,11 @@ abstract class AuthUserAttributeKey
     with AWSSerializable<String>
     implements Comparable<AuthUserAttributeKey> {
   /// {@macro amplify_core.auth_user_attribute_key}
-  const AuthUserAttributeKey();
+  const factory AuthUserAttributeKey(String key) = _AuthUserAttributeKey;
```
Jordan-Nelson, line 19: I'm not sure I understand the update here. Why add the factory constructor? I see it used in the converter, but you could use _AuthUserAttributeKey directly there, no?

Do we want to expose this for customers?

Jordan-Nelson (approved):

packages/amplify_core/lib/src/types/auth/session/auth_user.dart
```diff
@@ -5,34 +5,23 @@ import 'package:amplify_core/amplify_core.dart';
...
 
-@zAmplifyGenericSerializable
-class AuthUser<Details extends SignInDetails>
-    with AWSSerializable<Map<String, Object?>>, AWSDebuggable {
+@zAmplifySerializable
+class AuthUser with AWSSerializable<Map<String, Object?>>, AWSDebuggable {
   const AuthUser({
     required this.userId,
     required this.username,
-    required this.signInDetails,
```
NikaHsn, line 14: Q: why are we changing this?

NikaHsn (approved):

packages/amplify_core/lib/src/types/auth/session/auth_user.dart
dnys1: There are tradeoffs to having abstract types in data classes (SignInDetails). Generics are out. If this is included, a `fromJson` factory cannot be created since SignInDetails has no concrete type to call. I'll add it back to keep the signature the same and remove the `fromJson` factory.

dnys1: Actually, I'd rather keep it this way. I want to refine the type on `CognitoAuthUser` which means overriding the `signInDetails` field. Overriding a field with a field is a no-no, but overriding a getter with a field is okay.

packages/amplify_core/lib/src/types/auth/attribute/auth_user_attribute_key.dart
dnys1: Good catch - removed

PR #2809, opened 2023-04-03
chore(core): configurable analytics flush interval

fjnoyp:
*Issue #, if available:*

*Description of changes:*

Configurable analytics flush interval in amplifyConfiguration.dart

### DIFF
--- a/packages/amplify_core/lib/src/config/analytics/pinpoint_config.dart
+++ b/packages/amplify_core/lib/src/config/analytics/pinpoint_config.dart
@@ -50,15 +50,7 @@ class PinpointPluginConfig
 
   final int _autoFlushEventsInterval;
 
-  /// The duration in seconds (rounded up) between flushing analytics events to Pinpoint.
-  ///
-  /// A max of 100 events are included in each request. If more than 100 events
-  /// are recorded per interval, they will be temporarily queued and sent in a
-  /// later request. Consider reducing the duration for use cases where events
-  /// are recorded at a high rate for sustained periods of time.
-  ///
-  /// *Note:* The duration will be rounded down to the nearest whole second when
-  /// this value is serialized to json.
+  /// The duration in seconds between flushing analytics events to Pinpoint.
   @_DurationConverter()
   Duration get autoFlushEventsInterval =>
       Duration(seconds: _autoFlushEventsInterval);
--- a/packages/amplify_core/lib/src/config/analytics/pinpoint_config.dart
+++ b/packages/amplify_core/lib/src/config/analytics/pinpoint_config.dart
@@ -35,6 +35,7 @@ class PinpointPluginConfig
   const PinpointPluginConfig({
     required this.pinpointAnalytics,
     required this.pinpointTargeting,
+    this.autoFlushEventsInterval = 10000,
   });
 
   factory PinpointPluginConfig.fromJson(Map<String, Object?> json) =>
@@ -45,6 +46,7 @@ class PinpointPluginConfig
 
   final PinpointAnalytics pinpointAnalytics;
   final PinpointTargeting pinpointTargeting;
+  final int autoFlushEventsInterval;
 
   @override
   List<Object?> get props => [pinpointAnalytics, pinpointTargeting];
@@ -52,10 +54,13 @@ class PinpointPluginConfig
   PinpointPluginConfig copyWith({
     PinpointAnalytics? pinpointAnalytics,
     PinpointTargeting? pinpointTargeting,
+    int? autoFlushEventsInterval,
   }) {
     return PinpointPluginConfig(
       pinpointAnalytics: pinpointAnalytics ?? this.pinpointAnalytics,
       pinpointTargeting: pinpointTargeting ?? this.pinpointTargeting,
+      autoFlushEventsInterval:
+          autoFlushEventsInterval ?? this.autoFlushEventsInterval,
     );
   }
 
--- a/packages/amplify_core/lib/src/config/analytics/pinpoint_config.dart
+++ b/packages/amplify_core/lib/src/config/analytics/pinpoint_config.dart
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import 'package:amplify_core/amplify_core.dart';
+import 'package:json_annotation/json_annotation.dart';
 import 'package:meta/meta.dart';
 
 part 'pinpoint_config.g.dart';
@@ -35,7 +36,7 @@ class PinpointPluginConfig
   const PinpointPluginConfig({
     required this.pinpointAnalytics,
     required this.pinpointTargeting,
-    this.autoFlushEventsInterval = 10000,
+    this.autoFlushEventsInterval = const Duration(milliseconds: 10000),
   });
 
   factory PinpointPluginConfig.fromJson(Map<String, Object?> json) =>
@@ -46,15 +47,18 @@ class PinpointPluginConfig
 
   final PinpointAnalytics pinpointAnalytics;
   final PinpointTargeting pinpointTargeting;
-  final int autoFlushEventsInterval;
+
+  @DurationConverter()
+  final Duration autoFlushEventsInterval;
 
   @override
-  List<Object?> get props => [pinpointAnalytics, pinpointTargeting];
+  List<Object?> get props =>
+      [pinpointAnalytics, pinpointTargeting, autoFlushEventsInterval];
 
   PinpointPluginConfig copyWith({
     PinpointAnalytics? pinpointAnalytics,
     PinpointTargeting? pinpointTargeting,
-    int? autoFlushEventsInterval,
+    Duration? autoFlushEventsInterval,
   }) {
     return PinpointPluginConfig(
       pinpointAnalytics: pinpointAnalytics ?? this.pinpointAnalytics,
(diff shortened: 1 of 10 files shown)


packages/amplify_core/lib/src/config/analytics/pinpoint_config.dart
```diff
@@ -45,17 +46,21 @@ class PinpointPluginConfig
 
   final PinpointAnalytics pinpointAnalytics;
   final PinpointTargeting pinpointTargeting;
+  final int autoFlushEventsInterval;
 
   @override
   List<Object?> get props => [pinpointAnalytics, pinpointTargeting];
```
dnys1, line 52:
```suggestion
  final int _autoFlushEventsInterval;
  Duration get autoFlushEventsInterval => Duration(milliseconds: _autoFlushEventsInterval);

  @override
  List<Object?> get props => [pinpointAnalytics, pinpointTargeting, _autoFlushEventsInterval];
```

fjnoyp:
Hi @dnys1  good point about weirdness regarding millisecond units.  

Given that performance will be determined by platform, hardware, and internet speeds, I'm unsure what information a stress test will provide.  Ultimately with milliseconds units, the customer could feasibly set it to 1 millisecond which we'd expect to break whatever code we could put in place.  

I'm also unsure why we use millisecond units to begin with.  My approach here was to follow existing supported behavior.  I can check with native teams why as well, but unfortunately they are likely following existing behavior that is lost to time. 

I'll ask Abdallah to clarify since he's decided to steer the direction of Analytics.  I'll keep the existing code as is for now and then update in the PR what next steps we agree on.

dnys1:
How is it a breaking change if iOS was already using seconds? Where did we get milliseconds from?

fjnoyp:
> How is it a breaking change if iOS was already using seconds? Where did we get milliseconds from? Regardless, seconds makes more sense on just about every level. I would suggest adding some error handling as well for negative/zero values.

The field name 'autoFlushEventInterval' is the same between iOS and Android.  Therefore, when using the same configuration iOS will treat the number as seconds while Android will treat as milliseconds.  So perhaps not a breaking change then.  

Milliseconds is what Android uses but as agreed yesterday we'll use seconds like iOS. 

Following iOS, for zero values event flush is disabled.  I did the same for negative values as well.  Will update Analytics docs as well.

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -153,11 +153,14 @@ class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
       },
     );
 
-    _autoEventSubmitter = StoppableTimer(
-      duration: const Duration(seconds: 10),
-      callback: flushEvents,
-      onError: (e) => _logger.warn('Exception in events auto flush', e),
-    );
+    final autoFlushInterval = pinpointConfig.autoFlushEventsInterval;
+    if (autoFlushInterval.inSeconds > 0 || !autoFlushInterval.isNegative) {
```
dnys1, line 157: Huh? These conditions check the same thing, no? And shouldn't this use the default if it's invalid. This would mean it never auto flushes and the user gets no indication if I'm reading this correctly. Better to throw a `ConfigurationError`, I think, or print a warning and fallback to default.

packages/amplify_core/lib/src/config/analytics/pinpoint_config.dart
```diff
@@ -35,6 +36,7 @@ class PinpointPluginConfig
   const PinpointPluginConfig({
     required this.pinpointAnalytics,
     required this.pinpointTargeting,
+    this.autoFlushEventsInterval = const Duration(seconds: 30),
```
Jordan-Nelson, line 39: What is the current default in 0.6.x?

packages/amplify_core/lib/src/config/analytics/pinpoint_config.dart
```diff
@@ -119,3 +128,13 @@ class PinpointTargeting with AWSEquatable<PinpointTargeting>, AWSSerializable {
...
 }
+
+class DurationConverter implements JsonConverter<Duration, int> {
+  const DurationConverter();
+
+  @override
+  Duration fromJson(int json) => Duration(seconds: json);
+
+  @override
+  int toJson(Duration object) => object.inSeconds;
```
Jordan-Nelson, line 139: This will round down to the nearest whole second. Is that what we want? If so, can we document that on `autoFlushEventsInterval`? If a customer created their amplify config using `AmplifyConfig` class, they might be surprised to see to see it rounded.

packages/amplify_core/lib/src/config/analytics/pinpoint_config.dart
fjnoyp: For 0.6.x, we rely on Android and iOS.  On Android it is 30 seconds, on iOS it is 60 seconds.  

I asked the teams why it was 30 or 60 seconds and no one knew... 

I spoke with Abdallah this morning about this and we confirmed together that 30 seconds should be okay.  I can share the reasoning here if you'd like.

packages/amplify_core/lib/src/config/analytics/pinpoint_config.dart
```diff
@@ -46,16 +48,26 @@ class PinpointPluginConfig
   final PinpointAnalytics pinpointAnalytics;
   final PinpointTargeting pinpointTargeting;
 
+  /// The duration in seconds (rounded up) between flushing analytics events to Pinpoint.
+  ///
+  /// Max events per flush is 100.  Decrease duration to ensure you flush more events than you store over time.
```
Jordan-Nelson, line 53: > (rounded up)

I believe it will be rounded down, not up. The rounding behavior is only applicable if this is serialized to JSON.

> Max events per flush is 100. Decrease duration to ensure you flush more events than you store over time.

I'm not sure that you need to include this, but if you do, this probably needs a bit more detail. I could see someone not understanding that events will queue on disk. The duration only needs to be decreased if there will be a *sustained* high rate of events. Small bursts at a high rate would likely be fine for most use cases as they would just queue on disk and be sent shortly after.

```suggestion
  /// The duration between flushing analytics events to Pinpoint.
  ///
  /// A max of 100 events are included in each request. If more than 100 events 
  /// are recorded per interval, they will be temporarily queued and sent in a 
  /// later request. Consider reducing the duration for use cases where events 
  /// are recorded at a high rate for sustained periods of time.
  ///
  /// *Note:* The duration will be rounded down to the nearest whole second when 
  /// this value is serialized to json.
```

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -153,11 +153,22 @@ class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
...
     );
 
-    _autoEventSubmitter = StoppableTimer(
-      duration: const Duration(seconds: 10),
-      callback: flushEvents,
-      onError: (e) => _logger.warn('Exception in events auto flush', e),
-    );
+    final autoFlushEventsInterval = pinpointConfig.autoFlushEventsInterval;
+
+    if (autoFlushEventsInterval.isNegative) {
```
Jordan-Nelson, line 158: What about 0?

fjnoyp: 0 is implicitly handled - 0 interval means no autoflush.

Current conditions are:
if negative throw config error
if duration > 0 do autoflush
implicit case --> no autoflush, duration was 0

packages/amplify_core/lib/src/config/analytics/pinpoint_config.dart
```diff
@@ -119,3 +137,13 @@ class PinpointTargeting with AWSEquatable<PinpointTargeting>, AWSSerializable {
   @override
   Map<String, Object?> toJson() => _$PinpointTargetingToJson(this);
 }
+
+class DurationConverter implements JsonConverter<Duration, int> {
```
dnys1, line 141: Can be private

packages/amplify_core/lib/src/config/analytics/pinpoint_config.dart
dnys1: Might be just better to accept an `int` in the constructor and have Duration be the getter. If you do that, though, then I would rename the field in the ctor as `autoFlushEventsIntervalSecs` and then you can remove this converter class altogether

packages/analytics/amplify_analytics_pinpoint_dart/lib/src/analytics_plugin_impl.dart
```diff
@@ -153,11 +153,22 @@ class AmplifyAnalyticsPinpointDart extends AnalyticsPluginInterface {
...
+      throw ConfigurationError(
+        'The autoFlushEventsInterval field in amplifyconfiguration.dart must be positive.',
```
dnys1, line 160:
```suggestion
        'The autoFlushEventsInterval field in your Amplify configuration must be a positive integer or 0 to disable auto-flushing.',
```

Jordan-Nelson (approved):

dnys1 (approved):

PR #2814, opened 2023-04-04
fix(api): Support model definition target name

Equartey:
*Description of changes:*
Fixes failing api unit tests for legacy model schemas. A change in core [(#2798)](https://github.com/aws-amplify/amplify-flutter/pull/2798) removed the passing of a depreciated parameter to the underlying constructor. While we still aim to support older schema's this will remain in place. 


ragingsquirrel3 (approved):

dnys1 (approved):
Whoops - thanks!

PR #2815, opened 2023-04-04
feat(storage)!: update upload APIs options to add metadata field

NikaHsn:
*Issue #, if available:*

*Description of changes:*
- update uploadData and uploadFile APIs to move metadata from plugin options to category level options.

### DIFF
--- a/packages/amplify_core/lib/src/types/storage/upload_data_options.dart
+++ b/packages/amplify_core/lib/src/types/storage/upload_data_options.dart
@@ -15,14 +15,18 @@ class StorageUploadDataOptions extends StorageOperationOptions
   /// {@macro amplify_core.storage.upload_data_options}
   const StorageUploadDataOptions({
     super.accessLevel,
+    this.metadata = const {},
     this.pluginOptions,
   });
+
+  /// The metadata attached to the object to be uploaded.
+  final Map<String, String> metadata;
 
   /// {@macro amplify_core.storage.upload_data_plugin_options}
   final StorageUploadDataPluginOptions? pluginOptions;
 
   @override
-  List<Object?> get props => [accessLevel, pluginOptions];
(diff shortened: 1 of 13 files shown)


packages/amplify_core/lib/src/types/storage/upload_data_options.dart
```diff
@@ -15,21 +17,26 @@ class StorageUploadDataOptions extends StorageOperationOptions
   /// {@macro amplify_core.storage.upload_data_options}
   const StorageUploadDataOptions({
     super.accessLevel,
+    this.metadata,
     this.pluginOptions,
   });
 
+  /// The metadata attached to the object to be uploaded.
+  final Map<String, String>? metadata;
```
dnys1, line 25: Generally would recommend Maps always be non-nullable unless there is a specific reason to be able to distinguish an empty map from a null map.

```suggestion
  const StorageUploadDataOptions({
    super.accessLevel,
    this.metadata = const {},
    this.pluginOptions,
  });

  /// The metadata attached to the object to be uploaded.
  final Map<String, String> metadata;
```

packages/amplify_core/lib/src/types/storage/upload_data_options.dart
```diff
@@ -15,21 +17,26 @@ class StorageUploadDataOptions extends StorageOperationOptions
...
-  List<Object?> get props => [accessLevel, pluginOptions];
+  List<Object?> get props => [accessLevel, metadata, pluginOptions];
 
   @override
   String get runtimeTypeName => 'StorageUploadDataOptions';
 
   @override
   Map<String, Object?> toJson() => {
         'accessLevel': accessLevel?.name,
+        'metadata': json.encode(metadata),
```
HuiSF, line 39: Probably don't need the encoding?

HuiSF (approved):

PR #2819, opened 2023-04-05
fix(auth): Client secret support

dnys1:
Fixes some misses around client secret handling and adds e2e tests.


packages/auth/amplify_auth_cognito/example/integration_test/user_attributes_test.dart
```diff
@@ -22,7 +22,11 @@ void main() {
   IntegrationTestWidgetsFlutterBinding.ensureInitialized();
 
   group('User Attributes', () {
-    for (final environmentName in ['main', 'user-pool-only']) {
+    for (final environmentName in [
+      'main',
```
ragingsquirrel3, line 26: nit: This set of strings is repeated a lot so am wondering if could be a constant somewhere? That would reduce the possibility of someone doing an incomplete update in the future which might cause a miss on test coverage.

ragingsquirrel3 (approved):

packages/auth/amplify_auth_cognito/example/integration_test/user_attributes_test.dart
NikaHsn: +1

NikaHsn (approved):

packages/auth/amplify_auth_cognito/example/integration_test/user_attributes_test.dart
dnys1: Good idea 👍

PR #2821, opened 2023-04-06
chore(auth): remove private members from exports

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Remove private members from exports


packages/auth/amplify_auth_cognito/lib/amplify_auth_cognito.dart
```diff
@@ -5,6 +5,17 @@
 library amplify_auth_cognito;
 
 export 'package:amplify_auth_cognito_dart/amplify_auth_cognito_dart.dart'
-    hide AmplifyAuthCognitoDart;
+    hide
+        AmplifyAuthCognitoDart,
+        // TODO(Jordan-Nelson): determine if the following models should be
```
Jordan-Nelson, line 10: Note: I'm not sure all of these should be exported from the _dart package but I think we can evaluate that more closely at a later date. Left this as a todo for now.

dnys1 (approved):

ragingsquirrel3 (approved):

PR #2824, opened 2023-04-07
fix(notifications): Analytics naming updates, Android test fixes and token received bug fix

Samaritan1011001:
*Issue #, if available:*

*Description of changes:*
Includes the following changes,
- Renaming of some Analytics types such as AnalyticsUserProfile to UserProfile
- Fixed and added some Android test cases
- Added a second token channel for internal use only which solve a bug with race condition [Android native only in this PR] iOS native code needs to be updated to complete this resolution. 


packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationEventsStreamHandler.kt
```diff
@@ -13,11 +13,12 @@ private const val TAG = "PushNotificationEventsStreamHandler"
 
 @InternalAmplifyApi
 enum class NativeEvent {
-    TOKEN_RECEIVED, NOTIFICATION_OPENED, LAUNCH_NOTIFICATION_OPENED, FOREGROUND_MESSAGE_RECEIVED, BACKGROUND_MESSAGE_RECEIVED, ERROR;
+    TOKEN_RECEIVED,INTERNAL_TOKEN_RECEIVED, NOTIFICATION_OPENED, LAUNCH_NOTIFICATION_OPENED, FOREGROUND_MESSAGE_RECEIVED, BACKGROUND_MESSAGE_RECEIVED, ERROR;
```
HuiSF, line 16: nit: line breaks?

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationUtils.kt
```diff
@@ -168,5 +174,5 @@ fun NotificationPayload.toWritableMap(): Map<Any, Any?> {
             "channelId" to payload?.channelId,
         ),
 
-    )
+        )
```
HuiSF, line 177: nit: whiespace

packages/notifications/push/amplify_push_notifications/android/src/test/kotlin/com/amazonaws/amplify/amplify_push_notifications/InternalPushNotificationUtilsTest.kt
```diff
@@ -3,6 +3,7 @@ package com.amazonaws.amplify.amplify_push_notifications
 import android.content.ComponentName
 import android.content.Context
 import android.os.Bundle
+import androidx.core.app.ActivityCompat
```
HuiSF, line 6: sanity check - is this used?

packages/notifications/push/amplify_push_notifications/android/src/test/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationFirebaseMessagingServiceTest.kt
```diff
@@ -64,9 +66,15 @@ class PushNotificationFirebaseMessagingServiceTest {
         controller = buildService(PushNotificationFirebaseMessagingService::class.java, Intent())
     }
 
+    @After
+    fun tearDown(){
+        unmockkStatic(Bundle::getNotificationPayload)
+        unmockkStatic(NotificationPayload::toWritableMap)    }
```
HuiSF, line 72: nit: line break before `}`

packages/notifications/push/amplify_push_notifications/lib/src/native_push_notifications_plugin.g.dart
```diff
@@ -1,4 +1,4 @@
-//
+//
```
HuiSF, line 1: If you had to regenerate this file, please reformat.

packages/notifications/push/amplify_push_notifications_pinpoint/test/push_notifications_background_processing_test.dart
```diff
@@ -46,8 +41,36 @@ void main() {
...
+          isA<PushNotificationException>().having(
+            (e) => e.message,
+            'Cached config not found',
+            contains('config was not found'),
+          ),
+        ),
       );
     });
+
+    test('should pass', () async {
```
HuiSF, line 54:
```suggestion
    test('should configure Amplify plugins', () async {
```

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -233,6 +247,7 @@ abstract class AmplifyPushNotifications
       value: jsonEncode(config),
     );
     _isConfigured = true;
+    print('configured');
```
fjnoyp, line 250: remove

packages/notifications/push/amplify_push_notifications_pinpoint/lib/src/push_notifications_background_processing.dart
```diff
@@ -21,6 +25,7 @@ const _backgroundChannel = MethodChannel(
 @pragma('vm:entry-point')
 Future<void> amplifyBackgroundProcessing({
   @visibleForTesting AmplifySecureStorage? amplifySecureStorage,
+  @visibleForTesting AmplifyClass? mockAmplify,
```
fjnoyp, line 28: I don't think it's necessary to have mock or mockable prefix.  You can follow the existing pattern for amplifySecureStorage. 

The @visibleForTesting marker already indicates the purpose of the parameter. 

Furthermore I'm unaware of other parts of Amplify code using mock/mockable prefixes in the production code. 


Also, typo in comment above 'notificaiton'

fjnoyp:
Dillon's comments regarding buffer seems to make sense.  If we can buffer that initial token that was lost instead of duplicating to an internal channel seems better in the long run.  As an outside observer reading the code, it's not clear when/why we switch between internal and public channel.

packages/notifications/push/amplify_push_notifications/android/src/test/kotlin/com/amazonaws/amplify/amplify_push_notifications/InternalPushNotificationUtilsTest.kt
Samaritan1011001: Nope, removed it. Thanks!

packages/notifications/push/amplify_push_notifications/pubspec.yaml
```diff
@@ -12,6 +12,7 @@ dependencies:
   amplify_auth_cognito: ">=1.0.0-next.4 <1.0.0-next.5"
   amplify_core: ">=1.0.0-next.0 <1.0.0-next.1"
   amplify_secure_storage: ^0.1.0
+  async: ^2.11.0
```
HuiSF, line 15: Should be in sync of [this](https://github.com/aws-amplify/amplify-flutter/blob/next/aft.yaml#L9), can run `aft constraints` to align.

```suggestion
  async: ^2.10.0
```

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -142,6 +122,26 @@ void main() {
       ).called(1);
     });
 
+    test(
+        'should fire both internal token.first and external token received listener',
```
HuiSF, line 126: This test actually doesn't verify `token.first` though, can you add assertion for it?

packages/notifications/push/amplify_push_notifications_pinpoint/test/pinpoint_provider_test.dart
```diff
@@ -0,0 +1,213 @@
...
+
+    test(
+        'constructEventInfo should return journey data when there is journey details in the payload',
+        () async {
+      final res = pinpointProvider.constructEventInfo(
+        notification: PushNotificationMessage.fromJson(androidJourneyMessage),
+      );
+      final properties = res.properties;
+      final source = res.source;
+      expect(properties.attributes.containsKey('journey_id'), true);
```
HuiSF, line 81: nit:
```suggestion
      expect(properties.attributes.containsKey('journey_id'), isTrue);
```

packages/notifications/push/amplify_push_notifications_pinpoint/test/pinpoint_provider_test.dart
```diff
@@ -0,0 +1,213 @@
...
+      expect(source, PinpointEventSource.journey.name);
```
HuiSF, line 82: nit: could use matchers throughout
```suggestion
      expect(source, equals(PinpointEventSource.journey.name));
```

packages/notifications/push/amplify_push_notifications_pinpoint/test/pinpoint_provider_test.dart
```diff
@@ -0,0 +1,213 @@
...
+          pinpointAppId: anyNamed('pinpointAppId'),
+          region: anyNamed('region'),
+          authProvider: anyNamed('authProvider'),
+        ),
+      ).thenAnswer((realInvocation) async {});
+      await pinpointProvider.init(
+        config: notificationsPinpointConfig,
+        authProviderRepo: mockAmplifyAuthProviderRepository,
+        mockAnalyticsClient: mockAnalyticsClient,
+      );
```
HuiSF, line 115: Use assertion to verify function calls can complete without exceptions throughout. 

```suggestion
      expect(
        pinpointProvider.init(
          config: notificationsPinpointConfig,
          authProviderRepo: mockAmplifyAuthProviderRepository,
          mockAnalyticsClient: mockAnalyticsClient,
        ),
        completes,
      );
```

packages/notifications/push/amplify_push_notifications/pubspec.yaml
Samaritan1011001: Nice, updated other versions too based on `aft constraints`

HuiSF (approved):
LGTM!

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
```diff
@@ -84,7 +84,7 @@ class PushNotificationBackgroundService : JobIntentService(), MethodChannel.Meth
         }
     }
 
-    fun createAndRunBackgroundEngine(context: Context, loader: FlutterLoader ){
+    fun createAndRunBackgroundEngine(context: Context, loader: FlutterLoader, mockEngine: FlutterEngine? ){
```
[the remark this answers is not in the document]
fjnoyp: Okay, makes sense you need to mock the FlutterEngine, what about rewriting the code to be more clear as mentioned above: 

`backgroundFlutterEngine = mockEngine ?: FlutterEngineCache....`

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyLifecycleObserver.kt
```diff
@@ -11,8 +11,15 @@ private const val TAG = "AmplifyLifecycleObserver"
 
 @InternalAmplifyApi
 class AmplifyLifecycleObserver : DefaultLifecycleObserver {
+    private  var isCalledAtStart:Boolean = true
```
HuiSF, line 14:
```suggestion
    private  var isAppLaunch  = true
```

HuiSF (approved):

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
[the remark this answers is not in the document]
fjnoyp: What are you testing inside the if statement?  It seems you're just putting the engine within the FlutterEngineCache?

Since as you say you can mock the FlutterEngineCache, why not have that mock return the mockEngine and avoid the if statement?

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyLifecycleObserver.kt
fjnoyp: Formatting issue, double space.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyLifecycleObserver.kt
```diff
@@ -11,8 +11,15 @@ private const val TAG = "AmplifyLifecycleObserver"
...
     override fun onResume(owner: LifecycleOwner) {
+
+        // Need to null it only when the app resumes from background and not on app start.
```
fjnoyp, line 17: Only set launchNotification to null when app resumes from background.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
```diff
@@ -42,6 +42,12 @@ open class AmplifyPushNotificationsPlugin : FlutterPlugin, ActivityAware,
          * Flutter API interface.
          */
         var flutterApi: PushNotificationsFlutterApi? = null
+
+        /**
```
fjnoyp, line 46:
```suggestion
        /** Map of notification data when app was launched by tapping on the notification when the app is in a killed state, null otherwise. 
```

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -93,47 +73,47 @@ void main() {
...
-
-      testWidgetsFlutterBinding.defaultBinaryMessenger.setMockMethodCallHandler(
-        MethodChannel(tokenReceivedEventChannel.name),
-        handler,
-      );
+      // Sending the token after a delay so that token.first in configure gets the token.
+      // This is only required for tests since StreamQueue used for buffered stream in impl
+      // listens to the channel before token.first call making it useless to emit event when listened.
+      // This delay is replicated throughout the tests so token.first can fire.
+      Future.delayed(const Duration(microseconds: 1), () async {
```
dnys1, line 95: I don't get what this means or why it's needed. This is the equivalent of `Future.delayed(Duration.zero, ...` meaning "run in the next event loop". It's not clear why this is needed since the test plugin is creating the `StreamQueue` at the same time as the normal plugin (in the constructor).

If the test plugin is behaving differently than the normal plugin, then this test is really testing the test plugin which I don't think adds much value. The test plugin should behave as close as possible to the original plugin and we should be mocking one thing at a time (in this case the event channel).

packages/notifications/push/amplify_push_notifications_pinpoint/test/push_notifications_background_processing_test.dart
```diff
@@ -49,5 +44,27 @@ void main() {
...
+    test('should configure Amplify plugins', () async {
+      log.clear();
+      final mockStorage = MockAmplifySecureStorage();
+      when(mockStorage.read(key: anyNamed('key')))
+          .thenAnswer((_) async => amplifyconfig);
+      final mockAmplify = MockAmplifyClass();
+
+      when(mockAmplify.isConfigured).thenReturn(false);
+      when(mockAmplify.addPlugins(any))
+          .thenAnswer((realInvocation) => Future(() => null));
```
dnys1, line 57:
```suggestion
          .thenAnswer((realInvocation) async {});
```

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -226,6 +206,45 @@ void main() {
       );
     });
   });
+  test('should fail configure when registering device is unsuccessfull',
```
fjnoyp, line 209: typo unsuccessful

packages/notifications/push/amplify_push_notifications_pinpoint/lib/src/pinpoint_provider.dart
```diff
@@ -213,15 +226,15 @@ class PinpointProvider implements ServiceProviderClient {
...
+  ChannelType? _getChannelType() {
+    if (Platform.isAndroid) {
+      return ChannelType.gcm;
+    } else if (Platform.isIOS) {
+      if (zDebugMode) {
+        return ChannelType.apnsSandbox;
+      }
+      return ChannelType.apns;
+    }
+    return null;
```
dnys1, line 238: When would the platform not be iOS or Android? I would make this return `ChannelType`

packages/amplify_core/lib/src/types/notifications/push/push_notification_message.dart
```diff
@@ -55,16 +62,26 @@ class PushNotificationMessage
...
-  ApnsPlatformOptions? apnsOptions;
-  Map<Object?, Object?> data = {};
+  final String? title;
+  final String? body;
+  final String? imageUrl;
+  final String? deeplinkUrl;
+  final String? goToUrl;
+  final FcmPlatformOptions? fcmOptions;
+  final ApnsPlatformOptions? apnsOptions;
+  final Map<Object?, Object?> data;
```
dnys1, line 84: This will always be `Map<String, Object?>` if I'm not mistaken.
```suggestion
  final Map<String, Object?> data;
```

dnys1 (approved):

packages/notifications/push/amplify_push_notifications_pinpoint/lib/src/pinpoint_provider.dart
```diff
@@ -35,33 +42,39 @@ class PinpointProvider implements ServiceProviderClient {
   Future<void> init({
     required NotificationsPinpointPluginConfig config,
     required AmplifyAuthProviderRepository authProviderRepo,
-    // @visibleForTesting FlutterAnalyticsClient? mockAnalyticsClient,
+    @visibleForTesting AnalyticsClient? mockAnalyticsClient,
```
fjnoyp, line 45: You can just name it analyticsClient.  The visibleForTesting param already indicates it's just for testing purposes.

packages/notifications/push/amplify_push_notifications_pinpoint/lib/src/push_notifications_background_processing.dart
```diff
@@ -40,13 +45,14 @@ Future<void> amplifyBackgroundProcessing({
   }
   final notificationsPlugin = AmplifyPushNotificationsPinpoint();
   final authPlugin = AmplifyAuthCognito();
-  if (!Amplify.isConfigured) {
-    await Amplify.addPlugins([authPlugin, notificationsPlugin]);
-    await Amplify.configure(amplifyconfigStr);
+  final mockableAmplify = amplify ?? Amplify;
```
fjnoyp, line 48: I'm unsure what happened to my previous comments about including mockable prefix in the code.  
It's strange to see testing code mixed with production code here.  I haven't seen this anywhere else in our codebase and it confuses from the main purpose of your code. 

For example, is the below just for when you mock or for production cases as well?  What information are you conveying with the word 'mockable' that isn't already clear by adding @visibleForTesting above?

packages/notifications/push/amplify_push_notifications_pinpoint/test/pinpoint_provider_test.dart
```diff
@@ -0,0 +1,235 @@
...
+  final pinpointProvider = PinpointProvider();
+  final mockAmplifyAuthProviderRepository = MockAmplifyAuthProviderRepository();
+  const notificationsPinpointConfig = NotificationsPinpointPluginConfig(
+    appId: 'APP_ID',
+    region: 'REGION',
+  );
+  final awsIamAmplifyAuthProvider = MockAWSIamAmplifyAuthProvider();
+  final mockAnalyticsClient = MockAnalyticsClient();
+  group('PinpointProvider', () {
+    test('init fails when retreiving an Auth provider was not successfull', () {
```
fjnoyp, line 38: typo retrieving

fjnoyp: typo successful

Please go over your code changes for any other typos.

packages/notifications/push/amplify_push_notifications_pinpoint/test/push_notifications_background_processing_test.dart
```diff
@@ -1,40 +1,35 @@
...
+
+  Future<Map<String, dynamic>?>? handler(MethodCall methodCall) async {
+    log.add(methodCall);
+    return null;
+  }
+
+  setUp(() {
+    testWidgetsFlutterBinding.defaultBinaryMessenger.setMockMethodCallHandler(
+      backgroundChannel,
+      handler,
```
fjnoyp, line 30: This method can be inline since it's only used once.  (MethodCall methodCall) { log.add(......)

fjnoyp:
Approved.  Can you please validate your changes don't break the unit tests?  I know they're not working due to secure storage bug, but good to just double check nothing else is failing.

fjnoyp (approved):

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
Samaritan1011001: It's not the test plugin as I am just returning the constructed object with super values. 

Initially I was emitting the token event when a stream listens in on the event channel as you can see [here](https://github.com/aws-amplify/amplify-flutter/pull/2824/files#diff-2d8b06b2122317db94204b3791335562662be1056b2229fc3f543e0e26a8030dL62).

StreamQueue listens to the given stream internally and so the event would emit even before token.first gets called. 

But that was when I was using token.first but since I changed it to use bufferedStream.peek this wait is not required. Thanks for catching it, will change it back to prev state which works just fine.

packages/amplify_core/lib/src/types/notifications/push/push_notification_message.dart
Samaritan1011001: The data payload from iOS comes through as `Object?` which is why I kept it as that.

packages/notifications/push/amplify_push_notifications_pinpoint/lib/src/push_notifications_background_processing.dart
Samaritan1011001: It's only for testing purposes. I agree it's a bad idea to mix these. I'll update them.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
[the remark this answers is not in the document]
Samaritan1011001: I know there is no logic to test inside the if statement, I was more trying to cover that since it's an important if block for the functioning of killed state. 

I do return a mockEngine and skip the if block in [another test](https://github.com/aws-amplify/amplify-flutter/blob/fix/push-android-tests/packages/notifications/push/amplify_push_notifications/android/src/test/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundServiceTest.kt#L91).

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -431,37 +431,13 @@ void main() {
...
           );
-        await Future.delayed(const Duration(seconds: 3), () {});
+        await Future.delayed(const Duration(microseconds: 1), () {});
 
         expect(pref.containsKey(externalHandleKey), isTrue);
       },
     );
   });
 
-  test('should invoke the top-level or static external callback function',
```
Samaritan1011001, line 398: Removing this test as it's covered in the [flutter api test cases](https://github.com/aws-amplify/amplify-flutter/blob/fix/push-android-tests/packages/notifications/push/amplify_push_notifications/test/push_notification_flutter_api_test.dart#L71).

PR #2828, opened 2023-04-10
fix: prevent duplicate calls to onFieldSubmitted

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Adds a check to onFieldSubmitted to see if the authenticator is already in a busy state, and returns if so.

This resolves two potential issues:
1. Due to https://github.com/flutter/flutter/issues/124458, submitting an authenticator form with the enter/return key on web can lead to duplicate calls to onFieldSubmitted.
2. Even with the above issue resolved, an end user could submit the form with the enter/return, refocus on the text field, and then submit the form with the enter/return a second time while the first submission is pending.


packages/authenticator/amplify_authenticator/test/authenticated_view_test.dart
```diff
@@ -29,35 +38,62 @@ void main() {
...
+        await tester.sendKeyEvent(LogicalKeyboardKey.tab);
 
-      // Move focus to first next widget via keyboard (Sign In Button).
-      await tester.sendKeyEvent(LogicalKeyboardKey.tab);
+        // Move focus to first next widget via keyboard (Sign In Button).
+        await tester.sendKeyEvent(LogicalKeyboardKey.tab);
 
-      verifyNever(mockState.signIn);
+        // Submit form with Enter key.
+        await tester.sendKeyEvent(LogicalKeyboardKey.enter);
```
dnys1, line 63: Would tapping the button have the same effect?

dnys1 (approved):

Equartey (approved):

packages/authenticator/amplify_authenticator/test/authenticated_view_test.dart
dnys1: Okay, sounds good. I wouldn't mind it as a base case in these tests, but every other test in the package pretty much uses the button so I think it's okay.

PR #2832, opened 2023-04-10
docs(auth): Add inline examples

dnys1:
- Adds a mechanism for incorporating examples from Dart code inline in docs. Uses the [code_excerpter](https://github.com/dart-lang/site-shared/tree/main/packages/code_excerpter) library which is used internally by the Dart+Flutter teams and provides a very simple mechanism for orchestrating this flow.
- Adds inline code examples to the Auth category docs. To regenerate: `aft run build:docs --include=amplify_core`.

### DIFF
--- a/packages/amplify_core/lib/src/category/amplify_auth_category.dart
+++ b/packages/amplify_core/lib/src/category/amplify_auth_category.dart
@@ -49,6 +49,14 @@ class AuthCategory extends AmplifyCategory<AuthPluginInterface> {
   /// {@template amplify_core.amplify_auth_category.sign_up}
   /// Create a new user with the given [username] and [password].
   ///
+  /// Optionally accepts plugin [options] which allow customizing provider-specific
+  /// behavior, e.g. the Cognito User Pool.
+  ///
+  /// For more information, see the
+  /// [Amplify docs](https://docs.amplify.aws/lib/auth/signin/q/platform/flutter/#register-a-user).
+  ///
+  /// ## Examples
+  ///
   /// <?code-excerpt "doc/lib/auth.dart" region="imports"?>
   /// ```dart
   /// import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
@@ -57,6 +65,8 @@ class AuthCategory extends AmplifyCategory<AuthPluginInterface> {
   ///
   /// <?code-excerpt "doc/lib/auth.dart" region="signup"?>
   /// ```dart
+  /// /// Signs a user up with a username, password, and email. The required
+  /// /// attributes may be different depending on your app's configuration.
   /// Future<void> signUpUser({
   ///   required String username,
   ///   required String password,
@@ -102,14 +112,18 @@ class AuthCategory extends AmplifyCategory<AuthPluginInterface> {
   /// }
   /// ```
   ///
+  /// <?code-excerpt "doc/lib/auth.dart" region="handle-code"?>
+  /// ```dart
+  /// void _handleCodeDelivery(AuthCodeDeliveryDetails codeDeliveryDetails) {
+  ///   safePrint(
+  ///     'A confirmation code has been sent to ${codeDeliveryDetails.destination}. '
+  ///     'Please check your ${codeDeliveryDetails.deliveryMedium.name} for the code.',
+  ///   );
+  /// }
+  /// ```
+  ///
   /// If you need to resend the sign up code at any point in the sign up process,
   /// call [resendSignUpCode].
-  ///
-  /// Optionally accepts plugin [options] which allow customizing provider-specific
-  /// behavior, e.g. the Cognito User Pool.
-  ///
-  /// For more information, see the
-  /// [Amplify docs](https://docs.amplify.aws/lib/auth/signin/q/platform/flutter/#register-a-user).
   /// {@endtemplate}
   Future<SignUpResult> signUp({
     required String username,
@@ -125,6 +139,14 @@ class AuthCategory extends AmplifyCategory<AuthPluginInterface> {
   /// {@template amplify_core.amplify_auth_category.confirm_sign_up}
   /// Confirm the current sign up for [username] with the [confirmationCode]
   /// provided by the user.
+  ///
+  /// Optionally accepts plugin [options] which allow customizing provider-specific
+  /// behavior, e.g. the Cognito User Pool.
+  ///
+  /// For more information, see the
+  /// [Amplify docs](https://docs.amplify.aws/lib/auth/signin/q/platform/flutter/#register-a-user).
+  ///
+  /// ## Examples
   ///
   /// <?code-excerpt "doc/lib/auth.dart" region="imports"?>
   /// ```dart
@@ -180,12 +202,6 @@ class AuthCategory extends AmplifyCategory<AuthPluginInterface> {
   ///   );
   /// }
   /// ```
-  ///
-  /// Optionally accepts plugin [options] which allow customizing provider-specific
-  /// behavior, e.g. the Cognito User Pool.
-  ///
-  /// For more information, see the
-  /// [Amplify docs](https://docs.amplify.aws/lib/auth/signin/q/platform/flutter/#register-a-user).
   /// {@endtemplate}
   Future<SignUpResult> confirmSignUp({
     required String username,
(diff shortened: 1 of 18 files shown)


packages/aft/lib/src/models/package_selector.dart
```diff
@@ -1,12 +1,22 @@
...
+import 'dart:collection';
+
 import 'package:aft/aft.dart';
 import 'package:aws_common/aws_common.dart';
+import 'package:collection/collection.dart';
 import 'package:glob/glob.dart';
 import 'package:json_annotation/json_annotation.dart';
 import 'package:path/path.dart' as p;
 
+class _PathSet extends DelegatingSet<String> {
```
fjnoyp, line 13: Is this class the equivalent of the previous 'Set<String>.of` for removing duplicate entries in `Iterable<String> paths`?

Out of curiosity why did you switch away from `Set<String>.of`?  

Why can't we just use a HashSet directly since we're creating it and wrapping it with the DelegatingSet anyway?  Could we just declare a `Set` variable type and set it equal to the `HashSet`?

packages/amplify_core/lib/src/category/amplify_auth_category.dart
```diff
@@ -61,6 +125,67 @@ class AuthCategory extends AmplifyCategory<AuthPluginInterface> {
...
+  ///       confirmationCode: confirmationCode,
+  ///     );
+  ///     return _handleSignUpResult(result);
+  ///   } on AuthException catch (e) {
+  ///     safePrint('Error confirming user: ${e.message}');
+  ///   }
+  /// }
+  /// ```
+  ///
+  /// After confirming the sign up, check the [SignUpResult] object to check if
```
fjnoyp, line 153: Can we include the code description comments you have here directly in the auth docs?  I'm assuming that could make the `docs/lib/auth.dart` file more readable and better enforce separation on where we update the examples as current updates could feasibly be in `doc/lib/auth.dart` and `amplify_auth_category.dart`

packages/aft/lib/src/models/package_selector.dart
dnys1: Path comparison requires a special `==` method, which is why `Set.of` does not work. This delegates to a `HashSet` with `==` overridden correctly.

packages/amplify_core/lib/src/category/amplify_auth_category.dart
dnys1: Unfortunately, no. The goal of this is to make maintaining code snippets easier, not necessarily to reduce the burden of maintaining docs in general.

packages/amplify_core/lib/src/category/amplify_auth_category.dart
```diff
@@ -46,6 +48,68 @@ class AuthCategory extends AmplifyCategory<AuthPluginInterface> {
...
+  /// will be sent to the email or phone number provided during sign up. Prompt the
+  /// user to check their device for the code sent by Cognito.
+  ///
+  /// <?code-excerpt "doc/lib/auth.dart" region="handle-signup"?>
+  /// ```dart
+  /// Future<void> _handleSignUpResult(SignUpResult result) async {
+  ///   switch (result.nextStep.signUpStep) {
+  ///     case AuthSignUpStep.confirmSignUp:
+  ///       final codeDeliveryDetails = result.nextStep.codeDeliveryDetails!;
+  ///       _handleCodeDelivery(codeDeliveryDetails);
```
Jordan-Nelson, line 96: Did you intend for `_handleCodeDelivery` to be included in the doc comment as well? I think it is fine to exclude, but wasn't sure if you had intended it to be included here.

packages/amplify_core/lib/src/category/amplify_auth_category.dart
```diff
@@ -95,6 +256,186 @@ class AuthCategory extends AmplifyCategory<AuthPluginInterface> {
...
+  ///     // ···
+  ///     case AuthSignInStep.done:
+  ///       safePrint('Sign in is complete');
+  ///       break;
+  ///     // ···
+  ///   }
+  /// }
+  /// ```
+  ///
+  /// Optionally accepts plugin [options] which allow customizing provider-specific
```
Jordan-Nelson, line 434: The content here is sort of buried by examples. Do you think it is worth putting this above the example?

packages/amplify_core/lib/src/category/amplify_auth_category.dart
```diff
@@ -95,6 +256,186 @@ class AuthCategory extends AmplifyCategory<AuthPluginInterface> {
   ///
   /// For Cognito flows, including a [password] will initiate an SRP sign-in,
   /// while excluding the [password] will initiate a custom auth sign-in.
+  ///
+  /// <?code-excerpt "doc/lib/auth.dart" region="imports"?>
+  /// ```dart
+  /// import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
+  /// import 'package:amplify_flutter/amplify_flutter.dart';
```
Jordan-Nelson, line 263: Should there be a blank line between the imports and the first fn? It just looks a little crowded when viewing it in an IDE.

Screen Shot 2023-04-12 at 10 35 56 AM

packages/amplify_core/lib/src/category/amplify_auth_category.dart
```diff
@@ -46,6 +48,68 @@ class AuthCategory extends AmplifyCategory<AuthPluginInterface> {
 
   /// {@template amplify_core.amplify_auth_category.sign_up}
   /// Create a new user with the given [username] and [password].
+  ///
```
Jordan-Nelson, line 51: Might be worth adding a header or something to separate out the examples?

```suggestion
  ///
  /// ## Example
  ///
```

packages/amplify_core/lib/src/category/amplify_auth_category.dart
```diff
@@ -46,6 +48,68 @@ class AuthCategory extends AmplifyCategory<AuthPluginInterface> {
...
+  /// <?code-excerpt "doc/lib/auth.dart" region="imports"?>
+  /// ```dart
+  /// import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
+  /// import 'package:amplify_flutter/amplify_flutter.dart';
+  /// ```
+  ///
+  /// <?code-excerpt "doc/lib/auth.dart" region="signup"?>
+  /// ```dart
+  /// Future<void> signUpUser({
```
Jordan-Nelson, line 60: Might be worth calling out that this will vary based on the app's config.

```suggestion
  /// /// Signs a user up with a username, password, and email. The required 
  /// /// attributes may be different depending on your app's configuration.
  /// Future<void> signUpUser({
```

Jordan-Nelson (approved):
Very cool. Will be very helpful in ensuring the examples don't get out of date. I left some suggestions, but none are blockers.

packages/amplify_core/lib/src/category/amplify_auth_category.dart
dnys1: The spacing issue is weird - I played around with it and it's not so straightforward to solve unfortunately. I'll do a bit more digging if I have time but at the moment, this is currently the best I can do.

packages/amplify_core/lib/src/category/amplify_auth_category.dart
fjnoyp: Alright, what are the downsides to consolidating the doc comments + code examples into the `doc/lib/auth.dart` file?  

From my understanding, there wouldn't be a case where the comment and the code are used separately, so if they're intertwined wouldn't it be good to keep them together?

fjnoyp (approved):
Looks good, responded to one minor comment non blocking.

packages/amplify_core/lib/src/category/amplify_auth_category.dart
dnys1: There isn't any downside, it's just not possible.

The code snippets work by injecting Dart code into the Markdown doc comments. The only way to colocate the Dart code and documentation is to affix it as documentation to the Dart code, e.g.

```dart
// #docregion handle-signup
/// After confirming the sign up, check the [SignUpResult] object to check if
/// further confirmations are needed or if the sign up is complete.
Future<void> _handleSignUpResult(SignUpResult result) async {
  switch (result.nextStep.signUpStep) {
    case AuthSignUpStep.confirmSignUp:
      final codeDeliveryDetails = result.nextStep.codeDeliveryDetails!;
      _handleCodeDelivery(codeDeliveryDetails);
      break;
    case AuthSignUpStep.done:
      safePrint('Sign up is complete');
      break;
  }
}
// #enddocregion handle-signup
```

But this has two issues.

1. This code is no longer reusable. `handleSignUpResult` is injected in multiple places in the Auth category's doc comments. This general comment about confirming sign up is not applicable to each instance where we want the code to appear.
2. In the rendered HTML documentation, this comment will no longer appear as paragraph text, but as a code comment which is not appropriate in this case.

ragingsquirrel3 (approved):

PR #2833, opened 2023-04-10
chore(authenticator): missing cherry picks from main

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*

Cherry pick 672a7cf133b8d16bc05644118c56a8a36f5fdaa0 and 33df283fe17f206aa7b0e22484e83b60d4d4f789 from main to next.

### DIFF
--- a/packages/authenticator/amplify_authenticator/lib/src/utils/validators.dart
+++ b/packages/authenticator/amplify_authenticator/lib/src/utils/validators.dart
@@ -34,13 +34,10 @@ FormFieldValidator<String> simpleValidator(
 }
 
 FormFieldValidator<String> usernameValidator({
-  required bool isOptional,
   required BuildContext context,
   required InputResolver inputResolver,
 }) {
   return (String? input) {
-    if (isOptional) return null;
-
     if (input == null || input.isEmpty) {
       return inputResolver.resolve(
         context,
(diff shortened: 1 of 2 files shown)


dnys1 (approved):

packages/authenticator/amplify_authenticator/lib/src/utils/validators.dart
```diff
@@ -28,6 +33,32 @@ FormFieldValidator<String> simpleValidator(
   };
 }
 
+FormFieldValidator<String> usernameValidator({
+  required bool isOptional,
+  required BuildContext context,
+  required InputResolver inputResolver,
+}) {
+  return (String? input) {
+    if (isOptional) return null;
```
Equartey, line 42: Should we still validate the input if it's provided?

Equartey (approved):
Looks good! Only had one question.

packages/authenticator/amplify_authenticator/lib/src/utils/validators.dart
Jordan-Nelson: Username cannot be optional, so this line should never actually be hit. The code is confusing though. We can remove  `isOptional` as it is always false in this case.

Thanks for calling that out. I pushed an update.

PR #2842, opened 2023-04-11
chore(repo): Reorganize test packages

dnys1:
- Moves `amplify_test` to new `test` folder
- Adds `amplify_integration_test` package for example apps. This allows references to plugin packages since it is only ever imported by example apps. 

To avoid circular dependencies:
- `amplify_test` can be imported by any Amplify package except `amplify_core` (on which it depends) but can only reference AWS packages and `amplify_core`.
- `amplify_integration_test` can be imported **only** by example apps but is allowed to reference any AWS/Amplify package.


packages/test/amplify_integration_test/lib/amplify_integration_test.dart
```diff
@@ -1,17 +1,13 @@
-// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
fjnoyp, line 1: Should have copyright header still right?

packages/test/amplify_integration_test/README.md
```diff
@@ -0,0 +1,3 @@
+# amplify_integration_test
+
+Utilities for running Amplify integration tests.
```
fjnoyp, line 3: Can we give more details here, at minimum can you include the details regarding package dependencies mentioned in your Github comment?  I'm concerned we might loose track of that insight.

fjnoyp (commented):
What is the distinction for integration test related utils to exist in the test folder vs the plugin's `example/integration_test` folder?  For example, I see in auth cognito, `validation_utils, test_utils, setup_utils` in `integration_test` but `integration_test_auth_utils, create_mfa_code_response`, etc. in the new `test/integration_test` folder. 

Wouldn't it make sense for all util code for auth to exist in one package?  

Also, is this refactor work partially done?  Looking into `amplify_integration_test.dart` I only see exports related to Auth.  I assume we should refactor all other plugin code into the new `integration_test` folder as well.  How do you imagine splitting and planning out that additional followup work for the other plugins?

packages/test/amplify_test/lib/amplify_test.dart
```diff
@@ -0,0 +1,8 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+library amplify_test;
```
fjnoyp, line 4: For the different main test files, can we add some brief comments on what is being exported for each?  While it's clear now, I can imagine in the future where it might not be readily remembered what each test package is supposed to export.

dnys1: Updated 👍

dnys1:
> What is the distinction for integration test related utils to exist in the test folder vs the plugin's `example/integration_test` folder? For example, I see in auth cognito, `validation_utils, test_utils, setup_utils` in `integration_test` but `integration_test_auth_utils, create_mfa_code_response`, etc. in the new `test/integration_test` folder.

`example/integration_test` hosts tests related to that package. The utils in the Auth example package are specifically related to bootstrapping Auth integration tests. This package is intended to host code common to any integration test.

> Wouldn't it make sense for all util code for auth to exist in one package?

Potentially, but that is out of scope for this PR.
 
> Also, is this refactor work partially done? Looking into `amplify_integration_test.dart` I only see exports related to Auth. I assume we should refactor all other plugin code into the new `integration_test` folder as well. How do you imagine splitting and planning out that additional followup work for the other plugins?

The work in this PR is to split out code which previously existed in `amplify_test` but was only used in integration tests into its own package and to repurpose `amplify_test` for unit test code which can be directly imported into packages. No new code has been added.

Basically, the point of this PR is to prevent circular dependencies when publishing but it does also open the door for further optimizations and consolidations.

Jordan-Nelson (approved):

dnys1:
> Unsure if the failing package selector tests are related to this PR. Otherwise looks good to go.

Not related and fixed elsewhere

PR #2845, opened 2023-04-12
chore(notifications): additional tests for Dart and Android

Samaritan1011001:
*Issue #, if available:*

*Description of changes:*
- Added additional unit tests for the dart plugin and Android native code.
- Makes setBadgCount API sync method.
- Update mocks


packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -113,14 +118,83 @@ void main() {
...
+        notificationOpenedEventChannel.name,
+        notificationOpenedEventChannel.codec
+            .encodeSuccessEnvelope(standardAndroidPushMessage),
+        (_) {},
+      );
+
+      verify(
+        mockServiceProviderClient.recordNotificationEvent(
+          eventType: PinpointEventType.notificationOpened,
+          notification: anyNamed('notification'),
```
HuiSF, line 179: May be worth to restrictedly verify which message triggers the call of the listener?

HuiSF (approved):

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -460,4 +534,35 @@ void main() {
       },
     );
   });
+
+  test('get launchNotification is consumable', () async {
+    Future.delayed(const Duration(microseconds: 1), () async {
```
dnys1, line 539: Shouldn't be needed based off previous convo

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
```diff
@@ -84,6 +83,12 @@ void main() {
       });
     });
 
+    tearDown(() {
+      plugin.reset();
```
dnys1, line 87: Shouldn't be needed since you're creating a new plugin each test

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -296,15 +306,15 @@ abstract class AmplifyPushNotifications
   }
 
   @override
-  Future<void> setBadgeCount(int badgeCount) async {
+  void setBadgeCount(int badgeCount) {
```
dnys1, line 309: What's the reason for this change? If `_hostApi.setBadgeCount` throws it will be uncaught. You can change it to: `_hostApi.setBadgetCount(badgeCount).ignore()` to specifically ignore any exceptions (if we're sure this will always succeed) or change this back to Future

dnys1 (approved):

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
HuiSF: I think the reason behind this is that on iOS the `applicationBadgeIconNumber` setter is sync, and it doesn't throw, so on iOS it always succeeds; RN equivalent API is sync. Overall developers don't need to await for this action.

`_hostApi.setBadgeCount` impl on Android is guarded not to be called here, so we are not expecting it to throw any exception. But it may be a good idea to use `.ignore()` to ensure "developers don't need to await for this action".

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
Samaritan1011001: Yes, but without this, the execution gets stuck at bufferedStream.peek
Keeping this now and will update when I get to the root of it. Will add a todo to make sure it gets done.

packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.dart
Samaritan1011001: The content of the notification is not that important to this test plus the `PushNotificationMessage` does not have a unique id to rely the equality on. I am checking for the eventType as it's more important.

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
Samaritan1011001: Added` ignore()`

PR #2847, opened 2023-04-12
chore(authenticator): Enable lints

dnys1:
- Enables `library.yaml` lints in the main Authenticator package and `app.yaml` in its example
- Fixes ARB syntax errors


Equartey (approved):

packages/authenticator/amplify_authenticator/lib/src/models/authenticator_exception.dart
```diff
@@ -37,6 +38,12 @@ class AuthenticatorException extends AmplifyException {
     );
   }
 
+  const AuthenticatorException.customAuth()
```
Jordan-Nelson, line 41: Out of scope for this PR - I think this could be removed as it appears to be unused.

Jordan-Nelson (approved):

PR #2869, opened 2023-04-14
chore(core): Fix Dart 3 build error

dnys1:
Temporary workaround for Dart 3 build issue

### DIFF
--- a/packages/aws_common/lib/src/js/readable_stream.dart
+++ b/packages/aws_common/lib/src/js/readable_stream.dart
@@ -1,6 +1,9 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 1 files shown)


Jordan-Nelson (approved):

fjnoyp (approved):
Can we give some brief context on why we are forcing Dart 2.18? 

Something like:

```
// Dart 3 introduces several breaking changes to JS interop library
// that will require us to refactor this code. 
```

dnys1:
> Can we give some brief context on why we are forcing Dart 2.18?
> 
> Something like:
> 
> ```
> // Dart 3 introduces several breaking changes to JS interop library
> // that will require us to refactor this code. 
> ```

Anything I write now could be only part of the story tomorrow. The reality is Dart 3 is going to be a game of whack-a-mole until we migrate to it. Right now, it's only a class modifiers issue, but tomorrow it could easily be JS interop starts failing hard.

PR #2886, opened 2023-04-17
chore(repo): Clean up users in e2e tests

dnys1:
- Adds teardown calls to Authenticator e2e tests. Should resolve some flakiness with phone number tests.
- Adds a scheduled cron job to the e2e infra to delete all users every day.

### DIFF
--- a/infra/lib/cleanup-users.ts
+++ b/infra/lib/cleanup-users.ts
@@ -1,0 +1,42 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { CognitoIdentityProvider, UserType } from '@aws-sdk/client-cognito-identity-provider';
+import type * as lambda from 'aws-lambda';
+
+const CLIENT = new CognitoIdentityProvider({});
+const { USER_POOL_ID } = process.env;
+
+export const handler: lambda.Handler = async (event) => {
+    console.log(`Got event: ${JSON.stringify(event, null, 2)}`);
+
+    const users: UserType[] = [];
+    try {
+        console.log('Listing users...');
+
+        let nextToken: string | undefined;
+        do {
+            const usersResp = await CLIENT.listUsers({
+                UserPoolId: USER_POOL_ID!,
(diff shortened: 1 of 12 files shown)


NikaHsn (approved):

infra/lib/common.ts
```diff
@@ -136,6 +140,30 @@ export abstract class IntegrationTestStackEnvironment<
...
+      environment: {
+        USER_POOL_ID: userPool.userPoolId,
+      },
+      timeout: cdk.Duration.minutes(5),
+    });
+    userPool.grant(lambda, "cognito-idp:ListUsers", "cognito-idp:AdminDeleteUser");
+
+    const cronRule = new events.Rule(this, 'cleanup-users-rule', {
+      // Run daily at midnight
+      schedule: events.Schedule.expression('cron(0 0 ? * MON-FRI *)'),
```
ragingsquirrel3, line 163: nit: Could this run every day, not just mon-fri? In the future, am thinking of scheduling some very long running, high-dimension matrix suites on weekends to limit pool load, and might as well keep this running. IIUC, the lambda will run fast/bill low when pool is empty.

ragingsquirrel3 (approved):
approved w a few non-blocking comments

infra/lib/cleanup-users.ts
```diff
@@ -0,0 +1,42 @@
...
+export const handler: lambda.Handler = async (event) => {
+    console.log(`Got event: ${JSON.stringify(event, null, 2)}`);
+
+    const users: UserType[] = [];
+    try {
+        console.log('Listing users...');
+
+        let nextToken: string | undefined;
+        do {
+            const usersResp = await CLIENT.listUsers({
```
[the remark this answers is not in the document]
dnys1: That's why I scheduled it to run at midnight. I think keeping the logic simple here is sufficient.

infra/lib/common.ts
dnys1: Can we revisit then if it becomes an issue?

ragingsquirrel3: Sure, that's fine. We already run the integ tests daily including weekends but this should be fine for now.

PR #2898, opened 2023-04-19

dnys1:
- Updates repo dependencies
- Updates native `pigeon` code
- Clean up some Android code
- Regenerates smithy goldens + AWS smoke test SDK


packages/api/amplify_api_android/android/src/main/kotlin/com/amazonaws/amplify/amplify_api/NativeApiPluginBindings.java
```diff
@@ -1,7 +1,7 @@
 // 
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-// Autogenerated from Pigeon (v7.1.5), do not edit directly.
+// Autogenerated from Pigeon (v9.2.4), do not edit directly.
```
Jordan-Nelson, line 4: Files that end in `Pigeon.java`or `Messages.java` will be treated as generated. We could rename this, but this should be removed soon anyway, so not really a big deal.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationsHostApiBindings.java
```diff
@@ -1,7 +1,7 @@
 // 
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-// Autogenerated from Pigeon (v7.2.1), do not edit directly.
+// Autogenerated from Pigeon (v9.2.4), do not edit directly.
```
Jordan-Nelson, line 4: This one might be worth renaming or adjusting the `.gitattributes` for. Does not have to be in this PR though. Out of scope for the changes here.

At some point we should also look into using kotlin. Maybe we update it than as I think file naming in kotlin is less restrictive than in Java.

packages/notifications/push/amplify_push_notifications/ios/Classes/PushNotificationsNativePlugin.h
```diff
@@ -1,7 +1,7 @@
 // 
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-// Autogenerated from Pigeon (v7.2.1), do not edit directly.
+// Autogenerated from Pigeon (v9.2.4), do not edit directly.
```
Jordan-Nelson, line 4: These too. Again, out of scope for the PR. Just making a note. We put these in a `pigeons` dir for some of the other packages so that we could treat anything in that dir as generated.

```
**/pigeons/*.h linguist-generated=true
**/pigeons/*.m linguist-generated=true
```

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationsHostApiBindings.java
dnys1: Agreed and agreed.

Kotlin and Swift were both recently upgraded from "experimental" to stable in pigeon.

packages/amplify_native_legacy_wrapper/android/src/main/kotlin/com/amazonaws/amplify/amplify_native_legacy_wrapper/LegacyNativePluginPigeon.java
```diff
@@ -23,38 +23,64 @@
...
+    }
     return errorList;
   }
 
   public interface Result<T> {
+    @SuppressWarnings("UnknownNullness")
     void success(T result);
 
-    void error(Throwable error);
+    void error(@NonNull Throwable error);
```
fjnoyp, line 67: Question - why do we remove @NonNull in Kotlin yet the latest Pigeon is adding @NonNull to Java code?

dnys1: `@NonNull` is only relevant in Java where the type system has no concept of nullability. Adding them in Kotlin is like adding them in Dart... doesn't add much.

fjnoyp (approved):

Issue #748, opened 2021-07-26
Facing issues in running the app into IOS, while using this amplify_auth_cognito 


yashaeologic:
Hello guys,

Facing issues in running the app into IOS, while using this amplify_auth_cognito  in the pubspec.yaml.

Step to reproduce:
Run command for iOS app: flutter run

Errors:
--------------------------------------------------------------------------------
Resolving dependencies of `Podfile`
      CDN: trunk Relative path: CocoaPods-version.yml exists! Returning local because checking is only perfomed in repo update
    [!] CocoaPods could not find compatible versions for pod "amplify_auth_cognito":
      In Podfile:
        amplify_auth_cognito (from `.symlinks/plugins/amplify_auth_cognito/ios`)

    Specs satisfying the `amplify_auth_cognito (from `.symlinks/plugins/amplify_auth_cognito/ios`)` dependency were found, but they required a higher minimum deployment target.

--------------------------------------------------------------------------------
Resolving dependencies of `Podfile`
      CDN: trunk Relative path: CocoaPods-version.yml exists! Returning local because checking is only perfomed in repo update
      CDN: trunk Relative path: all_pods_versions_9_c_5.txt exists! Returning local because checking is only perfomed in repo update
      CDN: trunk Relative path: Specs/9/c/5/Amplify/1.10.1-unstable.6/Amplify.podspec.json exists! Returning local because checking is only perfomed in repo update
      CDN: trunk Relative path: all_pods_versions_5_1_8.txt exists! Returning local because checking is only perfomed in repo update
      CDN: trunk Relative path: Specs/5/1/8/AmplifyPlugins/1.10.1-unstable.6/AmplifyPlugins.podspec.json exists! Returning local because checking is only perfomed in repo update
      CDN: trunk Relative path: all_pods_versions_c_a_d.txt exists! Returning local because checking is only perfomed in repo update
      CDN: trunk Relative path: CocoaPods-version.yml exists! Returning local because checking is only perfomed in repo update
--------------------------------------------------------------------------------

Flutter Version: flutter 2.2.2
IDE: Android Studio
Amplify Version: 
       amplify_auth_cognito: ^0.2.0
       amplify_core: ^0.2.0
       amplify_flutter: ^0.2.0

Jordan-Nelson:
[...]

PR #2906, opened 2023-04-20
Update README.md

mkbsugita:
https://github.com/aws-amplify/amplify-flutter/issues/748

*Issue #, if available:*

#748

*Description of changes:*

For avoiding this error.

```
[!] CocoaPods could not find compatible versions for pod "amplify_auth_cognito_ios":
  In Podfile:
    amplify_auth_cognito_ios (from `.symlinks/plugins/amplify_auth_cognito_ios/ios`)

Specs satisfying the `amplify_auth_cognito_ios (from `.symlinks/plugins/amplify_auth_cognito_ios/ios`)` dependency were found, but they required a higher minimum deployment target.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,13 +145,19 @@ dependencies:
     sdk: flutter
 ```
 
-7. From the terminal run
+7. Open ios/Podfile file, add:
+
+```Podfile
+platform :ios, '13.0'
+```
+
+8. From the terminal run
 
 ```bash
 flutter pub get
 ```
 
-8. In your main.dart file, add:
+9. In your main.dart file, add:
 
 ```dart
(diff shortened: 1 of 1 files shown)


dnys1:
Hi @mkbsugita, thanks for the PR! And sorry for the delayed response.

I agree this helps solve the immediate build issue on iOS, however there are a couple more changes needed which are pointed out [here](https://docs.amplify.aws/lib/project-setup/platform-setup/q/platform/flutter/). Step 9 currently links to this page which also includes setup instructions for other platforms. 

Do you think it would be more clear if we moved Step 9 higher and maybe called it out better?

abdallahshaban557:
We have made some updates to our ReadME file, and no longer need this updated PR. Please let us know if you are seeing additional things that we should be updated!

Issue #2909, opened 2023-04-20
Amplify 1.0.0 does not actually support flutter 3.3


gianmarcocalbi:
### Description

I work with flutter 3.3.10 and I have just tried to upgrade amplify_auth_cognito to the stable 1.0.0 but, even though the documentation describes amplify 1.0.0 as compatible with Flutter >=3.3, I cannot make it work due to `async` package version compatibility between `amplify_auth_cognito` itself and the `flutter_test` from the flutter 3.3.10 sdk that is bind to `async 2.9.0`.

Here is the output of `flutter pub get`.

```
Because every version of flutter_test from sdk depends on async 2.9.0 and amplify_auth_cognito 1.0.0 depends on async ^2.10.0, flutter_test from sdk is incompatible with amplify_auth_cognito 1.0.0.
And because no versions of amplify_auth_cognito match >1.0.0 <2.0.0, flutter_test from sdk is incompatible with amplify_auth_cognito ^1.0.0.
So, because wayt_biz_web depends on both amplify_auth_cognito ^1.0.0 and flutter_test from sdk, version solving failed.
Running "flutter pub get" in wayt_biz_web...                            
pub get failed (1; So, because wayt_biz_web depends on both amplify_auth_cognito ^1.0.0 and flutter_test from sdk, version solving failed.)
```

I can get it work by overriding the dependencies

```
dependency_overrides: 
  async: ^2.9.0
```

but I don't know the implications of such override.

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [X] Auth
- [ ] Authenticator
- [ ] DataStore
- [ ] Storage

### Steps to Reproduce

1. Just define these dependencies in your pubspec.yaml
```
  amplify_auth_cognito: ^1.0.0
  amplify_flutter: ^1.0.0
```

2. run `flutter pub get`

### Platforms

- [X] iOS
- [X] Android
- [X] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Flutter Version

3.3.10

### Amplify Flutter Version

1.0.0

### Deployment Method

Amplify CLI

dnys1:
Thanks for bringing this to our attention, @gianmarcocalbi. There should be no harm setting the dependency override for now. We will have a proper fix available shortly.

PR #2913, opened 2023-04-20
fix(repo): Flutter 3.3 support

dnys1:
Fixes https://github.com/aws-amplify/amplify-flutter/issues/2909. Fixes dependency resolution when using Flutter 3.3

### DIFF
--- a/.github/workflows/flutter_vm.yaml
+++ b/.github/workflows/flutter_vm.yaml
@@ -18,6 +18,10 @@ jobs:
       matrix:
         sdk:
           - stable
+        flutter-version:
+          - ""
+          # Minimum supported version
+          - "3.3.0"
     steps:
       - name: Git Checkout
         uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # 3.1.0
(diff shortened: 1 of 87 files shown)


.github/workflows/flutter_vm.yaml
```diff
@@ -72,6 +81,6 @@ jobs:
         working-directory: ${{ inputs.working-directory }}
 
       - name: Run Tests
-        if: "always() && steps.bootstrap.conclusion == 'success' && steps.testCheck.outputs.hasTests == 'true'"
+        if: "always() && steps.bootstrap.conclusion == 'success' && steps.testCheck.outputs.hasTests == 'true' && matrix.flutter-version != '3.3.0'"
```
Jordan-Nelson, line 84: Is this to save time / compute, or would they fail?

Jordan-Nelson (approved):

.github/workflows/flutter_vm.yaml
dnys1: They sometimes fail and they're not a good indicator of our library's compatibility with Flutter 3.3. For example, the Authenticator UI goldens fail but that's because Flutter itself changed.

HuiSF (approved):

PR #2928, opened 2023-04-24
infra: Improve clean up performance

dnys1:
Even at 5 min, this is timing out. This batches/parallelizes the calls to improve performance while avoiding rate limiting issues.

### DIFF
--- a/infra/lib/cleanup-users.ts
+++ b/infra/lib/cleanup-users.ts
@@ -10,33 +10,35 @@ const { USER_POOL_ID } = process.env;
 export const handler: lambda.Handler = async (event) => {
     console.log(`Got event: ${JSON.stringify(event, null, 2)}`);
 
-    const users: UserType[] = [];
     try {
         console.log('Listing users...');
 
-        let nextToken: string | undefined;
-        do {
+        while (true) {
             const usersResp = await CLIENT.listUsers({
                 UserPoolId: USER_POOL_ID!,
-                PaginationToken: nextToken,
             });
-            for (const user of usersResp.Users ?? []) {
-                users.push(user);
+            const users: UserType[] = usersResp.Users ?? [];
+            if (users.length === 0) {
+                break;
             }
-            nextToken = usersResp.PaginationToken;
-        } while (nextToken);
+
+            console.log(`Got users: ${JSON.stringify(users, null, 2)}`);
+            await Promise.all(
+                users.map(async (user) => {
+                    console.log(`Deleting user: ${user.Username}`);
+                    await CLIENT.adminDeleteUser({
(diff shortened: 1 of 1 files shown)


HuiSF (approved):

infra/lib/cleanup-users.ts
```diff
@@ -10,33 +10,35 @@ const { USER_POOL_ID } = process.env;
...
-        } while (nextToken);
-    } catch (e) {
-        console.error(`Error listing users: ${e}`);
-        return;
-    }
 
-    console.log(`Got users: ${JSON.stringify(users, null, 2)}`);
+            console.log(`Got users: ${JSON.stringify(users, null, 2)}`);
+            await Promise.all(
+                users.map(async (user) => {
```
HuiSF, line 27: How many users would be listed in one batch, do we need to add throttling within each batch to avoid `TooManyRequestsException`?

infra/lib/cleanup-users.ts
```diff
@@ -10,33 +10,35 @@ const { USER_POOL_ID } = process.env;
...
+            );
 
-    for (const user of users) {
-        console.log(`Deleting user: ${user.Username}`);
-        await CLIENT.adminDeleteUser({
-            UserPoolId: USER_POOL_ID!,
-            Username: user.Username!,
-        });
+            // Prevents `TooManyRequestsException`
+            console.log('Waiting 5 seconds...');
```
HuiSF, line 37: 5 seconds is long... I guess in normal test cases, it won't enter the second loop as the number of testing users should be low, correct?

infra/lib/cleanup-users.ts
dnys1: One batch is like 100 users I think. No need to throttle the batch.

infra/lib/cleanup-users.ts
dnys1: Yeah, I tried 1 sec and it didn't work. 5 secs seems to work and is still quicker than deleting each one individually by like 10x. In most cases, this wouldn't be called. Tests should be doing clean up and if not, the number of users created in a single day is still low.

dnys1: I think the only reason this is an issue is that before enabling this there was literally thousands of users and I need to do a manual clean up to unclog the system.

HuiSF: Thanks for the explanation 👍
If in most cases, it doesn't enter the 2nd loop, then we could move the manual delay to where that the 2nd loop starts? So if there is only 1 loop we don't need to wait for the 5 seconds :D

HuiSF (approved):

ragingsquirrel3 (approved):

PR #2941, opened 2023-04-26
chore: remove native amplify deps from amplify_flutter, auth, and api

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*

- ~~Update AmplifyPluginInterface.configure() to accept a list of plugins~~
- Move authProviders from AmplifyAPIDart to APIPluginInterface to make it accessible to DataStore
- Move all amplify native bridge code from amplify_flutter, auth, and api to amplify_datastore
- Remove amplify_api_ios and amplify_api_android (their only purpose was the native bridge)
- Remove amplify_flutter_ios and amplify_flutter_android (all code other than native bridge code was only used in DS)


packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
```diff
@@ -8,16 +8,19 @@ import AmplifyPlugins
...
     private let bridge: DataStoreBridge
     private let modelSchemaRegistry: FlutterSchemaRegistry
     private let customTypeSchemaRegistry: FlutterSchemaRegistry
     private let dataStoreObserveEventStreamHandler: DataStoreObserveEventStreamHandler?
     private let dataStoreHubEventStreamHandler: DataStoreHubEventStreamHandler?
     private var channel: FlutterMethodChannel?
     private var observeSubscription: AnyCancellable?
+    private static var nativeAuthPlugin: NativeAuthPlugin?
+    static var nativeApiPlugin: NativeApiPlugin?
+    private var cognitoPlugin: CognitoPlugin?
```
dnys1, line 22: I would make private instance vars and initialize in `init`. For `FlutterAuthProviders`, we can pass in the instance var instead.

```suggestion
    private let nativeAuthPlugin: NativeAuthPlugin
    private let nativeApiPlugin: NativeApiPlugin
    private let cognitoPlugin: CognitoPlugin
```

dnys1 (approved):
Awesome! Pushed a commit with some cleanup. Otherwise, just a few comments.

packages/amplify_core/lib/src/plugin/amplify_api_plugin_interface.dart
```diff
@@ -37,6 +37,10 @@ abstract class APIPluginInterface extends AmplifyPluginInterface {
   /// Registers an [APIAuthProvider] with this plugin.
   void registerAuthProvider(APIAuthProvider authProvider);
 
+  /// The registered [APIAuthProvider] instances.
+  @protected
+  final Map<APIAuthorizationType, APIAuthProvider> authProviders = {};
```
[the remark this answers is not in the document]
Jordan-Nelson: Done

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/AmplifyDataStorePlugin.kt
```diff
@@ -770,4 +831,77 @@ class AmplifyDataStorePlugin : FlutterPlugin, MethodCallHandler {
...
+                }
+            } catch (e: Amplify.AlreadyConfiguredException) {
+                val flutterError = FlutterError(
+                    "AmplifyAlreadyConfiguredException",
+                    e.toString()
+                )
+                callback(kotlin.Result.failure(flutterError))
+            } catch (e: AmplifyException) {
+                val flutterError = FlutterError(
+                    "postExceptionToFlutterChannel",
```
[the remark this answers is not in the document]
Jordan-Nelson: I think this was a bad copy/paste 😅 . It should be "AmplifyException". I think I copied the method name instead of the error code. Thanks for catching it. Updated.

packages/amplify_datastore/ios/Classes/SwiftAmplifyDataStorePlugin.swift
Jordan-Nelson: Updated

dnys1 (approved):

packages/amplify_datastore/lib/amplify_datastore.dart
```diff
@@ -53,6 +57,114 @@ class AmplifyDataStore extends DataStorePluginInterface {
...
+      } on Exception {
+        await updateCurrentUser(null);
+      }
+      Amplify.Hub.listen(HubChannel.Auth, (AuthHubEvent event) {
+        updateCurrentUser(event.payload);
+      });
+    }
+
+    final apiPlugin = Amplify.API.plugins.firstOrNull;
+    if (apiPlugin != null) {
```
HuiSF, line 115: In addition to the API plugin presence check, we can also check if `amplifyconfiguration.dart` has a valid `GraphQL` typed API.

Although the native API plugin surely can handle missing config gracefully, I think it's better to not let the native layer throw an unmanaged exception. (especially that iOS doesn't print the exception info in Flutter debug console.)

HuiSF (commented):
Overall LGTM, thanks for cleaning this up.
Left a comment and I'm wondering what do you think.

packages/amplify_datastore/lib/amplify_datastore.dart
Jordan-Nelson: The API category checks that there is something present in the config for API. Where you thinking of something other than that?

https://github.com/aws-amplify/amplify-flutter/blob/81b8728b66c69f3be28f50521d4d07ece055783c/packages/api/amplify_api_dart/lib/src/api_plugin_impl.dart#L74-L85

HuiSF: Yes I meant to add more specific check for DataStore checking if a GraphQL API exists in the config while automatically adding/configuring the native API plug-in.

HuiSF (approved):
Thanks!

dnys1 (approved):

PR #2950, opened 2023-04-27
chore: Update `gitattributes`

dnys1:
Makes further use of gitattributes to define Windows-specific files (which should always be checked out/in with CRLF endings). Further clarifies which files are generated (e.g. by our tooling) vs. vendored (copied in from other repos).


HuiSF (approved):
Great, thanks!

Jordan-Nelson (approved):

PR #2955, opened 2023-04-28
test(auth): Improve auth user clean up

dnys1:
E2E infra costs are pretty high right now and cleanup relies on remembering to delete users after tests right now which is pretty error prone. This PR consolidates all that clean up code so that it automatically happens.

- Updates infra dependencies
- Migrates to L2 constructs for identity pools
- Adds `amplify_auth_integration_test` package
- Migrates Auth and Authenticator to use shared package


Jordan-Nelson (dismissed):

NikaHsn:
Q: we have `amplify_integration_test` package and it has some `auth utils`.  why do we want to add a new package for `amplify_auth_integration_test` ?

dnys1:
It's so that `amplify_integration_test` can be used in all non-Flutter/Flutter packages while `amplify_auth_integration_test` can only be used in `amplify_auth_cognito`/`amplify_authenticator` since it takes a dependency on `flutter` and `amplify_auth_cognito`.

NikaHsn (approved):

Jordan-Nelson (approved):

PR #2964, opened 2023-05-02
chore(canaries): Use 3.3 compatible flutter pub add command

dnys1:
Flutter 3.3 does not support the current syntax, which is causing tests to fail. It also fails when adding dependencies already present, so remove `flutter_test` cmd since it's already included in new flutter projects.

Updates `compileSdkVersion` and Kotlin version since Flutter 3.3 defaults do not work.


dnys1:
Passing workflow: https://github.com/aws-amplify/amplify-flutter/actions/runs/4864726974?pr=2964

ragingsquirrel3 (approved):

Jordan-Nelson (approved):

PR #2965, opened 2023-05-02
feat(aft): Docs commands

dnys1:
Adds `aft docs build` and `aft docs serve` to compile docs and serve docs on a local HTTP server with auto-rebuild, respectively.

### DIFF
--- a/.github/workflows/dart_vm.yaml
+++ b/.github/workflows/dart_vm.yaml
@@ -75,7 +75,7 @@ jobs:
       - name: Bootstrap
         id: bootstrap
         timeout-minutes: 20
-        run: aft bootstrap --fail-fast --include=${{ inputs.package-name }}
+        run: aft bootstrap --fail-fast --include=${{ inputs.package-name }} --verbose
 
       - name: Check Formatting
         if: "always() && steps.bootstrap.conclusion == 'success' && matrix.sdk == 'stable'"
--- a/.github/workflows/dart_vm.yaml
+++ b/.github/workflows/dart_vm.yaml
@@ -54,7 +54,7 @@ jobs:
         uses: subosito/flutter-action@dbf1fa04f4d2e52c33185153d06cdb5443aa189d # 2.8.0
         with:
           cache: true
-          channel: stable
+          channel: beta
 
       - name: Setup Dart
         uses: dart-lang/setup-dart@007c7cbb377cdbdf8922e2f8c4172dc60fbaaf9a # main
--- a/.github/workflows/dart_vm.yaml
+++ b/.github/workflows/dart_vm.yaml
@@ -44,10 +44,25 @@ jobs:
       - name: Git Submodules
         run: git submodule update --init
 
+      # Some Dart packages require Flutter to be available, for example `amplify_core`
+      # requires it for its `doc` package. Running this before "Setup Dart" means that
+      # the `dart` executable in PATH will always be the one setup by "Setup Dart".
+      #
+      # TODO(dnys1): Create mapping of Flutter->Dart versions so we can just use Dart
+      # vended by Flutter.
+      - name: Setup Flutter
+        uses: subosito/flutter-action@dbf1fa04f4d2e52c33185153d06cdb5443aa189d # 2.8.0
+        with:
+          cache: true
+          channel: stable
+
       - name: Setup Dart
         uses: dart-lang/setup-dart@007c7cbb377cdbdf8922e2f8c4172dc60fbaaf9a # main
         with:
           sdk: ${{ matrix.sdk }}
+
+      - name: Dart Version
+        run: dart --version
 
       - name: Setup aft
         shell: bash # Run in bash regardless of platform
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,8 @@ temp/
 .fvm/
 .packages
 pubspec_overrides.yaml
+doc/api/
+**/doc/api/
 
 # IDEs
 .idea/
--- a/packages/aft/lib/src/commands/docs_command.dart
+++ b/packages/aft/lib/src/commands/docs_command.dart
@@ -1,0 +1,389 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:async';
+import 'dart:io';
+
+import 'package:aft/aft.dart';
+import 'package:aft/src/options/glob_options.dart';
+import 'package:async/async.dart';
+import 'package:aws_common/aws_common.dart';
+import 'package:mustache_template/mustache_template.dart';
+import 'package:path/path.dart' as p;
+import 'package:shelf/shelf_io.dart' as io;
+import 'package:shelf_static/shelf_static.dart';
+
+/// Command to build and deploy API documentation.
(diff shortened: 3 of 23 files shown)


.github/workflows/dart_vm.yaml
```diff
@@ -39,11 +39,26 @@ jobs:
       - name: Git Submodules
         run: git submodule update --init
 
+      # Some Dart packages require Flutter to be available, for example `amplify_core`
+      # requires it for its `doc` package. Running this before "Setup Dart" means that
+      # the `dart` executable in PATH will always be the one setup by "Setup Dart".
+      #
+      # TODO(dnys1): Create mapping of Flutter->Dart versions so we can just use Dart
+      # vended by Flutter.
+      - name: Setup Flutter
```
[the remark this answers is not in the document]
dnys1: Yeah. Even though `amplify_core` is Dart-only, the examples we showcase in its docs all reference Flutter packages. So the `doc` package in `amplify_core` needs a dependency on Flutter so that we can have Flutter-specific samples in the `amplify_core` docs.

packages/aft/lib/src/commands/docs_command.dart
```diff
@@ -0,0 +1,363 @@
...
+            background-color: #eaeaea;
+        }
+
+    </style>
+</head>
+
+<body>
+    <section id="docs">
+        <h1>Amplify Docs</h1>
+        <table>
```
HuiSF, line 342: Made the first column sticky while scrolling horizontally :P 

```suggestion
            overflow-x: auto;
        }

        table {
            table-layout: auto;
            word-wrap: break-word;
            border-collapse: collapse;
        }

        th,
        td {
            padding: 8px;
            text-align: left;
            border-bottom: 1px solid #ddd;
        }

        thead th:first-child {
            background-color: #f2f2f2;
            position: sticky;
            left: 0;
        }

        td:first-child {
            position: sticky;
            left: 0;
        }

        th {
            background-color: #f2f2f2;
            font-weight: bold;
        }

        tr:nth-child(even) {
            background-color: #f9f9f9;
        }

        tr:nth-child(even) > td:first-child {
            background-color: #f9f9f9;
        }

        tr:nth-child(odd) {
            background-color: white;
        }

        tr:nth-child(odd) > td:first-child {
            background-color: white;
        }

        tr:hover {
            background-color: #eaeaea
        }

        tr:nth-child(even):hover > td:first-child, tr:nth-child(odd):hover > td:first-child {
            background-color: #eaeaea;
        }
    </style>
</head>

<body>
    <h1>Amplify Docs</h1>
    <section id="docs">
        <table>
```

HuiSF (approved):

packages/aft/lib/src/commands/docs_command.dart
dnys1: Lol good stuff - thanks!!

packages/aft/lib/src/commands/docs_command.dart
```diff
@@ -0,0 +1,389 @@
...
+      });
+    });
+  }
+
+  @override
+  Future<void> close() => _subscription.cancel();
+}
+
+/// Index page template to make navigation between package docs easier.
+const _indexTmpl = r'''
```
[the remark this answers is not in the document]
dnys1: Once this is built as a binary, locating an HTML file becomes difficult. Embedding the file in the binary would still need to happen, but since there are not really any good options for this in the Dart community, this is the best way right now.

fjnoyp:
Would the next step be to automatically call your aft build docs command to generate docs for each release?   

Is it possible for users navigating our Github to see the index.html of that package's doc output in the current package folder?  So if they're navigating our code and want to see docs for `datastore` or `aft`, they can directly see the docs/index.html and get the information they need directly within `AmplifyFlutter/packages/datastore`?

It seems right now we consolidate all docs in the output dir.

fjnoyp:
I appear to be running the command wrong? 

I was able to run `aft docs build` in aft and get a doc file generated, but when I tried running `aft docs build` at the `packages` or root level, I get the following output: 

➜  packages git:(feat/aft/docs) ✗ aft docs build
Building docs for amplified_todo...
Building docs for infra...
Building docs for aft...
Building docs for amplify_flutter...

Which then hangs indefinitely.  Looking at the code I'm not 100% sure why amplified_todo keeps getting selected as well.  Regardless it seems I cannot run the build command for all of amplify?

fjnoyp:
I appear to be running the command wrong? 

I was able to run `aft docs build` in aft and get a doc file generated, but when I tried running `aft docs build` at the `packages` or root level, I get the following output: 

```
➜  packages git:(feat/aft/docs) ✗ aft docs build
Building docs for amplified_todo...
Building docs for infra...
Building docs for aft...
Building docs for amplify_flutter...
```

Which then hangs indefinitely.  Looking at the code I'm not 100% sure why amplified_todo keeps getting selected as well.  Regardless it seems I cannot run the build command for all of amplify?  I've run aft clean and bootstrap.  I see that the docs files are generated for aft/amplified_todo/infra.

dnys1:
@fjnoyp 

> Would the next step be to automatically call your aft build docs command to generate docs for each release?

You mean to verify before publishing? 

Maybe. I think a next step could be to build docs for PRs and publish them to a temporary domain similar to the docs repo so that changes are more visible.

> Is it possible for users navigating our Github to see the index.html of that package's doc output in the current package folder? So if they're navigating our code and want to see docs for datastore or aft, they can directly see the docs/index.html and get the information they need directly within AmplifyFlutter/packages/datastore?

What would be the benefit of this over visiting the published docs on `pub.dev`?

> I appear to be running the command wrong?

You are running the command correctly. I've pushed some updates to make the error output more clear.

Running `aft docs build` will run it for _every_ package in the repo, though, which may take a while. I think the goal of these commands is to enable quick docs development, so running `aft docs serve` from a package directory will serve up that doc's packages and listen for changes which will auto-rebuild the docs.

`aft docs build` should probably only be used locally with an `--include` option which targets a specific subset of packages, unless you genuinely wish to see all the packages' docs.

packages/amplify_core/analysis_options.yaml
```diff
@@ -5,4 +5,3 @@ analyzer:
     public_member_api_docs: ignore
   exclude:
     - lib/**/*.g.dart
-    - doc/
```
Equartey, line 8: Can you verify we don't lose any pub points from analysis by changing this? Looks like we're currently [130/140](https://pub.dev/packages/amplify_core/score), but lets avoid a regression if possible.

dnys1: The goal is ultimately to not require any excludes which will help us catch analysis errors in CI which may show up in pub. Removing this exclude will help make sure our pub points remain high 👍

.gitignore
```diff
@@ -18,6 +18,8 @@ temp/
 .fvm/
 .packages
 pubspec_overrides.yaml
+doc/api/
+**/doc/api/
```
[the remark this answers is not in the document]
dnys1: No one is the top-level doc dir, the other applies to all sub-packages

fjnoyp (dismissed):
Thanks for the clarification there, I scoped the docs serve command to get a reasonable subset to run docs against.  Looking good so approved.

haverchuck (approved):

fjnoyp (approved):

Issue #2989, opened 2023-05-04
Amplify.configure() hangs on hot restart when using AmplifyPushNotificationsPinpoint


sbkomurcu:
### Description

I have been following the official documentation to configure both AmplifyAuthCognito and AmplifyPushNotificationsPinpoint for my Flutter app. However, I have encountered an issue where configuring both libraries simultaneously causes the Authenticator to become stuck on the loading screen. 

Interestingly, I discovered that when I comment out the line for AmplifyPushNotificationsPinpoint, the app runs without any issues.
 
Tested only on Android.

I have included the relevant code snippet below for reference:
```
import 'package:flutter/material.dart';
import 'package:amplify_analytics_pinpoint/amplify_analytics_pinpoint.dart';
import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
import 'package:amplify_authenticator/amplify_authenticator.dart';
import 'package:amplify_flutter/amplify_flutter.dart';
import 'package:amplify_push_notifications_pinpoint/amplify_push_notifications_pinpoint.dart';


import 'amplifyconfiguration.dart';

void main() async {
  WidgetsFlutterBinding.ensureInitialized();
  AmplifyLogger().logLevel = LogLevel.info;
  runApp(const MyApp());
}

class MyApp extends StatefulWidget {
  const MyApp({super.key});

  @override
  State<MyApp> createState() => _MyAppState();
}

class _MyAppState extends State<MyApp> {
  @override
  void initState() {
    super.initState();
    _configureAmplify();
  }

  void _configureAmplify() async {
    try {
      await Amplify.addPlugins([
        AmplifyAnalyticsPinpoint(),
        AmplifyAuthCognito(),

        // *** Commenting out the following line will cause the app to work as expected. ***
        //  AmplifyPushNotificationsPinpoint()
      ]);
      await Amplify.configure(amplifyconfig);
    } on AmplifyAlreadyConfiguredException {
      safePrint(
        'Tried to reconfigure Amplify; this can occur when your app restarts on Android.',
      );
    }
  }

  @override
  Widget build(BuildContext context) {
    return Authenticator(
      child: MaterialApp(
        builder: Authenticator.builder(),
        home: Scaffold(
          body: Center(
            child: Column(
              mainAxisAlignment: MainAxisAlignment.center,
              children: [
                const Text('You are logged in!'),
                ElevatedButton(
                  onPressed: () async {
                    await Amplify.Auth.signOut();
                  },
[...]

PR #2997, opened 2023-05-04
chore(authenticator): add warning for delayed asyncConfig

Jordan-Nelson:
*Issue #, if available:* Related to https://github.com/aws-amplify/amplify-flutter/issues/2989

*Description of changes:*
- Periodically log warnings from the authenticator while waiting on configuration

### DIFF
--- a/packages/authenticator/amplify_authenticator/lib/src/services/amplify_auth_service.dart
+++ b/packages/authenticator/amplify_authenticator/lib/src/services/amplify_auth_service.dart
@@ -316,13 +316,15 @@ class AmplifyAuthService
   Future<AmplifyConfig> waitForConfiguration() async {
     final timer = Timer.periodic(const Duration(seconds: 10), (timer) {
       logger.warn(
-        'Amplify is taking longer to configure than expected.'
+        'Amplify is taking longer than expected to configure.'
         ' Have you called `Amplify.configure()`?',
       );
     });
-    final config = await Amplify.asyncConfig;
-    timer.cancel();
-    return config;
+    try {
+      return await Amplify.asyncConfig;
+    } finally {
+      timer.cancel();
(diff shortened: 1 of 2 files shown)


Equartey (approved):
Nice QoL improvement, thanks!

packages/authenticator/amplify_authenticator/lib/src/services/amplify_auth_service.dart
```diff
@@ -311,13 +313,24 @@ class AmplifyAuthService implements AuthService {
   }
 
   @override
-  Future<AmplifyConfig> waitForConfiguration() {
-    return Amplify.asyncConfig;
+  Future<AmplifyConfig> waitForConfiguration() async {
+    final timer = Timer.periodic(const Duration(seconds: 10), (timer) {
+      logger.warn(
+        'Amplify is taking longer to configure than expected.'
```
dnys1, line 319: I think this reads slightly better

```suggestion
        'Amplify is taking longer than expected to configure.'
```

packages/authenticator/amplify_authenticator/lib/src/services/amplify_auth_service.dart
```diff
@@ -311,13 +313,24 @@ class AmplifyAuthService implements AuthService {
...
+        ' Have you called `Amplify.configure()`?',
+      );
+    });
+    final config = await Amplify.asyncConfig;
+    timer.cancel();
+    return config;
```
dnys1, line 325: `asyncConfig` can throw so we probably want to do:

```suggestion
    try {
      return await Amplify.asyncConfig;
    } finally {
      timer.cancel();
    }
```

Equartey (approved):

PR #3009, opened 2023-05-05
chore: include dev deps in workflows

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- remove example_common from the list of dev_deps of amplify_db_common_dart
- include dev dependencies in dependency graph when generating workflows, treat amplify_lints separately as it will not have dart code
- sort paths in workflow files

Changes to dev dependencies can result in built/test/lint failures. We should run CI when they change.


dnys1 (approved):

.github/workflows/amplify_api.yaml
```diff
@@ -7,30 +7,34 @@ on:
...
   pull_request:
     paths:
-      - '.github/workflows/flutter_vm.yaml'
       - '.github/workflows/amplify_api.yaml'
+      - '.github/workflows/flutter_vm.yaml'
+      - 'packages/amplify/amplify_flutter/lib/**/*.dart'
+      - 'packages/amplify/amplify_flutter/pubspec.yaml'
+      - 'packages/amplify_core/lib/**/*.dart'
+      - 'packages/amplify_core/pubspec.yaml'
+      - 'packages/amplify_datastore_plugin_interface/lib/**/*.dart'
```
ragingsquirrel3, line 16: I don't understand why this is here (realizing you did not introduce it here).

ragingsquirrel3 (approved):

.github/workflows/amplify_api.yaml
Jordan-Nelson: amplify_api depends on amplify_flutter and amplify_flutter depends on amplify_datastore_plugin_interface. I don't think amplify_flutter needs to depend on amplify_datastore_plugin_interface though. I opened a PR to remove it https://github.com/aws-amplify/amplify-flutter/pull/3015

PR #3012, opened 2023-05-05
chore(aws_common,3.0): Make AWSResult sealed

dnys1:
Adds pattern matching/functional capabilities to `AWSResult`.

### DIFF
--- a/packages/aws_common/lib/src/result/aws_result.dart
+++ b/packages/aws_common/lib/src/result/aws_result.dart
@@ -10,69 +10,141 @@ import 'package:aws_common/aws_common.dart';
 /// throw an exception if an exception occurred. See [exception] for more
 /// details.
 /// {@endtemplate}
-class AWSResult<T, E extends Exception>
-    with AWSEquatable<AWSResult<T, E>>, AWSDebuggable {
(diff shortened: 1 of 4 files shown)


packages/aws_common/lib/src/result/aws_result.dart
```diff
@@ -10,69 +10,141 @@ import 'package:aws_common/aws_common.dart';
...
 
   /// The value of the result, or null if there was an error retrieving it.
-  T? get valueOrNull {
-    switch (type) {
-      case AWSResultType.success:
-        // value will be non-null since it is required in AWSResult.success.
-        return _value!;
-      case AWSResultType.error:
-        return null;
+  V? get valueOrNull;
```
NikaHsn, line 41: given `V extends Object?` I'm not sure if we need this

dnys1: The `value` getter throws for error results whereas this returns `null`

NikaHsn (dismissed):

Jordan-Nelson (approved):

NikaHsn (approved):

PR #3013, opened 2023-05-05
fix(notifications): configure call hangs on Flutter app hot restart

HuiSF:
- add requestInitialToken native API to be called within configure and not relying on App life cycle, where when the configure is called again after Flutter app restarted, the plugin can receive token normally

*Issue #, if available:*

https://github.com/aws-amplify/amplify-flutter/issues/2989

*Description of changes:*

Added `requestInitialToken` native API to be called within configure and not relying on App life cycle, where when the configure is called again after Flutter app hot restarted, the plugin can still receive token normally.


packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
```diff
@@ -335,14 +335,26 @@ abstract class AmplifyPushNotifications
 
   Future<void> _registerDeviceWhenConfigure() async {
     late String deviceToken;
+
     try {
-      deviceToken = await _bufferedTokenStream.peek;
+      await _hostApi.requestInitialToken();
+      deviceToken =
+          await _bufferedTokenStream.peek.timeout(const Duration(seconds: 2));
```
dnys1, line 342: Is 2 seconds long enough? Can we make it 5 or 10 to be safe?

HuiSF: On iOS and Android token generation is client behavior, and doesn't require network round trip... it should be quick. But can increase the timeout surely to avoid surprises, as this timeout is for catching potential error without increasing latency for normal use cases. Will update to 5.

dnys1 (approved):

packages/notifications/push/amplify_push_notifications/lib/src/amplify_push_notifications_impl.dart
dnys1: Makes sense. Thanks!

Samaritan1011001 (approved):

Issue #2987, opened 2023-05-03
Bad state: No builder registered for SecureStorageInterface


khandaresuyash:
### Description

I have just followed the getting started guide for AWS Amplify

main.dart
`void main(List<String> args) async {
  WidgetsFlutterBinding.ensureInitialized();
  await Amplify.addPlugins([
    AmplifyAuthCognito(),
    AmplifyAPI(modelProvider: ModelProvider.instance),
  ]);
  await Amplify.configure(amplifyconfig); 
  runApp(ProviderScope(
      child: MaterialApp.router(
    routerConfig: router.config(),
  )));
}`

Now on the splash screen I am calling check login
`
class SplashPage extends ConsumerWidget {
  const SplashPage({super.key});

  @override
  Widget build(BuildContext context, WidgetRef ref) {
     ref.read(authService).checkLogin();
    return Scaffold();
  }
}
`

Check login is as follows
`class AuthService {
  final auth = AmplifyAuthCognito();
  final Ref ref;
  AuthService(this.ref);

  Future<String?> checkLogin() async {
    await Future.delayed(Duration.zero);
    try {
      final user = await auth.getCurrentUser();
      return user.userId;
    } on AuthException catch (e) {
      print(e.message);
      return null;
    }
  }
}`

I have tried many combinations I am getting the same error
`Bad state: No builder registered for SecureStorageInterface`

pubspec.yaml
` amplify_flutter: ^1.0.1
  amplify_auth_cognito: ^1.0.1
  amplify_api: ^1.0.1`

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [ ] Auth
- [ ] Authenticator
- [ ] DataStore
- [ ] Storage

### Steps to Reproduce

Mentioned above

### Platforms

- [X] iOS
- [ ] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Flutter Version

3.7.12

### Amplify Flutter Version

1.0.1

### Deployment Method

Amplify CLI

fjnoyp:
Hi @khandaresuyash thanks for posting this issue.  The Amplify categories are singletons that should only be instantiated once.  

In your `AuthService` please replace:

`final auth = AmplifyAuthCognito();`
with:
`final auth = Amplify.Auth;`

And let us know if that fixes your problem .

Jordan-Nelson:
[...]

PR #3018, opened 2023-05-09
fix: wait for addPlugin prior to configuration

Jordan-Nelson:
*Issue #, if available:* https://github.com/aws-amplify/amplify-flutter/issues/2987

*Description of changes:*
- wait for `.addPlugin()` calls to complete prior to configuring plugins

While user should await calls to addPlugins, we should gracefully handle the scenario where they do not. Currently if the .addPlugin() call is not awaited, the plugin does not get configured because the plugin has not been added at the time of configuration. However, the plugin later does get added and Amplify does successfully configure (but without the plugin). This is a weird state that we shouldn't let users get into.

### DIFF
--- a/packages/amplify_core/lib/src/amplify_class_impl.dart
+++ b/packages/amplify_core/lib/src/amplify_class_impl.dart
@@ -21,8 +21,16 @@ class AmplifyClassImpl extends AmplifyClass {
   final AmplifyAuthProviderRepository authProviderRepo =
       AmplifyAuthProviderRepository();
 
+  final _addPluginFutures = <Future<void>>[];
+
   @override
   Future<void> addPlugin(AmplifyPluginInterface plugin) {
+    final future = _addPlugin(plugin);
+    _addPluginFutures.add(future);
+    return future;
+  }
+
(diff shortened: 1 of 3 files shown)


packages/amplify_core/lib/src/amplify_class_impl.dart
```diff
@@ -64,6 +72,8 @@ class AmplifyClassImpl extends AmplifyClass {
     final amplifyConfig = AmplifyConfig.fromJson(
       (jsonDecode(config) as Map<Object?, Object?>).cast(),
     );
+    await Future.wait(_addPluginFutures);
+    _addPluginFutures.clear();
```
dnys1, line 76: Clear in `reset` as well

dnys1 (approved):

HuiSF (dismissed):

packages/amplify_core/lib/src/amplify_class_impl.dart
Jordan-Nelson: Updated.

dnys1 (approved):

HuiSF (approved):

PR #3032, opened 2023-05-11
chore(api): Model .fromJson() Refactor

Equartey:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-flutter/issues/816

*Description of changes:*
Flatens the nested data structure of `serializedData` across the method channel and the `.fromJson()` method found on codegen models. This PR is to be paired with changes in https://github.com/aws-amplify/amplify-codegen/pull/593


packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterSerializedModel.kt
```diff
@@ -23,12 +23,11 @@ data class FlutterSerializedModel(val serializedModel: SerializedModel) {
...
-        val cleanedSerializedData: Map<String, Any?> = serializedData.filterValues { it != null }
+        val cleanedSerializedData: Map<String, Any> = serializedData.filterValues { it != null } as Map<String, Any>
 
-        return mapOf(
-            "serializedData" to cleanedSerializedData,
-            "modelName" to modelName
-        )
+        val modelNameMap = mapOf("modelName" to modelName)
+
+        return cleanedSerializedData+modelNameMap
```
HuiSF, line 30: Hum, so the `modelName` key-value pair will blend in model fields and values? 
If I have a model schema

```
type AModel @model {
  id: ID!
  modelName: Int!
}
```

The `Int` value would be overridden, correct?

packages/amplify_datastore/example/integration_test/utils/setup_utils.dart
```diff
@@ -11,10 +11,10 @@ import 'package:amplify_flutter/amplify_flutter.dart';
 
 const ENABLE_CLOUD_SYNC =
     bool.fromEnvironment('ENABLE_CLOUD_SYNC', defaultValue: false);
-const DATASTORE_READY_EVENT_TIMEOUT = const Duration(minutes: 2);
-const DELAY_TO_START_DATASTORE = const Duration(milliseconds: 500);
-const DELAY_TO_CLEAR_DATASTORE = const Duration(seconds: 2);
-const DELAY_FOR_OBSERVE = const Duration(milliseconds: 100);
+const DATASTORE_READY_EVENT_TIMEOUT = Duration(minutes: 2);
```
HuiSF, line 14: 👍

packages/amplify_datastore/example/lib/main.dart
```diff
@@ -5,11 +5,12 @@ library sample_app;
 
 import 'dart:async';
 
+import 'package:amplify_core/amplify_core.dart';
 import 'package:amplify_datastore/amplify_datastore.dart';
 // Uncomment the below line to enable online sync
 // import 'package:amplify_api/amplify_api.dart';
 
-import 'package:amplify_flutter/amplify_flutter.dart';
+// import 'package:amplify_flutter/amplify_flutter.dart';
```
HuiSF, line 13: Intended?

packages/amplify_datastore/example/lib/main.dart
```diff
@@ -5,11 +5,12 @@ library sample_app;
 
 import 'dart:async';
 
+import 'package:amplify_core/amplify_core.dart';
```
HuiSF, line 8: I think we should import from `amplify_flutter` to ensure platform stuff can be correctly configured.

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/types/model/FlutterSerializedModel.kt
```diff
@@ -23,12 +23,11 @@ data class FlutterSerializedModel(val serializedModel: SerializedModel) {
 
     fun toMap(): Map<String, Any> {
 
-        val cleanedSerializedData: Map<String, Any?> = serializedData.filterValues { it != null }
+        val cleanedSerializedData: Map<String, Any> = serializedData.filterValues { it != null } as Map<String, Any>
```
HuiSF, line 26:
```suggestion
        val cleanedSerializedData: Map<String, Any> = serializedData.filterValues { it != null }.cast()
```

packages/amplify_core/lib/src/types/datastore/hub/hub_event_element.dart
```diff
@@ -29,10 +29,7 @@ Model _parseModelFromMap(
...
   final serializedElement = serializedHubEventElement['element'] as Map;
   final modelName = serializedHubEventElement['modelName'] as String;
-  final modelData = serializedElement['model'] as Map;
-  final serializedModelData =
-      (modelData['serializedData'] as Map).cast<String, dynamic>();
-  return provider
-      .getModelTypeByModelName(modelName)
-      .fromJson(serializedModelData);
+  final modelData =
+      Map<String, dynamic>.from(serializedElement['model'] as Map);
```
dnys1, line 33: Use `cast` to avoid allocating a new Map:

```suggestion
      (serializedElement['model'] as Map).cast<String, dynamic>();
```

packages/amplify_core/lib/src/types/datastore/models/subscription_event.dart
```diff
@@ -22,7 +22,7 @@ class SubscriptionEvent<T extends Model> {
 
     return SubscriptionEvent(
       item: modelType.fromJson(
-        Map<String, dynamic>.from(serializedItem['serializedData'] as Map),
+        Map<String, dynamic>.from(serializedItem),
```
dnys1, line 25:
```suggestion
        serializedItem,
```

packages/amplify_datastore/example/lib/models/BelongsToChildExplicit.dart
```diff
@@ -1,16 +1,27 @@
...
 // Generated files can be excluded from analysis in analysis_options.yaml
 // For more info, see: https://dart.dev/guides/language/analysis-options#excluding-code-from-analysis
 
-import 'package:amplify_core/amplify_core.dart';
-import 'package:flutter/foundation.dart';
-
 // ignore_for_file: public_member_api_docs, annotate_overrides, dead_code, dead_codepublic_member_api_docs, depend_on_referenced_packages, file_names, library_private_types_in_public_api, no_leading_underscores_for_library_prefixes, no_leading_underscores_for_local_identifiers, non_constant_identifier_names, null_check_on_nullable_type_parameter, prefer_adjacent_string_concatenation, prefer_const_constructors, prefer_if_null_operators, prefer_interpolation_to_compose_strings, slash_for_doc_comments, sort_child_properties_last, unnecessary_const, unnecessary_constructor_name, unnecessary_late, unnecessary_new, unnecessary_null_aware_assignments, unnecessary_nullable_for_final_variable_declarations, unnecessary_string_interpolations, use_build_context_synchronously
 
 import 'ModelProvider.dart';
+import 'package:amplify_core/amplify_core.dart';
```
dnys1, line 23: Can you regenerate these with Hui's codegen before merging?

packages/amplify_datastore/example/lib/models/Blog.dart
```diff
@@ -115,11 +126,10 @@ class Blog extends Model {
   Blog.fromJson(Map<String, dynamic> json)
       : id = json['id'],
         _name = json['name'],
-        _posts = json['posts'] is List
-            ? (json['posts'] as List)
-                .where((e) => e?['serializedData'] != null)
-                .map((e) => Post.fromJson(
-                    Map<String, dynamic>.from(e['serializedData'])))
+        _posts = json['posts'] != null
+            ? (json['posts']['items'] as List)
```
dnys1, line 130: Is `items` guaranteed to be there if `posts` is non-null?

You could add a safeguard if not:

```suggestion
            ? (json['posts']['items'] as List? ?? const [])
```

packages/amplify_datastore/example/lib/main.dart
dnys1: Yes, it's required. `amplify_core` should only be imported directly in Dart-only environments.

packages/amplify_datastore/example/lib/main.dart
```diff
@@ -177,6 +178,15 @@ class _MyAppState extends State<MyApp> {
     });
   }
 
+  getPost() async {
```
dnys1, line 181: Leaving the return type unspecified makes it implictly `dynamic`. The other code in this file sets a bad example. Also, is this supposed to be singular? This prints all the posts currently.

packages/amplify_datastore/ios/Classes/types/hub/FlutterHubElement.swift
```diff
@@ -50,7 +50,7 @@ public struct FlutterHubElement {
                 modelName: modelName
             )
             self.version = hubElement.version
-            let serializedData = model["serializedData"] as? [String: Any] ?? [:]
+            let serializedData = model as? [String: Any] ?? [:]
             self.deleted = serializedData["_deleted"] as? Bool ?? false
```
dnys1, line 54: Shouldn't be needed anymore since `self.model` is already `[String: Any]`


```suggestion
            self.deleted = self.model["_deleted"] as? Bool ?? false
```

packages/amplify_datastore/ios/Classes/types/model/FlutterSerializedModel.swift
```diff
@@ -230,30 +230,31 @@ public struct FlutterSerializedModel: Model, ModelIdentifiable, JSONValueHolder
                         continue
                     }
 
-                    result[key] = [
-                        "modelName": nextModelName,
-                        "serializedData": try generateSerializedData(
+                    let dataMap = try generateSerializedData(
```
dnys1, line 233: Similar comment to Hui's. Instead of merging `modelName` into the model (which may conflict with already-present data), it may be better to return a wrapped value or use `__typename` instead which is part of the GraphQL spec.

packages/amplify_datastore/pubspec.yaml
```diff
@@ -27,6 +27,9 @@ dev_dependencies:
   fake_async: ^1.2.0
   pigeon: ^9.2.4
 
+dependency_overrides:
+  async: ^2.10.0
```
dnys1, line 31: Shouldn't be needed

packages/api/amplify_api_dart/test/test_models/legacy/Blog.dart
```diff
@@ -15,10 +15,10 @@ import 'package:meta/meta.dart';
 
 import 'ModelProvider.dart';
 
-/** This is an auto generated class representing the Blog type in your schema. */
+/// This is an auto generated class representing the Blog type in your schema.
```
dnys1, line 18: What are we testing with these "legacy" models?

packages/api/amplify_api_dart/test/test_models/utils.dart
```diff
@@ -0,0 +1,9 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+String describeEnum(Object enumEntry) {
```
dnys1, line 4: Use `Enum.name` instead

dnys1 (commented):
Few comments - overall looking good

packages/amplify_datastore/example/lib/models/Blog.dart
Equartey: AppSync will pass an empty `items` array when `posts` is specified in the document. When `posts` is not in the document and therefore the response too, decoding returns `null` for posts.

packages/amplify_datastore/example/lib/main.dart
Equartey: Didn't mean to include this file in this PR, will be removed 👍

Jordan-Nelson (commented):
I left a few questions, but overall lgtm.

Equartey:
closing in favor of: https://github.com/aws-amplify/amplify-flutter/pull/4665

PR #3065, opened 2023-05-17
chore: remove CCI, update contributing to remove melos

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Remove Circle CI config
- Remove melos config, update docs and scripts to use aft instead of melos


.devcontainer/Dockerfile
```diff
@@ -44,5 +44,6 @@ RUN git clone --branch ${FLUTTER_CHANNEL} https://github.com/flutter/flutter.git
 	&& flutter config --no-analytics \
 	&& dart --disable-analytics
 
-# Activate melos
-RUN dart pub global activate melos
+# Activate aft
+# TODO(Jordan-Nelson): enable aft in containers
```
dnys1, line 48: What's blocking this?

dnys1 (approved):

.devcontainer/Dockerfile
Jordan-Nelson: It was failing to activate from path. I think the dir I have here is wrong, but debugging these are a pain (at least I am not aware of a good way to do it). I think ideally it would run aft bootstrap, but it isn't as necessary as it was when we were using melos since we can do a typical pub get now.

dnys1: Ah it's because the workspace is not mounted when building the Docker container, only when running it afterwards. You can set it in `devcontainer.json` as a `postStartCommand`:

```json5
{
  // ...
  "postStartCommand": "dart pub global activate -spath packages/aft"
}

Jordan-Nelson: That makes sense. Updated.

Equartey (approved):

dnys1 (approved):

PR #3119, opened 2023-05-27
chore(auth): Remove `package:oauth2` dependency

dnys1:
The Dart authors are deprecating `package:oauth2` (https://github.com/dart-lang/oauth2/issues/137). This means we can likely not expect to receive further updates. Currently, their `package:http` constraint doesn't allow the latest version (`1.0.0`) and this is the only package we use which brings `http` into our dependency tree.

Rather than force our customers to work around this, this PR has the minimum lift-and-shift required to remove `package:oauth2` and by extension, `package:http`, dependency. It does this by copying over the subset of `package:oauth2` we use and updating it to use `AWSHttpClient` instead.

Accuracy verified through CI and local e2e tests for all platforms.

### DIFF
--- a/packages/auth/amplify_auth_cognito_dart/lib/src/oauth/authorization_code_grant.dart
+++ b/packages/auth/amplify_auth_cognito_dart/lib/src/oauth/authorization_code_grant.dart
@@ -1,0 +1,353 @@
+// Copyright (c) 2012, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert';
+import 'dart:math';
+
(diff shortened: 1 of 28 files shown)


packages/auth/amplify_auth_cognito_dart/lib/src/oauth/authorization_code_grant.dart
Equartey, line 1: Should we also fork the unit tests for these files? I'm fine not taking them if you think our tests are sufficient.

Equartey (approved):
Good job reducing deps! Just had one non-blocking comment

Albert221:
`oauth2` package version `2.0.2` started to support `http v1.0.0` - https://pub.dev/packages/oauth2/changelog#202

dnys1:
Thanks for the heads up, @Albert221.

Closing in favor of https://github.com/aws-amplify/amplify-flutter/pull/3132. Will reopen when `oauth2` is officially deprecated.

PR #3134, opened 2023-06-01
fix(auth): User attribute serialization

dnys1:
Currently, the map entry is being serialized instead of the key.


Equartey (approved):
Nice find!

Jordan-Nelson (approved):

PR #3140, opened 2023-06-02
Update README.md

abdallahshaban557:
*Issue #, if available:*

updated readme slightly

*Description of changes:*

updated readme slightly

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,17 +52,13 @@ Amplify for Flutter is an open-source project and welcomes contributions from th
 - [Flutter](https://flutter.dev/docs/get-started/install)
 - [Amplify CLI](https://docs.amplify.aws/lib/project-setup/prereq/q/platform/flutter#option-1-watch-the-video-guide)
 
-#### Getting Started with Flutter app development and Amplify
-
-- Clone this repository
-- Install Amplify in a Flutter project
-- Add basic Amplify functionality to your project using one of the supported categories
+#### Getting Started Amplify Flutter
 
 1. Open your Flutter project. If you do not have an active Flutter project, you can create one after installing the [Flutter development tooling](https://flutter.dev/docs/get-started/install) and running `flutter create <project-name>` in your terminal.
 
 2. Using the Amplify CLI, run `amplify init` from the root of your project:
 
-See [Amplify CLI Installation](https://docs.amplify.aws/lib/project-setup/prereq/q/platform/flutter#install-and-configure-the-amplify-cli)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -143,57 +136,30 @@ class MyApp extends StatefulWidget {
...
-
-  // Send an event to Pinpoint
-  Future<void> _recordEvent() async {
-    final event = AnalyticsEvent('test');
-    event.customProperties.addBoolProperty('boolKey', true);
-    event.customProperties.addDoubleProperty('doubleKey', 10);
-    event.customProperties.addIntProperty('intKey', 10);
-    event.customProperties.addStringProperty('stringKey', 'stringValue');
-    await Amplify.Analytics.recordEvent(event: event);
+  void _configureAmplify() async {
```
dnys1, line 145:
```suggestion
  Future<void> _configureAmplify() async {
```

README.md
```diff
@@ -90,8 +86,7 @@ https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html
 4. Add Amplify categories (choose defaults for this example):
 
    ```bash
-   $ amplify add auth
-   $ amplify add analytics
+   $ amplify add auth   //choose default configuration after entering this command in your terminal.
```
dnys1, line 89:
```suggestion
   $ amplify add auth  # Choose default configuration after entering this command in your terminal.
```

README.md
dnys1: I would suggest we move away from the `StatefulWidget` approach to configuration and to directly configuring in `main`. I think this will save a lot of headache for users who are new to Flutter and state management.

```dart
Future<void> main() async {
  WidgetsFlutterBinding.ensureInitialized();
  await _configureAmplify();
  runApp(const MyApp());
}

Future<void> _configureAmplify() async {
  try {
    await Amplify.addPlugin(AmplifyAuthCognito());
    await Amplify.configure(amplifyconfig);
    safePrint('Successfully configured');
  } on Exception catch (e) {
    safePrint('Error configuring Amplify: $e');
  }
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
    return Authenticator(
      child: MaterialApp(
        builder: Authenticator.builder(),
        home: const Scaffold(
          body: Center(
            child: Text('You are logged in!'),
          ),
        ),
      ),
    );
  }
}
```

Jordan-Nelson: Correct me if I am mistaken, but I think you will see a white screen for about a second with this approach as configuration can take a while.

dnys1: Oh yeah... okay you're right. With Authenticator, the `StatefulWidget` approach is better.

dnys1 (approved):

Jordan-Nelson (approved):

Issue #3028, opened 2023-05-11
GraphQL subscription with AppSync custom domain fails; stream failed after AWS is configured with message "Exception from WebSocketService".


NajAkkad:
### Description

After running the application and configuring AmplifyAPI plugin, when subscribe method is called it fails and returns this exception:

```
ERROR | WebSocketBloc | Shutting down with exception: AmplifyException {
     "message": "Exception from WebSocketService.",
     "underlyingException": "WebSocketChannelException: WebSocketChannelException: WebSocketException: Connection to 
     '...'
```

### Categories

- [ ] Analytics
- [ ] API (REST)
- [X] API (GraphQL)
- [ ] Auth
- [ ] Authenticator
- [ ] DataStore
- [ ] Storage

### Steps to Reproduce

Mentioned in description

### Platforms

- [ ] iOS
- [X] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Flutter Version

3.7.11

### Amplify Flutter Version

1.0.1

### Deployment Method

Amplify CLI

fjnoyp:
Hi @NajAkkad can you please provide more information explaining what you've done so we can reproduce? 

1. Your `amplifyconfiguration.dart` file with sensitive keys removed
2. Your `graphql.schema` file. 
3. The code that causes the exception as well as relevant code before and after
4. What data is already stored online that you are trying to sync

NajAkkad:
> Hi @NajAkkad can you please provide more information explaining what you've done so we can reproduce?
> 
> 1. Your `amplifyconfiguration.dart` file with sensitive keys removed
> 2. Your `graphql.schema` file.
> 3. The code that causes the exception as well as relevant code before and after
> 4. What data is already stored online that you are trying to sync
> 5. Do you get this error on iOS as well?

Thanks for replying!

1. This is the setup method used to configure amplify plugins:

```
Future<void> setupAmplify({
  required AmplifyEnvironment environment,
}) async {
  if (Amplify.isConfigured) {
    return;
  }
  await Amplify.addPlugins([
    AmplifyAPI(),
    AmplifyAuthCognito(),
  ]);
  await Amplify.configure(environment.configuration);
}
```

2. I can't upload such file, but here's the subscription part of it

```
type Subscription {
  recordedEvents(userId: String!): EventsRecord @aws_subscribe(mutations : ["publishEvent"])
}
```

3. This is the function that is causing the issue.

```
[...]

PR #3148, opened 2023-06-06
fix(api): GraphQL subscription with custom domain formats URI correctly

ragingsquirrel3:
Fixes https://github.com/aws-amplify/amplify-flutter/issues/3028

There is a bug where GraphQL subscriptions do not work with custom domains. As documented on https://docs.aws.amazon.com/appsync/latest/devguide/custom-domain-name.html., AppSync custom domains require some runtime changes to the way the connection URI is formatted. This is automatically handled in Android and users reported this as a regression with version 1.

I tested this fix manually in Android with a real endpoint w custom domain

### DIFF
--- a/packages/api/amplify_api_dart/lib/src/decorators/web_socket_auth_utils.dart
+++ b/packages/api/amplify_api_dart/lib/src/decorators/web_socket_auth_utils.dart
@@ -13,6 +13,9 @@ import 'package:meta/meta.dart';
 
 const _appSyncHostPortion = 'appsync-api';
 const _appSyncRealtimeHostPortion = 'appsync-realtime-api';
+const _appSyncHostSuffix = 'amazonaws.com';
+const _appSyncPath = 'graphql';
+const _customDomainPath = 'graphql/realtime';
 
 // Constants for header values as noted in https://docs.aws.amazon.com/appsync/latest/devguide/real-time-websocket-client.html.
 const _requiredHeaders = {
@@ -48,17 +51,17 @@ Future<Uri> generateConnectionUri(
   var endpointUriHost = Uri.parse(config.endpoint).host;
   String path;
   if (endpointUriHost.contains(_appSyncHostPortion) &&
-      endpointUriHost.endsWith('amazonaws.com')) {
+      endpointUriHost.endsWith(_appSyncHostSuffix)) {
     // AppSync domain that contains "appsync-api" and ends with "amazonaws.com."
     // Replace "appsync-api" with "appsync-realtime-api," append "/graphql."
     endpointUriHost = endpointUriHost.replaceFirst(
       _appSyncHostPortion,
       _appSyncRealtimeHostPortion,
     );
-    path = 'graphql';
+    path = _appSyncPath;
   } else {
     // Custom domain, append "graphql/realtime" to the path like on https://docs.aws.amazon.com/appsync/latest/devguide/custom-domain-name.html.
-    path = 'graphql/realtime';
+    path = _customDomainPath;
   }
   // Return wss URI with auth query parameters.
   return Uri(
--- a/packages/api/amplify_api_dart/lib/src/decorators/web_socket_auth_utils.dart
+++ b/packages/api/amplify_api_dart/lib/src/decorators/web_socket_auth_utils.dart
@@ -11,6 +11,15 @@ import 'package:amplify_api_dart/src/graphql/web_socket/types/web_socket_types.d
 import 'package:amplify_core/amplify_core.dart';
 import 'package:meta/meta.dart';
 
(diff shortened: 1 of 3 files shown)


packages/api/amplify_api_dart/lib/src/decorators/web_socket_auth_utils.dart
```diff
@@ -36,10 +45,18 @@ Future<Uri> generateConnectionUri(
...
-  return Uri(scheme: 'wss', host: endpointUri.host, path: 'graphql').replace(
+  var path = 'graphql';
+  // Check for custom domain and format URL as documented on https://docs.aws.amazon.com/appsync/latest/devguide/custom-domain-name.html.
+  final appSyncDomainRegExp = RegExp(_appSyncEndpointPattern);
+  if (!appSyncDomainRegExp.hasMatch(config.endpoint)) {
+    endpointUri = Uri.parse(config.endpoint);
+    path = 'graphql/realtime';
+  }
+
+  return Uri(scheme: 'wss', host: endpointUri.host, path: path).replace(
```
dnys1, line 59: Don't think you need the regex, tbh. You could use `host.endsWith('.amazonaws.com')` or pattern matching:

```suggestion
  final endpointUri = Uri.parse(
    config.endpoint.replaceFirst('appsync-api', 'appsync-realtime-api'),
  );
  // Check for custom domain and format URL as documented on https://docs.aws.amazon.com/appsync/latest/devguide/custom-domain-name.html.
  final isCustomDomain = switch (endpointUri.host.split('.')) {
    [..._, 'amazonaws', 'com'] => false,
    _ => true,
  };
  return Uri(
    scheme: 'wss',
    host: endpointUri.host,
    path: isCustomDomain ? '/graphql/realtime' : '/graphql',
  ).replace(
```

dnys1: To be safe you could remove the `replaceAll` line as well for custom domains

Equartey: imo I think the regex is easier to read. Plus, since it's lifted from the android impl we know it covers existing customers.

ragingsquirrel3: I could go either way here but decided to err on side of simplicity in next revision here, tried to document w comments a little more, etc...

packages/api/amplify_api_dart/lib/src/decorators/web_socket_auth_utils.dart
```diff
@@ -36,14 +41,38 @@ Future<Uri> generateConnectionUri(
...
+    ).replace(
+      queryParameters: authQueryParameters,
+    );
+  }
+  // Custom domain (not AppSync); format URL as documented on https://docs.aws.amazon.com/appsync/latest/devguide/custom-domain-name.html.
+  // Appends "/graphql/realtime" path in addition to auth query parameters.
+  return Uri(
+    scheme: _wssScheme,
+    host: endpointUriHost,
+    path: 'graphql/realtime',
```
dnys1, line 73: This only differs from the above construction by `path` - can you consolidate the return so that's clear and there's no drift?

packages/api/amplify_api_dart/lib/src/decorators/web_socket_auth_utils.dart
```diff
@@ -36,14 +41,38 @@ Future<Uri> generateConnectionUri(
...
-      'header': encodedAuthHeaders,
-      'payload': base64.encode(utf8.encode(json.encode(_emptyBody))),
-    },
+  final authQueryParameters = {
+    'header': encodedAuthHeaders,
+    'payload': base64.encode(utf8.encode(json.encode(_emptyBody))),
+  };
+  // Conditionally format the URI for a) AppSync domain b) custom domain. In either
+  // case, add the auth query parameters.
+  final endpointUriHost = Uri.parse(config.endpoint).host;
```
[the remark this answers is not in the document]
ragingsquirrel3: refactored a bit to only call `replace` once

packages/api/amplify_api_dart/lib/src/decorators/web_socket_auth_utils.dart
ragingsquirrel3: tried to make that a bit clearer and only use the `Uri` constructor once with modified variables from if/else for appsync/custom domain.

Equartey (dismissed):

dnys1 (approved):

dnys1 (approved):

Equartey (approved):

PR #3150, opened 2023-06-06
fix(core): Disallowed JS tearoff

dnys1:
Dart 3.1 disallows tearoffs of JS methods.


dnys1:
Tearoff is on line 241. This fixes the issue by making the tearoff of a Dart method instead a JS method.

HuiSF (approved):

NikaHsn (approved):

PR #3153, opened 2023-06-07
chore(core): User agent enhancements

dnys1:
Tracks the category and method from which an API call originated for the purpose of engagement analytics.

### DIFF
--- a/packages/amplify_core/lib/src/http/amplify_category_method.dart
+++ b/packages/amplify_core/lib/src/http/amplify_category_method.dart
@@ -1,0 +1,159 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+@internal
+library;
+
+import 'dart:async';
+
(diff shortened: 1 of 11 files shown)


packages/amplify_core/lib/src/http/amplify_category_method.dart
```diff
@@ -0,0 +1,159 @@
...
+
+  @override
+  Category get category => Category.analytics;
+
+  @override
+  final String method;
+}
+
+enum StorageCategoryMethod with AmplifyCategoryMethod {
+  copy('37'),
```
HuiSF, line 105: A quick sanity check, the operation identifier numbers are platform specific or remain consistent across platform? Asking as I remember seeing JS has a different set.

dnys1: Yeah, these are Flutter-specific

HuiSF (approved):

Jordan-Nelson (approved):

Issue #3147, opened 2023-06-05
Access Path Style Host option in S3 getUrl


ChristopherLinnett:
The s3 bucket I need to work with is using periods "." in its name. 

I have found this is incompatible with the virtualized-style path output by Flutter's Amplify.Storage.getUrl method. Requesting an option in this method to specify the user wishes to produce the older path-style host in the signed url.

HuiSF:
Hi @ChristopherLinnett thanks for opening this issue.

Bucket name that contains dots cannot be used virtual hosted-style URL, this a known issue to S3 service (Details please see [here](https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/) (the highlighted callout)). As the Path style access is in the plan of deprecation, we highly recommend to use bucket name without dots to get around of this known issue. Any reason that you can't rename the S3 bucket?

HuiSF:
Oops didn't mean to close the issue before the actual release.

dnys1:
The fix for this has been released.

PR #3163, opened 2023-06-07
feat(storage): using path style URLs when bucket name contains dots

HuiSF:
*Issue #, if available:*

https://github.com/aws-amplify/amplify-flutter/issues/3147

*Description of changes:*

* Update the infra stack to add a bucket that has dots in its name for integration tests
* Add support of using path style URLs when need to
* throwing exceptions when attempt to use acceleration endpoint when use path style URLs

### DIFF
--- a/infra/lib/stack.ts
+++ b/infra/lib/stack.ts
@@ -249,9 +249,13 @@ export class AmplifyFlutterIntegStack extends cdk.Stack {
 
     // The Storage stack
     const storage = new StorageIntegrationTestStack(this, [
-      { environmentName: "main" },
+      {
+        environmentName: "main",
+        enableTransferAcceleration: true,
+      },
       {
         environmentName: "custom-prefix",
+        enableTransferAcceleration: true,
         prefixResolver(accessLevel, identityId) {
           switch (accessLevel) {
             case StorageAccessLevel.public:
@@ -267,6 +271,11 @@ export class AmplifyFlutterIntegStack extends cdk.Stack {
           [StorageAccessLevel.protected]: "shared",
           [StorageAccessLevel.private]: "private",
         },
+      },
+      {
+        environmentName: "dots-in-name",
+        enableTransferAcceleration: false,
+        bucketName: "amplify.integ-test.stack.com",
       },
     ]);
 
--- a/infra/lib/storage/stack.ts
+++ b/infra/lib/storage/stack.ts
@@ -1,10 +1,10 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import * as cognito_identity from "@aws-cdk/aws-cognito-identitypool-alpha";
 import * as cdk from "aws-cdk-lib";
 import { RemovalPolicy } from "aws-cdk-lib";
 import * as cognito from "aws-cdk-lib/aws-cognito";
-import * as cognito_identity from "@aws-cdk/aws-cognito-identitypool-alpha";
 import * as iam from "aws-cdk-lib/aws-iam";
 import * as lambda from "aws-cdk-lib/aws-lambda";
 import * as lambda_nodejs from "aws-cdk-lib/aws-lambda-nodejs";
@@ -50,6 +50,13 @@ interface StorageIntegrationTestEnvironmentProps
   ) => string;
 
   prefixOverrides?: Record<StorageAccessLevel, String>;
+
+  /**
+   * The name of the bucket. If not provided, it will be auto-generated.
+   */
+  bucketName?: string;
+
+  enableTransferAcceleration?: boolean;
 }
 
 export class StorageIntegrationTestStack extends IntegrationTestStack<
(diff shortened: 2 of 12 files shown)


packages/smithy/smithy_aws/lib/src/s3/s3_client_config.dart
```diff
@@ -44,7 +44,7 @@ class S3ClientConfig {
     S3ServiceConfiguration? signerConfiguration,
   }) {
     return S3ClientConfig(
-      usePathStyle: usePathStyle ?? this.useAcceleration,
+      usePathStyle: usePathStyle ?? this.usePathStyle,
```
dnys1, line 47: Oops 😅

infra/lib/storage/stack.ts
```diff
@@ -50,6 +50,10 @@ interface StorageIntegrationTestEnvironmentProps
   ) => string;
 
   prefixOverrides?: Record<StorageAccessLevel, String>;
+
+  bucketNameOverride?: string;
```
dnys1, line 54:
```suggestion
  /**
   * The name of the bucket. If not provided, it will be auto-generated.
   */
  bucketName?: string;
```

packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
```diff
@@ -22,6 +22,12 @@ const s3ControllableOperationCanceledException =
       ' exception allows you to take further action when an operation is canceled.',
 );
 
+/// The exception thrown when attempt to use acceleration endpoint with a bucket
+/// that has dots in its name.
+const accelerateEndpointUnusable = UnknownException(
```
dnys1, line 27: Do we have a service exception class? `UnknownException` doesn't seem correct

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
```diff
@@ -29,21 +29,56 @@ const testDateTimeNowOverride = #_testDateTimeNowOverride;
...
     required S3PrefixResolver prefixResolver,
     required AWSIamAmplifyAuthProvider credentialsProvider,
     required AWSLogger logger,
+    required DependencyManager dependencyManager,
     String? delimiter,
+  }) {
+    final usePathStyle = s3PluginConfig.bucket.contains('.');
+
+    if (usePathStyle) {
+      logger.warn('Your bucket name contains dots (`"."`), the StorageS3 plugin'
```
dnys1, line 43:
```suggestion
      logger.warn('Since your bucket name contains dots (`"."`), the StorageS3 plugin'
```

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
```diff
@@ -29,21 +29,56 @@ const testDateTimeNowOverride = #_testDateTimeNowOverride;
...
+          ' will use path style URLs to communicate with the S3 service. S3 Transfer'
+          ' acceleration is not supported by path style URLs. Details refer to:'
```
dnys1, line 45:
```suggestion
          ' acceleration is not supported for path style URLs. For more information, refer to:'
```

packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
HuiSF: There is no service exception for this, as `s3-accelerate` subdomain within a path style URL is unresolvable, in this case the API call returns a `NetworkException` which doesn't describe the actual issue. In addition, in the use case of `getUrl`, we need to detect this exception before an end user actually hits the non-working URL, where `NetworkException` is out of context.

Hence, we need to throw an exception before the non-working URL gets generated and used.

`UnknownException` implements `StorageException` so it can be caught by `on StorageException`, I think it's sufficient for this case.

NikaHsn (dismissed):

infra/lib/stack.ts
```diff
@@ -6,7 +6,6 @@ import { CfnOutput } from "aws-cdk-lib";
 import * as cognito from "aws-cdk-lib/aws-cognito";
 import * as wafv2 from "aws-cdk-lib/aws-wafv2";
 import { Construct } from "constructs";
-import { AnalyticsIntegrationTestStack } from "./analytics/stack";
```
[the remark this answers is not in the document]
HuiSF: Ahh I think the eslint did something to remove it.

packages/storage/amplify_storage_s3_dart/lib/src/exception/s3_storage_exception.dart
HuiSF: `ConfigurationError` is appropriate to be thrown when calling `plugin.configure`. 

`accelerateEndpointUnusable` will be thrown when calling `getUrl`, upload and download APIs (where `useAccelerateEndpoint` option is determinable). Excepting to catch only `StorageException` while calling APIs should be more straightforward without worry about a sorta unrelated exception. What do you think?

dnys1: Right, my point is it shouldn't be caught because it will always fail and the only recourse is to correct their usage of the API

HuiSF: Ahh I got what you meant. Yea making the exception failing hard is clearer.

dnys1 (approved):

NikaHsn (approved):

PR #3173, opened 2023-06-09
feat(logging): logger can register one plugin per type in a logger hierarchy

NikaHsn:
*Issue #, if available:*

*Description of changes:*

- Update registerPlugin API to allow registering one plugin per type in a logger hierarchy.
- Add getPlugin API to get plugin of an specified type registered in a logger hierarchy.

### DIFF
--- a/packages/aws_common/lib/src/logging/aws_logger.dart
+++ b/packages/aws_common/lib/src/logging/aws_logger.dart
@@ -5,6 +5,7 @@ import 'dart:async';
 
 import 'package:aws_common/aws_common.dart';
 import 'package:aws_common/src/logging/logging_ext.dart';
+import 'package:collection/collection.dart';
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
 
@@ -22,7 +23,7 @@ const zDefaultLogLevel = LogLevel.info;
 ///
 /// By default, a [SimpleLogPrinter] is registered on the root [AWSLogger]
 /// which impacts all child loggers.
-class AWSLogger implements Closeable {
+class AWSLogger with AWSDebuggable implements Closeable {
   /// Creates a top-level [AWSLogger].
   ///
   /// {@macro aws_common.logging.aws_logger}
@@ -40,6 +41,7 @@ class AWSLogger implements Closeable {
   @protected
   AWSLogger.protected(String namespace) : _logger = Logger(namespace) {
     _init(this);
+    _parent?._children.add(this);
   }
 
   static bool _initialized = false;
@@ -63,6 +65,24 @@ class AWSLogger implements Closeable {
 
   final Logger _logger;
 
+  /// Parent of this logger in the logger hierarchy.
+  AWSLogger? get _parent {
+    return activeLoggers[_logger.parent?.fullName];
+  }
+
+  /// Children of this logger in the logger hierarchy.
+  final List<AWSLogger> _children = [];
+
+  Never _pluginAlreadyRegistered(String pluginType) {
+    final loggerInstance = '$runtimeTypeName($namespace)';
+    throw StateError(
+      'A plugin of type "$pluginType" is already registered to'
+      ' "$loggerInstance" in the same logging hierarchy. Unregister the'
+      ' existing plugin from "$loggerInstance" first and then register the'
+      ' new plugin.',
+    );
+  }
+
   /// The namespace of this logger.
   String get namespace => _logger.fullName;
 
@@ -72,10 +92,31 @@ class AWSLogger implements Closeable {
     return AWSLogger('$namespace.$name');
   }
 
+  /// Returns a plugin of type [Plugin] registered to this
+  /// logger hierarchy or `null`.
+  Plugin? getPlugin<Plugin extends AWSLoggerPlugin>() {
+    final registeredPlugin = _subscriptions.keys
+            .firstWhereOrNull((element) => element.runtimeType == Plugin)
+        as Plugin?;
+    return registeredPlugin ?? _parent?.getPlugin<Plugin>();
+  }
+
   /// Registers an [AWSLoggerPlugin] to handle logs emitted by this logger
   /// instance.
-  void registerPlugin(AWSLoggerPlugin plugin) {
-    unregisterPlugin(plugin);
+  ///
+  /// Throws [StateError] if a plugin with same type is registered to this
+  /// logger hierarchy.
+  void registerPlugin<T extends AWSLoggerPlugin>(
+    T plugin,
+  ) {
+    bool hasPlugin(AWSLogger logger) =>
+        logger._subscriptions.keys.any((element) => element.runtimeType == T) ||
+        logger._children.any(hasPlugin);
+
+    if (getPlugin<T>() != null || _children.any(hasPlugin)) {
+      _pluginAlreadyRegistered(T.toString());
+    }
+
     _subscriptions[plugin] = _logger.onRecord
         .map((record) => record.toLogEntry())
         .listen(plugin.handleLogEntry);
@@ -141,6 +182,10 @@ class AWSLogger implements Closeable {
 
   @override
   void close() => unregisterAllPlugins();
+
+  @override
+  @mustBeOverridden
+  String get runtimeTypeName => 'AWSLogger';
 }
 
 /// {@template aws_common.logging.aws_logger_plugin}
--- a/packages/aws_common/test/logging/aws_logger_test.dart
+++ b/packages/aws_common/test/logging/aws_logger_test.dart
@@ -19,6 +19,8 @@ class MockLoggerPlugin extends AWSLoggerPlugin {
   }
 }
 
+class MockLoggerPlugin2 extends MockLoggerPlugin {}
+
 void main() {
   final logRecords = {
     LogLevel.verbose: LogEntry(
@@ -97,7 +99,7 @@ void main() {
 
     test('Multiple LoggerPlugins register and unregister properly', () async {
       final loggerPlugin1 = MockLoggerPlugin();
-      final loggerPlugin2 = MockLoggerPlugin();
+      final loggerPlugin2 = MockLoggerPlugin2();
       final logger = AWSLogger()
         ..registerPlugin(loggerPlugin1)
         ..registerPlugin(loggerPlugin2)
@@ -219,6 +221,77 @@ void main() {
 
       expect(loggerPlugin.timesCalled, 5);
       expect(globalLoggerPlugin.timesCalled, 0);
+    });
+
+    test('getPlugin returns registered plugin in the logger hierarchy', () {
+      final plugin = MockLoggerPlugin();
+      final logger = AWSLogger()..registerPlugin(plugin);
+      expect(logger.getPlugin<MockLoggerPlugin>(), plugin);
+
+      final childLogger = logger.createChild('child');
+      expect(childLogger.getPlugin<MockLoggerPlugin>(), plugin);
+
+      logger.unregisterPlugin(plugin);
+      expect(logger.getPlugin<MockLoggerPlugin>(), isNull);
+      expect(childLogger.getPlugin<MockLoggerPlugin>(), isNull);
+
+      childLogger.registerPlugin(plugin);
+      expect(logger.getPlugin<MockLoggerPlugin>(), isNull);
+      expect(childLogger.getPlugin<MockLoggerPlugin>(), plugin);
+    });
+
+    test(
+        'registerPlugin handles registering one plugin per type'
+        ' in the logger hierarchy', () {
+      final plugin = MockLoggerPlugin();
+      final childLogger = AWSLogger()
+          .createChild('child')
+          .createChild('grandChild')
+          .createChild('grandGrandChild');
+      expect(() => childLogger.registerPlugin(plugin), returnsNormally);
+      expect(() => childLogger.registerPlugin(plugin), throwsStateError);
+      expect(() => AWSLogger().registerPlugin(plugin), throwsStateError);
+
+      final newPlugin = MockLoggerPlugin();
+      expect(() => AWSLogger().registerPlugin(newPlugin), throwsStateError);
+      expect(() => childLogger.registerPlugin(newPlugin), throwsStateError);
+
+      expect(() => childLogger.unregisterPlugin(plugin), returnsNormally);
+      expect(childLogger.getPlugin<MockLoggerPlugin>(), isNull);
+
+      expect(() => AWSLogger().registerPlugin(newPlugin), returnsNormally);
+      expect(() => childLogger.registerPlugin(newPlugin), throwsStateError);
+      expect(AWSLogger().getPlugin<MockLoggerPlugin>(), newPlugin);
+      expect(childLogger.getPlugin<MockLoggerPlugin>(), newPlugin);
+    });
+
+    group('registerPlugin allows registering plugins of same superclass', () {
+      test('can register to the root logger', () {
+        final logger = AWSLogger();
+        final plugin1 = MockLoggerPlugin();
+        final plugin2 = MockLoggerPlugin2();
+
+        expect(() => logger.registerPlugin(plugin2), returnsNormally);
+        expect(() => logger.registerPlugin(plugin1), returnsNormally);
+
+        expect(logger.getPlugin<MockLoggerPlugin>(), plugin1);
+        expect(logger.getPlugin<MockLoggerPlugin2>(), plugin2);
+      });
+
+      test('can register to the root logger and its child', () {
+        final logger = AWSLogger();
+        final plugin1 = MockLoggerPlugin();
+        final plugin2 = MockLoggerPlugin2();
+        final childLogger = logger.createChild('child');
+
+        expect(() => childLogger.registerPlugin(plugin2), returnsNormally);
+        expect(() => logger.registerPlugin(plugin1), returnsNormally);
+
+        expect(logger.getPlugin<MockLoggerPlugin>(), plugin1);
+        expect(logger.getPlugin<MockLoggerPlugin2>(), isNull);
+        expect(childLogger.getPlugin<MockLoggerPlugin>(), plugin1);
+        expect(childLogger.getPlugin<MockLoggerPlugin2>(), plugin2);
+      });
     });
   });
 }
--- a/packages/amplify_core/lib/src/logger/amplify_logger.dart
+++ b/packages/amplify_core/lib/src/logger/amplify_logger.dart
@@ -37,6 +37,9 @@ class AmplifyLogger extends AWSLogger {
     assert(name.isNotEmpty, 'Name should not be empty');
     return AmplifyLogger('$namespace.$name');
   }
+
+  @override
+  String get runtimeTypeName => 'AmplifyLogger';
 }
 
 /// {@template amplify_core.logger.amplify_logger_plugin}
--- a/packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.mocks.dart
+++ b/packages/notifications/push/amplify_push_notifications/test/amplify_push_noitfications_impl_test.mocks.dart
@@ -57,6 +57,9 @@ class _FakeAWSLogger_2 extends _i1.SmartFake implements _i4.AWSLogger {
           parent,
           parentInvocation,
         );
+
+  @override
+  String get runtimeTypeName => '_FakeAWSLogger_2';
 }
 
 /// A class which mocks [PushNotificationsHostApi].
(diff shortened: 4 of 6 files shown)


packages/aws_common/lib/src/logging/aws_logger.dart
```diff
@@ -72,10 +87,38 @@ class AWSLogger implements Closeable {
     return AWSLogger('$namespace.$name');
   }
 
+  /// Returns a plugin of type [Plugin] registered to this
+  /// logger hierarchy or null.
```
dnys1, line 91:
```suggestion
  /// logger hierarchy or `null`.
```

packages/aws_common/test/logging/aws_logger_test.dart
```diff
@@ -220,5 +222,76 @@ void main() {
...
+    });
+
+    test('Get children returns children in the logger hierarchy', () {
+      final logger = AWSLogger();
+      final childLogger1 = logger.createChild('child-1');
+      final childLogger2 = logger.createChild('child-2');
+      final grandChildLogger = childLogger1.createChild('grandChild');
+
+      expect(logger.children.contains(childLogger1), isTrue);
+      expect(logger.children.contains(childLogger2), isTrue);
```
dnys1, line 240:
```suggestion
      expect(logger.children, unorderedEquals([childLogger1, childLogger2]));
```

packages/aws_common/test/logging/aws_logger_test.dart
```diff
@@ -220,5 +222,76 @@ void main() {
...
+      expect(childLogger1.children.contains(grandChildLogger), isTrue);
```
dnys1, line 241:
```suggestion
      expect(childLogger1.children, equals([grandChildLogger]));
```

packages/aws_common/test/logging/aws_logger_test.dart
```diff
@@ -220,5 +222,76 @@ void main() {
...
+      final loggerPlugin = MockLoggerPlugin();
+      final logger = AWSLogger()..registerPlugin(loggerPlugin);
+      expect(logger.getPlugin<MockLoggerPlugin>(), loggerPlugin);
+
+      final childLogger = logger.createChild('child');
+      expect(childLogger.getPlugin<MockLoggerPlugin>(), loggerPlugin);
+
+      childLogger.unregisterPlugin(loggerPlugin);
+      expect(childLogger.getPlugin<MockLoggerPlugin>(), loggerPlugin);
+      expect(logger.getPlugin<MockLoggerPlugin>(), loggerPlugin);
```
dnys1, line 255: It's not clear to me what this is testing. Can you add a comment or use the `reason` parameter to `expect`?

packages/aws_common/test/logging/aws_logger_test.dart
```diff
@@ -220,5 +222,76 @@ void main() {
...
+
+      logger.unregisterPlugin(loggerPlugin);
+      expect(logger.getPlugin<MockLoggerPlugin>(), isNull);
+      expect(childLogger.getPlugin<MockLoggerPlugin>(), isNull);
+
+      childLogger.registerPlugin(loggerPlugin);
+      expect(logger.getPlugin<MockLoggerPlugin>(), isNull);
+      expect(childLogger.getPlugin<MockLoggerPlugin>(), loggerPlugin);
+    });
+
```
dnys1, line 265: Can you add a test for the following:

```dart
final logger = AWSLogger();
final childLogger = logger.createChild('child');
final plugin1 = MockLoggerPlugin();
final plugin2 = MockLoggerPlugin2();

// Test 1
logger.registerPlugin(plugin1);
childLogger.registerPlugin(plugin2);

expect(logger.getPlugin<MockLoggerPlugin>(), ...);
expect(logger.getPlugin<MockLoggerPlugin2>(), ...);
expect(childLogger.getPlugin<MockLoggerPlugin>(), ...);
expect(childLogger.getPlugin<MockLoggerPlugin2>(), ...);

// Test 2 (switch registration)
logger.registerPlugin(plugin2);
childLogger.registerPlugin(plugin1);

// same expectations
```

packages/aws_common/test/logging/aws_logger_test.dart
```diff
@@ -220,5 +222,76 @@ void main() {
...
+        ..registerPlugin(plugin);
+      expect(childLogger.getPlugin<MockLoggerPlugin>(), plugin);
+
+      final newPlugin = MockLoggerPlugin();
+      AWSLogger().registerPlugin(newPlugin);
+      final parentPlugin = AWSLogger().getPlugin<MockLoggerPlugin>();
+      final childPlugin = childLogger.getPlugin<MockLoggerPlugin>();
+      expect(parentPlugin, newPlugin);
+      expect(childPlugin, newPlugin);
+      expect(newPlugin, isNot(plugin));
```
dnys1, line 283: I don't find it very intuitive that calling registerPlugin on a parent would deregister and close a plugin on the child.

packages/aws_common/lib/src/logging/aws_logger.dart
```diff
@@ -72,10 +87,38 @@ class AWSLogger implements Closeable {
...
+  ) {
+    final registeredPlugin = getPlugin<T>();
+    if (registeredPlugin != null) {
+      throw StateError(
+        'A plugin with same type is already registered in the logger hierarchy.'
+        ' Unregister the existing plugin first before registering a new plugin.',
+      );
+    }
+    // Unregisters plugin for all childeren to avoid duplicate logging.
+    for (final element in children) {
```
dnys1, line 116: Why only two levels deep?

What was the rationale for having only one plugin per type per hierarchy? I had remembered something more like the following:

```dart
final logger = AWSLogger();
final childLogger = logger.createChild('child');

final plugin1 = LoggerPlugin();
final plugin2 = LoggerPlugin();

logger.registerPlugin(plugin1);
childLogger.registerPlugin(plugin2);

logger.getPlugin<LoggerPlugin>(); // plugin1
childLogger.getPlugin<LoggerPlugin>(); // plugin2

childLogger.unregisterPlugin(plugin2);
childLogger.getPlugin<LoggerPlugin>(); // plugin1
```

packages/aws_common/lib/src/logging/aws_logger.dart
```diff
@@ -72,10 +87,38 @@ class AWSLogger implements Closeable {
...
+
   /// Registers an [AWSLoggerPlugin] to handle logs emitted by this logger
   /// instance.
-  void registerPlugin(AWSLoggerPlugin plugin) {
-    unregisterPlugin(plugin);
+  ///
+  /// Throws [StateError] if a plugin with same type is registered to this
+  /// logger hierarchy.
+  /// Unregisteres plugins with same type from all children in the hierarchy
+  /// to avoid duplicate logging.
```
dnys1, line 104:
```suggestion
  /// Throws [StateError] if a plugin with same type is registered to this
  /// logger hierarchy.
  ///
  /// Unregisters plugins with same type from all children in the hierarchy
  /// to avoid duplicate logging.
```

packages/aws_common/lib/src/logging/aws_logger.dart
NikaHsn: ops, will fix the loop.

We can do what you suggested with these considerations:
- logger.getPlugin<T>() and logger.configure(PluginKey, (plugin) {...}) returns/configures a plugin which is closest to the logger in the logger hierarchy. this can be a source of confusion for developers/customers.  
- When registering a plugin we should check if same plugin is registered in the hierarchy and not allow registering it again otherwise the plugin records same logs multiple times which is not expected/wanted.
  
We decided to allow at most on plugin per type in a logger hierarchy as we think this is the common way of using logger and logger plugin. This way getPlugin<T>() and logger.configure(PluginKey, (plugin) {...}) behavior won't cause any confusion for developers/customers.
wdyt?

dnys1: Oh yeah. I think the concession we made was that the plugin would be responsible for tracking hierarchical settings so that the settings on the root logger could differ from the child settings. Basically, we want to ensure that `AmplifyLogger()` and `AmplifyLogger.category(Category.auth)` can maintain different settings even though they're in the same hierarchy. 

If that's still the plan, then I'm okay with keeping the 1 logger type per/hierarchy but you'll need to address the subclassing issues I mentioned below e.g. when one logger plugin inherits from another. These are technically different classes but will conflict due to the implementation (i.e. checking `is Plugin`).

I would also prefer if `registerPlugin` did not automatically de-register existing plugins anymore, but instead threw an `Error` and required an explicit call to `unregisterPlugin` first since the new implicit behavior is no longer intuitive.

NikaHsn: re `Basically, we want to ensure that AmplifyLogger() and AmplifyLogger.category(Category.auth) can maintain different settings even though they're in the same hierarchy.`:  this is correct. plugin is responsible to keep track of settings for hierarchical loggers. e.g. the CW logger plugin tracks the logLevel for each logger it should send to CW.

re `subclassing issues`: that should be fixed using `.runtimeType is Plugin`.

re `registerPlugin does not automatically de-register existing plugins`: agree will update the code.

packages/aws_common/lib/src/logging/aws_logger.dart
```diff
@@ -2,14 +2,20 @@
...
 
 import 'package:aws_common/aws_common.dart';
 import 'package:aws_common/src/logging/logging_ext.dart';
+import 'package:collection/collection.dart';
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
 
 /// The default log level used by [AWSLogger].
 const zDefaultLogLevel = LogLevel.info;
+String _stateErorrMsgPluginIsRegistered(AWSLogger logger) =>
```
dnys1, line 15: Can you make this static on the logger class

dnys1:
```suggestion
String _stateErrorMsgPluginIsRegistered(AWSLogger logger) =>
```

packages/aws_common/lib/src/logging/aws_logger.dart
```diff
@@ -141,6 +207,14 @@ class AWSLogger implements Closeable {
 
   @override
   void close() => unregisterAllPlugins();
+
+  /// String value of [runtimeType]
```
dnys1, line 211:
```suggestion
  /// The string value of [runtimeType], used for debugging purposes.
```

dnys1: Or mixin `AWSDebuggable`

packages/aws_common/lib/src/logging/aws_logger.dart
```diff
@@ -72,10 +92,56 @@ class AWSLogger implements Closeable {
...
+    while (queue.isNotEmpty) {
+      final logger = queue.removeFirst();
+      if (logger._subscriptions.keys
+          .any((element) => element.runtimeType == T)) {
+        throw StateError(_stateErorrMsgPluginIsRegistered(logger));
+      }
+      if (logger._children.isNotEmpty) {
+        queue.addAll(logger._children);
+      }
+    }
```
dnys1, line 143:
```suggestion
    bool hasPlugin(AWSLogger logger) =>
        logger.getPlugin<T>() != null || logger._children.any(hasPlugin);
    if (hasPlugin(this)) {
      throw StateError(_stateErorrMsgPluginIsRegistered(this));
    }
```

packages/aws_common/test/logging/aws_logger_test.dart
dnys1: Can you add this? I don't see any tests which will evaluate the same thing (`is T` vs. `runtimeType == T`)

dnys1:
Make sure to follow Dart documentation guidelines throughout: https://dart.dev/effective-dart/documentation

packages/aws_common/lib/src/logging/aws_logger.dart
```diff
@@ -63,6 +69,20 @@ class AWSLogger implements Closeable {
 
   final Logger _logger;
 
+  /// Parent of this logger in the logger hierarchy.
+  AWSLogger? get _parent {
+    return activeLoggers[_logger.parent?.fullName];
+  }
+
+  /// Children of this logger in the logger hierarchy.
+  List<AWSLogger> get _children {
```
dnys1, line 78: This can probably be optimized -- if every time you create logger, you add the instance to its parent's `children` array, this could be static instead of built each time it's needed

packages/aws_common/lib/src/logging/aws_logger.dart
NikaHsn: the queue implementation has a fixed o(n) n = logger hierarchy depth. the recursive version is less effiecient (I think should be like o(n2)) think of a scenario that a plugin is already registered to a node in the logger hierarchy and then calling registere to the root logger.

dnys1: You're discounting the cost of creating/manipulating a Queue (which is backed by an array) which is O(n) in each usage here.

**My version**
Time complexity: `logger.getPlugin` is O(depth) for each child + children iteration is O(n) = O(n + dn) = O((d + 1) * n), which would simplify to O(n) for shallow trees.
Space complexity: O(1) (no data structures are initialized)

**Your version**
Time complexity: Each logger in the tree is visited, thus O(n + d) + time to manipulate queues (creation, adding, removing, and garbage collection) O(n) = O(n + d) but with a large factor on n.
Space complexity: O(n) for the queues

In most use cases, I would expect my version to perform better. And even if it were marginally slower, this method is not expected to be called often and the complexity of your code is not warranted.

packages/aws_common/test/logging/aws_logger_test.dart
NikaHsn: there is a test case below `registerPlugin allows registering plugins of same superclass` that covers the testing for (is T vs. runtimeType == T). 
what do we want to test with the above test case?

packages/aws_common/lib/src/logging/aws_logger.dart
NikaHsn: thanks for hashing out the details. this is the summary of comparing the two:

**Time complexity**
n= number of loggers in the hierarchy
Queue implementation: o(n)
Recursive implementation: o(nlogn)

**Space complexity**
Queue implementation can be improved by using `ListQueue<AWSLogger>(activeLoggers.length);` instead of `Queue<AWSLogger>()`. 
Recusrsive version uses Stack space.

I have a slight preference for using queue. happy to change the code if you think we should do recursive.

dnys1: The recursive code is much simpler to read and understand and avoids duplication of logic. If you want to clean up and optimize the iterative approach, then please do so, but it is currently slower in most cases and reads worse.

The recursive can trivially be made O(n), thus being quicker in all cases than the iterative as well.

```dart
bool hasPlugin(AWSLogger logger) =>
    logger._subscriptions.keys.any((element) => element.runtimeType == T) ||
    logger._children.any(hasPlugin);
if (getPlugin<T>() != null || _children.any(hasPlugin)) {
  throw StateError(_stateErorrMsgPluginIsRegistered(this));
}
```

packages/aws_common/lib/src/logging/aws_logger.dart
NikaHsn: it can't be static bc need instance data.

dnys1: However the current usage only ever uses `this` for `logger`

packages/aws_common/test/logging/aws_logger_test.dart
dnys1: If you change `runtimeType == T` to `is T` in your code all your current tests pass. The tests I'm proposing do not pass.

packages/aws_common/lib/src/logging/aws_logger.dart
NikaHsn: updated to private property and use Never. 
why do you suggest to use static method and passing logger as parameter?

packages/aws_common/test/logging/aws_logger_test.dart
NikaHsn: changing the `runtimeType == T` to `is T` cause the `registerPlugin handles registering one plugin per type in the logger hierarchy` to fail. so this use case is covered.

packages/aws_common/lib/src/logging/aws_logger.dart
dnys1: I copied over what you had. If you're only ever reporting on the current logger then you obviously don't need to pass the logger parameter.

packages/aws_common/test/logging/aws_logger_test.dart
dnys1: That test shouldn't fail when swapping the check. (nor does it fail for me - so I'm not sure why it would fail for you).

The test that should fail, if any, is the one you mentioned above: `registerPlugin allows registering plugins of same superclass`. However, this test passes when changing the type check.

Please double check. I cannot make tests fail by changing to `is T` and would recommend you add the tests mentioned to prevent a regression.

packages/aws_common/lib/src/logging/aws_logger.dart
dnys1: The point was about keeping it within `AWSLogger` not in the global namespace, and consolidating code to include the `throw` statement.

packages/aws_common/test/logging/aws_logger_test.dart
NikaHsn: `registerPlugin allows registering plugins of same superclass` did not fail with changing `== T` to `is T` because `MockLoggerPlugin2 extends MockLoggerPlugin` and registering plugin1 and then plugin2 does not make it fail. 
updated the test case to registere plugin2 first and then plugin1 and now it fails if changing `== T` to `is T`. also extended the test case to test registering to a logger and also registering to logger and its child.
now the test case fails if changing `== T` to `is T` or changing `== Plugin` to `is Plugin`.

dnys1: Yes, that's correct. Glad we're on the same page.

packages/aws_common/lib/src/logging/aws_logger.dart
```diff
@@ -63,6 +65,22 @@ class AWSLogger implements Closeable {
...
+
+  /// Children of this logger in the logger hierarchy.
+  final List<AWSLogger> _children = [];
+
+  Never get _pluginAlreadyRegistered {
+    final loggerInstance = '$runtimeTypeName($namespace)';
+    throw StateError(
+        'A plugin with same type is already registered to "$loggerInstance"'
+        ' in the same logging hierarchy. Unregister the existing plugin from'
+        ' "$loggerInstance" first and then register the new plugin.');
```
dnys1, line 81: Typically would add a trailing comma here - not sure why the analyzer didn't catch it

packages/aws_common/lib/src/logging/aws_logger.dart
```diff
@@ -63,6 +65,22 @@ class AWSLogger implements Closeable {
...
 
+  /// Parent of this logger in the logger hierarchy.
+  AWSLogger? get _parent {
+    return activeLoggers[_logger.parent?.fullName];
+  }
+
+  /// Children of this logger in the logger hierarchy.
+  final List<AWSLogger> _children = [];
+
+  Never get _pluginAlreadyRegistered {
```
dnys1, line 76:
```suggestion
  Never _pluginAlreadyRegistered() {
```

Invoking a getter for side effects looks weird.

dnys1: Consider printing `T` as well

packages/aws_common/test/logging/aws_logger_test.dart
```diff
@@ -220,5 +222,72 @@ void main() {
...
+      // test registering plugins to the logger and its child.
+      final childLogger = logger.createChild('child');
+      expect(() => childLogger.registerPlugin(plugin2), returnsNormally);
+      expect(() => logger.registerPlugin(plugin1), returnsNormally);
+
+      expect(logger.getPlugin<MockLoggerPlugin>(), plugin1);
+      expect(logger.getPlugin<MockLoggerPlugin2>(), isNull);
+      expect(childLogger.getPlugin<MockLoggerPlugin>(), plugin1);
+      expect(childLogger.getPlugin<MockLoggerPlugin2>(), plugin2);
+    });
```
dnys1, line 291: Trailing comma will help with formatting. Generally, if test names go over 80chars it's best to wrap them using adjacent strings.

dnys1 (approved):

Jordan-Nelson (dismissed):

dnys1 (approved):

Jordan-Nelson (approved):

Issue #3122, opened 2023-05-30
Amplify.signIn() is taking longer than expected in Amplify Flutter v1


tushkaty:
### Description

Hello Team,

We upgraded the amplify to latest version. The authentication with username and password taking alot of time. It was very fast earlier. 

Is it new normal?

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [X] Auth
- [ ] Authenticator
- [ ] DataStore
- [ ] Storage

### Platforms

- [X] iOS
- [X] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Flutter Version

3.3.10

### Amplify Flutter Version

1.1.0

### Deployment Method

Amplify CLI

Jordan-Nelson:
Hello @tushkaty - I see you indicated that you are targeting iOS and Android. Are you noticing a significant difference between v0 and v1 for both platforms? Are there specific methods that (signIn(), signUp(), etc.) that seem to be slower compared to v0?

tushkaty:
@Jordan-Nelson Thanks for the quick response. I see a significant delay in `      await _auth.signIn(
        username: userName,
        password: password,
      );`

Jordan-Nelson:
Okay. 6 seconds would be unusually long. I have a few follow up questions:
- Are the logs from iOS or Android?
- Is this on a physical device or a simulator / emulator?
- Are you running the app in debug or profile?
- Are you sure that the device has a stable internet connection?

tushkaty:
> * Are the logs from iOS or Android? - iOS
> * Is this on a physical device or a simulator / emulator? - Simulator
> * Are you running the app in debug or profile? - debug
> * Are you sure that the device has a stable internet connection? - Stable (Works great with 0.6 version)

Jordan-Nelson:
@tushkaty Thanks for the info.

I am seeing 1-2 seconds on a simulator (in debug mode), but I do see much longer times on a physical iPhone 13 though (~6s in profile mode and ~8 in debug). Almost all of time seems to be spent on network requests (see screenshot from Flutetr dev tools below), which will depend on the network speed of your device and how far you are from the region your Cognito back end is located in.

I am going to mark this as a bug while we investigate further.

[...]

PR #3175, opened 2023-06-10
fix(auth, authenticator): prevent fetch session during sign up

Jordan-Nelson:
*Issue #, if available:* https://github.com/aws-amplify/amplify-flutter/issues/3122

*Description of changes:*
- prevent `Auth.signIn()` from fetching the session by using `loadCredentials()` insead of `getUserPoolTokens()` and by using an AnonymousCredentialsProvider on sign in calls.
- Update the Authenticator to use getCurrentUser to check the sign in state prior to calling sign in. This aligns with the updated behavior in the auth library and will prevent unnecessary network calls to fetch the session.

*Background:*
- The sign in flow can result in unnecessary network calls. At multiple points in the sign in flow there is an attempt to retrieve credentials. If there is no unauthenticated identity at this point, the auth lib attempts to create an unauthenticated identity. This results in 2 extra calls in the case where unuathenticated identities are allowed (one to get ID and one to get credentials) and results in several additional calls in the case that unuathenticated identities are not allowed (each attempt will call get ID which will fail).

### DIFF
--- a/packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_in_state_machine.dart
+++ b/packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_in_state_machine.dart
@@ -23,6 +23,7 @@ import 'package:amplify_auth_cognito_dart/src/sdk/cognito_identity_provider.dart
 import 'package:amplify_auth_cognito_dart/src/sdk/sdk_bridge.dart';
 import 'package:amplify_auth_cognito_dart/src/state/cognito_state_machine.dart';
 import 'package:amplify_auth_cognito_dart/src/state/state.dart';
+import 'package:amplify_auth_cognito_dart/src/util/credentials_providers.dart';
 import 'package:amplify_core/amplify_core.dart';
 import 'package:async/async.dart';
 import 'package:built_collection/built_collection.dart';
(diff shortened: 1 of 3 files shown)


packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_in_state_machine.dart
```diff
@@ -593,8 +594,13 @@ final class SignInStateMachine
       }
     });
 
-    final initResponse =
-        await cognitoIdentityProvider.initiateAuth(initRequest).result;
+    final initResponse = await cognitoIdentityProvider
+        .initiateAuth(
+          initRequest,
+          // initiateAuth does not require credentials
```
dnys1, line 600:
```suggestion
          // initiateAuth does not require credentials and using the default fetchAuthSession
          // credentials provider will trigger unnecessary calls to Cognito
```

packages/auth/amplify_auth_cognito_test/test/state/sign_in_state_machine_test.dart
```diff
@@ -136,6 +136,15 @@ void main() {
         ).ignore();
 
       final signInStateMachine = stateMachine.expect(SignInStateMachine.type);
+
+      final fetchAuthSessionStateMachine = stateMachine.getOrCreate(
+        FetchAuthSessionStateMachine.type,
+      );
+
+      fetchAuthSessionStateMachine.stream.listen(
+        (_) => throw StateError('.signIn() should not fetch auth session.'),
```
dnys1, line 145:
```suggestion
        (_) => fail('.signIn() should not fetch auth session.'),
```

dnys1 (approved):

PR #3223, opened 2023-06-14
fix(analytics): Disallowed unauth access

dnys1:
- Adds category dependencies tracking to ensure categories are configured in topological order which allows plugins to be guaranteed dependent categories are already configured in their `configure` methods.
- Fixes an issue where backends which disallow unauthenticated access fail to configure but in a way which is not catchable and which allows `Amplify.configure` to succeed when it should throw an error.


packages/amplify_core/lib/src/category/amplify_push_notifications_category.dart
```diff
@@ -9,6 +9,10 @@ class PushNotificationsCategory
   @nonVirtual
   Category get category => Category.pushNotifications;
 
+  @override
+  @nonVirtual
+  Set<Category> get categoryDependencies => const {Category.analytics};
```
fjnoyp, line 14: Does push notifications need analytics to function?  I thought it was setup to work independently by importing in analytics code but not relying on the analytics category itself.

fjnoyp (approved):
Looks good, thanks for handling this.

Jordan-Nelson (approved):

PR #3225, opened 2023-06-14
ci: Add default Android workflow

dnys1:
A few things I noticed from the dependabot PRs.

1. For packages without Android unit tests, we don't run any checks when files in the `android/` folder changes. This adds a fallback workflow which just runs `flutter build` when those files change so we have *some* assurance it won't break things.
2. The integ tests workflow should run when it itself changes (which is what we do for all other workflows)


.github/workflows/amplify_db_common.android.yaml
```diff
@@ -0,0 +1,28 @@
+# Generated with aft. To update, run: `aft generate workflows`
+name: amplify_db_common Android
```
Jordan-Nelson, line 2: Can we remove the /android folder from amplify_db_common?

.github/workflows/flutter_android.build.yaml
```diff
@@ -0,0 +1,49 @@
+name: Flutter Android
```
Jordan-Nelson, line 1:
```suggestion
name: Flutter Android Build
```

Jordan-Nelson (approved):

.github/workflows/amplify_db_common.android.yaml
Jordan-Nelson: Oh I see there is a native android dep for eu.simonbinder:sqlite3-native-library:3.40.0. nvm.

PR #3232, opened 2023-06-16
ci: report canary failures to cloudwatch

fjnoyp:
*Issue #, if available:*

*Description of changes:*

Related CDK update PR: https://github.com/aws-amplify/amplify-ci-support/pull/146

Github action for sending canary failure/success to cloudwatch. 


.github/composite_actions/fetch_backends/action.yaml
```diff
@@ -25,7 +25,7 @@ runs:
         role-duration-seconds: 900
 
     - name: Create AWS profile
-      run: ./build-support/create_integration_test_profile.sh
+      run: ./build-support/create_aws_test_profile.sh
```
fjnoyp, line 28: renamed since shells script is used in cloudwatch reporting now too.

.github/workflows/amplify_canaries.yaml
```diff
@@ -1,6 +1,7 @@
 name: Amplify Canaries
 on:
   schedule:
+    # TODO(kylechen): DO NOT MERGE until clarified runtime frequency
```
fjnoyp, line 4: We need to discuss how often the canary should run. 

The UI team canaries run every 15 minutes 24/7. 

### UI team canary rough cost estimate: 
1. Each canary action takes ~5 minutes: https://github.com/aws-amplify/amplify-ui/actions/runs/5292184239
2. 24 hours / 15 minutes => 96 runs / day 
3. Unix cost multiplier = x1 
4. Total cost / day = 5 * 96 * 1 = 480 

### Amplify Flutter canaries cost estimate: 
1. Each canary action takes ~55 minutes: https://github.com/aws-amplify/amplify-flutter/actions/runs/5290286906
2. 24 hours / 15 minutes => 96 runs / day 
3. MacOS cost multiplier = x10 
4. Total cost / day = 55 * 96 * 10 = 52800

If we were to match UI team behavior our costs would be ~ x110 theirs. 

If we were to run actions at a frequency that matches their "480" cost it would be: 

48/55 runs a day (vs their 96 runs a day)

.github/workflows/amplify_canaries.yaml
```diff
@@ -156,3 +157,35 @@ jobs:
         run: |
           flutter build ios --simulator --target=integration_test/main_test.dart
           flutter test -d test integration_test/main_test.dart
+
+  log-failure-metric:
+    # Send failure data point to Cloudwatch metric CanaryTestFailure on failure
+    # See repo amplify-ci-support/github-cloudwatch-cdk for related CDK Cloudwatch setup
+    runs-on: macos-latest
```
fjnoyp, line 164: Maybe we can run on ubuntu-latest to save on costs.

dnys1: Yeah, I don't see any reason we'd run this on macos

.github/workflows/amplify_canaries.yaml
dnys1: Daily seems appropriate for our canaries. Our repo doesn't change that frequently and half the tests run on published versions.

We may actually want to split out the tests that use published versions to run more infrequently. Getting constant reminders that a published library is broken doesn't seem super helpful, especially if this is to somehow create SIM tickets.

.github/composite_actions/log_metric/action.yaml
```diff
@@ -0,0 +1,52 @@
...
+description: 'Log data point to a metric with the provided value. If the metric is not there, it will create one.'
+inputs:
+  aws-region:
+    required: true
+    description: The AWS region
+  role-to-assume:
+    required: true
+    description: The role to assume in the STS session
+  metric-name:
+    description: 'target metric name'
```
dnys1, line 11: Use consistent formatting, e.g. no quotes, descriptions as sentences

.github/composite_actions/log_metric/action.yaml
```diff
@@ -0,0 +1,52 @@
...
+      run: ./build-support/create_aws_test_profile.sh
+      shell: bash
+
+    - name: Put metric data
+      shell: bash
+      # Run bash script to put metric data.
+      # First trim the input by removing trailing or leading spaces, then removing any new line characters.
+      # Then, we do regex test to match against expected input shape. Finally, put metric data using aws cli.
+      run: |
+        metric_name_trimmed="$(echo -e "${metric_name}" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//' | tr -d '\n')"
```
dnys1, line 36: Maybe use Dart for this? This is very dense and IIUC is just doing the equivalent of `string.trim()`

.github/workflows/amplify_canaries.yaml
```diff
@@ -156,3 +157,35 @@ jobs:
...
+    environment: ci
+    needs: build
+    if: ${{ success() }}
+    steps:
+      - name: Log success data point to metric CanaryTestFailure
+        uses: ./.github/composite_actions/log_metric
+        with:
+          metric-name: CanaryTestFailure
+          value: 0
```
dnys1, line 189: Shouldn't this be more like:

```suggestion
          metric-name: CanaryTestSuccess
          value: 1
```

What does logging a `0` value do?

dnys1: Oh I see. Maybe link to this: https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/publishingMetrics.html#publishingZero

.github/composite_actions/log_metric/action.yaml
```diff
@@ -0,0 +1,52 @@
...
+  metric-name:
+    description: 'target metric name'
+    required: true
+  value:
+    description: 'the value to log to the metric'
+    required: true
+runs:
+  using: 'composite'
+  steps:
+    - name: Configure AWS credentials
```
ragingsquirrel3, line 19: What if we want to use this in a job that already has creds?

ragingsquirrel3:
Has this been tested anywhere? If so, are there example run logs we can see?

.github/composite_actions/log_metric/action.yaml
```diff
@@ -0,0 +1,52 @@
...
+      uses: aws-actions/configure-aws-credentials@67fbcbb121271f7775d2e7715933280b06314838 # 1.7.0
+      with:
+        role-to-assume: ${{ inputs.role-to-assume }}
+        aws-region: ${{ inputs.aws-region }}
+        role-duration-seconds: 900
+
+    - name: Create AWS profile
```
ragingsquirrel3, line 26: I think this can be removed here as creating the profile only needed for running amplify pull in headless env (as CLI doesn't accept token). For calls to aws sdk, the creds available as env variables should be sufficient.

ragingsquirrel3: thought about this more, I guess not a blocker here as we could add a bool param to this action to skip getting creds and could set that for some other consumer that already has creds if/when needed in future

.github/composite_actions/log_metric/action.yaml
ragingsquirrel3: thought about this more, I guess not a blocker here as we could add a bool param to this action to skip getting creds and could set that for some other consumer that already has creds if/when needed in future

.github/composite_actions/log_metric/action.yaml
```diff
@@ -0,0 +1,52 @@
+name: 'Log Metric'
+description: 'Log data point to a metric with the provided value. If the metric is not there, it will create one.'
+inputs:
```
ragingsquirrel3, line 3: request: on a high level, this should support some way for additional dimensions on the metric. At the very least, this could be hardcoded in this PR for flutter channel and version (as the canaries test across this metric) and the sent metrics should include these as dimensions so we can make the alarms on specific channels/versions if desired, or aggregate them. As written, the alarm will include all the data with no way to separate it later.

.github/workflows/amplify_canaries.yaml
```diff
@@ -156,3 +157,35 @@ jobs:
...
           flutter test -d test integration_test/main_test.dart
+
+  log-failure-metric:
+    # Send failure data point to Cloudwatch metric CanaryTestFailure on failure
+    # See repo amplify-ci-support/github-cloudwatch-cdk for related CDK Cloudwatch setup
+    runs-on: macos-latest
+    environment: ci
+    needs: build
+    if: ${{ failure() }}
+    steps:
```
ragingsquirrel3, line 168: I get this error so I think you need to call checkout here as it will not have context/env from prereq job

```
Error: Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under '/Users/runner/work/amplify-flutter/amplify-flutter/.github/composite_actions/log_metric'. Did you forget to run actions/checkout before running your local action?
```

.github/workflows/amplify_canaries.yaml
fjnoyp: Added link.

.github/composite_actions/log_metric/action.yaml
fjnoyp: Yeah, thought about this more too - in what case would a job already have credentials?

Since jobs run in their own virtual image wouldn't they not be sharing credentials anyway?  

Our current actions that require credentials are pull backend and put metric, seems to make sense to keep them self contained and not having the parent jobs having credentials which could have additional security implications.

.github/workflows/amplify_canaries.yaml
fjnoyp: That's strange, been checking the UI team's implementation and they don't need to do that.. Perhaps it's because they use environment: ci but I don't see us using environment in our setup.

.github/composite_actions/log_metric/action.yaml
ragingsquirrel3: If we wanted to use this in the integ test jobs after pulling the backends, it would already have creds in that case.

.github/composite_actions/log_metric/action.yaml
fjnoyp: Update dimensions data as agreed on yesterday.

.github/composite_actions/log_metric/action.yaml
```diff
@@ -0,0 +1,44 @@
...
+  using: 'composite'
+  steps:
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@67fbcbb121271f7775d2e7715933280b06314838 # 1.7.0
+      with:
+        role-to-assume: ${{ inputs.role-to-assume }}
+        aws-region: ${{ inputs.aws-region }}
+        role-duration-seconds: 900
+
+    - name: Setup Dart
```
ragingsquirrel3, line 34: nit/question: Can we skip this as used here bc used in a context where flutter already installed?

ragingsquirrel3 (dismissed):

.github/composite_actions/log_metric/action.yaml
```diff
@@ -0,0 +1,44 @@
+name: Log Metric
+description: Log data point to a metric with the provided value. If the metric is not there, it will create one
+# To avoid 'Credentials could not be loaded' calling workflows must include:
```
dnys1, line 3: Why is this commented and not removed?

.github/composite_actions/log_metric/action.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+    - name: Setup Dart
+      uses: dart-lang/setup-dart@196f54580e9eee2797c57e85e289339f85e6779d # main
+      with:
+        sdk: stable
+
+    - name: Run Dart script
+      # Run a Dart script to put metric data.
+      # First trim the input by removing trailing or leading spaces, then removing any new line characters.
+      # Then, we do regex test to match against expected input shape. Finally, put metric data using aws cli.
```
dnys1, line 42: This doesn't belong here. It should be in the Dart file.

.github/composite_actions/log_metric/action.yaml
fjnoyp: This entire comment is meant as a reminder/notice that the calling Github workflow must include the following permissions.  

# To avoid 'Credentials could not be loaded' calling workflows must include:
# permissions:
#  id-token: write
#  contents: read

.github/composite_actions/log_metric/action.yaml
fjnoyp: yes not necessary, good call.

.github/workflows/amplify_canaries.yaml
```diff
@@ -46,6 +50,26 @@ jobs:
...
 
+      - name: Log failing builds
+        if: ${{ failure() }}
+        uses: ./.github/composite_actions/log_metric
+        with:
+          role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
+          aws-region: ${{ secrets.AWS_REGION }}
+          metric-name: BuildCanaryTestFailure
+          value: 1
+          dimensions: channel=${{ matrix.channel }},label=CanaryBuildFailures
```
dnys1, line 61: Did you add these labels since last review? I don't understand why they would necessary. The metric name already groups values.

ragingsquirrel3: +1

ragingsquirrel3 (approved):

.github/workflows/amplify_canaries.yaml
dnys1: Missing platform too

fjnoyp: I'm a bit confused by this comment then: 
https://github.com/aws-amplify/amplify-ci-support/pull/146#discussion_r1240541017

It seems to suggest we should add a label property.

fjnoyp: Here's an internal link to the Slack message from last week in which I mentioned why we currently do not track platform for the build step: https://amzn-aws.slack.com/archives/C02JC797CGK/p1687551527260999

We can split up the build step but perhaps that's not necessary yet.

fjnoyp: Thanks, removed the label from dimensions.

ragingsquirrel3 (approved):

dnys1 (approved):

PR #3242, opened 2023-06-19
chore: bump build_value_generator

ragingsquirrel3:
Bumps `build_value_generator` to 8.6.1 in `aft.yaml` and includes result of `aft constraints udpate`.


dnys1 (approved):
Typically we would run the build throughout but I don't see anything in the changelog which would affect us.

PR #3252, opened 2023-06-20
chore(deps): Bump `drift` and `sqlite`

dnys1:
Bumps `drift` and `sqlite` deps


packages/common/amplify_db_common/android/build.gradle
```diff
@@ -41,13 +41,13 @@ android {
     }
 
     defaultConfig {
-        minSdkVersion 16
+        minSdkVersion 19
```
Equartey, line 44: Just double checking we're okay to bump this too

Equartey (approved):

packages/common/amplify_db_common/android/build.gradle
dnys1: Yeah it’s 24 everywhere else

NikaHsn (approved):

PR #3303, opened 2023-06-22
chore(infra): update aws-cdk version

ragingsquirrel3:
Addresses dependabot alert https://github.com/ragingsquirrel3/amplify-flutter/security/dependabot/3

### DIFF
--- a/infra/package.json
+++ b/infra/package.json
@@ -17,7 +17,7 @@
     "@types/jest": "^27.5.2",
     "@types/node": "^18.16.2",
     "@types/prettier": "^2.7.2",
-    "aws-cdk": "2.77.0",
+    "aws-cdk": ">=2.80.0",
     "esbuild": "^0.15.18",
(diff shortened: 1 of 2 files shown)


infra/package.json
```diff
@@ -17,7 +17,7 @@
     "@types/jest": "^27.5.2",
     "@types/node": "^18.16.2",
     "@types/prettier": "^2.7.2",
-    "aws-cdk": "2.77.0",
+    "aws-cdk": ">=2.80.0",
```
dnys1, line 20: The identity pool construct version needs to match this, which is why I was pinning. I get totally lost with npm version constraints, but I wasn't able to unpin this without getting version issues with that.

dnys1: The `aws-cdk-lib` dependency should also match this.

ragingsquirrel3: Thanks for info on both comments, will have to look into this more if this is breaking.

PR #3315, opened 2023-06-27
chore(canaries): Improve flakiness

dnys1:
- Adds timeout and retry logic for booting the iOS simulator
- Uses iOS 14 for iOS runs. iOS 13 is flaky for some reason (to do with the flutter tooling itself). With iOS 14 we are still testing a pre-concurrency version of iOS which should catch any issues related to our use of async/await. Otherwise, we don't expect any difference between iOS 13/14.


Equartey (approved):

fjnoyp (approved):
Thanks for tackling this!

PR #3317, opened 2023-06-28
ci: Speed up bootstrap

dnys1:
Reduces `aft bootstrap` time in CI by only bootstrapping necessary packages.

This improves average workflow speed by 80-90% (seriously quick now!) and should help flakiness since we're no longer DDoS'ing pub as bad in our workflows.

### DIFF
--- a/packages/aft/lib/src/commands/bootstrap_command.dart
+++ b/packages/aft/lib/src/commands/bootstrap_command.dart
@@ -91,31 +91,27 @@ const amplifyEnvironments = <String, String>{};
       for (final package in bootstrapPackages) _createEmptyConfig(package),
     ]);
     if (build) {
-      // Packages which must be built because they vendor assets required for
-      // running all downstream packages.
-      final mustBuild = [
-        'amplify_auth_cognito_dart',
-        'amplify_secure_storage_dart'
-      ].map((pkgName) => repo.allPackages[pkgName]!);
-      final buildPackages = [
-        ...bootstrapPackages,
-
-        // Include "must build" packages if any of the command packages depend
-        // on them.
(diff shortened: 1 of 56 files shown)


Equartey (dismissed):
🚀

packages/aft/lib/src/commands/bootstrap_command.dart
```diff
@@ -88,7 +88,17 @@ const amplifyEnvironments = <String, String>{};
       [for (final package in bootstrapPackages) _createEmptyConfig(package)],
     );
     if (build) {
-      for (final package in bootstrapPackages) {
+      // Packages which must be built because they vendor assets required for
```
fjnoyp: typo: because the vendor

packages/aft/lib/src/commands/bootstrap_command.dart
```diff
@@ -88,7 +88,17 @@ const amplifyEnvironments = <String, String>{};
...
+      // running all downstream packages.
```
fjnoyp: nit: Comment should be capitalized

fjnoyp (approved):
Looks good, thanks for tackling this.

packages/aft/lib/src/commands/bootstrap_command.dart
dnys1: Vendor is a verb here

packages/aft/lib/src/commands/bootstrap_command.dart
dnys1: It is capitalized?

Equartey (approved):

PR #3321, opened 2023-06-28
fix(auth): Remove `nonce` usage

dnys1:
`nonce` was previously used to protect against replay attacks, but these are primarily a concern with the implicit OAuth grant which we do not support. To unblock potential issues with custom OIDC providers, we're removing the nonce checks.

### DIFF
--- a/packages/auth/amplify_auth_cognito_dart/lib/src/jwt/src/cognito.dart
+++ b/packages/auth/amplify_auth_cognito_dart/lib/src/jwt/src/cognito.dart
@@ -103,6 +103,10 @@ extension CognitoIdToken on JsonWebToken {
   /// standards.
   ///
   /// [Reference](https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-the-id-token.html)
+  @Deprecated('''
+  Usage of the nonce claim is no longer supported in the Amplify Flutter libraries.
+  To avoid confusion, this getter will be removed in a future version of the library.
+  ''')
   String? get nonce => claims.customClaims['nonce'] as String?;
 
   /// The Cognito username.
--- a/packages/auth/amplify_auth_cognito_dart/lib/src/model/session/cognito_user_pool_tokens.dart
+++ b/packages/auth/amplify_auth_cognito_dart/lib/src/model/session/cognito_user_pool_tokens.dart
@@ -75,18 +75,12 @@ abstract class CognitoUserPoolTokens
   String get username => CognitoIdToken(idToken).username;
 
   /// Validates the tokens against the client state.
-  void validate({
-    String? nonce,
-  }) {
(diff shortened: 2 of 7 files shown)


packages/auth/amplify_auth_cognito_dart/lib/src/jwt/src/cognito.dart
```diff
@@ -89,22 +89,6 @@ extension CognitoAccessToken on JsonWebToken {
...
-  /// If you do not provide a nonce value in your request, Amazon Cognito
-  /// automatically generates and validates a nonce when you authenticate
-  /// through a third-party identity provider, then adds it as a nonce claim to
-  /// the ID token.
-  ///
-  /// The implementation of the nonce claim in Amazon Cognito is based on OIDC
-  /// standards.
-  ///
-  /// [Reference](https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-the-id-token.html)
-  String? get nonce => claims.customClaims['nonce'] as String?;
```
Jordan-Nelson, line 106: Should this be deprecated instead of removed?

dnys1: Hmm yeah, it should. Good catch.

dnys1: I actually just put it back. Customers can use it if they want.

Jordan-Nelson (approved):

packages/auth/amplify_auth_cognito_dart/lib/src/model/session/cognito_user_pool_tokens.dart
```diff
@@ -75,18 +75,12 @@ abstract class CognitoUserPoolTokens
...
-    // Missing nonce values or nonce mismatches should throw exceptions,
-    // indicating the tokens cannot be used.
-    final idTokenNonce = idToken.nonce;
-    if (nonce == null || idTokenNonce == null) {
-      throw const InvalidStateException('Missing nonce');
-    }
-    if (nonce != idTokenNonce) {
-      throw const InvalidStateException('Nonce values do not match');
-    }
+  void validate() {
```
Equartey, line 78: What is the purpose of keeping an empty function? 

Is this still being called somewhere else?

dnys1: It was recommended by Tim so that the context was not lost in the codebase and so that it wasn't reimplemented.

dnys1: It is still being called, yes.

Equartey (approved):

Issue #2592, opened 2023-01-22
Enable configuration/customization of the default country/dial code for phone number and the country selection dialog UI


mithunkavullakandiyil:
### Description

Require a better solution to choose the default country/dial code for phone number based username field.
The Country class the countryCodes constants from 'package:amplify_authenticator/src/utils/country_code.dart'' are not exported.

As a workaround I use the below code:
```
// ignore: implementation_imports
import 'package:amplify_authenticator/src/utils/country_code.dart';
...
if (state.username.isEmpty &&
    state.password.isEmpty &&
    state.country.value != '91') {
  Future.delayed(const Duration(milliseconds: 200), () {
    if (state.username.isEmpty && state.country.value != '91') {
      state.country = const Country('IN1', '91', CountryResolverKey.in1$);
    }
  });
}
```

Also provide a solution to customize the Country selection dialog. The Country selection dialog throws RenderFlex overflow error in landscape mode.

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [ ] Auth
- [X] Authenticator
- [ ] DataStore
- [ ] Storage

### Platforms

- [ ] iOS
- [ ] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Environment

```bash
flutter doctor
Doctor summary (to see all details, run flutter doctor -v):
[√] Flutter (Channel stable, 3.3.10, on Microsoft Windows [Version 10.0.22621.1105], locale en-IN)
[√] Android toolchain - develop for Android devices (Android SDK version 33.0.1)
[√] Chrome - develop for the web
[√] Visual Studio - develop for Windows (Visual Studio Community 2022 17.4.3)
[√] Android Studio (version 2021.3)
[√] VS Code, 64-bit edition (version 1.74.3)
[√] Connected device (4 available)
[√] HTTP Host Availability

• No issues found!
```

### Dependencies

```bash
flutter pub deps --no-dev --style=compact
Dart SDK 2.18.6
Flutter SDK 3.3.10
edga 1.0.3-alpha7+29

dependencies:
- amplify_analytics_pinpoint 0.6.11 [amplify_analytics_pinpoint_android amplify_analytics_pinpoint_ios amplify_core aws_common flutter meta]
- amplify_api 0.6.11 [amplify_api_android amplify_api_ios amplify_core amplify_flutter aws_common collection flutter meta plugin_platform_interface]
[...]

PR #3354, opened 2023-07-03
feat(authenticator): default dial code

khatruong2009:
*Issue #, if available:*
Issue #2592
https://github.com/aws-amplify/amplify-flutter/issues/2592

*Description of changes:*
- Changed the Country class to CountryCode to make it clearer that the class contains country codes.
- Added all the CountryCodes inside the country_code.dart file
- added defaultDialCode to the parameters in the amplify_authenticator.dart file
- added defaultDialCode to the parameters in the authenticator_state.dart file
- 


packages/authenticator/amplify_authenticator/example/lib/resolvers/localized_country_resolver.dart
```diff
@@ -6,7 +6,7 @@ import 'package:flutter/material.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 
 /// Our custom country string resolver which calls into our localizations class.
-class LocalizedCountryResolver extends CountryResolver {
+class LocalizedCountryResolver extends DialCodeResolver {
```
dnys1, line 9: Since there may be people using this, to avoid a breaking change we need to introduce a forwarding declaration like so (next to the resolver type definition):

```dart
@Deprecated('Use DialCodeResolver instead')
typedef CountryResolver = DialCodeResolver;
```

packages/authenticator/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -419,6 +425,12 @@ class Authenticator extends StatefulWidget {
   /// method.
   final AuthenticatorStep initialStep;
 
+  /// The default dial code code to use for phone number sign in or sign up.
+  /// The default is set to US (+1).
+
+  /// The options to configure the dial code selector.
```
dnys1, line 431: Use a `///` to separate lines:

```suggestion
  /// The default dial code code to use for phone number sign in or sign up.
  /// The default is set to US (+1).
  ///
  /// The options to configure the dial code selector.
```

dnys1: I would rephrase this a bit, since it's about the options as a whole now.

```suggestion
  /// The options for displaying dial codes in popup modals.
  ///
  /// Phone number fields in the Authenticator allow the user to select a dial code
  /// via a popup modal different than the default. By default, the US dial code (+1)
  /// is chosen and a list of all available dial codes is shown in the popup.
  ///
  /// Use these options to select which dial code is selected by default and, optionally,
  /// limit which options are shown in the modal.
```

packages/authenticator/amplify_authenticator/lib/src/l10n/auth_strings_resolver.dart
```diff
@@ -39,7 +35,7 @@ class AuthStringResolver {
   final ButtonResolver buttons;
 
   /// The resolver class for countries
-  final CountryResolver countries;
+  final DialCodeResolver countries;
```
dnys1, line 38: We should probably rename this too with a forwarding def as mentioned above

packages/authenticator/amplify_authenticator/lib/src/mixins/authenticator_phone_field.dart
```diff
@@ -13,26 +13,26 @@ import 'package:flutter/material.dart';
 mixin AuthenticatorPhoneFieldMixin<FieldType extends Enum,
         T extends AuthenticatorFormField<FieldType, String>>
     on AuthenticatorFormFieldState<FieldType, String, T>
-    implements SelectableConfig<CountryResolverKey, Country> {
-  late final CountryResolver _countriesResolver = stringResolver.countries;
+    implements SelectableConfig<DialCodeResolverKey, DialCode> {
+  late final DialCodeResolver _countriesResolver = stringResolver.countries;
```
dnys1, line 17:
```suggestion
  late final DialCodeResolver _dialCodeResolver = stringResolver.countries;
```

packages/authenticator/amplify_authenticator/lib/src/utils/dial_code_options.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'package:amplify_authenticator/amplify_authenticator.dart';
+
+class DialCodeOptions {
```
dnys1, line 6: Can we add comments here? You can copy over the comments mentioned above using `@template`/`@macro` tags

packages/authenticator/amplify_authenticator/lib/src/utils/dial_code.dart
```diff
@@ -0,0 +1,511 @@
...
+  DialCode._('VN', '84', DialCodeResolverKey.vn$),
+  DialCode._('VG', '1284', DialCodeResolverKey.vg$),
+  DialCode._('VI', '1340', DialCodeResolverKey.vi$),
+  DialCode._('WF', '681', DialCodeResolverKey.wf$),
+  DialCode._('YE', '967', DialCodeResolverKey.ye$),
+  DialCode._('ZM', '260', DialCodeResolverKey.zm$),
+  DialCode._('ZW', '263', DialCodeResolverKey.zw$)
+];
+
+const initialDialCode = DialCode.us;
```
dnys1, line 511: We can remove this

dnys1:
Great work! Please make sure to follow [conventional commit style](https://www.conventionalcommits.org/en/v1.0.0/) for PR titles 🚀

packages/authenticator/amplify_authenticator/example/lib/resolvers/localized_country_resolver.dart
```diff
@@ -6,7 +6,11 @@ import 'package:flutter/material.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 
 /// Our custom country string resolver which calls into our localizations class.
-class LocalizedCountryResolver extends CountryResolver {
+
+@Deprecated('Use DialCodeResolver instead')
+typedef CountryResolver = DialCodeResolver;
```
dnys1, line 11: This needs to be defined next to the `DialCodeResolver` as well so that's visible to user apps. Placing it here in the example app would not make it usable for end users.

packages/authenticator/amplify_authenticator/lib/src/l10n/auth_strings_resolver.dart
```diff
@@ -25,12 +21,12 @@ class AuthStringResolver {
...
+    DialCodeResolver? countries,
     InputResolver? inputs,
     MessageResolver? messages,
     TitleResolver? titles,
   })  : buttons = buttons ?? const ButtonResolver(),
-        countries = countries ?? const CountryResolver(),
+        countries = countries ?? const DialCodeResolver(),
         inputs = inputs ?? const InputResolver(),
         titles = titles ?? const TitleResolver(),
         messages = messages ?? const MessageResolver();
```
dnys1, line 32: As discussed, this should be:

```dart
const AuthStringResolver({
  ButtonResolver? buttons,
  // ignore: deprecated_member_use_from_same_package
  @Deprecated('Use areas instead') CountryResolver? countries,
  DialCodeResolver? areas,
  InputResolver? inputs,
  MessageResolver? messages,
  TitleResolver? titles,
})  : buttons = buttons ?? const ButtonResolver(),
      areas = areas ?? countries ?? const DialCodeResolver(),
      inputs = inputs ?? const InputResolver(),
      titles = titles ?? const TitleResolver(),
      messages = messages ?? const MessageResolver();
```

packages/authenticator/amplify_authenticator/lib/src/l10n/auth_strings_resolver.dart
dnys1: As discussed, this should be:

```dart
  /// The resolver class for area codes.
  final DialCodeResolver areas;

  /// The resolver class for area codes.
  @Deprecated('Use areas instead')
  DialCodeResolver get countries => areas;
```

packages/authenticator/amplify_authenticator/lib/src/state/authenticator_state.dart
```diff
@@ -168,9 +167,9 @@ class AuthenticatorState extends ChangeNotifier {
   String _newPassword = '';
 
   /// The value for the country code portion of the phone number field
-  Country get country => _country;
+  DialCode get dialCode => _country;
 
-  set country(Country newCountry) {
+  set dialCode(DialCode newCountry) {
     final oldCountry = _country;
```
Jordan-Nelson, line 173: I would rename this to `oldDialCode` for consistency throughout.

packages/authenticator/amplify_authenticator/lib/src/state/authenticator_state.dart
```diff
@@ -185,7 +184,8 @@ class AuthenticatorState extends ChangeNotifier {
     notifyListeners();
   }
 
-  Country _country = initialCountryCode;
+  late DialCode _country = defaultDialCode;
```
Jordan-Nelson, line 187:
```suggestion
  late DialCode _dialCode = defaultDialCode;
```

You can rename _country as part of these changes so that it is consistently called dial code everywhere.

packages/authenticator/amplify_authenticator/lib/src/utils/dial_code.dart
```diff
@@ -0,0 +1,509 @@
...
+  static const zm = DialCode._('ZM', '260', DialCodeResolverKey.zm$);
+  static const zw = DialCode._('ZW', '263', DialCodeResolverKey.zw$);
+}
+
+/// List of country information.
+///
+/// Countries are listed in alphabetical order according to the default
+/// translations, not the country code. However, countryCodes will be ordered
+/// according to the i18n content at runtime to account for custom translations.
+const dialCodes = [
```
Jordan-Nelson, line 264: I think we should make DialCode an enum. You could then remove `dialCodes` in favor of `DialCode.values`. It shouldn't change much else.

DialCode as an enum.
```dart
enum DialCode {
  us$('US', '1', DialCodeResolverKey.us$),
  af$('AF', '93', DialCodeResolverKey.af$);
  // remaining codes

  const DialCode(this.dialCode, this.value, this.key);

  final String dialCode;
  final String value;
  final DialCodeResolverKey key;
}
```

Example use of `DialCodes`.values instead of `dialCodes`
```dart
myFunction() {
  print(DialCodes.values.length);
}

File: packages/authenticator/amplify_authenticator/lib/src/utils/dial_code.dart [hunk 12]
```diff
@@ -0,0 +1,265 @@
...
+import 'package:amplify_authenticator/src/l10n/dial_code_resolver.dart';
+
+enum DialCode {
+  /// List of country information.
+  ///
+  /// Countries are listed in alphabetical order according to the default
+  /// translations, not the dial code. However, DialCodes will be ordered
+  /// according to the i18n content at runtime to account for custom translations.
+  /// Note: India, Iceland, and the Dominican Republic dial codes are slightly modified
+  /// due to their codes containing conflicts in the dart language
```
Jordan-Nelson, line 13:
```suggestion
  /// A list of [ISO 3166-1](https://en.wikipedia.org/wiki/ISO_3166-1) dial codes.
  ///
  /// The enum name will be equal to the [alpha 2 code](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Officially_assigned_code_elements)
  /// except in cases where this results in a conflict with Dart reserved
  /// keywords. In those cases, a "1" is added as a suffix.
  enum DialCode {
```

A comment that starts with `///` is a [doc comment](https://dart.dev/language/comments#documentation-comments) in dart. It will show up in customer's IDEs and will be in the generated docs on pub (example: https://pub.dev/documentation/amplify_authenticator/latest/amplify_authenticator/Authenticator-class.html).

It will be associated with whatever comes immediately after it. Currently, that is `DialCode.us`. We want this comment associated with `DialCode` though. You can move it up above `DialCode` to resolve that.

I would include links to the ISO standard that this list is based on. I would also remove the comment about how they are ordered. It isn't really important now that this is an enum.

packages/authenticator/amplify_authenticator/example/lib/main.dart
```diff
@@ -83,7 +83,7 @@ class _MyAppState extends State<MyApp> {
     // automatically configures the default for the others.
     const stringResolver = AuthStringResolver(
       buttons: LocalizedButtonResolver(),
-      countries: LocalizedCountryResolver(),
+      areas: LocalizedDialResolver(),
```
Jordan-Nelson, line 86: I don't think `areas` is correct. See https://github.com/aws-amplify/amplify-flutter/pull/3354#discussion_r1267128723.

packages/authenticator/amplify_authenticator/lib/src/l10n/auth_strings_resolver.dart
```diff
@@ -25,21 +21,27 @@ class AuthStringResolver {
   /// {@macro amplify_authenticator.auth_string_resolver}
   const AuthStringResolver({
     ButtonResolver? buttons,
-    CountryResolver? countries,
+    // ignore: deprecated_member_use_from_same_package
+    @Deprecated('Use areas instead') CountryResolver? countries,
+    DialCodeResolver? areas,
```
Jordan-Nelson, line 26: I don't think `areas` is correct. See https://github.com/aws-amplify/amplify-flutter/pull/3354#discussion_r1267128723.

packages/authenticator/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -454,7 +463,19 @@ class Authenticator extends StatefulWidget {
...
       )
-      ..add(DiagnosticsProperty<EdgeInsets>('padding', padding));
+      ..add(DiagnosticsProperty<EdgeInsets>('padding', padding))
+      ..add(
+        DiagnosticsProperty<DialCode>(
+          'defaultDialCode',
+          dialCodeOptions.defaultDialCode,
+        ),
+      );
+    properties.add(
```
dnys1, line 473: Please cascade as above

packages/authenticator/amplify_authenticator/lib/src/mixins/authenticator_phone_field.dart
```diff
@@ -97,34 +98,37 @@ mixin AuthenticatorPhoneFieldMixin<FieldType extends Enum,
...
+                    _dialCodeResolver
                         .resolve(context, country.key)
                         .toLowerCase()
                         .contains(controller.text.toLowerCase()),
               )
-              .sortedBy(
-                (country) => _countriesResolver.resolve(context, country.key),
-              );
+              .toList()
+            ..sort(
```
dnys1, line 112: Any reason you chose this over `sortedBy`? They do the same, I think, but `sortedBy` is a lot cleaner and harder to mess up.

packages/authenticator/amplify_authenticator/lib/src/state/authenticator_state.dart
```diff
@@ -20,7 +19,7 @@ typedef BlocEventPredicate = bool Function(AuthState state);
 ///
 /// Intended to be used within custom UIs for the Amplify Authenticator.
 class AuthenticatorState extends ChangeNotifier {
-  AuthenticatorState(this._authBloc) {
+  AuthenticatorState(this._authBloc, {this.defaultDialCode = DialCode.us}) {
```
dnys1, line 22: Should be `required` with no default. We don't want to accidentally not pass this.

packages/authenticator/amplify_authenticator/lib/src/utils/dial_code.dart
```diff
@@ -0,0 +1,262 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import 'package:amplify_authenticator/src/l10n/dial_code_resolver.dart';
+
+/// A list of [ISO 3166-1](https://en.wikipedia.org/wiki/ISO_3166-1) dial codes.
+///
+/// The enum name will be equal to the [alpha 2 code](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Officially_assigned_code_elements)
+/// except in cases where this results in a conflict with Dart reserved
+/// keywords. In those cases, a "1" is added as a suffix.
+enum DialCode {
```
dnys1, line 11: Awesome! 🤩

dnys1 (commented):
Agree with Jordan's comment. Let's use something besides `areas`.

packages/authenticator/amplify_authenticator/lib/src/state/authenticator_state.dart
```diff
@@ -168,24 +167,25 @@ class AuthenticatorState extends ChangeNotifier {
...
       );
     }
-    _country = newCountry;
+    _dialCode = newDialCode;
     notifyListeners();
   }
 
-  Country _country = initialCountryCode;
+  late DialCode _dialCode = defaultDialCode;
+  DialCode defaultDialCode;
```
dnys1, line 188: This value should never change once set.

```suggestion
  final DialCode defaultDialCode;
```

packages/authenticator/amplify_authenticator/lib/src/mixins/authenticator_phone_field.dart
```diff
@@ -13,51 +13,51 @@ import 'package:flutter/material.dart';
...
             .resolve(context, country.key)
             .toLowerCase()
             .contains(_searchVal.toLowerCase()),
       )
-      .sortedBy(
-        (country) => _countriesResolver.resolve(context, country.key),
-      )
-      .toList();
+      .toList()
+    ..sortedBy(
```
dnys1, line 43: Unlike `sort` which returns `void` and acts in-place on the list, `sortedBy` returns the sorted list and, thus, cannot be cascaded.

dnys1: And since `sortedBy` works on any `Iterable`, you do not need to call `iterable.toList().sortedBy(...)` which would allocate two lists. Instead, you can just call `iterable.sortedBy(...)` and this will return the sorted list.

packages/authenticator/amplify_authenticator/lib/src/l10n/auth_strings_resolver.dart
```diff
@@ -25,21 +21,27 @@ class AuthStringResolver {
   /// {@macro amplify_authenticator.auth_string_resolver}
   const AuthStringResolver({
     ButtonResolver? buttons,
-    CountryResolver? countries,
+    // ignore: deprecated_member_use_from_same_package
+    @Deprecated('Use areas instead') CountryResolver? countries,
```
dnys1, line 25:
```suggestion
    @Deprecated('Use dialCodes instead') CountryResolver? countries,
```

dnys1 (approved):

Jordan-Nelson (approved):

dnys1 (approved):

PR #3387, opened 2023-07-07
chore(repo): Improve analysis speed

dnys1:
Ignores generated SDK files in the Dart analyzer which hog resources when opening the monorepo. To get analysis on these files, we can open the packages individually in VSCode, which will ignore the top-level settings.


Equartey (approved):
🚀

Issue #3452, opened 2023-07-20
Not all datastore hub events have corresponding DataStoreHubEvent classes and OutboxMutationFailed no longer exists


dgagnon:
### Description

When building switch cases to handle all possible DataStoreHubEvent, many of the underlying events do not have corresponding DataStoreHubEvent classes:

see amplify_datastore-1.2.0-supports-only-mobile+1\lib\src\amplify_datastore_stream_controller.dart:
- ready
- subscriptionsEstablished
- syncQueriesReady

This forces developper to instead check for the eventName string.  This not ideal for 2 reasons:
- Impossible to check for exhaustiveness
- Different from the Authentication hub implementation, see https://docs.amplify.aws/lib/auth/auth-events/q/platform/flutter/

Also, Amplify-Android has an event called OutboxMutationFailed, which is very useful is it allows reacting right away when an error occurs.  I see that swift does not seem to have an equivalent.  I would be great to have this in amplify-flutter.  
See https://docs.amplify.aws/lib/datastore/datastore-events/q/platform/android/#outboxmutationfailed

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [ ] Auth
- [ ] Authenticator
- [X] DataStore
- [ ] Notifications (Push)
- [ ] Storage

### Steps to Reproduce

1. create a demo app
2. Add auth and datastore categories
3. use code from the doc

### Screenshots

```dart
  /// It is important to configure the HubChannel before configuring Amplify
  Future<DatastoreService> _configureDatastoreService() async {
    /// Avoid piling on subscriptions
    if (_isDatastoreConfigured == false) {
      /// This is a critical process and the user UI is not built yet
      /// no controllers exist yet to catch this error, so we catch it here
      try {
        hubDatastoreSubscription = Amplify.Hub.listen(HubChannel.DataStore, (hubEvent) {
          dbg("datastore hub event received: ${hubEvent.eventName.toString()}", cat: 'DATASTORE');
          switch (hubEvent.eventName) {
            /// We received a ready event, emitting [readyState]
            case "ready":
              isDatastoreReady = true;
              readyState.enter();

            /// We received a ready event, emitting [readyState]
            case "outboxMutationFailed":
              // isDatastoreReady = true;
              err("Outbox Mutation Error", null, cat: 'DATA');
              outboxMutationFailedState.enter();

            /// We received a networkStatus event, emitting [networkState] or [noNetworkState]
[...]

PR #3454, opened 2023-07-21
feat(datastore): Adds DataStoreHubEventType to DataStoreHubEvents

Equartey:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-flutter/issues/3452

*Description of changes:*
Adds `DataStoreHubEventType` to `DataStoreHubEvents` to allow for more discrete event filtering. 


packages/amplify_core/lib/src/types/datastore/hub/datstore_hub_event.dart
```diff
@@ -3,13 +3,74 @@
 
 import 'package:amplify_core/amplify_core.dart';
 
+/// {@template datastore_hub_event}
```
dnys1, line 6: These template tags must be globally unique across all packages. Always prefix with the package name by habit.

```suggestion
/// {@template amplify_core.datastore_hub_event}
```

dnys1 (dismissed):
Is there a way to add a test that ensures exhaustiveness?

packages/amplify_core/lib/src/types/datastore/hub/datstore_hub_event.dart
```diff
@@ -3,13 +3,74 @@
...
 import 'package:amplify_core/amplify_core.dart';
 
+/// {@template amplify_core.datastore_hub_event}
+/// DataStore periodically publishes state notifications onto Amplify's Hub.
+/// You can subscribe to the Hub to gain insight into the internal state of the DataStore.
+///
+/// The following example shows how to listen to DataStore Hub events:
+/// ```dart
+/// Amplify.Hub.listen(HubChannel.DataStore, (msg) {
+///  if (msg.type case DataStoreHubType.ready) {
```
dnys1, line 13: It's a good idea, but for simple enum compares `==` is better, IME. I've found a lot of places where `if case` does not provide help as to whether or not the check is correct (e.g. whether it would always fail)

dnys1 (approved):
Thanks for the quick turnaround on this!

Jordan-Nelson (approved):

PR #3455, opened 2023-07-21
chore(datastore): removed analysis options

Equartey:
*Description of changes:*
Removes the analysis options that only excluded the example app. This was causing friction when trying to use the example app and resolve local deps. 


dnys1 (approved):
Let's make sure analysis passes but LGTM 🚀

dnys1 (approved):

Jordan-Nelson (approved):

PR #3475, opened 2023-07-27
ci: Introduce Dart-based GH Action package

dnys1:
Introduces a Dart package which wraps the GitHub Action Node.js packages. This work is largely based on Dart's `setup-dart` repo (https://github.com/dart-lang/setup-dart). 

The "launch iOS simulator" action was particularly flaky and using the GH Action API should help identify the root cause a lot easier. It also provides a lot more control over the output of the actions and provides a much better debugging experience vs. shell scripts.

In order to do this, the compiled JS artifacts must be tracked in the repo which will add a bit of bloat. But overall, this should open the door to much less flaky GitHub actions as we gain better insight into the reasons why actions are failing and much better reporting.


actions/README.md
```diff
@@ -0,0 +1,20 @@
...
+- `bin/<action-name>.yaml`: The `action.yaml` file of the action which specifies the inputs
+
+Create both of these files to register a new action with name `action-name`. When building, the action will be compiled
+and copied to `.github/composite_actions/<action-name>`.
+
+## Building
+
+Building requires a dev release of the Dart compiler.
+
+To build, run `tool/build.dart`.
```
ragingsquirrel3, line 20: question: When would someone have to rebuild? Would it be when changing the dart layer or the JS sources to get recompiled? Any other times?

ragingsquirrel3 (approved):

actions/README.md
dnys1: Any time the actions themselves change. The build command runs dart2js and copies the JS output to the .github directory so they can be used in CI.

NikaHsn: would you please add these to the readme

NikaHsn (approved):

Issue #3368, opened 2023-07-07
Amplify Configuration Issue during hot restart when using Authenticator & DataStore


afnx:
### Description

As it's discussed in https://github.com/aws-amplify/amplify-flutter/issues/1658, when you do a hot restart, `Amplify.isConfigured` returns false, causing `Amplify.configure()` to throw an AmplifyAlreadyConfiguredException. Some workarounds have been suggested like muting all exceptions, but no matter which solution you choose, the Authenticator will still check if Amplify is configured by itself and think that it's not. This causes Authenticator to get stuck in the loading state and continuously print the following warning message on the console:

```
flutter: WARN  | AmplifyAuthService | Amplify is taking longer than expected to configure. Have you called `Amplify.configure()`?
```

What temporary solutions can we implement to address the Authenticator part of the issue until a PR is released to fully resolve the situation? It's currently not possible to work in this way.

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [ ] Auth
- [X] Authenticator
- [x] DataStore
- [ ] Notifications (Push)
- [ ] Storage

### Steps to Reproduce

1. Run the code below.
2. Try to perform a hot restart.

```dart
import 'package:amplify_api/amplify_api.dart';
import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
import 'package:amplify_authenticator/amplify_authenticator.dart';
import 'package:amplify_datastore/amplify_datastore.dart';
import 'package:amplify_flutter/amplify_flutter.dart';
import 'package:flutter/material.dart';
import 'package:test_app/models/ModelProvider.dart';

import 'amplifyconfiguration.dart';

void main() async {
  WidgetsFlutterBinding.ensureInitialized();

  try {
    if (!Amplify.isConfigured) {
      await Amplify.addPlugins([
        AmplifyAuthCognito(),
        AmplifyDataStore(modelProvider: ModelProvider.instance),
        AmplifyAPI(),
      ]);
      await Amplify.configure(amplifyconfig);
    }
  } on AmplifyAlreadyConfiguredException {
    debugPrint('Amplify configuration failed.');
  }

  runApp(const MyApp());
}

class MyApp extends StatefulWidget {
  const MyApp({super.key});

  @override
  State<MyApp> createState() => _MyAppState();
}

class _MyAppState extends State<MyApp> {
  @override
  Widget build(BuildContext context) {
    return Authenticator(
      onException: (exception) {
[...]

PR #3497, opened 2023-08-01
fix(datastore): hot restart

fjnoyp:
*Issue #, if available:*

https://github.com/aws-amplify/amplify-flutter/issues/3368

*Description of changes:*

Ignore platform exceptions so that authenticator no longer freezes app with datastore on hot restart.

### DIFF
--- a/packages/amplify_datastore/lib/amplify_datastore.dart
+++ b/packages/amplify_datastore/lib/amplify_datastore.dart
@@ -15,7 +15,8 @@ import 'package:meta/meta.dart';
 export 'package:amplify_core/src/types/datastore/datastore_types.dart'
     hide DateTimeParse;
 
-class AmplifyDataStore extends DataStorePluginInterface {
+class AmplifyDataStore extends DataStorePluginInterface
+    with AWSDebuggable, AmplifyLoggerMixin {
   /// Constructs an AmplifyDataStore plugin with mandatory [modelProvider]
   /// and optional datastore configuration properties including
   ///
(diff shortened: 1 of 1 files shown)


packages/amplify_datastore/lib/amplify_datastore.dart
```diff
@@ -154,6 +149,7 @@ class AmplifyDataStore extends DataStorePluginInterface {
           Map<String, String>.from(e.details as Map),
         );
       } else if (e.code == 'AmplifyAlreadyConfiguredException') {
+        logger.info('Ignoring Amplify Native already configured exception');
```
dnys1, line 152: I would make all logs debug unless there's a good reason customers should be aware of them.

```suggestion
        logger.debug('Ignoring Amplify Native already configured exception');
```

packages/amplify_datastore/lib/amplify_datastore.dart
```diff
@@ -77,11 +78,8 @@ class AmplifyDataStore extends DataStorePluginInterface {
       } on PlatformException catch (e) {
         if (e.code.contains('AmplifyAlreadyConfiguredException') ||
             e.code.contains('AlreadyConfiguredException')) {
-          throw const AmplifyAlreadyConfiguredException(
-            AmplifyExceptionMessages.alreadyConfiguredDefaultMessage,
-            recoverySuggestion:
-                AmplifyExceptionMessages.alreadyConfiguredDefaultSuggestion,
-          );
+          logger.info('Ignoring Amplify Native already configured exception');
+          return;
```
dnys1, line 82: Is it correct to return? I think Hub listeners and statements like `NativeApiPlugin.setup` must be set up in the case of a hot restart.

dnys1 (approved):

Jordan-Nelson (approved):

PR #3511, opened 2023-08-03
ci: Add setup-chromedriver action

dnys1:
The action we've been using is very flaky and asks customers to use workarounds constantly.


.github/composite_actions/setup_chromedriver/dist/index.mjs
```diff
@@ -0,0 +1,6889 @@
+import { createRequire as __WEBPACK_EXTERNAL_createRequire } from "module";
```
Jordan-Nelson, line 1: Where is this code generated or pulled from? Apologies if I am missing something obvious.

Jordan-Nelson: I see. Could it be generated prior to use rather than committed?

Jordan-Nelson (approved):

dnys1:
@Jordan-Nelson 
> I see. Could it be generated prior to use rather than committed?

The way JS actions work requires that the JS be available statically in the repo, unfortunately. Kind of annoying but these files have appropriate `.gitattributes` entries to prevent being tracked in repo stats.

https://docs.github.com/en/actions/creating-actions/creating-a-javascript-action#commit-tag-and-push-your-action-to-github

NikaHsn (approved):

PR #3534, opened 2023-08-09
ci: Fix iOS launch timeouts

dnys1:
Solves iOS launch timeouts by using macOS XL runners to improve performance.

Experimented with caching downloaded runtimes using built-in GH cache but it did not provide any performance gain to directly downloading the runtime from Apple and installing it. Left the infra though if we want to experiment more.


NikaHsn (approved):

actions/lib/bootstrap.mjs
```diff
@@ -1,6 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+// import * as cache from '@actions/cache';
```
ragingsquirrel3, line 4: nit/question: Why is this commented? Can it be removed?

ragingsquirrel3 (approved):

ragingsquirrel3 (approved):

NikaHsn (approved):

PR #3535, opened 2023-08-09
chore: Global Metric Tracking 

fjnoyp:
*Issue #, if available:*

*Description of changes:*

Track the fields of global metrics that we discussed. 

Add code to track the failing step.  Identifying the failing job was a bit tricky, so we send all the matrix values comma delimited to parse the jobs json response from github to get the job id for getting the failing step in another step. 

**Please Note**
I am unable to test on this repo because the github secrets needs to be updated with a Github Token for making the requests to Github API for which step has failed on the job. 

Thus this PR cannot be pushed as is.  Furthermore you'll see that the repo URL in the requests are pointing to my github repo (fnoyp/amplify-flutter) instead of (aws-amplifiy/amplify-flutter). 


fjnoyp:
The tests will fail because it requires secrets such as AWS Region and Github Token to be set.

PR #3568, opened 2023-08-16
feat(aft): Validate local constraints in `aft constraints` and `aft publish`

dnys1:
There are bespoke, often confusing, rules to how inter-repo constraints must be handled such that `pub` allows packages to be published. See #2885, #2366, #2865.

This introduces a checker into both `aft constraints` and `aft publish` which validates that package constraints are configured to allow publishing and updates them (or reports an error) if not.

### DIFF
--- a/packages/aft/lib/src/constraints_checker.dart
+++ b/packages/aft/lib/src/constraints_checker.dart
@@ -1,0 +1,367 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'package:aft/aft.dart';
+import 'package:aws_common/aws_common.dart';
+import 'package:collection/collection.dart';
+import 'package:pub_semver/pub_semver.dart';
+import 'package:pubspec_parse/pubspec_parse.dart';
+
+typedef MismatchedDependency = ({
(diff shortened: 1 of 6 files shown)


packages/aft/lib/src/constraints_checker.dart
```diff
@@ -0,0 +1,367 @@
...
+        mismatchedDependencies.add(
+          (
+            package: package,
+            dependencyName: dependencyPath.last,
+            message: errorMessage,
+          ),
+        );
+      case ConstraintsAction.apply:
+      case ConstraintsAction.update:
+        package.pubspecInfo.pubspecYamlEditor.update(
```
fjnoyp, line 53: When a package update is automatically updated, do we print that out anywhere to inform the user like we do for mistmatchedDependencies?

fjnoyp (dismissed):

packages/aft/lib/src/constraints_checker.dart
dnys1: No, but since the changes would be evident, I'm not sure it adds much value.

Equartey (dismissed):

Equartey (approved):

fjnoyp (approved):

Equartey (approved):

PR #3569, opened 2023-08-16
chore(aft): Better handling of pre-release Dart SDK usage

dnys1:
Some packages in the repo (for example, the `actions` pkg) set a pre-release Dart SDK constraint so that preview features of Dart can be leveraged. This updates `aft` so that:
1. `bootstrap` skips packages which set this constraint if the current Dart SDK doesn't support it.
2.  The constraints checker ignores preview constraints since these packages do not need to conform to the global setting.

### DIFF
--- a/packages/aft/lib/src/commands/bootstrap_command.dart
+++ b/packages/aft/lib/src/commands/bootstrap_command.dart
@@ -79,6 +79,24 @@ const amplifyEnvironments = <String, String>{};
           // command is significantly newer/older than the embedded one.
           (pkg) => pkg.name != 'aft',
         )
+        .where(
+          // Skip bootstrapping packages which set incompatible Dart SDK constraints,
+          // e.g. packages which are leveraging preview features.
+          //
+          // The problem of packages setting incorrect constraints, for example setting
+          // `^3.0.5` when the current repo constraint is `^3.0.0` and we're running
+          // `aft` with `3.0.1` is a different issue handled by the constraints commands.
+          (pkg) {
+            final compatibleWithActiveSdk = pkg.compatibleWithActiveSdk;
+            if (!compatibleWithActiveSdk) {
+              logger.info(
+                'Skipping package ${pkg.name} since it sets an incompatible Dart SDK constraint: '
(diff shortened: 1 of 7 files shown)


Equartey (approved):

packages/aft/lib/src/commands/bootstrap_command.dart
```diff
@@ -79,6 +79,15 @@ const amplifyEnvironments = <String, String>{};
...
           (pkg) => pkg.name != 'aft',
         )
+        .where(
+          // Skip bootstrapping packages which set incompatible Dart SDK constraints,
+          // e.g. packages which are leveraging preview features.
+          //
+          // The problem of packages setting incorrect constraints, for example setting
+          // `^3.0.5` when the current repo constraint is `^3.0.0` and we're running
+          // `aft` with `3.0.1` is a different issue handled by the constraints commands.
+          (pkg) => pkg.compatibleWithCurrentSdk,
```
fjnoyp, line 89: Should we print the packages that are ignored as a result?  In the future should someone change to a incompatible Dart SDK for whatever reason, perhaps it would be good to notify it's being ignored in bootstrap should some confusion result.

packages/aft/lib/src/config/config.dart
```diff
@@ -330,9 +327,23 @@ class PackageInfo
...
+    );
+  }
+
+  static final _whitespace = RegExp(r'\s+');
+
   @override
   List<Object?> get props => [name];
 
+  @override
+  String get runtimeTypeName => 'PackageInfo';
```
fjnoyp, line 345: General programming approach question - why do we combine multiple classes into the same file?  As someone unfamiliar with your aft code it's a bit confusing to have these large files to scroll through and also not being able to search by classes by file name that match the class contained.

packages/aft/lib/src/commands/bootstrap_command.dart
dnys1: Done 👍

fjnoyp (approved):

PR #3577, opened 2023-08-17
feat(aft): Allow `aft link` to work outside the repo

dnys1:
The use case for this is improving the experience of working on issues and features from apps which live outside the repo. It adds a `--root/-r` option to `aft link` which can overrides the default logic for finding the root directory.

Assume my `amplify-flutter` repo is cloned at `~/amplify-flutter`. I can now do the following:

```console
$ flutter create my_app
$ cd my_app && flutter pub add amplify_flutter amplify_auth_cognito
$ aft link -r ~/amplify-flutter
```

This will create a `pubspec_overrides.yaml` file in `my_app` which links all Amplify Flutter dependencies back to the repo so changes can be worked on live.

### DIFF
--- a/packages/aft/lib/src/commands/amplify_command.dart
+++ b/packages/aft/lib/src/commands/amplify_command.dart
@@ -78,8 +78,21 @@ abstract class AmplifyCommand extends Command<void>
     return Directory(directory);
   }();
 
+  /// The override for the Amplify Flutter root directory.
+  ///
+  /// Useful for running commands `aft link` from outside the repo.
+  ///
+  /// **NOTE**: Not all commands support this.
+  late final _rootDirectoryOverride = () {
+    if (argParser.findByNameOrAlias('root') != null) {
+      return argResults?['root'] as String?;
+    }
+    return null;
+  }();
+
   late final AftConfigLoader aftConfigLoader = AftConfigLoader(
     workingDirectory: workingDirectory,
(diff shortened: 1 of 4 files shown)


packages/aft/lib/src/commands/link_command.dart
```diff
@@ -13,6 +13,16 @@ import 'package:yaml_edit/yaml_edit.dart';
 /// Command to link all Dart/Flutter packages in the repo together using
 /// `pubspec_overrides.yaml`.
 class LinkCommand extends AmplifyCommand {
+  LinkCommand() {
+    argParser.addOption(
+      'root',
+      abbr: 'r',
+      help:
+          'The root of the Amplify Flutter repo. Defaults to automatic determination '
+          'but must be set when linking outside the repo.',
```
Equartey, line 22: Consider adding an example usage here.

packages/aft/lib/src/commands/amplify_command.dart
```diff
@@ -78,8 +78,21 @@ abstract class AmplifyCommand extends Command<void>
     return Directory(directory);
   }();
 
+  /// The override for the Amplify Flutter root directory.
+  ///
+  /// Useful for running commands `aft link` from outside the repo.
+  ///
+  /// **NOTE**: Not all commands support this.
```
Equartey, line 85: nit: I'd include what commands it works with

PR #3582, opened 2023-08-18
chore(logging): Queued Store Item

khatruong2009:
*Issue #, if available:*

*Description of changes:*

- Added isFull and getAll methods
- Added amplify_logging_cloudwatch package
- Brought queued_item_store into amplify_logging_cloudwatch


.flutter-plugins-dependencies
```diff
@@ -0,0 +1 @@
...":"/Users/ktruon/.pub-cache/hosted/pub.dev/package_info_plus-4.1.0/","native_build":false,"dependencies":[]}],"windows":[{"name":"connectivity_plus","path":"/Users/ktruon/.pub-cache/hosted/pub.dev/connectivity_plus-4.0.2/","native_build":true,"dependencies":[]},{"name":"device_info_plus","path":"/Users/ktruon/.pub-cache/hosted/pub.dev/device_info_plus-9.0.3/","native_build":false,"dependencies":[]},{"name":"package_info_plus","path":"/Users/ktruon/.pub-cache/hosted/pub.dev/package_info_plus-4.1.0/","native_build":false,"dependencies":[]}],"web":[{"name":"connectivity_plus","path":"/Users/ktruon/.pub-cache/hosted/pub.dev/connectivity_plus-4.0.2/","dependencies":[]},{"name":"device_info_plus","path":"/Users/ktruon/.pub-cache/hosted/pub.dev/device_info_plus-9.0.3/","dependencies":[]},{"name":"package_info_plus","path":"/Users/ktruon/.pub-cache/hosted/pub.dev/package_info_plus-4.1.0/","dependencies":[]}]},"dependencyGraph":[{"name":"connectivity_plus","dependencies":[]},{"name":"device_info_plus","dependencies":[]},{"name":"flutter_plugin_android_lifecycle","dependencies":[]},{"name":"package_info_plus","dependencies":[]}],"date_created":"2023-08-17 17:11:41.266673","version":"3.10.6"}
```
dnys1, line 1: This file should not be tracked. Needs to be removed from the cache (`git rm --cached .flutter-plugin-dependencies`) and added to the gitignore.

.flutter-plugins
```diff
@@ -0,0 +1,5 @@
+# This is a generated file; do not edit or check into version control.
```
dnys1, line 1: This file should not be tracked. Needs to be removed from the cache (`git rm --cached .flutter-plugin-dependencies`) and added to the gitignore.

.flutter-plugins-dependencies
dnys1: The problem I think is that you've run `flutter pub get` in this project, even though it is Dart only, and the Dart `gitignore` and Flutter `gitignore` vary.

packages/logging_cloudwatch/amplify_logging_cloudwatch/example/pubspec.yaml
```diff
@@ -0,0 +1,15 @@
+name: amplify_logging_cloudwatch_example
+description: An example app showcasing the amplify_logging_cloudwatch package.
+version: 1.0.0
+
```
dnys1, line 4: I need to fix the template, but this should be here.

```suggestion
publish_to: none
```

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/index_db/indexed_db_adapter.dart
```diff
@@ -0,0 +1,176 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:async';
```
dnys1, line 4: Can we copy over tests for these classes as well?

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/queued_item_store/in_memory_queued_item_store.dart
```diff
@@ -14,11 +15,13 @@ class InMemoryQueuedItemStore implements QueuedItemStore {
 
   /// The next ID that should be used when adding an item in the DB.
   int _nextId = 0;
+  int _currentTotalByteSize = 0;
   final LinkedHashMap<int, QueuedItem> _database =
       LinkedHashMap<int, QueuedItem>();
 
   @override
   void addItem(String string, String timestamp) {
+    _currentTotalByteSize += utf8.encode(string).length;
```
dnys1, line 24: Discuss with Nika. This logic is already running in her plugin class.

dnys1:
Looking good. Few overall notes:
- I would push branches to this repo instead of your fork. It makes review and CI easier.
- Go ahead and run `aft generate workflows` so that this new package is incorporated into the CI pipeline.

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/aws_logging_cloudwatch.dart
```diff
@@ -3,3 +3,6 @@
 
 /// aws_logging_cloudwatch
 library aws_logging_cloudwatch;
+
+export './src/queued_item_store/in_memory_queued_item_store.dart';
+export './src/queued_item_store/queued_item_store.dart';
```
NikaHsn, line 8:
```suggestion
export 'src/queued_item_store/in_memory_queued_item_store.dart';
export 'src/queued_item_store/queued_item_store.dart';
```

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/index_db/indexed_db_adapter.dart
```diff
@@ -0,0 +1,177 @@
...
+    return readValues;
+  }
+
+  @override
+  Future<void> deleteItems(Iterable<QueuedItem> items) async {
+    if (items.isEmpty) return;
+
+    _currentTotalByteSize -= utf8.encode(items.first.value).length;
+    _currentTotalByteSize -= utf8.encode(items.first.timestamp).length;
+    _currentTotalByteSize -= 8; // 8 bytes for the id
```
dnys1, line 107: Needs a TODO here and above. Running `utf8.encode` twice for every log entry is a poor use of CPU cycles and should not be done on the main thread. Improvement options are:
- Offload to another thread with worker bee (several examples in the repo)
- Use `String.length` instead, which gives the size in UTF-16 code points

It's also not clear to me if this is even correct. Here's a screenshot from Firefox showing how it stores a JSON blob in IndexedDB.

Screenshot 2023-08-21 at 4 25 38 PM

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/index_db/indexed_db_adapter.dart
```diff
@@ -0,0 +1,177 @@
...
+  @override
+  Future<bool> isFull(int maxSizeInMB) async {
+    final maxBytes = maxSizeInMB * 1024 * 1024;
+    return _currentTotalByteSize >= maxBytes;
+  }
+
+  /// Clear the database.
+  @override
+  Future<void> clear() async {
+    _currentTotalByteSize = 0;
```
dnys1, line 154: This should be the last line, as it's only true if the two futures below succeed

packages/logging_cloudwatch/amplify_logging_cloudwatch/test/amplify_logging_cloudwatch_test.dart
```diff
@@ -0,0 +1,7 @@
+import 'package:test/test.dart';
+
+void main() {
```
dnys1, line 3: Can remove this file

packages/logging_cloudwatch/amplify_logging_cloudwatch/test/queued_item_store_test.dart
```diff
@@ -0,0 +1,227 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import 'package:aws_logging_cloudwatch/src/queued_item_store/in_memory_queued_item_store.dart';
+import 'package:aws_logging_cloudwatch/src/queued_item_store/queued_item_store.dart';
+import 'package:test/test.dart';
+
+void main() {
+  late QueuedItemStore db;
+
+  group('InMemoryQueuedItemStore ', () {
```
dnys1, line 11: Can we copy over the IDB tests as well?

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/drift/drift_queued_item_store.dart
```diff
@@ -0,0 +1,125 @@
...
+        // Note: From schemaVersion 1->2 we renamed DriftJsonStrings to DriftQueuedItems before the GA release
+        await m.createAll();
+      },
+    );
+  }
+
+  @override
+  Future<void> addItem(String value, String timestamp) async {
+    _currentTotalByteSize += utf8.encode(value).length;
+    _currentTotalByteSize += utf8.encode(timestamp).length;
```
dnys1, line 63: The encoded timestamp's byte count will always be the same. Hardcode it in a static var.

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/index_db/indexed_db_adapter.dart
```diff
@@ -0,0 +1,177 @@
...
+    );
+    final store = transaction.objectStore(storeName);
+    return store;
+  }
+
+  @override
+  Future<void> addItem(String string, String timestamp) async {
+    _currentTotalByteSize += utf8.encode(string).length;
+    _currentTotalByteSize += utf8.encode(timestamp).length;
+    _currentTotalByteSize += 8; // 8 bytes for the id
```
dnys1, line 74: This logic is duplicated in multiple places. Best to consolidate in it a helper method somewhere.

dnys1 (commented):
Few comments, but looking good!

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/index_db/indexed_db_adapter.dart
```diff
@@ -0,0 +1,186 @@
...
+    );
+    final store = transaction.objectStore(storeName);
+    return store;
+  }
+
+  /// Calculates the size of a queued item by adding the length of the UTF-8 encoded [value],
+  /// 23 bytes for the timestamp, and 8 bytes for the ID.
+  ///
+  /// Returns the total size of the queued item in bytes.
+  int sizeOfQueuedItem(String value) {
```
NikaHsn, line 73: ditto

packages/logging_cloudwatch/amplify_logging_cloudwatch/test/queued_item_store_test.dart
```diff
@@ -0,0 +1,232 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import 'package:amplify_logging_cloudwatch/src/queued_item_store/dart_queued_item_store.dart';
+import 'package:aws_logging_cloudwatch/src/queued_item_store/queued_item_store.dart';
+import 'package:test/test.dart';
+
+void main() {
+  late DartQueuedItemStore db;
+
+  group('InMemoryQueuedItemStore ', () {
```
NikaHsn, line 11: isn't it testing `DartQueuedItemStore` ?

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/queued_item_store/in_memory_queued_item_store.dart
```diff
@@ -14,11 +14,25 @@ class InMemoryQueuedItemStore implements QueuedItemStore {
...
   int _nextId = 0;
+  int _currentTotalByteSize = 0;
   final LinkedHashMap<int, QueuedItem> _database =
       LinkedHashMap<int, QueuedItem>();
 
+  /// Calculates the size of a queued item by adding the length of the UTF-8 encoded [value],
+  /// 23 bytes for the timestamp, and 8 bytes for the ID.
+  ///
+  /// Returns the total size of the queued item in bytes.
+  int sizeOfQueuedItem(String value) {
```
NikaHsn, line 25: ditto

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/drift/drift_queued_item_store.dart
```diff
@@ -0,0 +1,132 @@
...
+        await m.createAll();
+      },
+    );
+  }
+
+  /// Calculates the size of a queued item by adding the length of the UTF-8 encoded [value],
+  /// 23 bytes for the timestamp, and 8 bytes for the ID.
+  ///
+  /// Returns the total size of the queued item in bytes.
+  int sizeOfQueuedItem(String value) {
```
NikaHsn, line 63: nit: I would get the timestamp in the function params and use timestamp.value instead of assuming it is 23 bytes always bc it is 23 bytes if it is local time and in milliseconds.)

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/drift/drift_queued_item_store.dart
```diff
@@ -0,0 +1,132 @@
...
+        await m.createAll();
+      },
+      onUpgrade: (Migrator m, int from, int to) async {
+        // Note: From schemaVersion 1->2 we renamed DriftJsonStrings to DriftQueuedItems before the GA release
+        await m.createAll();
+      },
+    );
+  }
+
+  /// Calculates the size of a queued item by adding the length of the UTF-8 encoded [value],
```
NikaHsn, line 59: update the docs, it is not using utf8 encoded value

.github/dependabot.yaml
```diff
@@ -785,6 +785,16 @@ updates:
           - "version-update:semver-patch"
       - dependency-name: "example_common"
       - dependency-name: "amplify_lints"
+  - package-ecosystem: "pub"
```
NikaHsn, line 788: is it generated by aft? I think it is missing a section for `packages/logging_cloudwatch/amplify_logging_cloudwatch`. also no GH workflow is added for this package.

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/queued_item_store/queued_item_store.dart
```diff
@@ -38,6 +38,9 @@ class QueuedItem {
     required this.timestamp,
   });
 
+  /// Getter for byte size of the item.
```
dnys1, line 41:
```suggestion
  /// The size of the item, in bytes, when stored in the database.
```

Mentioning it's a getter doesn't add much value. Prefer using more descriptive comments: https://dart.dev/effective-dart/documentation#prefer-starting-a-non-boolean-variable-or-property-comment-with-a-noun-phrase

dnys1 (approved):

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/queued_item_store/queued_item_store.dart
```diff
@@ -38,6 +38,9 @@ class QueuedItem {
     required this.timestamp,
   });
 
+  /// The size of the item, in bytes, when stored in the database.
+  int get byteSize => value.length + timestamp.length + 8; // 8 bytes for id
```
NikaHsn, line 42: instead of calculating the size each time on the getter, isn't it better to use a private prop and return it from the getter? 

```suggestion
  const QueuedItem({
    required this.id,
    required this.value,
    required this.timestamp,
  }) : _byteSize = value.length + timestamp.length + 8;
  final int _byteSize;
  int get byteSize => _byteSize;
```

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/drift/drift_queued_item_store.dart
```diff
@@ -0,0 +1,125 @@
...
+    );
+  }
+
+  @override
+  Future<void> addItem(String value, String timestamp) async {
+    final queuedItem = QueuedItem(
+      id: 0, // temporary value used to calculate byte size
+      value: value,
+      timestamp: timestamp,
+    );
```
NikaHsn, line 65: nit: it feels wired to create a temp QueuedItem only to use the byteSize instead of `value.length + timestamp.length + 8`.

.github/workflows/amplify_logging_cloudwatch.yaml
NikaHsn, line 1: should we expect some workflow to run on this PR?

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/drift/drift_queued_item_store.dart
khatruong2009: I agree but the alternative would be to repeat the code of value.length + timestamp.length + 8 multiple times throughout multiple files instead of just using the property and we can change the calculation in the future in one place if needed.

NikaHsn: we create temp QueuedItem in two places, instead we can use `value.length + timestamp.length + 8`.  in other places we can use the `QueuedItem.byteSize`. wdyt?

NikaHsn (approved):

PR #3613, opened 2023-08-25
ci: Fix AWS credentials retrieval

dnys1:
Bumps the `configure-aws-credentials` action so that we no longer receive this error:

```
Assuming role with OIDC
Error: Could not assume role with OIDC: SecretAccessKey contains special characters.
```


Equartey (approved):
Great find 🚀

Jordan-Nelson (approved):

Issue #3625, opened 2023-08-28
Traling slash added at the end of endpoint


dagovalsusa:
### Description

Hello, 

A slash is added to the end of the url endpoint which then causes a bad response (404) from the endpoint itself.
Our setup is as follows

` "api": {
      "plugins": {
        "awsAPIPlugin": {
          "APPGQL": {
            "endpointType": "GraphQL",
            "endpoint": "https://xxxx.xxx.xx/graphql",
            "region": "us-east-1",
            "authorizationType": "AMAZON_COGNITO_USER_POOLS"
          }
        }
      }
    }`

but the http call url has an extra "/" at the end as seen from the picture

<img width="1439" alt="Screenshot 2023-08-28 alle 16 52 22" src="https://github.com/aws-amplify/amplify-flutter/assets/11755587/e52135dc-4eeb-4136-9f64-c995b5a701ce">

we debugged and we noticed that in the amplify-flutter/packages/api/amplify_api_dart/lib/src/util/amplify_api_config.dart file in the getUri method the path '' adds one slash too many

`Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.13.1, on macOS 13.4.1 22F770820d darwin-arm64, locale it-IT)
[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 14.3.1)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2021.2)
[✓] VS Code (version 1.81.1)
[✓] Connected device (4 available)
[✓] Network resources`

### Categories

- [ ] Analytics
- [ ] API (REST)
- [X] API (GraphQL)
- [ ] Auth
- [ ] Authenticator
- [ ] DataStore
- [ ] Notifications (Push)
- [ ] Storage

### Steps to Reproduce

1. Configure amplify
2. run a query 

### Platforms

- [X] iOS
- [X] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Flutter Version

3.13.1

### Amplify Flutter Version

1.3.3 (latest)

### Deployment Method

Custom Pipeline

dagovalsusa:
Hello, the pull request memtioned up doesn't works
Anyone have a suggestions or know who o where the "/" is added?

dagovalsusa:
I tried the pull request mentioned up as @dnys1 suggested.
[...]

PR #3626, opened 2023-08-28
fix: removed trailing slash at the end of uri

alegos27:
*Issue #3625, if available:*

*Description of changes:*
Added an if in path segments replacing due to prevent the add of the slash at the end of the URI.


dnys1:
Thank you for the PR, @alegos27! Would you be able to share the issues you were having previously?

dnys1:
Never mind. I see you created https://github.com/aws-amplify/amplify-flutter/issues/3625. Can you add a small test to ensure that your use case is covered by this change?

packages/api/amplify_api_dart/lib/src/util/amplify_api_config.dart
```diff
@@ -37,7 +37,7 @@ class EndpointConfig with AWSEquatable<EndpointConfig> {
     return parsed.replace(
       pathSegments: [
         ...parsed.pathSegments,
-        ...pathSegmentsFromPath,
+        if (pathSegmentsFromPath.isNotEmpty && pathSegmentsFromPath.first.isNotEmpty) ...pathSegmentsFromPath,
       ],
```
dnys1, line 41: Would this be better?

```dart
pathSegments: [
  ...parsed.pathSegments,
  ...pathSegmentsFromPath,
].where((segment) => segment.isNotEmpty),
```

Since this method is shared between REST and GraphQL, could there be use cases where a trailing slash is desired?

alegos27:
We tried our fix but it not works (the slash at the end of url is still present). We can't understand where was build the uri used by the function 'Amplify.API.query'. After a small test seems that the function getUri in amplify_api_config.dart was never called.

We can't find the implementation of this function:
 ```
GraphQLOperation<T> query<T>({required GraphQLRequest<T> request}) =>
      identifyCall(
        ApiCategoryMethod.query,
        () => defaultPlugin.query(request: request),
      );
```

File: packages/api/amplify_api_dart/lib/src/util/amplify_api_config.dart (hunk 1 above)
Comment (alegos27) [reply packages/api/amplify_api_dart/lib/src/util/amplify_api_config.dart:41]: Yes this could be better and has the same effect. But unfortunately it doesn't solve our problem.

**Comment by dagovalsusa** [comment]
this fix works, it's only hard to test it including libs into pubspec file

File: packages/api/amplify_api_dart/lib/src/util/amplify_api_config.dart (hunk 1 above)
Comment (alegos27) [reply packages/api/amplify_api_dart/lib/src/util/amplify_api_config.dart]: I have updated the PR with your suggested code. We test our code and this fix works fine!

**Comment by alegos27** [comment]
We test the fix by modify the code directly in pub cache (and it works fine) because we can't find a way to import the forked branch in pubspec.

We tried this:
`dependencies:
  amplify_api:
    git:
      url: https://github.com/alegos27/amplify-flutter
      ref: fix/api/trailing-slash
      path: packages/api/amplify_api
  amplify_auth_cognito: ^1.3.3
  amplify_flutter: ^1.3.3`
  
But it seems that when we build the app the file used is from hosted package and not from git. Can you help us? So we can add the test you required. Thanks!

**Comment by dnys1** [comment]
Hi @alegos27, the best way to test a change is with path overrides instead of git overrides, for example:

```yaml
dependency_overrides:
  amplify_api_dart:
    path: path/to/repo/packages/api/amplify_api_dart
```

alegos27:
Thanks @dnys1! Your suggestion works fine! The fix solve our problem. Do you think it could be merged? If the trailing slash would be useful for other reasons or for REST standard, maybe it's possible to add a configuration flag to enable or disable trailing slash in URI.
What do you think about it? Thanks a lot!

dagovalsusa (approved):

packages/api/amplify_api_dart/test/amplify_api_config_test.dart
```diff
@@ -6,71 +6,72 @@ import 'package:amplify_core/amplify_core.dart';
...
-
-      expect(uri, equals(expected));
+      const expected = '$endpoint/$path?foo=bar&bar=baz';
+      expect(uri.toString(), equals(expected));
     });
 
-    test('should handle a leading slash', () async {
+    test('should handle a path with a leading slash', () async {
+      const endpoint = 'https://abc123.amazonaws.com/test';
       const path = '/path/to/nowhere';
```
dnys1, line 46: Can this be `/path/to/nowhere/` to test the behavior when `path` has a trailing slash?

Jordan-Nelson: Yeah. This is an interesting case. Similar to my comment above, I think the trailing slash should be kept if the consumer explicitly included it. The full uri becomes `https://abc123.amazonaws.com/test/path/to/nowhere/?foo=bar&bar=baz` (with a `/` immediately before the `?`). Is that your expectation?

dnys1: That's what I would expect, yeah

Jordan-Nelson: Updated

dnys1 (approved):
LGTM

Equartey (approved):

PR #3646, opened 2023-08-30
feat(logging): log stream provider to cache log stream

NikaHsn:
*Issue #, if available:*

*Description of changes:*
- add caching to log stream provider
- cleanup the sync logic
- TODO: add log rotation to queued item store

### DIFF
--- a/packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/cloudwatch_logger_plugin.dart
+++ b/packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/cloudwatch_logger_plugin.dart
@@ -6,8 +6,6 @@ import 'dart:math';
 
 import 'package:aws_common/aws_common.dart';
 import 'package:aws_logging_cloudwatch/aws_logging_cloudwatch.dart';
-import 'package:aws_logging_cloudwatch/src/queued_item_store/in_memory_queued_item_store.dart';
-import 'package:aws_logging_cloudwatch/src/queued_item_store/queued_item_store.dart';
 import 'package:aws_logging_cloudwatch/src/sdk/cloud_watch_logs.dart';
 import 'package:aws_logging_cloudwatch/src/stoppable_timer.dart';
 import 'package:fixnum/fixnum.dart';
@@ -32,10 +30,11 @@ import 'package:meta/meta.dart';
 // The maximum number of log events in a batch is 10,000.
 
 const int _maxNumberOfLogEventsInBatch = 10000;
-const int _maxLogEventsTimeSpanInBatch = 24 * 3600;
 const int _maxLogEventsBatchSize = 1048576;
 const int _baseBufferSize = 26;
 const int _maxLogEventSize = 256000;
+final int _maxLogEventsTimeSpanInBatch =
+    const Duration(hours: 24).inMilliseconds;
 
 typedef _LogBatch = (List<QueuedItem> logQueues, List<InputLogEvent> logEvents);
 
@@ -138,51 +137,28 @@ class CloudWatchLoggerPlugin extends AWSLoggerPlugin
 
   Future<void> _startSyncingIfNotInProgress() async {
     Future<void> startSyncing() async {
-      String logStream;
-      try {
-        logStream = await _logStreamProvider.logStream;
-      } on Exception catch (e) {
-        if (await _logStore.isFull(_pluginConfig.localStoreMaxSizeInMB)) {
-          await _logStore.clear();
-          logger.warn(
-            'Reached local store max size of: '
-            '${_pluginConfig.localStoreMaxSizeInMB}.Hence logs are deleted from '
-            'local store.',
-          );
+      final batchStream = _getLogBatchesToSync();
+      await for (final (logs, events) in batchStream) {
+        final response = await _sendToCloudWatch(events);
+        // TODO(nikahsn): handle tooOldLogEventEndIndex
+        // and expiredLogEventEndIndex.
+        if (response.rejectedLogEventsInfo?.tooNewLogEventStartIndex != null) {
+          // TODO(nikahsn): throw and exception to enable log rotation if the
+          // log store is full.
+          break;
         }
-        logger.error('Failed to create CloudWatch log stream', e);
-        return;
-      }
-
-      final batcheStream = _getLogBatchesToSync();
-      await for (final (logs, events) in batcheStream) {
-        try {
-          final response = await _sendToCloudWatch(events, logStream);
-          if (response.rejectedLogEventsInfo?.tooNewLogEventStartIndex !=
-              null) {
-            break;
-          }
-          await _logStore.deleteItems(logs);
-        } on Exception catch (e) {
-          if (await _logStore.isFull(_pluginConfig.localStoreMaxSizeInMB)) {
-            await _logStore.deleteItems(logs);
-            logger.warn(
-              'Reached local store max size of: '
-              '${_pluginConfig.localStoreMaxSizeInMB}.Hence logs are deleted '
-              'from local store.',
-            );
-          }
-          logger.error('Failed to sync batched logs to CloudWatch', e);
-        }
+        await _logStore.deleteItems(logs);
       }
     }
 
     if (!_syncing) {
+      // TODO(nikahsn): disable log rotation.
       _syncing = true;
       try {
         await startSyncing();
       } on Exception catch (e) {
         logger.error('Failed to sync logs to CloudWatch.', e);
+        // TODO(nikahsn): enable log rotation if the log store is full
       } finally {
         _syncing = false;
       }
@@ -200,20 +176,25 @@ class CloudWatchLoggerPlugin extends AWSLoggerPlugin
 
   Future<PutLogEventsResponse> _sendToCloudWatch(
     List<InputLogEvent> logEvents,
-    String logStreamName,
   ) async {
+    final logStreamName = await _logStreamProvider.defaultLogStream;
     final request = PutLogEventsRequest(
       logGroupName: _pluginConfig.logGroupName,
       logStreamName: logStreamName,
       logEvents: logEvents,
     );
-    return _client.putLogEvents(request).result;
+    try {
+      return await _client.putLogEvents(request).result;
+    } on ResourceNotFoundException {
+      await _logStreamProvider.createLogStream(logStreamName);
+      return _client.putLogEvents(request).result;
+    }
   }
 
   Stream<_LogBatch> _getLogBatchesToSync() async* {
     final queuedLogs = (await _logStore.getAll()).toList();
     if (queuedLogs.isEmpty) {
-      yield ([], []);
+      return;
     }
     final logEvents = <InputLogEvent>[];
     final logQueues = <QueuedItem>[];
--- a/packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/log_stream_provider.dart
+++ b/packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/log_stream_provider.dart
@@ -1,6 +1,8 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import 'dart:async';
+
 import 'package:aws_logging_cloudwatch/src/sdk/cloud_watch_logs.dart';
 import 'package:intl/intl.dart';
 
@@ -9,17 +11,19 @@ import 'package:intl/intl.dart';
 /// [CloudWatchLogStreamProvider]
 /// {@endtemplate}
 abstract class CloudWatchLogStreamProvider {
-  /// Returns CloudWatch log stream name to use for sending logs to CloudWatch.
-  ///
-  /// It creates the log stream if it does not exists.
-  Future<String> get logStream;
+  /// Returns the default log stream name from cache. if cache is missing it
+  /// calls [createLogStream] and return the log stream name.
+  FutureOr<String> get defaultLogStream;
+
+  /// Creates the log stream and add it to the cache.
+  Future<void> createLogStream(String logStreamName);
 }
 
 /// {@template aws_logging_cloudwatch.default_cloudwatch_logstream_provider}
 /// The default implementaion of [CloudWatchLogStreamProvider].
 ///
-/// It uses `logStreamName` if provided otherwise uses `yyyy-MM-dd` date format
-/// of UTC time now for the `logStreamName`.
+/// It uses `defaultLogStreamName` if provided otherwise uses `yyyy-MM-dd`
+/// date format of UTC time now for the `defaultLogStreamName`.
 /// {@endtemplate}
 class DefaultCloudWatchLogStreamProvider
     implements CloudWatchLogStreamProvider {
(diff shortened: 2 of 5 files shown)


packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/cloudwatch_logger_plugin.dart
```diff
@@ -200,43 +173,49 @@ class CloudWatchLoggerPlugin extends AWSLoggerPlugin
...
       logEvents: logEvents,
     );
-    return _client.putLogEvents(request).result;
+    try {
+      return _client.putLogEvents(request).result;
+    } on ResourceNotFoundException {
+      await _logStreamProvider.createLogStream(logStreamName);
+      return _client.putLogEvents(request).result;
+    } on Exception {
+      rethrow;
```
dnys1, line 189: This is redundant and not needed.

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/cloudwatch_logger_plugin.dart
```diff
@@ -200,43 +173,49 @@ class CloudWatchLoggerPlugin extends AWSLoggerPlugin
...
   ) async {
+    final logStreamName = await _logStreamProvider.logStream;
     final request = PutLogEventsRequest(
       logGroupName: _pluginConfig.logGroupName,
       logStreamName: logStreamName,
       logEvents: logEvents,
     );
-    return _client.putLogEvents(request).result;
+    try {
+      return _client.putLogEvents(request).result;
```
dnys1, line 184: If it's not awaited here, this will only catch synchronous exceptions

```suggestion
      return await _client.putLogEvents(request).result;
```

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/cloudwatch_logger_plugin.dart
```diff
@@ -200,43 +173,49 @@ class CloudWatchLoggerPlugin extends AWSLoggerPlugin
...
+    } on ResourceNotFoundException {
+      await _logStreamProvider.createLogStream(logStreamName);
+      return _client.putLogEvents(request).result;
```
dnys1, line 187:
```suggestion
      return await _client.putLogEvents(request).result;
```

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/cloudwatch_logger_plugin.dart
```diff
@@ -200,43 +173,49 @@ class CloudWatchLoggerPlugin extends AWSLoggerPlugin
...
-      if (totalByteSize + size >= _maxLogEventsBatchSize ||
-          logEvents.length >= _maxNumberOfLogEventsInBatch ||
-          (logEvents.length > 1 &&
-              currentLogEvent.timestamp - logEvents.first.timestamp >=
-                  _maxLogEventsTimeSpanInBatch)) {
-        yield (logQueues, logEvents);
-        totalByteSize = 0;
-        logEvents.clear();
-        logQueues.clear();
+    if (queuedLogs.isNotEmpty) {
```
dnys1, line 195: Prefer exit early to improve readability and decrease average indent level

```suggestion
    if (queuedLogs.isEmpty) {
      return;
    }
```

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/cloudwatch_logger_plugin.dart
```diff
@@ -200,43 +173,49 @@ class CloudWatchLoggerPlugin extends AWSLoggerPlugin
...
+      var totalByteSize = 0;
+
+      for (final currentLog in queuedLogs) {
+        final currentLogEvent = currentLog.toInputLogEvent();
+        final size = currentLogEvent.message.length + _baseBufferSize;
+        if (totalByteSize + size >= _maxLogEventsBatchSize ||
+            logEvents.length >= _maxNumberOfLogEventsInBatch ||
+            (logEvents.length > 1 &&
+                currentLogEvent.timestamp - logEvents.first.timestamp >=
+                    _maxLogEventsTimeSpanInBatch)) {
```
dnys1, line 207: Please break these out into variables

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/log_stream_provider.dart
```diff
@@ -55,10 +66,9 @@ class DefaultCloudWatchLogStreamProvider
           )
           .result;
     } on ResourceAlreadyExistsException {
-      return logStreamName;
+      return;
     } on Exception {
       rethrow;
     }
```
dnys1, line 72: Remove

packages/logging_cloudwatch/aws_logging_cloudwatch/test/cloudwatch_logger_plugin_test.dart
```diff
@@ -132,25 +131,163 @@ void main() {
...
       );
 
+      when(() => mockCloudWatchLogsClient.putLogEvents(any())).thenAnswer(
+        (_) => mockPutLogEventsOperation,
+      );
+
       when(() => mockCloudWatchLogStreamProvider.logStream)
           .thenAnswer((_) async => Future<String>.value('log stream name'));
 
+      when(() => mockQueuedItemStore.addItem(any(), any()))
```
dnys1, line 141: Is there any benefit to mocking over using the in-mem store? I would generally mock a single layer per test, in this case just the CloudWatch SDK.

packages/logging_cloudwatch/aws_logging_cloudwatch/test/cloudwatch_logger_plugin_test.dart
```diff
@@ -132,25 +131,163 @@ void main() {
...
+        (_) async => Future<Iterable<QueuedItem>>.value(queuedItems),
+      );
+
+      await expectLater(
+        plugin.flushLogs(),
+        completes,
+      );
+
+      verify(
+        () => mockCloudWatchLogStreamProvider.logStream,
```
dnys1, line 157: Same question. Why not use the real log stream provider here?

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/cloudwatch_logger_plugin.dart
```diff
@@ -32,7 +30,7 @@ import 'package:meta/meta.dart';
 // The maximum number of log events in a batch is 10,000.
 
 const int _maxNumberOfLogEventsInBatch = 10000;
-const int _maxLogEventsTimeSpanInBatch = 24 * 3600;
+const int _maxLogEventsTimeSpanInBatch = 24 * 3600 * 1000;
```
dnys1, line 33: Prefer typed values to avoid errors like this

```suggestion
const Duration _maxLogEventsTimeSpanInBatch = Duration(hours: 24);
```

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/cloudwatch_logger_plugin.dart
```diff
@@ -200,43 +173,49 @@ class CloudWatchLoggerPlugin extends AWSLoggerPlugin
...
+    } on ResourceNotFoundException {
+      await _logStreamProvider.createLogStream(logStreamName);
+      return _client.putLogEvents(request).result;
+    } on Exception {
+      rethrow;
+    }
   }
 
   Stream<_LogBatch> _getLogBatchesToSync() async* {
     final queuedLogs = (await _logStore.getAll()).toList();
```
dnys1, line 194: These can be streamed

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/log_stream_provider.dart
```diff
@@ -36,15 +40,22 @@ class DefaultCloudWatchLogStreamProvider
...
   static final DateFormat _dateFormat = DateFormat('yyyy-MM-dd');
+  final _cachedLogStreams = <String>{};
 
-  /// Creates CloudWatch log stream if does not exists and returns
-  /// the log stream name.
-  ///
-  /// Throws an [Exception] if fails to create the log stream.
   @override
-  Future<String> get logStream async {
+  FutureOr<String> get logStream async {
```
dnys1, line 46: Making this `FutureOr` is a decent idea, but by marking this `async` it will always return a `Future`.

In the cached case, this is the same as:

```dart
Future<String> get logStream {
   if (_cachedLogStreams.contains(logStreamName)) {
      return Future.value(logStreamName);
    }
}
```

Further, if you await the return of a `FutureOr` it is implicitly wrapped in `Future.value` and you cannot use the result synchronously.

That is to say, to fully leverage `FutureOr`, you need to switch on the return and treat it as a sealed class over `T` and `Future<T>`.

https://dartpad.dev/?id=dc5c257693d0d454c819ffcf38b626f0

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/cloudwatch_logger_plugin.dart
```diff
@@ -138,51 +136,26 @@ class CloudWatchLoggerPlugin extends AWSLoggerPlugin
...
-              'Reached local store max size of: '
-              '${_pluginConfig.localStoreMaxSizeInMB}.Hence logs are deleted '
-              'from local store.',
-            );
-          }
-          logger.error('Failed to sync batched logs to CloudWatch', e);
+        final response = await _sendToCloudWatch(events);
+        if (response.rejectedLogEventsInfo?.tooNewLogEventStartIndex != null) {
+          throw _TooNewLogEventException(
+            'logs can not be more than 2 hours in the future',
```
dnys1, line 144:
```suggestion
            'Logs cannot be more than 2 hours in the future',
```

dnys1: What is the value of a private exception type for this? Will it be caught and handled separately in the future?

dnys1: Also, what is the value of throwing? What is a user meant to do?

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/cloudwatch_logger_plugin.dart
NikaHsn: it is unnecessary await in return: https://dart.dev/tools/linter-rules/unnecessary_await_in_return

dnys1: It's a bug: https://github.com/dart-lang/linter/issues/2357

dnys1: Move the return to the next line to avoid lint

dnys1: But the return is actually not needed in this case

dnys1: The await is

packages/logging_cloudwatch/aws_logging_cloudwatch/test/cloudwatch_logger_plugin_test.dart
NikaHsn: mocking dependencies make the tests faster, reliable and can isolate the code under test from its dependencies.

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/cloudwatch_logger_plugin.dart
dnys1: Hmm yes, sorry, I thought we were speaking about the `try` block. My mistake, there is no semantic difference for this line.

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/cloudwatch_logger_plugin.dart
```diff
@@ -138,51 +137,26 @@ class CloudWatchLoggerPlugin extends AWSLoggerPlugin
...
-            await _logStore.deleteItems(logs);
-            logger.warn(
-              'Reached local store max size of: '
-              '${_pluginConfig.localStoreMaxSizeInMB}.Hence logs are deleted '
-              'from local store.',
-            );
-          }
-          logger.error('Failed to sync batched logs to CloudWatch', e);
+        final response = await _sendToCloudWatch(events);
+        if (response.rejectedLogEventsInfo?.tooNewLogEventStartIndex != null) {
```
dnys1, line 143: There are two other properties to this, `expiredLogEventEndIndex` and `tooOldLogEventEndIndex`, which will need to be handled. I'm guessing later iterations will take into account partial completion as well since currently it is an all-or-nothing approach for a batch?

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/log_stream_provider.dart
```diff
@@ -9,10 +11,12 @@ import 'package:intl/intl.dart';
 /// [CloudWatchLogStreamProvider]
 /// {@endtemplate}
 abstract class CloudWatchLogStreamProvider {
-  /// Returns CloudWatch log stream name to use for sending logs to CloudWatch.
-  ///
-  /// It creates the log stream if it does not exists.
-  Future<String> get logStream;
+  /// Returns log stream name from cache. if cache is missing it
+  /// calls [createLogStream] and return the log stream name.
+  FutureOr<String> get logStream;
```
dnys1, line 16: I would maybe change the name to `currentLogStream` or similar to distinguish from `createLogStream`. Is this class going to be public?

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/cloudwatch_logger_plugin.dart
```diff
@@ -138,51 +137,26 @@ class CloudWatchLoggerPlugin extends AWSLoggerPlugin
...
-            'Reached local store max size of: '
-            '${_pluginConfig.localStoreMaxSizeInMB}.Hence logs are deleted from '
-            'local store.',
-          );
-        }
-        logger.error('Failed to create CloudWatch log stream', e);
-        return;
-      }
-
       final batcheStream = _getLogBatchesToSync();
```
dnys1, line 140: Typo

packages/logging_cloudwatch/aws_logging_cloudwatch/test/cloudwatch_logger_plugin_test.dart
dnys1: Then there should be tests for the log stream provider as well since that's currently not being tested.

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/log_stream_provider.dart
```diff
@@ -36,15 +40,22 @@ class DefaultCloudWatchLogStreamProvider
   final String _logGroupName;
   final CloudWatchLogsClient _client;
   static final DateFormat _dateFormat = DateFormat('yyyy-MM-dd');
+  final _cachedLogStreams = <String>{};
```
dnys1, line 43: Maybe `_createdLogStreams` is a better name since your keeping track of what's been created.

khatruong2009 (approved):

packages/logging_cloudwatch/aws_logging_cloudwatch/test/cloudwatch_logger_plugin_test.dart
NikaHsn: agree, will add them in a seperate PR.

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/cloudwatch_logger_plugin.dart
NikaHsn: correct, added a TODO

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/log_stream_provider.dart
NikaHsn: updated to Future<String>

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/log_stream_provider.dart
NikaHsn: the DefaultCloudWatchLogStreamProvider is not going to be public. should I move it to the `cloudwatch_logger_plugin.dart` and make it private?

PR #3655, opened 2023-08-30
chore(authenticator_test): Updated Golden tests to run on any screen

Equartey:
*Description of changes:*
- Updated Golden tests to run on any screen & expands the golden test suite to test every screen, except for `AuthenticatorStep.loading`. 
- Fixed Beta channel failures
- Golden test failures are now saved as artifacts for download 
- `amplify_authenticator.ymal` workflow uses a MacOS runner to ensure more consistent testing


packages/authenticator/amplify_authenticator/lib/amplify_authenticator.dart
```diff
@@ -437,6 +438,9 @@ class Authenticator extends StatefulWidget {
   /// {@macro amplify_authenticator_dial_code_options}
   final DialCodeOptions dialCodeOptions;
 
+  @visibleForTesting
+  final AuthenticatorState? mockAuthenticatorState;
```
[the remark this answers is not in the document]
Equartey: These were added to populate the state with required properties depending on the desired authenticator step during testing. For example, some screens such as the TOTP setup screen expect TotpDetails to be non-null. 

Perhaps naming it `authenticatorStateOverride` would make its purpose more explicit?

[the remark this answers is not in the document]
Jordan-Nelson: Do we need setState as well though? Did setting the initial state via the mocked AuthenticatorState and state machine not work, or is there a need to change the state inside a test?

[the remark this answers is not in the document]
Equartey: > Do we need setState as well though? is there a need to change the state inside a test?
Yes -- `setState` is required. The step `continueSignInWithMfaSelection` requires the actual state to be typed as `ContinueSignInWithMfaSelection` in order to retrieve it's required values.

Additionally, having one state for all tests was brittle. Any changes to the business logic of state inputs would need to be duplicated in tests and the bloc, ie transforming the Cognito provided `TotpSetupDetails` into a `setupURI` string. Instead, we only need to supply the required inputs for a given state. Then we are able to reuse the business logic in the bloc and discretely represent the proper state of each screen of the application.

packages/aft/lib/src/commands/generate/generate_workflows_command.dart
```diff
@@ -244,6 +246,8 @@ jobs:
     with:
       package-name: ${package.name}
       working-directory: $repoRelativePath
+      ${hasGoldens ? 'runner-os: macos-latest' : ''}
+      ${hasGoldens ? 'has-goldens: true' : ''}
```
Equartey, line 250: This creates empty lines in other generated workflows when running the generate command. 

Couldn't find a simple way to trim the extra lines when not needed...

@dnys1 would it be better to have them always defined?

.github/workflows/amplify_authenticator.yaml
```diff
@@ -9,46 +9,10 @@ on:
     paths:
       - '.github/workflows/amplify_authenticator.yaml'
       - '.github/workflows/flutter_vm.yaml'
-      - 'packages/amplify/amplify_flutter/lib/**/*.dart'
```
dnys1, line 12: What happened here?

packages/aft/lib/src/commands/generate/generate_workflows_command.dart
```diff
@@ -244,6 +246,8 @@ jobs:
     with:
       package-name: ${package.name}
       working-directory: $repoRelativePath
+      ${hasGoldens ? 'runner-os: macos-latest' : ''}
```
dnys1, line 249: Remove, it's not a parameter you can pass anyway. But only `has-goldens` is needed to know which runner to use.

.github/workflows/amplify_authenticator.yaml
Equartey: Running the generate workflows command removes several paths in all the workflow files. I only included this one though.

.github/workflows/flutter_vm.yaml
```diff
@@ -82,6 +86,14 @@ jobs:
...
 
       - name: Run Tests
+        id: testJob
         if: "always() && steps.bootstrap.conclusion == 'success' && steps.testCheck.outputs.hasTests == 'true'"
         run: flutter test
         working-directory: ${{ inputs.working-directory }}
+
+      - name: Archive golden failures
+        if: "always() && inputs.has-goldens && steps.testJob.conclusion == 'failure'"
+        uses: actions/upload-artifact@65d862660abb392b8c4a3d1195a2108db131dd05
```
dnys1, line 96: Add the version or branch in a comment at the end
```suggestion
        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # 3.1.3
```

packages/authenticator/amplify_authenticator/lib/amplify_authenticator.dart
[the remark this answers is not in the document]
dnys1: I'd rather make `stateMachineBloc` visible for testing than this and call `setState` on it than expose a parameter like this

dnys1 (dismissed):

packages/aft/lib/src/commands/generate/generate_workflows_command.dart
```diff
@@ -214,6 +214,8 @@ $groupPubPackages
     // TODO(dnys1): Enable E2E runs for Dart packages
     final needsE2ETest = package.flavor == PackageFlavor.flutter &&
         package.integrationTestDirectory != null;
+    final hasGoldens = package.flavor == PackageFlavor.flutter &&
+        package.goldensTestDirectory != null;
```
Jordan-Nelson, line 218: An alternative to checking for a directory would be to look for a dev dep of `golden_toolkit`. It isn't required to run golden tests, but I think we would be likely to use it in any future packages that have golden tests. Looking for a ui directory is fine too. If this is still in place when future UI components are built, we will just have to confirm to that dir.

packages/authenticator/amplify_authenticator_test/lib/src/mock_authenticator_app.dart
```diff
@@ -1,17 +1,28 @@
...
 import 'package:amplify_authenticator_test/src/configs/email_config.dart';
-import 'package:amplify_authenticator_test/src/finders/authenticated_app_finder.dart';
-import 'package:amplify_authenticator_test/src/finders/authenticator_finder.dart';
 import 'package:amplify_flutter/amplify_flutter.dart';
 import 'package:amplify_integration_test/amplify_integration_test.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart';
 
+/// Use for testing the [MockAuthenticatorApp] widget locally.
+void main() {
```
Jordan-Nelson, line 17: if you would like to commit this, can you move it to an /example dir and add all the platforms (not just iOS)

Jordan-Nelson (approved):

packages/aft/lib/src/commands/generate/generate_workflows_command.dart
Equartey: Understood, this for sure could be different. Since we don't have any other UI packages or other established conventions, we can readdress this down the road.

dnys1 (approved):

PR #3656, opened 2023-08-30
feat(logging): enable queue rotation

NikaHsn:
*Issue #, if available:*

*Description of changes:*
- add queue rotation
- next: to use queue rotation when log store is full

### DIFF
--- a/packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/dart_queued_item_store.stub.dart
+++ b/packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/dart_queued_item_store.stub.dart
@@ -23,7 +23,11 @@ class DartQueuedItemStore implements QueuedItemStore, Closeable {
   }
 
   @override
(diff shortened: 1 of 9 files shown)


dnys1 (commented):
I'm not sure this is the right interface for this. `enableQueueRotation` implies a persistent change IMO vs. something like `rotate`. But I feel more strongly that this should be a new method, maybe `shiftItem` which is this logic and addItem, or `pop`/`rotate` which is called independently before `addItem`.

NikaHsn (commented):
if the name of `enableQueueRotation` implies a persistant change It can be renamed to `rotate` or `withRotation`. the way I see it is a parameter for the `addItem` method and passed for each method invocation then it only applies to that method call and is not a persistent change.
re adding a new method: let's assume we add a new method called `shiftItem` then callers need to apply a logic to decide if use `addItem` or `shiftItem`. by adding a flag to `addItem ` method the use is easier as they just need to pass the flag to the addItem. so I'm not sure if should add a new method.

dnys1 (approved):

PR #3671, opened 2023-08-31
chore: do not sign optional requests

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*


packages/smithy/smithy_aws/lib/src/http/interceptors/with_sig_v4.dart
```diff
@@ -31,16 +31,9 @@ class WithSigV4 extends HttpRequestInterceptor {
...
-    // for optional auth requests only.
-    try {
-      await credentialsProvider.retrieve();
-    } on Exception {
-      if (isOptional) {
-        return request;
-      }
-      rethrow;
-    }
+    // Do not attempt to sign requests where auth is optional.
```
dnys1, line 34:
```suggestion
    // Do not attempt to sign requests where auth is optional.
    //
    // This is only set in Cognito and SSO services where the trait indicates
    // that signing is strictly unnecessary and that signing the request does
    // not impact the behavior of the APIs. 
```

Wondering if we should put a check in place in smithy_codegen which throws if we encounter an `@optionalAuth` trait which is not in one of these services 🤔

dnys1: For example, we could add this to `packages/smithy_codegen/lib/src/generate.dart`:

```dart

/// Asserts that the `@optionalAuth` trait is only present in Cognito and SSO
/// services, where we know the behavior is that signing is strictly unnecessary.
///
/// Any other services which onboard this trait must be validated for the same behavior
/// before we can generate for them.
class _AssertOptionalAuthVisitor extends CategoryShapeVisitor<Shape> {
  const _AssertOptionalAuthVisitor();

  /// The list of known services where `@optionalAuth` is present and its behavior has
  /// been verified.
  static const _knownServices = [
    'com.amazonaws.cognitoidentity',
    'com.amazonaws.cognitoidentityprovider',
    'com.amazonaws.sso',
    'com.amazonaws.ssooidc',
  ];

  @override
  EnumShape enumShape(EnumShape shape, [Shape? parent]) => shape;

  @override
  ListShape listShape(ListShape shape, [Shape? parent]) => shape;

  @override
  MapShape mapShape(MapShape shape, [Shape? parent]) => shape;

  @override
  MemberShape memberShape(MemberShape shape, [Shape? parent]) => shape;

  @override
  OperationShape operationShape(OperationShape shape, [Shape? parent]) {
    if (shape.hasTrait<OptionalAuthTrait>() &&
        !_knownServices.contains(shape.shapeId.namespace)) {
      throw StateError(
        'Shape has @optionalAuth but it is not from a validated service: '
        '${shape.shapeId}',
      );
    }
    return shape;
  }

  @override
  ResourceShape resourceShape(ResourceShape shape, [Shape? parent]) => shape;

  @override
  ServiceShape serviceShape(ServiceShape shape, [Shape? parent]) => shape;

  @override
  SetShape setShape(SetShape shape, [Shape? parent]) => shape;

  @override
  SimpleShape simpleShape(SimpleShape shape, [Shape? parent]) => shape;

  @override
  StructureShape structureShape(StructureShape shape, [Shape? parent]) => shape;

  @override
  UnionShape unionShape(UnionShape shape, [Shape? parent]) => shape;
}
```

Jordan-Nelson: Updated the comment.

As for the check, I think we have to be able to assume traits mean the same thing across services. In this case "optionalAuth" means that signing the request is not required, has no impact on the behavior.

packages/authenticator/amplify_authenticator/lib/src/services/amplify_auth_service.dart
```diff
@@ -191,9 +201,8 @@ class AmplifyAuthService
   Future<bool> get isLoggedIn async {
     return _withUserAgent(() async {
       try {
-        final result = await Amplify.Auth.fetchAuthSession();
-
-        return result.isSignedIn;
+        await Amplify.Auth.getCurrentUser();
+        return true;
```
Equartey, line 205: Does this throws if it can't get the user?

packages/auth/amplify_auth_cognito_test/test/state/sign_in_state_machine_test.dart
```diff
@@ -136,6 +136,15 @@ void main() {
         ).ignore();
 
       final signInStateMachine = stateMachine.expect(SignInStateMachine.type);
+
+      final fetchAuthSessionStateMachine = stateMachine.getOrCreate(
+        FetchAuthSessionStateMachine.type,
+      );
+
+      fetchAuthSessionStateMachine.stream.listen(
+        (_) => throw StateError('.signIn() should not fetch auth session.'),
```
dnys1, line 145: nb: 
```suggestion
        (_) => fail('.signIn() should not fetch auth session.'),
```

dnys1 (approved):

Equartey (dismissed):

Equartey (approved):

dnys1 (approved):

PR #3692, opened 2023-09-05
fix(authenticator): Wrap social button text

dnys1:
Prevent overflow of social button text in constrained layouts by shrinking the text to fit the available width.

## Before
Screenshot_1693936730

## After
Screenshot_1694019438


Equartey (dismissed):

Equartey (approved):

Equartey (dismissed):

packages/authenticator/amplify_authenticator/lib/src/widgets/social/social_button.dart
```diff
@@ -46,6 +45,8 @@ class SocialSignInButtons extends StatelessAuthenticatorComponent {
                   ?.resolve({}) ??
               Theme.of(context).textTheme.labelLarge;
           final tp = TextPainter(
+            // ignore: deprecated_member_use
```
Jordan-Nelson, line 48:
```suggestion
            // TODO: replace with textScaleFactor when min flutter version is bumped to x.y.z
            // see: https://docs.flutter.dev/release/breaking-changes/deprecate-textscalefactor#migrating-code-that-consumes-textscalefactor
            // ignore: deprecated_member_use
```

packages/authenticator/amplify_authenticator/lib/src/widgets/social/social_button.dart
```diff
@@ -205,24 +217,32 @@ class _SocialSignInButtonState
           builder: (context, constraints) {
             final padding = calculatePadding(constraints);
             return Padding(
-              padding: EdgeInsets.symmetric(horizontal: padding),
+              padding: EdgeInsets.symmetric(
+                vertical: 2,
```
Jordan-Nelson, line 221: I think this is increasing the height of the button. Is that what we want? It seems like the social buttons have a larger height than other buttons here: https://github.com/aws-amplify/amplify-flutter/pull/3692/files#diff-19731debe5953de4ce3baae91c6a663bf423a41677f989d42e1448b480a54353

Jordan-Nelson (approved):

PR #3708, opened 2023-09-07
refactor(ci): Allow region split

dnys1:
Refactors the E2E stack to allow splitting the stacks between regions. Some features are not available in `eu-south-1` so we will be splitting the categories between `us-west-2` (analytics, storage) and `eu-south-1` (auth).

In particular this makes each category's stack a true `Stack` instead of a `NestedStack` so they can have their own region.

**Before**: Root (`Stack`) -> Category (`NestedStack`) -> Category Environment (`NestedStack`)
**After**: Root (`Stack`) -> Category (`Stack`) -> Category Environment (`NestedStack`)

This also migrates a lot of the configuration management to the backend. A `Trigger` runs after every deployment to upload the consolidated `amplifyconfiguration.dart` file. The local Dart file simply downloads the file instead of uploading it.

This also moves the GitHub OIDC and secrets management to CDK so that manual management of the secrets is no longer needed - they are updated automatically after each deployment. To protect Amplify app IDs, a custom resource pulls app IDs on deployment using hard-coded app names (which are not secret).

Overall, this provides cleaner separation of resources and improves deployment speed by a few minutes on average. It also means categories can be deployed independently again and rollbacks happen per category, not for the whole backend.

### DIFF
--- a/.github/workflows/e2e_android.yaml
+++ b/.github/workflows/e2e_android.yaml
@@ -15,6 +15,7 @@ on:
         required: true
         type: boolean
 
+# TEST
 jobs:
   e2e-test:
     runs-on:
--- a/.github/workflows/e2e_android.yaml
+++ b/.github/workflows/e2e_android.yaml
@@ -15,7 +15,6 @@ on:
         required: true
         type: boolean
 
-# TEST
 jobs:
   e2e-test:
     runs-on:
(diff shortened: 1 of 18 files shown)


dnys1:
See last "fixes" commit for passing workflows

Equartey (dismissed):

Equartey (approved):

Jordan-Nelson (approved):

PR #3733, opened 2023-09-12
chore(aft): Package `aft` should not have `pubspec_overrides.yaml`

dnys1:
Unlike every other package, aft must work without `pubspec_overrides`. It's counter productive to generate it and and its presence can interfere with local development.

### DIFF
--- a/packages/aft/lib/src/commands/link_command.dart
+++ b/packages/aft/lib/src/commands/link_command.dart
@@ -59,6 +59,16 @@ Map<String, String> _collectDependencies(
   return dependencyPaths;
 }
 
+/// Removes the `pubspec_overrides.yaml` file for [package].
+Future<void> _removePubspecOverride(PackageInfo package) async {
+  final pubspecOverrides = File(
+    path.join(package.path, 'pubspec_overrides.yaml'),
+  );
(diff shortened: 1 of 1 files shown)


packages/aft/lib/src/commands/link_command.dart
```diff
@@ -59,13 +59,29 @@ Map<String, String> _collectDependencies(
...
   PackageInfo package,
   Map<String, String> dependencyPaths,
   // To be merged in
   Map<String, Dependency> existingDependencyOverrides,
 ) async {
+  // Unlike every other package, aft must work without pubspec_overrides.
+  // It's counter productive to generate it and its presence can interfere
+  // with local development.
+  if (package.name == 'aft') {
+    await _removePubspecOverride(package);
```
Equartey, line 83: Could we prevent the override file from being created in the first place?

Jordan-Nelson (approved):

packages/aft/lib/src/commands/link_command.dart
dnys1: Whoops. Should've been a `return`. Fixed 👍

Jordan-Nelson (approved):

Equartey (approved):

PR #3770, opened 2023-09-20
feat(logging): add LoggingConfig to AmplifyConfig

NikaHsn:
*Issue #, if available:*

*Description of changes:*
- add LoggingConfig to AmplifyConfig
- logger package to use core logging config


khatruong2009 (approved):

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/cloudwatch_logger_plugin.dart
```diff
@@ -273,7 +276,15 @@ class CloudWatchLoggerPlugin extends AWSLoggerPlugin
       return false;
     }
     final loggingConstraint = _getLoggingConstraint();
-    return logEntry.level >= loggingConstraint.defaultLogLevel;
+    LogLevel defaultLogLevel;
+    try {
+      defaultLogLevel = LogLevel.values
+          .byName(loggingConstraint.defaultLogLevel.toLowerCase());
```
dnys1, line 282: Nah I would throw here - we shouldn't accept invalid configurations. Users may have had a typo or something silly.

dnys1 (approved):

packages/logging_cloudwatch/aws_logging_cloudwatch/lib/src/cloudwatch_logger_plugin.dart
dnys1: Throw a `ConfigurationError` type

NikaHsn: updated the logging config to use enum instead of string. so if the config file is not valid it throws at the configuration.

PR #3807, opened 2023-09-25
fix(aft): removed 'has-goldens' check from dart_vm.yaml workflows

Equartey:
*Description of changes:*
Fix for workflows trying to pass `has-goldens: false` to `dart_vm.yaml` runner. Dart runner should never have golden tests so we should not be passing this flag.


dnys1 (approved):
LGTM. Fix for failing test: https://github.com/aws-amplify/amplify-flutter/pull/3808

fjnoyp (approved):

Issue #3805, opened 2023-09-25
Trim Username/Email fields before validation


denxorz:
### Description

When our users fill the email and username field (automatically) they might leave a space at the end of a field. This now leads to the error "Invalid email format" or "Username must only contain alphanumeric characters and symbols". This confuses our users.

It would be nice to trim the fields before running the validation.

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [X] Auth
- [X] Authenticator
- [ ] DataStore
- [ ] Notifications (Push)
- [ ] Storage

### Steps to Reproduce

1. Enter "username " in a username field
2. Try to login

or

1. Enter "email@test.com " in an email field
2. Try to sign up

### Platforms

- [X] iOS
- [X] Android
- [ ] Web
- [ ] macOS
- [X] Windows
- [ ] Linux

### Flutter Version

3.13.1

### Amplify Flutter Version

1.4.0

### Deployment Method

Custom Pipeline

khatruong2009:
The fix for this was included in v1.6.0. Please let us know if you face any other issues.

PR #3809, opened 2023-09-25
fix(authenticator): fix validate before trim bug

khatruong2009:
*Issue #, if available:*
#3805
*Description of changes:*
- Trim username/email before it is validated


packages/authenticator/amplify_authenticator/test/sign_up_form_test.dart
```diff
@@ -172,6 +219,60 @@ void main() {
...
+          final signInPage = SignUpPage(tester: tester);
+
+          await signInPage.enterUsername('user@example.com ');
+          await signInPage.enterPassword('Password123');
+
+          await signInPage.submitSignUp();
+
+          final usernameFieldError = find.descendant(
+            of: signInPage.usernameField,
+            matching: find.text('Invalid email format.'),
```
Equartey, line 242: question: This could cannot catch a validation error if we ever change the error message. Is there another assertion that can be made here? It's okay if not -- but worth looking into imo.

packages/authenticator/amplify_authenticator/test/sign_in_form_test.dart
```diff
@@ -1,22 +1,59 @@
...
+
   group('Sign In View', () {
     group('navigation', () {
       testWidgets('via TabBar', (tester) async {
         await tester.pumpWidget(const MockAuthenticatorApp());
         await tester.pumpAndSettle();
 
         final signInPage = SignInPage(tester: tester);
-        // ignore: cascade_invocations
-        signInPage.expectStep(AuthenticatorStep.signIn);
```
Jordan-Nelson, line 19: *Question:* Was this removed on purpose?

Equartey (approved):

packages/authenticator/amplify_authenticator/lib/src/utils/validators.dart
```diff
@@ -173,7 +173,9 @@ FormFieldValidator<String> validateEmail({
         context,
         InputResolverKey.emailEmpty,
       );
-    } else if (!emailRegex.hasMatch(email)) {
+    }
+    email = email.trim();
```
Jordan-Nelson, line 177: Should we do this for the phone validator too?

Jordan-Nelson (approved):

PR #3874, opened 2023-09-29
Feat/mine/finalized dart action v3

fjnoyp:
*Issue #, if available:*

*Description of changes:*

Add logic to send Cloudwatch metrics from all Github Action workflows. 

Update aft to generate workflows that track metrics. 
Update base github workflows to track metrics. 

WIP: 
Null matrix causes log metric to crash. 
Null value error unknown cause 


fjnoyp:
Please ignore: 

1. Switched test runners (switched to perform local tests on my account) 
2. Any workflow ending in 'copy' they are forced to run to check behavior 

Desired review: 
1. aft changes
2. base github action workflow changes (flutter_vm.yaml dart_vm.yaml etc)

.github/workflows/amplify_analytics_pinpoint copy.yaml
```diff
@@ -0,0 +1,26 @@
+# Generated with aft. To update, run: `aft generate workflows`
```
dnys1, line 1: Why all these copies?

packages/aft/lib/src/commands/generate/generate_workflows_command.dart
```diff
@@ -593,3 +592,9 @@ jobs:
     workflowFile.writeAsStringSync(content);
   }
 }
+
+const permissionsBlock = '''
```
dnys1, line 596: Move back to defining inline. Can keep `const` if all workflows need secrets now.

.github/workflows/flutter_vm.yaml
```diff
@@ -97,3 +97,20 @@ jobs:
...
+          role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
+          aws-region: ${{ secrets.AWS_REGION }}
+
+          job-status: ${{ job.status }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+          test-type: unit
+          working-directory: ${{ inputs.working-directory }}
+
+          framework: flutter
```
dnys1, line 114: The other way to do this instead of passing this in for each workflow is just keep a map in the action of workflowName -> values.

PR #3891, opened 2023-10-02
chore(repo): Dials scheduled GH Actions to 1x/day

haverchuck:
*Description of changes:*
Alters codegenerated action files so that scheduled actions now run 1x per day instead of 2x.


Equartey (approved):
LGTM

PR #3912, opened 2023-10-07
chore(repo): Expand CW Metric logging to all tests

fjnoyp:
Issue #, if available:

Description of changes:

Add logic to send Cloudwatch metrics from all Github Action workflows.

Update aft to generate workflows that track metrics.
Update base github workflows to track metrics.


fjnoyp:
Please note that all dart unit tests on beta have been failing since an unknown time due to a compile issue with Object on a smithy http_operation.  This will need to be fixed separately. 

Failing scheduled dart test (same for all category dart tests):

https://github.com/aws-amplify/amplify-flutter/actions/runs/6439416299/job/17487065066

.github/composite_actions/log_cw_metric_wrapper/action.yaml
```diff
@@ -85,8 +85,8 @@ runs:
 
         metric-name: "github_metric_1.0"
         test-type: ${{ inputs.test-type }}
-        package-name: ${{ inputs.package-name }}
-        matrix: ${{ toJson(matrix) }}
+        working-directory: ${{ inputs.working-directory }}
+        matrix: ${{ toJson(matrix) != 'null' && toJson(matrix) || '' }}
```
Jordan-Nelson, line 89: I'm not following this. Is there a way to write this more clearly in code? If not, can you add a comment.

.github/workflows/amplify_analytics_pinpoint.yaml
```diff
@@ -78,7 +78,10 @@ defaults:
   run:
     shell: bash
 
-permissions: read-all
+# These permissions are needed to interact with GitHub's OIDC Token endpoint.
+permissions:
+  id-token: write
```
Jordan-Nelson, line 83: Why is write access needed?

.github/composite_actions/log_cw_metric/action.yaml
```diff
@@ -22,9 +21,6 @@ inputs:
   test-type:
     description: canary, integration, unit testType.
     required: true
-  category:
```
fjnoyp, line 25: This is now computed based on working-directory

.github/composite_actions/log_cw_metric_wrapper/action.yaml
fjnoyp, line 89: Updated to handle null matrix cases

.github/workflows/amplify_analytics_pinpoint_android.yaml
```diff
@@ -18,7 +18,10 @@ on:
 defaults:
   run:
     shell: bash
-permissions: read-all
+# These permissions are needed to interact with GitHub's OIDC Token endpoint.
+permissions:
+  id-token: write
+  contents: read
```
fjnoyp, line 24: All workflows need these permissions now in order to send the Cloudwatch Metric

.github/workflows/dart_ddc.yaml
```diff
@@ -12,8 +12,7 @@ on:
         type: string
 
 jobs:
-  test:
-    name: Test
+  test-dart-ddc:
```
fjnoyp, line 15: NOTE: in order to properly distinguish jobs within a given workflow run, the jobs must not have a name field set and must have a unique identifiable name.  Please see the comments added to log_cw_metric.dart for more information.

.github/composite_actions/log_cw_metric_wrapper/action.yaml
fjnoyp: Good point, this code can be improved.  Updated to:

```
# Sending a null value will cause the action to fail so we must send an empty string here.
        matrix: ${{ matrix == '' && '{}' || toJSON(matrix) }}
```
        
If the matrix value is empty we send a blank {} else we convert to json and send back.

.github/workflows/amplify_analytics_pinpoint.yaml
fjnoyp: Setting the permissions to write is required in order to request an OpenID Connect JWT.

https://stackoverflow.com/questions/72183048/what-is-the-permission-scope-of-id-token-in-github-action

Removing this permission causes logging to fail to assume the token.

Jordan-Nelson (approved):

.github/composite_actions/log_cw_metric_wrapper/action.yaml
```diff
@@ -85,8 +85,10 @@ runs:
 
         metric-name: "github_metric_1.0"
         test-type: ${{ inputs.test-type }}
-        package-name: ${{ inputs.package-name }}
-        matrix: ${{ toJson(matrix) }}
+        working-directory: ${{ inputs.working-directory }}
+
+        # Sending a null value will cause the action to fail so we must send an empty string here.
+        matrix: ${{ matrix == '' && '{}' || toJSON(matrix) }}
```
haverchuck, line 91: Question asked out of ignorance: Where does the `matrix` var come from, and why wouldn't we do ${{ toJSON(matrix) || matrix == '' && '{}'  }}?

.github/workflows/amplify_analytics_pinpoint.yaml
haverchuck: We need a token in order to post the log to cloudwatch, I assume?

fjnoyp: Yes that's correct.

fjnoyp: This utilizes the same framework that AppSec signed off several months ago when we just started sending Cloudwatch logs from our canaries.

.github/composite_actions/log_cw_metric_wrapper/action.yaml
fjnoyp: `matrix` is a github context variable see: https://docs.github.com/en/actions/learn-github-actions/contexts#matrix-context

I guess we could do it that way as it's just an inversion of the logic.

haverchuck: I read this statement as:

> if `matrix` equals "{}", assign the matrix yaml value to "true".
> ELSE
> parse the matrix yaml value from json

I don't understand that logic. Why would we set the matrix yaml value to "true"? Or am I misreading this? I'm very new to GH actions so it's entirely possible I am.

fjnoyp: We don't have ternary operators in Github Actions but that's essentially what we're trying to do here. If matrix is null we want to send {} otherwise we send a json version of matrix.  

To do that we first check if matrix environment variable is empty or not.  If it's empty we send {} this is achieved via &&, otherwise (||) we send toJSON(matrix). 

Yes it is a javascript style logical operators.   It is not evaluating to both '' and '{}' at the same time, it is using the '' to check matrix value before returning {} in that case. 

`${{ matrix == '' && '{}' || toJSON(matrix) }}`

haverchuck (dismissed):

Jordan-Nelson (approved):

haverchuck (approved):

PR #3943, opened 2023-10-12
chore(logging): fixes and improvement to logging example app

fjnoyp:
*Issue #, if available:*

*Description of changes:*

Fix and clean logger example app. 

1. Initialize Amplify in stateful widget like our other example apps
2. Fix the navigation error
3. Remove redundant code by storing category loggers in a map
4. Fix UI issues with clipping overflow content by refactoring UI elements 

### DIFF
--- a/packages/logging_cloudwatch/amplify_logging_cloudwatch/example/lib/amplifyconfiguration_logging.dart
+++ b/packages/logging_cloudwatch/amplify_logging_cloudwatch/example/lib/amplifyconfiguration_logging.dart
@@ -2,14 +2,10 @@ const loggingconfig = '''{
       "plugins": {
         "cloudWatchLoggerPluginConfiguration": {
             "enable": true,
-            "logGroupName": "<log-group>",
-            "region": "<region>",
+            "logGroupName": "test-logging-log-group",
+            "region": "us-west-2",
             "localStoreMaxSizeInMB": 5,
             "flushIntervalInSeconds": 60,
-            "defaultRemoteConfiguration": {
-                "endpoint": "https://<restApiId>.execute-api.<region>.amazonaws.com/prod/loggingconstraints",
-                "refreshIntervalInSeconds": 1200
-            },
             "loggingConstraints": {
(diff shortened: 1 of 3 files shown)


packages/logging_cloudwatch/amplify_logging_cloudwatch/example/lib/amplifyconfiguration_logging.dart
NikaHsn, line 1: can you not check in changes to this file.

packages/logging_cloudwatch/amplify_logging_cloudwatch/example/lib/homepage.dart
```diff
@@ -10,40 +10,51 @@ class MyHomePage extends StatefulWidget {
...
-  final _analyticsLogger = AmplifyLogger.category(Category.analytics);
-  LogLevel? _analyticsLogLevel;
-  final TextEditingController _analyticsLogMsgController =
-      TextEditingController();
+class LoggerInfo {
+  LoggerInfo(
+    this.category,
+    this.logger,
+    this.logMsgController, {
+    this.logLevel = LogLevel.verbose,
```
NikaHsn, line 18: nit: better to not set a default value so the log level would be null if not set by user.

fjnoyp: Is that desired behavior though?  Since if it's null the logic prevents any log from being sent while wiping the user's log message that they wrote?

NikaHsn (approved):

haverchuck (approved):

packages/logging_cloudwatch/amplify_logging_cloudwatch/example/lib/homepage.dart
NikaHsn: good observation. they way I see it is if the log level is not set then user would see it and set the value. given the app does not show error if user does not set the log level but put a log message and wipes out the log message it is a good idea to set the log level to vebose.

PR #3961, opened 2023-10-16
fix(logging): use indexedDB on web if it is supported

NikaHsn:
*Issue #, if available:*

*Description of changes:*
- indexedDb adapter to wait for the test DB open result when checking if indexedDB is supported.
- DartQueuedItemStore web implementation to wait for indexedDb adapter and fall back to in memory if indexedDb is not supported
- indexedDb adapter getAll method to read result as Object instead of Map

### DIFF
--- a/packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/dart_queued_item_store.web.dart
+++ b/packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/dart_queued_item_store.web.dart
@@ -3,6 +3,8 @@
 
 // ignore_for_file: implementation_imports
 
+import 'dart:async';
+
 import 'package:amplify_core/amplify_core.dart';
 import 'package:amplify_logging_cloudwatch/src/queued_item_store/index_db/indexed_db_adapter.dart';
 import 'package:aws_logging_cloudwatch/src/queued_item_store/in_memory_queued_item_store.dart';
@@ -17,8 +19,8 @@ class DartQueuedItemStore
   // ignore: avoid_unused_constructor_parameters
   DartQueuedItemStore(String? storagePath);
 
-  late final QueuedItemStore _database = () {
-    if (IndexedDbAdapter.checkIsIndexedDBSupported()) {
+  late final Future<QueuedItemStore> _database = () async {
+    if (await IndexedDbAdapter.checkIsIndexedDBSupported()) {
       return IndexedDbAdapter();
     }
     logger.warn(
@@ -40,7 +42,8 @@ class DartQueuedItemStore
     String timestamp, {
     bool enableQueueRotation = false,
   }) async {
-    await _database.addItem(
+    final db = await _database;
+    await db.addItem(
       string,
       timestamp,
       enableQueueRotation: enableQueueRotation,
@@ -49,29 +52,34 @@ class DartQueuedItemStore
 
   @override
   Future<void> deleteItems(Iterable<QueuedItem> items) async {
-    await _database.deleteItems(items);
+    final db = await _database;
+    await db.deleteItems(items);
   }
 
   @override
   Future<Iterable<QueuedItem>> getCount(int count) async {
-    return _database.getCount(count);
+    final db = await _database;
+    return db.getCount(count);
   }
 
   @override
   Future<Iterable<QueuedItem>> getAll() async {
-    return _database.getAll();
+    final db = await _database;
+    return db.getAll();
   }
 
   @override
-  bool isFull(int maxSizeInMB) {
-    return _database.isFull(maxSizeInMB);
+  Future<bool> isFull(int maxSizeInMB) async {
+    final db = await _database;
+    return db.isFull(maxSizeInMB);
   }
 
   /// Clear IndexedDB data.
   @override
   @visibleForTesting
   Future<void> clear() async {
-    return _database.clear();
+    final db = await _database;
+    return db.clear();
   }
 
   @override
--- a/packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/index_db/indexed_db_adapter.dart
+++ b/packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/index_db/indexed_db_adapter.dart
@@ -97,10 +97,10 @@ class IndexedDbAdapter implements QueuedItemStore {
     for (final elem in request.result) {
       final value = elem as Object;
       final id = getProperty<int>(value, 'id');
-      final string = getProperty<String>(value, 'value');
+      final itemValue = getProperty<String>(value, 'value');
       final timestamp = getProperty<String>(value, 'timestamp');
       readValues.add(
-        QueuedItem(id: id, value: string, timestamp: timestamp),
+        QueuedItem(id: id, value: itemValue, timestamp: timestamp),
       );
     }
     return readValues;
@@ -129,25 +129,7 @@ class IndexedDbAdapter implements QueuedItemStore {
 
   @override
   Future<Iterable<QueuedItem>> getAll() async {
-    final readValues = <QueuedItem>[];
-
-    await _databaseOpenEvent;
-    final store = _getObjectStore();
-    final request = store.getAll(null, null);
-
-    await request.future;
-
-    for (final elem in request.result) {
-      final value = elem as Map<String, dynamic>;
-      final id = value['id'] as int;
-      final itemValue = value['value'] as String;
-      final timestamp = value['timestamp'] as String;
-      readValues.add(
-        QueuedItem(id: id, value: itemValue, timestamp: timestamp),
-      );
-    }
-
-    return readValues;
+    return getCount();
   }
 
   @override
@@ -168,14 +150,14 @@ class IndexedDbAdapter implements QueuedItemStore {
   void close() {}
 
   /// Check that IndexDB will work on this device.
-  static bool checkIsIndexedDBSupported() {
+  static Future<bool> checkIsIndexedDBSupported() async {
     if (indexedDB == null) {
       return false;
     }
     // indexedDB will be non-null in Firefox private browsing,
     // but will fail to open.
     try {
-      indexedDB!.open('test', 1).result;
+      await indexedDB!.open('test', 1).future;
       return true;
     } on Object {
       return false;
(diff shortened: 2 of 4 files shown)


khatruong2009 (approved):

Equartey (approved):

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/dart_queued_item_store.web.dart
```diff
@@ -15,18 +17,23 @@ class DartQueuedItemStore
     implements QueuedItemStore, Closeable {
   /// {@macro amplify_logging_cloudwatch.index_db_queued_item_store}
   // ignore: avoid_unused_constructor_parameters
-  DartQueuedItemStore(String? storagePath);
+  DartQueuedItemStore(String? storagePath) {
```
fjnoyp, line 20: Shouldn't this be an async constructor?  Also doesn't this introduce race conditions if you call the constructor and then call one of the async instance methods?  The instance methods do not check if _init() was completed.

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/index_db/indexed_db_adapter.dart
```diff
@@ -138,15 +138,14 @@ class IndexedDbAdapter implements QueuedItemStore {
     await request.future;
 
     for (final elem in request.result) {
-      final value = elem as Map<String, dynamic>;
-      final id = value['id'] as int;
-      final itemValue = value['value'] as String;
-      final timestamp = value['timestamp'] as String;
+      final value = elem as Object;
```
fjnoyp, line 141: This code appears to be a duplicate of `getCount` - can we please refactor the code to avoid duplication?

fjnoyp, line 141: Also, why are we treating value as Object now?  We should prefer Dart's typing safety instead of casting objects to Object or dynamic.

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/index_db/indexed_db_adapter.dart
```diff
@@ -138,15 +138,14 @@ class IndexedDbAdapter implements QueuedItemStore {
...
+      final id = getProperty<int>(value, 'id');
+      final string = getProperty<String>(value, 'value');
```
fjnoyp, line 143: We should avoid naming variables as data types.  String and string can be used interchangeably in some languages making this more confusing.  Why did we rename from itemValue?

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/index_db/indexed_db_adapter.dart
```diff
@@ -168,14 +167,15 @@ class IndexedDbAdapter implements QueuedItemStore {
...
-  static bool checkIsIndexedDBSupported() {
+  static Future<bool> checkIsIndexedDBSupported() async {
     if (indexedDB == null) {
       return false;
     }
     // indexedDB will be non-null in Firefox private browsing,
     // but will fail to open.
     try {
-      indexedDB!.open('test', 1).result;
+      final openRequest = indexedDB!.open('test', 1);
```
fjnoyp, line 177: Shouldn't we just await indexedDB!.open('test', 1); instead of making 2 lines?

fjnoyp (changes requested):

fjnoyp:
@NikaHsn looking through related code, I think we have a similar problem in: `dart_queued_item_store.vm.dart` 

In the below code we appear to also be 'hiding' the async initialization of DartQueuedItemStore.  We have async code in the constructor that we do not await.  If the call to `getApplicationSupportDirectory` returns a future, we then call the constructor of DriftQueuedItemStore with a future object which could cause an exception/crash. 

```
class DartQueuedItemStore implements QueuedItemStore, Closeable {
  /// {@macro amplify_logging_cloudwatch.dart_queued_item_store}
  factory DartQueuedItemStore(String? storagePath) {
    final FutureOr<String> path;
    if (storagePath == null) {
      path = getApplicationSupportDirectory().then((value) => value.path);
    } else {
      path = storagePath;
    }
    final database = DriftQueuedItemStore(path);
    return DartQueuedItemStore._(database);
  }
```

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/index_db/indexed_db_adapter.dart
NikaHsn: not casting as Object throws exception.

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/dart_queued_item_store.web.dart
NikaHsn: it should not cause race conditions because Dart event loop handles microtasks queue first and then events queue. 
to do it async and wait for _database to init means all instance methods needs to be async. currently the `isFull` method is a sync method. can make all instance method async and wait for _database if that's the right way to do it.

NikaHsn:
> @NikaHsn looking through related code, I think we have a similar problem in: `dart_queued_item_store.vm.dart`
> 
> In the below code we appear to also be 'hiding' the async initialization of DartQueuedItemStore. We have async code in the constructor that we do not await. If the call to `getApplicationSupportDirectory` returns a future, we then call the constructor of DriftQueuedItemStore with a future object which could cause an exception/crash.
> 
> ```
> class DartQueuedItemStore implements QueuedItemStore, Closeable {
>   /// {@macro amplify_logging_cloudwatch.dart_queued_item_store}
>   factory DartQueuedItemStore(String? storagePath) {
>     final FutureOr<String> path;
>     if (storagePath == null) {
>       path = getApplicationSupportDirectory().then((value) => value.path);
>     } else {
>       path = storagePath;
>     }
>     final database = DriftQueuedItemStore(path);
>     return DartQueuedItemStore._(database);
>   }
> ```

`factory DriftQueuedItemStore(FutureOr<String> storagePath)` :  DriftQueuedItemStore accepts Future as input parameter. so no need to wait for `getApplicationSupportDirectory`. `packages/common/amplify_db_common_dart/lib/src/connect_io.dart` wait for async call to get path if it is a Future.

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/index_db/indexed_db_adapter.dart
fjnoyp: What exception do you get though?  Casting to the wrong type will get an exception but doesn't mean we should cast to Object.

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/dart_queued_item_store.web.dart
fjnoyp: Thanks that makes a lot more sense then.  We haven't really used microtasks in the code before and I'm unfamiliar with them but that makes sense.  

But yes, the async Future methods will be good but sync method `isFull` will likely not work.

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/index_db/indexed_db_adapter.dart
NikaHsn: it was casting to Map<String, dynamic> and the exception was something like `javascript object is not Map ...`. 
updated the code to call getCount method to reuse the implementation as you suggested.

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/dart_queued_item_store.web.dart
NikaHsn: make sense. I will update the code to make the _database be a Future<QueuedItemStore> and the isFull to be async.

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/index_db/indexed_db_adapter.dart
```diff
@@ -129,25 +129,7 @@ class IndexedDbAdapter implements QueuedItemStore {
...
-      final id = value['id'] as int;
-      final itemValue = value['value'] as String;
-      final timestamp = value['timestamp'] as String;
-      readValues.add(
-        QueuedItem(id: id, value: itemValue, timestamp: timestamp),
-      );
-    }
-
-    return readValues;
+    return getCount();
```
fjnoyp, line 132: Why do we have 2 separate methods if they do the same exact thing?

fjnoyp:
Regarding the `as Object` cast - there's a better way to do this using the JSONSerializable package instead. 

See the [github_jobs.dart file](https://github.com/aws-amplify/amplify-flutter/blob/main/actions/lib/src/githubJobs/github_jobs.dart) for an example of how to do it.  In the future we should refactor code to use that.

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/index_db/indexed_db_adapter.dart
fjnoyp: Yeah it's tricky with JSInterop.  I think you need to call '`SON.decode(` method to convert from a javascript object to a map.  Better would be to use codegen described here: https://github.com/aws-amplify/amplify-flutter/pull/3961#issuecomment-1767296183

packages/logging_cloudwatch/amplify_logging_cloudwatch/lib/src/queued_item_store/index_db/indexed_db_adapter.dart
NikaHsn: > Why do we have 2 separate methods if they do the same exact thing?

this is the case only for IDBObjectStore. other implementations of QueuedItemStore have different impl for getCount and getAll.

fjnoyp: Ah ok thanks for clarifying

fjnoyp (approved):
Looks good thanks for addressing those comments!

PR #3963, opened 2023-10-17
fix(smithy): casts output to fix beta tests

haverchuck:
*Description of changes:*
Casts output to fix the following `beta` test failure:

```
Failed to build aft:aft:
packages/smithy/smithy/lib/src/http/http_operation.dart:271:16: Error: A value of type 'Object?' can't be assigned to a variable of type 'Output?'.
 - 'Object' is from 'dart:core'.
      output = switch (payload) {
               ^
```
            

### DIFF
--- a/packages/smithy/smithy/lib/src/http/http_operation.dart
+++ b/packages/smithy/smithy/lib/src/http/http_operation.dart
@@ -268,7 +268,6 @@ abstract class HttpOperation<InputPayload, Input, OutputPayload, Output>
     var successCode = this.successCode();
     try {
       final payload = await protocol.deserialize(response.split());
-      // ignore: unnecessary_cast
       output = switch (payload) {
         final Output p => p,
         _ => buildOutput(payload, response),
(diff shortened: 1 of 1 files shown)


packages/smithy/smithy/lib/src/http/http_operation.dart
```diff
@@ -268,10 +268,11 @@ abstract class HttpOperation<InputPayload, Input, OutputPayload, Output>
     var successCode = this.successCode();
     try {
       final payload = await protocol.deserialize(response.split());
+      // ignore: unnecessary_cast
       output = switch (payload) {
         Output _ => payload,
         _ => buildOutput(payload, response),
-      };
+      } as Output;
```
Jordan-Nelson, line 275: I think this is the preferred fix.

```dart
      output = switch (payload) {
        final Output p => p,
        _ => buildOutput(payload, response),
      };
```

The issue with the statement currently is that line in line 273 `payload` is type `OutputPayload`. We know it is type `Output` (line 273 would not execute if it were not) so the cast here should be safe, but I think we can avoid it by returning `p` which the dart compilers knows is type `Output`.

NikaHsn (approved):

Jordan-Nelson (approved):

Issue #3949, opened 2023-10-14
[Request] Indicate current device on AuthDevice


dtodt:
### Description

I'm implementing MFA SMS, and after confirm a user and try to remember the device, I get a DeviceNotTrackedException.

<details>
<summary>Console error</summary>

```sh
AmplifyAuthCognitoDart.rememberDevice (package:amplify_auth_cognito_dart/src/auth_plugin_impl.dart:938:7)
```

</details>

<details>
<summary>Exception message</summary>

```json
{"message":"This device does not have an id, either it was never tracked or previously forgotten."}
```

</details>

Debugging the code I found out that the following code is returning null.

<details>
<summary>auth_plugin_impl.dart @ rememberDevice</summary>

```dart
Future<void> rememberDevice() async {
  final tokens = await stateMachine.getUserPoolTokens();
  final username = tokens.username;
  final deviceSecrets = await _deviceRepo.get(username); // << HERE
  final deviceKey = deviceSecrets?.deviceKey; // << HERE
  if (deviceSecrets == null || deviceKey == null) {
    throw const DeviceNotTrackedException();
  }
...
```

</details>

I'm not sure why, this should work normally.

> [Update 1]

Context: 
We have as a business rule, to allow only one mobile device at a time, so on every new device, we forget the old ones.

Debugging the package, I found out that the local device secrets is created when I confirm sign in with the provided code. And when I run `forgetDevice`, the `auth_plugin_impl.dart` removes the local device secrets along with the remote ones, which causes the failure and exception at `rememberDevice`.

Is there any way to forget other devices only?

> [Update 2]

After some search, I found out that, maybe during forgetDevice a fix could be done, so that when you are forgetting a device you doesn't accidentally forget the local device.

<details>
<summary>Possible fix</summary>

```dart
Future<void> forgetDevice([AuthDevice? device]) async {
  final tokens = await stateMachine.getUserPoolTokens();
  final username = tokens.username;
  final deviceSecrets = await _deviceRepo.get(username);
  final deviceKey = device?.id ?? deviceSecrets?.deviceKey;
  if (deviceKey == null) {
    throw const DeviceNotTrackedException();
  }
>>> current
  await _deviceRepo.remove(username);

>>> fix
  if (device == null || device.id == deviceSecrets?.deviceKey) {
    await _deviceRepo.remove(username);
  }
...
```

</details>

### Categories

- [X] Auth

### Steps to Reproduce

1. signIn
[...]

PR #3966, opened 2023-10-18
fix(auth): forget local device

dtodt:
*Issue #3949 :*

*Description of changes:*
- auth device current property added;
- cognito device current property added;
- fetchDevices method marks the current device (when exists);
- forgetDevice only removes the local device if matches;

The fetchDevices changes allow to conditionally forget remote devices, ex:
```dart
final devices = await _cognito().fetchDevices();
for (final device in devices) {
  if (device.current ?? false) continue;
  await _cognito().forgetDevice(device);
}
```


dtodt:
Split into two:

https://github.com/aws-amplify/amplify-flutter/pull/4060
https://github.com/aws-amplify/amplify-flutter/pull/4061

PR #399, opened 2021-02-20
fix(amplify_datastore): avoid tearing down iOS subscription after clear()

Amplifiyer:
*Issue #, if available:* fixes #395

*Description of changes:* fix(amplify_datastore): avoid tearing down iOS subscription after clear()

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3987, opened 2023-10-23
fix(analytics): event client flush events to do not discard events from cache on auth exception

NikaHsn:
*Issue #, if available:*
- when unauthenticated access is disabled, Analytics can record events when user is not logged-in but flush events fails due to auth exception and events are cleared from cache.

*Description of changes:*
- if flush events (PutEvents request) fails due to Auth exception, do not clear cache as these events are always retryable


fjnoyp (approved):

khatruong2009 (approved):

NikaHsn:
closing in favour of #399

PR #4016, opened 2023-10-26
fix(actions): no longer send empty string values

haverchuck:
*Description of changes:*
Stops sending empty strings in cw metric dimensions.


actions/bin/log_cw_metric.dart
```diff
@@ -60,8 +62,8 @@ Future<void> logMetric() async {
   */
 
   // Inputs for Metric
-  final metricName = core.getRequiredInput('metric-name');
-  final testType = core.getRequiredInput('test-type');
+  final metricName = core.getRequiredInput('metric-name') ?? defaultValue;
```
fjnoyp, line 65: We shouldn't have defaultValue for requiredInput as an exception should be thrown if a value is not supplied here.  Also for testType, we throw if an improper value is given anyway below.

haverchuck: Fixed.

PR #4054, opened 2023-11-01
feat(logging): add a new logging category and plugin interface

NikaHsn:
*Issue #, if available:*

*Description of changes:*
- add a new logging category
- add a new logging plugin interface
- add a new logger interface

### DIFF
--- a/packages/amplify_core/lib/src/category/amplify_logging_category.dart
+++ b/packages/amplify_core/lib/src/category/amplify_logging_category.dart
@@ -38,21 +38,27 @@ class LoggingCategory extends AmplifyCategory<LoggingPluginInterface> {
   }
 
   /// {@template amplify_core.amplify_logging_category.enable}
-  /// Starts recording logs emitted by [logger].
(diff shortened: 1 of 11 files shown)


packages/amplify_core/lib/src/category/amplify_logging_category.dart
```diff
@@ -0,0 +1,60 @@
...
+
+  /// {@template amplify_core.amplify_logging_category.enable}
+  /// Starts recording logs emitted by [logger].
+  /// {@endtemplate}
+  void enable() {
+    return defaultPlugin.enable();
+  }
+
+  /// {@template amplify_core.amplify_logging_category.disable}
+  /// Stops recording logs emitted by [logger].
```
Equartey, line 48: suggestion: I'd add a call out that this will drop logs and `flush()` should be called if they want to preserve them.

Equartey (approved):

khatruong2009 (approved):

PR #4058, opened 2023-11-02
fix(actions): send metrics on pr and send none values

haverchuck:
*Issue #, if available:*
Fixes metrics issues

*Description of changes:*


actions/pubspec.yaml
```diff
@@ -7,7 +7,7 @@ environment:
   sdk: ^3.2.0-150.0.dev
 
 dependencies:
```
haverchuck, line 9: For some reason, the github action was pulling in a very old version of aws_common that was missing a Stream Extension we were relying on. Bumping the version to avoid a slew of "method not supported on Stream" exceptions.

NikaHsn (dismissed):

.github/composite_actions/log_cw_metric_wrapper/action.yaml
```diff
@@ -53,14 +53,6 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - name: Exit if not scheduled
```
fjnoyp, line 56: Are you sure you want to track metrics for every single test trigger?

.github/composite_actions/log_cw_metric_wrapper/action.yaml
```diff
@@ -83,7 +75,7 @@ runs:
         job-status: ${{ inputs.job-status }}
         github-token: ${{ inputs.github-token }}
 
-        metric-name: "github_metric_2.0"
```
fjnoyp, line 86: If this is not a major change why not use 2.1 or 2.01 instead?

fjnoyp (commented):
Thanks looks good left some comments

.github/composite_actions/log_cw_metric_wrapper/action.yaml
haverchuck: Yes, at least for now.

.github/composite_actions/log_cw_metric_wrapper/action.yaml
haverchuck: Kind of wanted to use a whole integer bump until this is stabilized and our dashboards are solidified.

fjnoyp: Right so if there are more changes incoming shouldn't we stay at 2 and then bump to a major version 3 when things are stabilized?

.github/composite_actions/log_cw_metric_wrapper/action.yaml
fjnoyp: If we do this shouldn't we track the trigger type? Otherwise we have no means to separate schedules from PR tests that are failing because the PR is half done.

haverchuck: Oh that's a great point - let me think on it

.github/composite_actions/log_cw_metric_wrapper/action.yaml
haverchuck: Fixed

.github/composite_actions/log_cw_metric_wrapper/action.yaml
haverchuck: Done - sending event type now.

fjnoyp: Where are you sending this event type or other means to distinguish events based on whether they were triggered by a PR / scheduled / etc. ?

haverchuck: @fjnoyp line 139 of [log_cw_metric.dart](https://github.com/aws-amplify/amplify-flutter/pull/4058/files#diff-b295e4ced78ed86faa0d0198c1be5404053c1f96a6c1f35a7bbf96a42768fd9c)

fjnoyp (approved):

NikaHsn (approved):

PR #4061, opened 2023-11-03
feat: auth-device-current prop

dtodt:
Issue #3949 :

Description of changes:

- auth device current property added;
- cognito device current property added;
- fetchDevices method marks the current device (when exists);

The fetchDevices changes allow to conditionally forget remote devices, ex:

```dart
final devices = await _cognito().fetchDevices();
for (final device in devices) {
  if (device.current ?? false) continue;
  await _cognito().forgetDevice(device);
}
```


tyllark:
Sorry for the nearly 2 year late reply. We appreciate your interest in implementing this feature, but we ended up using a different approach as described [here](https://docs.amplify.aws/flutter/build-a-backend/auth/manage-users/manage-devices/#fetch-the-current-device). Thank you!

Issue #3649, opened 2023-08-30
Update Local Subscriptions with GraphQL Response Upon Outbox Mutations Processed


martin-minovski:
### Description

I've encountered an issue with the DataStore plugin when working with a GraphQL API. I have set up an AWS Amplify app, and for two of my models, I've added a custom resolver function to the create mutation pipeline resolver just before the final resolver function that saves data to DynamoDB. This custom resolver function modifies some of the model attributes.

The problem I'm observing is that while the GraphQL create mutation response correctly reflects the modified data, the DataStore subscriptions are not updating the Flutter widgets to match this response. Instead, they persist the unmodified data from the saved model. To force the UI to correctly display the data that's stored on the cloud, I have to restart the app.

Currently, I have a workaround which looks like this:

```dart
StreamSubscription<DataStoreHubEvent>? _hubSubscription;

void observeEvents() {
  _hubSubscription = Amplify.Hub.listen(HubChannel.DataStore, (hubEvent) async {
    if (hubEvent.eventName == 'outboxMutationProcessed') {
      final status = hubEvent.payload as OutboxMutationEvent?;
      if (status!.modelName == 'ChatMessage') {
        await stopObservingMessages();
        observeMessages();
      }
      if (status!.modelName == 'ChatThread') {
        await stopObservingThreads();
        observeThreads();
      }
    }
  });
}

void observeMessages() {
  _messageStream = Amplify.DataStore.observeQuery(
    ChatMessage.classType,
    sortBy: [ChatMessage.SENTAT.descending()],
  ).listen((QuerySnapshot<ChatMessage> snapshot) {
    _chatMessages = snapshot.items;
    notifyListeners();
  });
}

void observeThreads() {
  _threadStream = Amplify.DataStore.observeQuery(
      ChatThread.classType
  ).listen((QuerySnapshot<ChatThread> snapshot) {
    _chatThreads = snapshot.items;
    notifyListeners();
  });
}

Future<void> stopObservingMessages() async {
  if (_messageStream == null) return;
  await _messageStream?.cancel();
  _messageStream = null;
}

Future<void> stopObservingThreads() async {
  if (_threadStream == null) return;
  await _threadStream?.cancel();
  _threadStream = null;
}
```

However, this solution is far from ideal, as it triggers a scan on all items with every sent message.

[...]

PR #4084, opened 2023-11-07
fix(datastore): emit observeQuery snapshot when model create mutation results in an updated model

Jordan-Nelson:
*Issue #, if available:* https://github.com/aws-amplify/amplify-flutter/issues/3649

*Description of changes:*
- When event type is `EventType.create` and the model already exists in the list, evaluate if the model has changed

### DIFF
--- a/packages/amplify_core/lib/src/types/datastore/models/query_snapshot.dart
+++ b/packages/amplify_core/lib/src/types/datastore/models/query_snapshot.dart
@@ -71,37 +71,60 @@ class QuerySnapshot<T extends Model> {
   QuerySnapshot<T> withSubscriptionEvent({
     required SubscriptionEvent<T> event,
   }) {
-    final sortedListCopy = SortedList<T>.from(_sortedList);
     SortedList<T>? updatedSortedList;
 
     final newItem = event.item;
-    final newItemMatchesPredicate = where == null || where!.evaluate(newItem);
-    final currentItemIndex =
+    final matchesPredicate = where == null || where!.evaluate(newItem);
+    final currentIndex =
         // TODO(HuiSF): remove the ignore when merging CPK feature commits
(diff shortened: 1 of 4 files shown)


packages/amplify_core/lib/src/types/datastore/models/query_snapshot.dart
```diff
@@ -71,37 +71,60 @@ class QuerySnapshot<T extends Model> {
...
+          // exists in the list with a different value. This is the result of
+          // the item being created on this device and App Sync returning an
+          // updated item during the create mutation. This can happen when using
+          // custom resolvers.
+          updatedSortedList = _sortedList.copy()
+            ..updateAtSorted(
+              currentIndex,
+              newItem,
+            );
+        }
```
Jordan-Nelson, line 103: *note:* I refactored this and added comments to make it easier to understand but this code block is the only logic change. When the event type is `EventType.create` and the item value has changed (currentItem != newItem) the item in the list is updated with the new value. Previously this was ignored. This is problem when the model create response from App Sync contains an updated model, which is the case when using a custom resolver.

Equartey: question: is `currentItem != newItem` a sufficient comparison, like is it a deep equals?

Jordan-Nelson: Yes, item is of type Model and we override the equality operator on models. For example, a typical Blog model would look like:

```dart
  @override
  bool operator ==(Object other) {
    if (identical(other, this)) return true;
    return other is Blog &&
        id == other.id &&
        _name == other._name &&
        DeepCollectionEquality().equals(_posts, other._posts);
  }
```

Equartey (approved):

NikaHsn (approved):

Issue #4079, opened 2023-11-06
Flutter: Custom Phone Number Form Field accepts empty value still after setting `required` argument as `true`


kavienanj:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

Flutter

### Which UI component?

Authenticator

### How is your app built?

Flutter

### What browsers are you seeing the problem on?

iOS (Flutter), Android (Flutter)

### Please describe your bug.

When adding the phone number field to a form (eg: SignUpForm) with the `required` argument set as `true`, the validator function still treats the field as optional. Which allows the form to be submitted with an empty phone number. The field hint seems to work perfectly. So I assume the `required` argument is not being passed into the `validatePhoneNumber(bool isOptional, ... )` function properly. This assumption is because when I copied the default `validatePhoneNumber` from the library, and set `isOptional` argument as `true'. Then I passed this rewritten validator function to the form field, it worked.

### What's the expected behaviour?

When adding the phone number field to a form (eg: SignUpForm) with the `required` argument set as `true`, if a user trys to submit the form leaving the phone number field empty, the validator should return the 'Phone Number field must not be blank' error text and stop the user from submitting.

### Help us reproduce the bug!

In main.dart file
```dart
class MyApp extends StatelessWidget {
....
    return Authenticator(
      dialCodeOptions: const DialCodeOptions(defaultDialCode: DialCode.im),
      signUpForm: SignUpForm.custom(
        fields: [
          SignUpFormField.username(),
          SignUpFormField.phoneNumber(required: true),
          SignUpFormField.password(),
          SignUpFormField.passwordConfirmation(),
        ],
      ),
      child: MaterialApp.router( ... )
}
```
the `AWSAmplifyService.configure()` runs the necessary Amplify/Cognito configurations. `AppRouterConfig.router` is GoRouter Configurations.
Run the Flutter app and Try submitting the `SignUpForm`. It Should submit with empty phone number.
[...]

PR #4106, opened 2023-11-10
fix(authenticator): required phone number validator

Jordan-Nelson:
*Issue #, if available:* https://github.com/aws-amplify/amplify-flutter/issues/4079

*Description of changes:*
- Drop the country code prefix before performing validation

The country code is populated from a drop down list of known valid numbers so there is no need to validate it. Including it in validation makes it challenging to check for a blank phone number since the country code is always present.

### DIFF
--- a/packages/authenticator/amplify_authenticator/lib/src/utils/validators.dart
+++ b/packages/authenticator/amplify_authenticator/lib/src/utils/validators.dart
@@ -17,7 +17,7 @@ final emailRegex = RegExp(r'^\S+@\S+$');
 /// The "+" prefix is excluded since validation is performed against the number
 /// without the prefix. That is, if the full number is "+1-123-555-7890", the
 /// "+1" is dropped prior to validation.
-final phoneNumberRegex = RegExp(r'^d+$');
+final phoneNumberRegex = RegExp(r'^\d+$');
 final _codeRegex = RegExp(r'^\d{6}$');
 final _uppercase = RegExp(r'[A-Z]');
 final _lowercase = RegExp(r'[a-z]');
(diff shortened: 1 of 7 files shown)


khatruong2009 (approved):

packages/authenticator/amplify_authenticator/lib/src/utils/validators.dart
```diff
@@ -11,7 +11,13 @@ import 'package:flutter/material.dart';
...
 final usernameRegex = RegExp(r'^\S+$');
 final emailRegex = RegExp(r'^\S+@\S+$');
-final phoneNumberRegex = RegExp(r'^\+\d+$');
+
+/// Regex for a valid phone number.
+///
+/// The "+" prefix is excluded since validation is performed against the number
+/// without the prefix. That is, if the full number is "+1-123-555-7890", the
+/// "+1" is dropped prior to validation.
+final phoneNumberRegex = RegExp(r'^d+$');
```
Equartey, line 20: What is this meant to match? Should it not be `^[0-9]+$`?

Equartey: or `^\d+$`?

packages/authenticator/amplify_authenticator/test/sign_up_form_test.dart
```diff
@@ -118,6 +118,44 @@ void main() {
         },
       );
 
+      testWidgets(
+        'displays message when submitted with empty phone number if the field is required',
```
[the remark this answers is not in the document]
Jordan-Nelson: Yes. Line 149 would fail today on main as the phone number is currently never considered empty as there is always a dial code.

packages/authenticator/amplify_authenticator/lib/src/utils/validators.dart
Jordan-Nelson: Yes, the `\` should not have been removed. Updated.

Equartey (approved):

khatruong2009 (approved):

Issue #4189, opened 2023-11-27
Future Request: Search and filter data with Amplify.API without custom GraphQL Document


MarlonJD:
### Description

Hello, on amplify-flutter there is no search option like ModelQueries.list, do we planning to add this?
I would like to try to add this if it's possible. 

Query predicate operations for search will be one of these:
```
matchPhrasePrefix, matchPhrase, match, multiMatch, ne, range, regexp, eq, exists, gt, gte, lt, lte, wildcard
```

for example search operation can be like this example,
it can be build on the ModelQueries.list,

example ModelQueries.list was like this:
```
final request = ModelQueries.list(
  Student.classType,
  where: Student.NAME.eq("John"),
);
```

to 

```
// eq example
final request = ModelQueries.search(
  Student.classType,
  where: Student.NAME.eq("John"),
  limit: 10,
);

// on search matchPhrasePrefix operation will be important
final request = ModelQueries.search(
  Student.classType,
  where: Student.NAME.matchPhrasePrefix("Jo"),
  limit: 10,
);
```

### Categories

- [ ] Analytics
- [ ] API (REST)
- [X] API (GraphQL)
- [ ] Auth
- [ ] Authenticator
- [ ] DataStore
- [ ] Notifications (Push)
- [ ] Storage

### Platforms

- [X] iOS
- [X] Android
- [X] Web
- [X] macOS
- [X] Windows
- [X] Linux

### Flutter Version

3.3.10

### Amplify Flutter Version

0.6.10

### Deployment Method

Amplify CLI

### Schema

```GraphQL
type Student @model @searchable {
  name: String
  dateOfBirth: AWSDate
  email: AWSEmail
  examsCompleted: Int
}
```

Equartey:
Hi @MarlonJD, thanks for sharing the proactive solution. 

It's unclear to me why you would need a separate `ModelQueries.serach()` method. Could you please explain why your use case could not be solved with a new query predicate passed to the existing `ModelQueries.list()` method?

MarlonJD:
Hi @Equartey, thanks. It's for @searchable directive, see https://docs.amplify.aws/flutter/build-a-backend/graphqlapi/search-and-result-aggregations/ here, it's creating new query in graphql, for example `listTodos`is for normal query, but if you use OpenSearch and get search results for this with `searchTodos`. If I use listTodos, 

1. I have to create customDocument and replace listTodos with searchTodos in created document in `ModelQueries.list()`
[...]

PR #4207, opened 2023-12-04
Feature: search in ModelQueries

MarlonJD:
**Issue:**
#4189 

**Description of changes:**
I created ModelQueries.search which based on ModelQueries.list, it's the same with .list except where operations, please check if it's looks good for you, I only added matchPhrasePrefix as a operation, I noted as  `/// It's only for GraphQL API with Amplify.API.query()`

There are DataStore operations like `QuerySortBy` and it's not a creating any issue that only using Amplify.API users. So I just added `matchPhrasePrefix` as a `QueryPredicateOperation`.

If it's looks good, I can add other operations using on search in GraphQL. These are: `matchPhrase, match, multiMatch, ne, range, regexp, eq, exists, gt, gte, lt, lte, wildcard`

but I think it's already enough with `matchPhrasePrefix`.

### DIFF
--- a/packages/amplify_core/lib/src/types/query/query_field.dart
+++ b/packages/amplify_core/lib/src/types/query/query_field.dart
@@ -285,4 +285,26 @@ class QueryField<T> {
   QuerySortBy descending() {
     return QuerySortBy(field: fieldName, order: QuerySortOrder.descending);
   }
+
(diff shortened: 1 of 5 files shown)


MarlonJD:
I created new pr for tests fails, I cloned and made changes and run `ft run test:unit:flutter` #4210

Issue #4432, opened 2024-02-05
java.io.FileNotFoundException: can't read keyset while initializing Amplify


danielgomezrico:
### Description

# Issue
A FileNotFoundException and stack trace is logged while initializing Amplify. The full log looks like:

It was reported on https://github.com/aws-amplify/amplify-flutter/issues/2640 as fixed but it is still happening on:
- amplify_auth_cognito: 1.6.1
- amplify_flutter: 1.6.1

```console
W/AndroidKeysetManager( 1231): keyset not found, will generate a new one
W/AndroidKeysetManager( 1231): java.io.FileNotFoundException: can't read keyset; the pref value __androidx_security_crypto_encrypted_prefs_key_keyset__ does not exist
W/AndroidKeysetManager( 1231): 	at com.google.crypto.tink.integration.android.SharedPrefKeysetReader.readPref(SharedPrefKeysetReader.java:71)
W/AndroidKeysetManager( 1231): 	at com.google.crypto.tink.integration.android.SharedPrefKeysetReader.readEncrypted(SharedPrefKeysetReader.java:89)
W/AndroidKeysetManager( 1231): 	at com.google.crypto.tink.KeysetHandle.read(KeysetHandle.java:105)
W/AndroidKeysetManager( 1231): 	at com.google.crypto.tink.integration.android.AndroidKeysetManager$Builder.read(AndroidKeysetManager.java:311)
W/AndroidKeysetManager( 1231): 	at com.google.crypto.tink.integration.android.AndroidKeysetManager$Builder.readOrGenerateNewKeyset(AndroidKeysetManager.java:287)
W/AndroidKeysetManager( 1231): 	at com.google.crypto.tink.integration.android.AndroidKeysetManager$Builder.build(AndroidKeysetManager.java:238)
W/AndroidKeysetManager( 1231): 	at androidx.security.crypto.EncryptedSharedPreferences.create(EncryptedSharedPreferences.java:123)
W/AndroidKeysetManager( 1231): 	at com.amazonaws.amplify.amplify_secure_storage.amplify_secure_storage.EncryptedKeyValueRepository$sharedPreferences$2.invoke(EncryptedKeyValueRepository.kt:31)
W/AndroidKeysetManager( 1231): 	at com.amazonaws.amplify.amplify_secure_storage.amplify_secure_storage.EncryptedKeyValueRepository$sharedPreferences$2.invoke(EncryptedKeyValueRepository.kt:21)
[...]

PR #4434, opened 2024-02-05
fix: `google.crypto.tink` version constraint

Jordan-Nelson:
*Issue #, if available:* https://github.com/aws-amplify/amplify-flutter/issues/4432

*Description of changes:*
- Fix the `google.crypto.tink` version constraint syntax

Android Security depends on an old version of `google.crypto.tink`. This version logs a long and confusing warning when a new keyset is created. It is not harmful, but it is confusing and clutters logs. We previously resolved this by adding a dependency on v1.8 which contains a fix (https://github.com/tink-crypto/tink/issues/534). However, this dependabot PR (https://github.com/aws-amplify/amplify-flutter/pull/3220) used the incorrect syntax for the version constraint, causing it to be ignored.

### DIFF
--- a/packages/secure_storage/amplify_secure_storage/android/build.gradle
+++ b/packages/secure_storage/amplify_secure_storage/android/build.gradle
@@ -54,7 +54,7 @@ dependencies {
     implementation 'androidx.security:security-crypto:1.0.0'
     // TODO(Jordan-Nelson): remove once security-crypto:1.1.0 is stable.
     // See https://github.com/aws-amplify/amplify-flutter/issues/2640
-    implementation 'com.google.crypto.tink:tink-android:[1.8.0'
+    implementation 'com.google.crypto.tink:tink-android:[1.8.0, )'
     testImplementation 'junit:junit:4.13.2'
     testImplementation 'org.robolectric:robolectric:4.10.3'
     testImplementation 'androidx.test:core-ktx:1.5.0'


Equartey (approved):

packages/secure_storage/amplify_secure_storage/android/build.gradle
```diff
@@ -54,7 +54,7 @@ dependencies {
     implementation 'androidx.security:security-crypto:1.0.0'
     // TODO(Jordan-Nelson): remove once security-crypto:1.1.0 is stable.
     // See https://github.com/aws-amplify/amplify-flutter/issues/2640
-    implementation 'com.google.crypto.tink:tink-android:[1.8.0'
+    implementation 'com.google.crypto.tink:tink-android:[1.8.0, )'
```
NikaHsn, line 57: shall we use an upper bound like [1.8.0, 2.0.0)?

Jordan-Nelson: We don't have any need for an upper bound. We do not use this dependency directly. It is a transitive dependency via `androidx.security:security-crypto`. `androidx.security:security-crypto` depends on an older version (~1.6.0) which logs a confusing message every time a key set is created. We need a version greater than or equal to 1.8 which is when that log was removed but we do not need to have nay constraint on the upper bound.

NikaHsn (approved):

Issue #4490, opened 2024-02-28
Exception: There is no active TOTP setup


dkliss:
### Description
-EDIT, Correction on sync

Hi, 

When setting up TOTP, after Amplify.Auth.setUpTotp() has been setup, we will be  at Amplify.Auth.verifyTotpSetup(totpCode). The verifyTotpSetup() only allows **single attempt** to enter correct code. And if second attempt is made I get below exception. Is this how it is expected to work (i.e. only one attempt)?  

```
flutter: Error confirming attribute update: AuthPreconditionException {
  "message": "There is no active TOTP setup",
  "recoverySuggestion": "Call Amplify.Auth.setUpTotp first"
}
```

```
amplify_flutter: ^1.6.1
amplify_auth_cognito: ^1.6.2

```
To reproduce: 
1. run Amplify.Auth.setUpTotp() 
2. Connect with Authenticator.
3. Run Amplify.Auth.verifyTotpSetup(totpCode) with incorrect code --> **Exception: "Code mismatch"** 
4. Run  Amplify.Auth.verifyTotpSetup(totpCode) with correct code--> **Exception: "There is no active TOTP setup."**
5. Re-run Amplify.Auth.setUpTotp(), which will create new code with different sync secret. 
6. Re-scan new qr code with Authenticator.
7. Re-Run Amplify.Auth.verifyTotpSetup(totpCode) with correct code  -> **This function is void and no success  response is received* --> **No exception means success.**  

In  above scenario, even if I do not remove (i.e. execute step 6) authenticator scanned in step 2 above (Connect with Authenticator), Amplify.Auth.verifyTotpSetup() still works even after a new Amplify.Auth.setUpTotp() is generated in step 5. 

To reproduce (Step 6 removed from above scenario): 

1. run Amplify.Auth.setUpTotp() 
2. Connect with Authenticator.
3. Run Amplify.Auth.verifyTotpSetup(totpCode) with incorrect code --> **Exception: "Code mismatch"** 
4. Run  Amplify.Auth.verifyTotpSetup(totpCode) with correct code--> **Exception: "There is no active TOTP setup."**
5. Re-run Amplify.Auth.setUpTotp(), which will create new code with different sync secret. 
6 Re-Run Amplify.Auth.verifyTotpSetup(totpCode) with correct code  -> **This function is void and no success  response is received* --> **No exception means success.**  

Whats happening in this seems like old QR code is still in Amplify and hence when I add verify code, that one is used.  

[...]

PR #4532, opened 2024-03-06
fix(auth): Allow retries with verifyTotpSetup()

Equartey:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-flutter/issues/4490

*Description of changes:*
Catches and correctly handles missed `EnableSoftwareTokenMFAException` when entering a wrong code during TOTP setup. 


packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/totp_setup_state_machine.dart
```diff
@@ -59,29 +62,52 @@ final class TotpSetupStateMachine
...
+        emit(
+          TotpSetupState.requiresVerification(
+            _details!,
           ),
-        )
-        .result;
+        );
+        return;
+      }
+      logger.error(
```
Jordan-Nelson, line 103: We are catching and swallowing all exceptions other than `EnableSoftwareTokenMfaException`, correct? Is that what we want to do? I was expecting the exception to be rethrown.

Equartey: Good point. I made the exception set the state to failure which should bubble up the concern.

Jordan-Nelson (dismissed):

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/totp_setup_state_machine.dart
NikaHsn, line 1: do we need to do similar thing in `sign_in_state_machine.dart` in `Future<void> verifySoftwareToken` method?

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/totp_setup_state_machine.dart
```diff
@@ -59,29 +62,59 @@ final class TotpSetupStateMachine
...
+        'Failed to verify TOTP code. Please try again. If the problem persists, contact support.',
```
NikaHsn, line 111: not sure if we should include `If the problem persists, contact support.` in the error message

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/totp_setup_state_machine.dart
```diff
@@ -59,29 +62,59 @@ final class TotpSetupStateMachine
...
+              session: _session,
+              userCode: event.code,
+              friendlyDeviceName: event.friendlyDeviceName,
+            ),
+          )
+          .result;
+    } on Exception catch (e, st) {
+      // Handle mismatch code exception that may occur during TOTP verification.
+      // See: https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_VerifySoftwareToken.html#API_VerifySoftwareToken_Errors
+      // TODO(equartey): Investigate when CodeMismatchException is thrown and handle accordingly.
```
NikaHsn, line 93: reading the exceptions' discription seems the `CodeMismatchException` is thrown when the totp is already setup and later when signing up the user using a wrong code. I'm not very familiar with our auth state machine so not sure if we have different flow/state machine for setting up the totp and then using totp code in the sign-in flow

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/totp_setup_state_machine.dart
Equartey: I don't believe this is a concern. The sign in flow currently allows multiple attempts. That being said, looking at that function I don't know offhand what catches the `EnableSoftwareTokenMfaException` that enables the retries.

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/totp_setup_state_machine.dart
Equartey: I was curious about this too, though I could not get that exception to throw. Our state machine blocks calls to `Amplify.Auth.verifyTotpSetup()` when the client is not in the setup process. After an authenticator app is setup, the totp code gets submitted during sign. As it's setup now, I don't think this is a concern.

Equartey: note: I omitted the TODO comment after reviewing it more.

NikaHsn (approved):

khatruong2009 (approved):

PR #4549, opened 2024-03-14
feat!: Update storage APIs to accept `StoragePath`

khatruong2009:
*Issue #, if available:*

*Description of changes:*
- Update all Storage APIs to accept a new type - StoragePath
- Remove key/accessLevel from storage APIs
- Add new internal logic for resolving storage paths

Note: this PR does not make any changes to prefix resolver. That will occur in a separate PR.


packages/storage/amplify_storage_s3/example/integration_test/use_case_test.dart
```diff
@@ -90,6 +89,12 @@ void main() {
         await Amplify.addPlugins([authPlugin, storagePlugin]);
         await Amplify.configure(amplifyEnvironments[entry.key]!);
 
+        try {
```
Jordan-Nelson, line 92:
```suggestion
        // confirm no user is signed in before starting tests
        try {
```

packages/storage/amplify_storage_s3_dart/lib/src/platform_impl/download_file/download_file_html.dart
```diff
@@ -40,14 +40,13 @@ S3DownloadFileOperation downloadFile({
...
-  required String key,
+  required StoragePath path,
   required AWSFile localFile,
   required StorageDownloadFileOptions options,
   required S3PluginConfig s3pluginConfig,
   required StorageS3Service storageS3Service,
 }) async {
-  final s3PluginOptions = options.pluginOptions as S3DownloadFilePluginOptions;
-  final targetIdentityId = s3PluginOptions.targetIdentityId;
+  final s3PluginOptions = options.pluginOptions as S3DownloadFilePluginOptions?;
```
Jordan-Nelson, line 49: @khatruong2009 do you recall if there was a reason that this was changed to nullable?

packages/storage/amplify_storage_s3/example/integration_test/use_case_test.dart
```diff
@@ -56,17 +56,16 @@ void main() {
...
   final testObjectKey3 = 'user1-private-object-${uuid()}';
   final testObject3CopyKey = 'user1-private-object-copy-${uuid()}';
   final testObject3CopyMoveKey = 'user1-private-object-copy-move-${uuid()}';
   const testObjectFileName1 = 'user1Guest.txt';
   const testObjectFileName2 = 'user1Protected.jpg';
   const testObjectFileName3 = 'user1Private.large';
 
-  for (final entry in amplifyEnvironments.entries) {
+  for (final entry in amplifyEnvironments.entries
+      .where((element) => element.key != 'custom-prefix')) {
```
khatruong2009, line 68: are we going to eventually add back in the custom-prefix environment?

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
```diff
@@ -297,7 +257,7 @@ class StorageS3Service {
     final urlRequest = AWSHttpRequest.raw(
       method: AWSHttpMethod.get,
       host: host,
-      path: keyToGetUrl,
+      path: '/$resolvedPath',
```
khatruong2009, line 260: is this supposed to get a "/" at the beginning?

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/storage_s3_service_test.dart
```diff
@@ -402,7 +369,7 @@ void main() {
 
         expect(
           storageS3Service.list(
-            path: 'a path',
+            path: const StoragePath.fromString('apath'),
```
khatruong2009, line 372: I made "a path" one word because I thought it couldn't have a space in it. I'm not sure if this was on purpose or not or if I should change it back

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_upload_task_test.dart
```diff
@@ -91,7 +88,7 @@ void main() {
     group('Uploading S3DataPayload', () {
       final testDataPayload = S3DataPayload.string('Upload me please!');
       final testDataPayloadBytes = S3DataPayload.bytes([101, 102]);
-      const testKey = 'object-upload-to';
+      const testPath = '/object-upload-to';
```
khatruong2009, line 91: remove /

packages/storage/amplify_storage_s3/example/integration_test/use_case_test.dart
Jordan-Nelson: I don't see any benefit to adding it back. With the changes in this PR an environment that has paths `public/`, `protected/`, and `private/` is not different that an env that has path `foo/`, `bar/`, and `baz/`.

There are an infinite number of backends that could be created. We cannot test against all of them. The standard config ("main") already covers an env that has paths with and without identity ids.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
Jordan-Nelson: Yes. In this case we are building a url in the form of `htps://<bucket>/<path>`. It would not be a valid URL without a `/`. This does not become part of the key.

This `/` was already included by the way, I just moved the location. It was added on line 281 or 289 in the removed diff.

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/storage_s3_service_test.dart
Jordan-Nelson:
```suggestion
            path: const StoragePath.fromString('a path'),
```

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_upload_task_test.dart
Jordan-Nelson: updated

Jordan-Nelson (approved):

packages/storage/amplify_storage_s3_dart/test/path_resolver/path_resolver_test.dart
```diff
@@ -7,6 +7,13 @@ import 'package:test/test.dart';
 
 class MockTokenIdentityAmplifyAuthProvider
     implements TokenIdentityAmplifyAuthProvider {
+  MockTokenIdentityAmplifyAuthProvider({
+    this.getIdException,
+  });
+
+  Exception? getIdException;
+  int getIdCount = 0;
```
Equartey, line 15: Q: Is this being used?

Jordan-Nelson: Yes - in the test "should resolve multiple paths with identity Id". It is used to confirm getId is only called once when resolving multiple paths.

Equartey (approved):

packages/storage/amplify_storage_s3/example/integration_test/content_type_infer/content_type_infer_html.dart
```diff
@@ -38,10 +38,13 @@ void testContentTypeInferTest({
   group('content type inferring within upload', () {
     tearDownAll(() async {
       await Amplify.Storage.removeMany(
-        keys: testUploadKeys,
-        options: const StorageRemoveManyOptions(
-          accessLevel: StorageAccessLevel.private,
-        ),
+        paths: testUploadKeys
```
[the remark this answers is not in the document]
Jordan-Nelson: We are rewriting the entire integration test suite so we I don't think it makes sense to spend effort updating these tests. We updated them to get them passing, but follow up PRs will remove these.

packages/storage/amplify_storage_s3/example/integration_test/content_type_infer/content_type_infer_io.dart
```diff
@@ -34,10 +34,13 @@ void testContentTypeInferTest({
 
     tearDownAll(() async {
       await Amplify.Storage.removeMany(
-        keys: testUploadKeys,
-        options: const StorageRemoveManyOptions(
-          accessLevel: StorageAccessLevel.private,
-        ),
+        paths: testUploadKeys
```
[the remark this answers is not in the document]
Jordan-Nelson: See comment above

packages/storage/amplify_storage_s3/example/integration_test/issues/get_url_special_characters_test.dart
```diff
@@ -41,10 +41,12 @@ void main() {
     final key = 'Test - ${DateTime.now()}';
     await Amplify.Storage.uploadData(
       data: S3DataPayload.bytes('hello'.codeUnits),
-      key: key,
+      path: StoragePath.fromString('public/$key'),
```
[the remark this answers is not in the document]
Jordan-Nelson: See comment above

packages/amplify_core/lib/src/types/storage/storage_path_from_identity_id.dart
```diff
@@ -0,0 +1,25 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'package:amplify_core/amplify_core.dart';
+import 'package:meta/meta.dart';
+
+/// {@template amplify_core.storage.storage_path_from_identity_id}
+/// Creates a [StoragePath] from the current user's identityId.
+/// {@endtemplate}
+class StoragePathFromIdentityId implements StoragePath {
```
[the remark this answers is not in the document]
Jordan-Nelson: This is private. The type isn't exported from the public API so it would have to be imported from the full file path to be used and consumers would get a lint warning they would need to ignore (`ignore: implementation_imports`).

The only benefit that I can see of marking it as internal is in case we accidentally add it to the public exports. The downside of marking it as internal is that we have to ignore the lint warnings where it is used in the storage package. If we are concerned about accidentally exporting something that should not be public I think a better alternative is to look into tools that can check for API changes on each PR. 

The pattern that Amplify Flutter seems to follow currently is to only mark something as internal if it is being exported as part of the public API. If you search for `ignore: implementation_imports` you can find ~100 references to classes/constants that are not exported from the public API and are not marked as internal.

re: combining into a single class - We would need a way to tell what type of StoragePath this is so that we can only fetch identity Id for paths that require it. Currently we use to the class type to tell what type of StoragePath it is. So we would need to add a public (marked as @internal) attribute to indicate what type of storage path this is. I think a separate class is more scalable. It is possible/likely that we will have other types of StoragePaths in the future. Combining them all into a single class will be hard to maintain.

packages/amplify_core/lib/src/types/storage/storage_path.dart
```diff
@@ -0,0 +1,47 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'package:amplify_core/src/types/storage/storage_path_from_identity_id.dart';
+import 'package:meta/meta.dart';
+
+/// And Object that represents the full remote path of a storage item.
```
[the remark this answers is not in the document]
Jordan-Nelson: Updated

packages/amplify_core/lib/src/types/storage/storage_path_from_identity_id.dart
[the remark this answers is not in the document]
NikaHsn: Thank you, Jordan, for the clarification. I overlooked the fact that it wasn't exported from storage_type.dart.

packages/storage/amplify_storage_s3_dart/lib/src/path_resolver/s3_path_resolver.dart
```diff
@@ -0,0 +1,91 @@
...
+      }
+      throw UnknownException(
+        'An unknown error occurred while fetching the identity Id. '
+        'See underlyingException for more info.',
+        underlyingException: e,
+      );
+    }
+  }
+
+  /// Resolve the given [path].
```
NikaHsn, line 41:
```suggestion
  /// Resolves the given [path].
```

packages/storage/amplify_storage_s3_dart/lib/src/path_resolver/s3_path_resolver.dart
```diff
@@ -0,0 +1,91 @@
...
+    if (resolvedPath.startsWith('/')) {
+      throw const StoragePathValidationException(
+        'StoragePath cannot start with a leading "/"',
+        recoverySuggestion: 'Remove the leading "/" from the StoragePath',
+      );
+    }
+    return resolvedPath;
+  }
+
+  /// Resolve multiple [StoragePath] objects.
```
NikaHsn, line 73:
```suggestion
  /// Resolves multiple [StoragePath] objects.
```

NikaHsn (approved):

PR #4561, opened 2024-03-15
chore: update download apis to use path

khatruong2009:
*Issue #, if available:*

*Description of changes:*


packages/storage/amplify_storage_s3/example/integration_test/use_case_test.dart
```diff
@@ -320,13 +320,14 @@ void main() {
         });
 
         testWidgets(
-            skip: true,
+            // skip: true,
```
Jordan-Nelson, line 323: You can probably remove this if you are commenting it out

packages/storage/amplify_storage_s3_dart/example/bin/example.dart
```diff
@@ -207,7 +207,8 @@ Future<void> downloadDataOperation() async {
 
   final s3Plugin = Amplify.Storage.getPlugin(AmplifyStorageS3Dart.pluginKey);
   final downloadDataOperation = s3Plugin.downloadData(
-    key: key,
+    // key: key,
```
Jordan-Nelson, line 210: todo: remove

packages/storage/amplify_storage_s3_dart/test.txt
```diff
@@ -0,0 +1 @@
+你好世界！
```
Jordan-Nelson, line 1: ?

packages/storage/amplify_storage_s3_dart/test/test_utils/mocks.dart
```diff
@@ -1,15 +1,31 @@
...
 import 'package:amplify_storage_s3_dart/src/storage_s3_service/storage_s3_service.dart';
 import 'package:amplify_storage_s3_dart/src/storage_s3_service/transfer/transfer.dart';
 import 'package:aws_signature_v4/aws_signature_v4.dart';
 import 'package:mocktail/mocktail.dart';
 import 'package:smithy/smithy.dart';
 
-class MockStorageS3Service extends Mock implements StorageS3Service {}
+class MockStorageS3Service extends Mock implements StorageS3Service {
+  @override
+  S3DownloadTask downloadData({
```
Jordan-Nelson, line 17: Why does this need an override?

khatruong2009: I added this because I was getting an error when testing and thought adding that might help but it hasn't. Here's the error:

```
Bad state: No method stub was called from within `when()`. Was a real method called, or perhaps an extension method?
package:mocktail/src/mocktail.dart 262:7            When._completeWhen
package:mocktail/src/mocktail.dart 257:12           When.thenAnswer
test/platform_impl/download_file_io_test.dart 63:9  main.<fn>.<fn>
```

PR #4569, opened 2024-03-18
feat(storage): copy and move APIs

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*


packages/storage/amplify_storage_s3/example/integration_test/use_case_test.dart
```diff
@@ -696,52 +657,6 @@ void main() {
...
-              ),
-              options: const StorageCopyOptions(
-                pluginOptions: S3CopyPluginOptions(
-                  getProperties: true,
-                ),
-              ),
-            ).result;
-
-            expect(result.copiedItem.eTag, isNotEmpty);
-          });
```
khatruong2009, line 743: Why is this test being removed?

packages/storage/amplify_storage_s3_dart/lib/src/amplify_storage_s3_dart_impl.dart
```diff
@@ -363,12 +363,12 @@ class AmplifyStorageS3Dart extends StoragePluginInterface
...
-    required StorageItemWithAccessLevel<StorageItem> source,
-    required StorageItemWithAccessLevel<StorageItem> destination,
+    required StoragePath source,
+    required StoragePath destination,
     StorageCopyOptions? options,
   }) {
-    final s3Source = S3ItemWithAccessLevel.from(source);
-    final s3Destination = S3ItemWithAccessLevel.from(destination);
+    // final s3Source = S3ItemWithAccessLevel.from(source);
+    // final s3Destination = S3ItemWithAccessLevel.from(destination);
```
khatruong2009, line 371: Are we leaving this in the code?

Jordan-Nelson: No. I removed these.

packages/storage/amplify_storage_s3/example/integration_test/use_case_test.dart
Jordan-Nelson: I think I removed this because it depended on a test I was removing and it seemed redundant to other e2e tests we have. I added it back though and updated it to no longer be dependent on the other test. We can decide if a test like this is needed or not when we refactor the full test suite.

khatruong2009 (approved):

PR #4615, opened 2024-03-27
chore(storage): add e2e tests for `uploadData`, `uploadFile`, `getUrl`, `getProperties`

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Add e2e test for `uploadData`, `uploadFile`, `getUrl`, and `getProperties` APIs

### DIFF
--- a/packages/storage/amplify_storage_s3/example/integration_test/get_properties_test.dart
+++ b/packages/storage/amplify_storage_s3/example/integration_test/get_properties_test.dart
@@ -1,0 +1,66 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'package:amplify_core/amplify_core.dart';
+import 'package:amplify_storage_s3_example/amplifyconfiguration.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:integration_test/integration_test.dart';
+
+import 'utils/configure.dart';
+import 'utils/sign_in_new_user.dart';
+import 'utils/tear_down.dart';
+
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  group('getProperties()', () {
+    final path = 'public/get-properties-${uuid()}';
+    final data = 'get properties data'.codeUnits;
+    final metadata = {'foo': 'bar'};
+    setUpAll(() async {
+      await configure(amplifyEnvironments['main']!);
+      addTearDownPath(StoragePath.fromString(path));
+      await Amplify.Storage.uploadData(
+        data: HttpPayload.bytes(data),
+        path: StoragePath.fromString(path),
+        options: StorageUploadDataOptions(metadata: metadata),
+      ).result;
+    });
+
+    test('String StoragePath', () async {
+      final result = await Amplify.Storage.getProperties(
+        path: StoragePath.fromString(path),
+      ).result;
+      expect(result.storageItem.path, path);
+      expect(result.storageItem.metadata, metadata);
+      expect(result.storageItem.eTag, isNotNull);
+      expect(result.storageItem.size, data.length);
+    });
+
+    test('with identity ID', () async {
+      final userIdentityId = await signInNewUser();
+      final name = 'get-properties-with-identity-id-${uuid()}';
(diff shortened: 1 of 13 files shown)


khatruong2009 (approved):

packages/storage/amplify_storage_s3/example/integration_test/utils/tear_down.dart
```diff
@@ -0,0 +1,31 @@
...
+final _logger = AmplifyLogger().createChild('StorageTests');
+
+/// Adds a tear down to remove the object at [path].
+void addTearDownPath(StoragePath path) {
+  addTearDown(
+    () {
+      try {
+        return Amplify.Storage.remove(path: path).result;
+      } on Exception catch (e) {
+        _logger.warn('Failed to remove file after test', e);
```
Equartey, line 16: Q: wonder if we should fail the test when we can't tear down the resource. If there's a failure, we need to investigate the cause, rather than let it happen silently.

packages/storage/amplify_storage_s3/example/integration_test/utils/tear_down.dart
```diff
@@ -0,0 +1,31 @@
...
+  );
+}
+
+/// Adds a tear down to delete the current user.
+void addTearDownCurrentUser() {
+  addTearDown(() {
+    try {
+      return Amplify.Auth.deleteUser();
+    } on Exception catch (e) {
+      _logger.warn('Failed to delete user after test', e);
```
Equartey, line 28: Q: Same here, although probably less important for this test suite

packages/storage/amplify_storage_s3/example/integration_test/get_properties_test.dart
```diff
@@ -0,0 +1,73 @@
...
+    testWidgets('with identity ID', (_) async {
+      final userIdentityId = await signInNewUser();
+      final name = 'get-properties-with-identity-id-${uuid()}';
+      final data = 'with identity ID'.codeUnits;
+      final expectedResolvedPath = 'private/$userIdentityId/$name';
+      addTearDownPath(StoragePath.fromString(expectedResolvedPath));
+      await Amplify.Storage.uploadData(
+        data: HttpPayload.bytes(data),
+        path: StoragePath.fromIdentityId(
+          (identityId) => 'private/$identityId/$name',
```
Equartey, line 49: suggestion/nit: everywhere we call `(identityId) => '/some/path'` could be `(identityId) => expectedResolvedPath`.

packages/storage/amplify_storage_s3/example/integration_test/utils/sign_in_new_user.dart
```diff
@@ -0,0 +1,30 @@
...
+
+/// Creates a new user and signs them in.
+///
+/// Returns the users identityId.
+///
+/// A tear down will be added to delete the user after the test completes.
+Future<String> signInNewUser() async {
+  addTearDownCurrentUser();
+  final username = 'test-user-1-${uuid()}';
+  const password = 'TestUser1!';
```
Equartey, line 18: fix: we should be generating the username and password, we have a util for this: https://github.com/aws-amplify/amplify-flutter/blob/main/packages/test/amplify_test/lib/src/mock_data.dart#L60-L65

Jordan-Nelson: Updated

packages/storage/amplify_storage_s3/example/integration_test/get_properties_test.dart
Jordan-Nelson: I updated in some of the locations. I want to us the `fromIdentityId` constructor when invoking the API under test (getProperties) but using `expectedResolvedPath` for the setup makes sense.

packages/storage/amplify_storage_s3/example/integration_test/utils/tear_down.dart
Jordan-Nelson: I added `rethrow` to these. During development it can be nice to catch and log to focus on the test rather than the teardown, but I agree that these failing could indicate an issue.

Equartey (approved):
Nice work! Great improvement over gen1 E2E suite.

PR #4619, opened 2024-03-30
feat: update plus plugins version

NikaHsn:
*Issue #, if available:*

*Description of changes:*
- update dart sdk to ^3.3.0 (required for package_info_plus ^6.0.0, device_info_plus: ^10.0.1, connectivity_plus: ^6.0.1)
- update flutter sdk to ^3.19.0 (required for package_info_plus ^6.0.0, device_info_plus: ^10.0.1, connectivity_plus: ^6.0.1)
- run dart fix


packages/api/amplify_api/lib/src/connectivity_plus_platform.dart
```diff
@@ -16,12 +16,13 @@ class ConnectivityPlusPlatform extends ConnectivityPlatform {
...
-        switch (connectivityResult) {
-          case ConnectivityResult.none:
-            return ConnectivityStatus.disconnected;
-          default:
-            return ConnectivityStatus.connected;
-        }
+          .map((List<ConnectivityResult> connectivityResults) {
+        // Check if any of the results indicate a connection
+        final isConnected = connectivityResults
+            .any((result) => result != ConnectivityResult.none);
```
Jordan-Nelson, line 22: Question: What should the behavior be if the result contains ConnectivityResult.none AND ConnectivityResult.wifi? Is that possible?

___

nit: I think you can simplify this:
```dart
final isConnected = connectivityResults.contains(ConnectivityResult.none);
```
___

This is out of scope for this PR, but it seems like ConnectivityResult.bluetooth would be treated like an active connection. Same with ConnectivityResult.other. I am not sure if we want to treat those like an active connection.

NikaHsn: based on their doc it is not possible to have ConnectivityResult.none AND ConnectivityResult.wifi.

> In case of no connectivity, the list contains a single element of [ConnectivityResult.none]. Note also that this is the only case where [ConnectivityResult.none] is present.

NikaHsn: do you mean 
```
final isNotConnected = connectivityResults.contains(ConnectivityResult.none);
```
I think this may be less efficent as it checks all the elements in the list

Jordan-Nelson: Yes, that is what I meant.

`.contains()` will exit at the first element that matches.

[Here](https://api.dart.dev/stable/3.3.3/dart-core/Iterable/contains.html) is the implementation.

```dart
  bool contains(Object? element) {
    for (E e in this) {
      if (e == element) return true;
    }
    return false;
  }
```

NikaHsn: right, this is the case where the list contains ConnectivityResult.none (it happens when the device is not conntected and the list supposed to only have one element and it is ConnectivityResult.none). I meant the senarios when the device is connected and so the list may have multiple elements and none of them is ConnectivityResult.none

Equartey (approved):

Jordan-Nelson (approved):

packages/api/amplify_api/lib/src/connectivity_plus_platform.dart
Jordan-Nelson: I see. Makes sense.

.github/workflows/actions.yaml
```diff
@@ -114,10 +114,10 @@ jobs:
       - name: Git Checkout
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # 3.5.3
 
-      - uses: actions/setup-java@cd89f46ac9d01407894225f350157564c9c7cee2 # 3.12.0
+      - uses: actions/setup-java@v3
```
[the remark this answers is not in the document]
NikaHsn: changed to use the same commit sha as before

qwertylolman:
```
Because amplify_auth_cognito >=1.8.0 depends on amplify_analytics_pinpoint >=1.8.0 <1.9.0 which depends on package_info_plus ^6.0.0, amplify_auth_cognito >=1.8.0 requires package_info_plus ^6.0.0.
So, because global_pass depends on both amplify_auth_cognito 1.8.0 and package_info_plus 7.0.0, version solving failed.
```

is it an expected behavior?

PR #4630, opened 2024-04-01
chore(aft): Added generate amplify-swift

Equartey:
*Description of changes:*
Adds `aft generate amplify-swift` command
- `-b` to change target branch
- `-o` to change output
- `-d` to run a diff of current vs incoming
- Sets cloned files to read-execute. This prevents modifications, but allows `pod install`
- Added internal definitions for pods

### DIFF
--- a/packages/aft/lib/src/commands/generate/generate_amplify_swift_command.dart
+++ b/packages/aft/lib/src/commands/generate/generate_amplify_swift_command.dart
@@ -16,7 +16,11 @@ class PluginConfig {
     required this.name,
     required this.remotePathToSource,
   });
+
+  /// The name of the plugin.
   final String name;
+
+  /// The path to the plugin source files in the Amplify Swift repo.
   final String remotePathToSource;
 }
 
@@ -160,7 +164,7 @@ class GenerateAmplifySwiftCommand extends AmplifyCommand with GlobOptions {
   Future<void> _generatePlugin(PluginConfig plugin) async {
     logger.info('Selecting source files for ${plugin.name}...');
 
-    final repoDir = await _pluginDirForPath(plugin.remotePathToSource);
+    final remotePluginDir = await _pluginDirForPath(plugin.remotePathToSource);
 
     final outputDir = Directory(p.join(outputPath, plugin.name));
 
@@ -177,8 +181,8 @@ class GenerateAmplifySwiftCommand extends AmplifyCommand with GlobOptions {
     // Copy the files from the repo to the plugin directory.
     logger
       ..info('Copying plugin files for ${plugin.name}...')
-      ..verbose('From $repoDir to $outputDir');
-    await copyPath(repoDir.path, outputDir.path);
+      ..verbose('From $remotePluginDir to $outputDir');
+    await copyPath(remotePluginDir.path, outputDir.path);
   }
 
   Future<void> checkDiff(PluginConfig plugin) async {
--- a/packages/aft/lib/src/commands/generate/generate_amplify_swift_command.dart
+++ b/packages/aft/lib/src/commands/generate/generate_amplify_swift_command.dart
@@ -1,0 +1,288 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:async';
+import 'dart:io';
+
+import 'package:aft/aft.dart';
+import 'package:aft/src/options/glob_options.dart';
+import 'package:async/async.dart';
+import 'package:git/git.dart';
+import 'package:path/path.dart' as p;
+
+enum PermissionScope {
+  readExecuteWrite('777'),
+  readExecute('555');
+
+  const PermissionScope(this.value);
+  final String value;
+}
+
+class PluginConfig {
+  const PluginConfig({
+    required this.name,
+    required this.remotePathToSource,
+  });
+  final String name;
+  final String remotePathToSource;
+}
+
+/// Command for generating the Amplify Swift plugins for the DataStore plugin.
+class GenerateAmplifySwiftCommand extends AmplifyCommand with GlobOptions {
+  GenerateAmplifySwiftCommand() {
+    argParser
+      ..addOption(
+        'branch',
+        abbr: 'b',
+        help: 'The branch of Amplify Swift to target',
+        defaultsTo: 'release',
+      )
+      ..addOption(
+        'output',
+        abbr: 'o',
+        help: 'The lib/-relative output directory for the plugins',
+        defaultsTo: 'packages/amplify_datastore/ios/internal',
+      )
+      ..addFlag(
+        'diff',
+        abbr: 'd',
+        help: 'Show the diff of the generated files',
+        negatable: false,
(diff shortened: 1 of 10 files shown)


packages/aft/lib/src/commands/generate/generate_amplify_swift_command.dart
```diff
@@ -0,0 +1,288 @@
...
+  /// Sets up the Amplify Swift repo for use later
+  Future<void> _setupRepo() async {
+    if (_repoCache[branchTarget] != null) {
+      return;
+    }
+    Directory repoDir;
+    repoDir = await _downloadRepository();
+    repoDir = await _checkoutRepositoryRef(repoDir, branchTarget);
+
+    _repoCache[branchTarget] = repoDir;
```
Jordan-Nelson, line 151: Suggestion: You can make `repoDir` final if you refactor this a bit

```suggestion
    final repoDir = await _checkoutRepositoryRef(await _downloadRepository(), branchTarget);

    _repoCache[branchTarget] = repoDir;
```

Jordan-Nelson: Alternatively, something like this would work as well.

```suggestion
    final repoDir = await _downloadRepository();
    final repoRef = await _checkoutRepositoryRef(repoDir, branchTarget);

    _repoCache[branchTarget] = repoRef;
```

packages/aft/lib/src/commands/generate/generate_amplify_swift_command.dart
```diff
@@ -0,0 +1,288 @@
...
+
+    final outputDir = Directory(p.join(outputPath, plugin.name));
+
+    // Clear out the directory if it already exists.
+    // This is to ensure that we don't have any stale files.
+    if (await outputDir.exists()) {
+      logger.verbose(
+        'Clearing out existing plugin directory for ${plugin.name}...',
+      );
+      await _setPermissions(outputDir.path, PermissionScope.readExecuteWrite);
```
Jordan-Nelson, line 180: Q: Do you need to set permissions to execute if you are just removing the files?

Jordan-Nelson: Q: What permissions exist without this?

packages/aft/lib/src/commands/generate/generate_amplify_swift_command.dart
```diff
@@ -0,0 +1,288 @@
...
+      );
+    }
+    logger.info(
+      'No differences between incoming and current for ${plugin.name}.',
+    );
+  }
+
+  /// Sets the permissions of all files in [path]  to
+  Future<void> _setPermissions(String path, PermissionScope scope) async {
+    const scope = '777';
```
Jordan-Nelson, line 223: Q: Does this need to be updated?

packages/aft/lib/src/commands/generate/generate_amplify_swift_command.dart
```diff
@@ -0,0 +1,288 @@
...
+      ..info('Copying plugin files for ${plugin.name}...')
+      ..verbose('From $repoDir to $outputDir');
+    await copyPath(repoDir.path, outputDir.path);
+    await _setPermissions(outputDir.path, PermissionScope.readExecute);
+  }
+
+  Future<void> checkDiff(PluginConfig plugin) async {
+    logger.info('Checking diff for ${plugin.name}...');
+    final incoming = (await _pluginDirForPath(plugin.remotePathToSource)).path;
+    // Set the permissions of the incoming files to readExecute (fails otherwise)
```
Jordan-Nelson, line 196: Q: What fails when the permissions are not updated?

packages/aft/lib/src/util.dart
```diff
@@ -115,3 +116,34 @@ void sortPackagesTopologically<T>(
         .compareTo(ordered.indexOf(getPubspec(a).name));
   });
 }
+
+// Copyright 2017, the Dart project authors.
+// Copied from io.dart https://pub.dev/documentation/io/latest/io/copyPath.html
```
Jordan-Nelson, line 121: Q: Is this comment/link correct?

packages/aft/lib/src/util.dart
```diff
@@ -115,3 +116,34 @@ void sortPackagesTopologically<T>(
...
+  }
+  if (p.isWithin(from, to)) {
+    throw ArgumentError('Cannot copy from $from to $to');
+  }
+  return false;
+}
+
+// Copyright 2017, the Dart project authors.
+// Copied from io.dart https://pub.dev/documentation/io/latest/io/copyPath.html
+Future<void> copyPath(String from, String to) async {
```
Jordan-Nelson, line 134: Q: Why not use the copyPath method from Dart io?

packages/aft/lib/src/util.dart
```diff
@@ -115,3 +116,34 @@ void sortPackagesTopologically<T>(
...
+  if (p.canonicalize(from) == p.canonicalize(to)) {
+    return true;
+  }
+  if (p.isWithin(from, to)) {
+    throw ArgumentError('Cannot copy from $from to $to');
+  }
+  return false;
+}
+
+// Copyright 2017, the Dart project authors.
```
Jordan-Nelson, line 132: Q: Is this a standard way to provide attribution?

packages/aft/lib/src/commands/generate/generate_amplify_swift_command.dart
Equartey: Good suggestion, took the later.

packages/aft/lib/src/commands/generate/generate_amplify_swift_command.dart
Equartey: diff command catches the literal difference of permissions. Ie, 555 local vs 755 remote.

packages/aft/lib/src/commands/generate/generate_amplify_swift_command.dart
Equartey: Good catch, will remove.

packages/aft/lib/src/util.dart
Equartey: Discussed offline: Added io package as a dep.

packages/aft/lib/src/commands/generate/generate_amplify_swift_command.dart
Equartey: After looking into this more, turns out git doesn't preserve file permissions. Will remove this logic.

khatruong2009 (approved):

khatruong2009 (approved):

packages/amplify_datastore/example/ios/Podfile
```diff
@@ -31,6 +31,11 @@ target 'Runner' do
   use_frameworks!
   use_modular_headers!
 
+  # Point pods to internal source files
```
Jordan-Nelson, line 34: What is the purpose of these changes? Will customers have to make changes in their app?

Jordan-Nelson (approved):

PR #4685, opened 2024-04-09
chore(datastore): Stubbed Swift migration

Equartey:
*Description of changes:*
This PR does three things: 

1. Runs `aft generate amplify-swift` to get the files from our WIP amplify-swift branch `flutter-datastore-v2`
2. Updates Swift plugin imports & types
3. Stubs the API's that need to be migrated/implemented 

Xcode project builds, but tests are expected to fail. 


packages/amplify_datastore/ios/Classes/CognitoPlugin.swift
```diff
@@ -22,259 +22,140 @@ public class CognitoPlugin: AuthCategoryPlugin {
...
+//                let result = NativeAWSAuthCognitoSession(from: session)
+//                operation.dispatch(result: .success(result))
+//            }
+//        }
+//        return operation
+//    }
+//
+//    public func getCurrentUser() -> AuthUser? {
+//        return currentUser
+//    }
```
Jordan-Nelson, line 58: fetchAuthSession and getCurrentUser are commented out here, but implemented below with TODO statements. Do you need to keep this commented out code?

packages/amplify_datastore/ios/Classes/CognitoPlugin.swift
```diff
@@ -22,259 +22,140 @@ public class CognitoPlugin: AuthCategoryPlugin {
...
     public let key: PluginKey = "awsCognitoAuthPlugin"
     
     public func configure(using configuration: Any?) throws {}
     
-    public func reset(onComplete: @escaping BasicClosure) {
-        onComplete()
+// TODO: Migrate to Async Swift v2
+//    public func reset(onComplete: @escaping BasicClosure) {
+//        onComplete()
+//    }
```
Jordan-Nelson, line 32: *nit:* Could this function throw an exception until it is migrated rather than committing commented out code?

packages/amplify_datastore/ios/internal/AWSDataStorePlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
```diff
@@ -0,0 +1,601 @@
+//
```
Jordan-Nelson, line 1: *suggestion:* Can we update `.gitattributes` to treat everything under /internal as generated code, similar to what we do for generated pigeon code? This will make it so it is collapsed by default in code review. It won't be something we typically want to review in detail in code reviews.

It would also be nice if we could add something like this to the top of each file, but this is just a nice to have `// generated code. do not edit. - run 'aft ...' to pull in latest from amplify-swift`

packages/amplify_datastore/ios/Classes/CognitoPlugin.swift
Equartey: Long term no. I left it here as a reference for the migration task. It should be removed then.

packages/amplify_datastore/ios/Classes/CognitoPlugin.swift
Equartey: Looks like this is only used in testing, but I added a stub for it.

packages/amplify_datastore/ios/internal/AWSDataStorePlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
Equartey: Good suggestion, I added the directory to `.gitattributes`. 

I also agree it would be nice to have that header call out. After a quick glance, it appears that adding data at the beginning of a file is more involved. Which seems to add more complexity than its worth for this use case.

Jordan-Nelson (approved):

khatruong2009 (approved):

PR #4717, opened 2024-04-16
fix: properly handle 16-bit code units

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Update AWS File to *not* convert code units to a Blob on web as this is 1/ unnecessary and 2/ problematic for code units that are not 8-bit (for example, a UTF-16 encoded string)
- Add tests for a UTF-16 encoded string


khatruong2009 (approved):

Equartey (commented):
Q: Can we add this to a integ test too?

Equartey (approved):

PR #4764, opened 2024-04-23
chore!: remove deprecated auth types

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Removes deprecated auth types


packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_in_state_machine.dart
```diff
@@ -56,15 +56,6 @@ final class SignInStateMachine
...
 
-  /// The default flow used to sign in.
-  late final AuthFlowType defaultAuthFlowType = () {
-    // Get the flow from the plugin config
-    final pluginFlowType =
-        expect<CognitoPluginConfig>().auth?.default$?.authenticationFlowType ??
-            AuthenticationFlowType.userSrpAuth;
-    return pluginFlowType.sdkValue;
-  }();
-
```
NikaHsn, line 67: I acknowledge that this field is not present in CLI Gen2. However, with these modifications, we are bypassing the configuration value for users who are still using CLI Gen1.

Jordan-Nelson: That is correct. Customers can specify this via SignInOptions though.

The current behavior is to use the value from SignInOptions if it exists. If not, use the value from the config if it exists. If neither exist, `AuthenticationFlowType.userSrpAuth` is the default.

The new behavior is to use the value from SignInOptions if it exists. If not, `AuthenticationFlowType.userSrpAuth` is the default.

It is a breaking change that will be called out in the change log for v2.

Equartey (approved):

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_in_state_machine.dart
NikaHsn: I think we may need to add a note in our Gen1/V2 documentation indicating the value from the configuration file is not used by the library. Instead, users should use the SignInOptions.

NikaHsn (dismissed):

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_in_state_machine.dart
Jordan-Nelson: Agreed. I meant to say this will be called out in the *migration guide*

Equartey (approved):

NikaHsn (approved):

PR #4772, opened 2024-04-24
chore(api)!: Removed deprecated members …

Equartey:
*Description of changes:*
Removed GraphQLRequest.serializeAsMap(), QueryFieldOperator.evaluateSerialized(), and RestException.

### DIFF
--- a/packages/amplify_datastore/test/query_predicate_test.dart
+++ b/packages/amplify_datastore/test/query_predicate_test.dart
@@ -8,33 +8,6 @@ import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_inte
 import 'package:amplify_test/test_models/ModelProvider.dart';
 import 'package:flutter_test/flutter_test.dart';
 
-// TODO(Jordan-Nelson): Remove at next major version bump.
-/// A `Post` model that does not have a `toMap()` impl.
-class PostWithNoToMap extends Post {
-  PostWithNoToMap({
-    String? id,
-    required String title,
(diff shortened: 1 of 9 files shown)


Jordan-Nelson (dismissed):

NikaHsn (approved):

packages/amplify_datastore/test/query_predicate_test.dart
```diff
@@ -474,53 +342,6 @@ void main() {
...
-      final postWithNoToMap3 = PostWithNoToMap(
-        title: 'post 3',
-        rating: 1,
-        blog: blog2,
-      );
-
-      // TODO(Jordan-Nelson): Remove at next major version bump.
-      // can be removed when `getId()` is removed.
-      test('equals (id)', () {
-        final testPredicate = Post.BLOG.eq(blog1.id);
```
Jordan-Nelson, line 496: I think rather than remove this test, and the one immediately after, we can update them to use `modelIdentifier` instead of `id`. I don't see other tests that cover that case.

We should call this out in the v2 migration guide.

```dart
final testPredicate = Post.BLOG.eq(blog1.modelIdentifier);
```

Jordan-Nelson (approved):

packages/amplify_datastore/test/query_predicate_test.dart
Jordan-Nelson: Discussed offline. We have other tests that cover this.

Issue #4693, opened 2024-04-12
Type comparison with UnderlyingExceptions might not be working as expected.


jmartins-sh:
### Description

Hello there! I'm currently testing scenarios where Cognito triggers (Lambdas) might return validation errors. While catching the `LambdaException`, I noticed that the underlying exception has a prefix of "_$".  

For example, I encountered `_$UserLambdaValidationException`, whereas the exception provided by amplify-flutter is simply `UserLambdaValidationException` (without the "_$").  

Is this expected behavior? Did I misunderstand something?

Appreciate the help. :)

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [X] Auth
- [ ] Authenticator
- [ ] DataStore
- [ ] Notifications (Push)
- [ ] Storage

### Steps to Reproduce

1. Create a lambda that throws an Error("Not authorized to use the system.");
2. Catch the exception in flutter code
3. Compare the type of underlying exception (supposed to be the _$UserLambdaValidationException) inside the LambdaException with UserLambdaValidationException
4. The comparison returns false.

### Screenshots

![image (1)](https://github.com/aws-amplify/amplify-flutter/assets/16046723/de2d835a-95a7-4c7c-a675-cba73503d50b)

![image](https://github.com/aws-amplify/amplify-flutter/assets/16046723/5f4d5da3-f44f-42a9-bd84-1f8c3ced60c3)

### Platforms

- [ ] iOS
- [X] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Flutter Version

3.19.1

### Amplify Flutter Version

1.7.0

### Deployment Method

Amplify CLI

Jordan-Nelson:
Hello @joaoantoniomartinsfilho 

_$UserLambdaValidationException extends UserLambdaValidationException. [Here](https://github.com/aws-amplify/amplify-flutter/blob/main/packages/auth/amplify_auth_cognito_dart/lib/src/sdk/src/cognito_identity_provider/model/user_lambda_validation_exception.g.dart) is the class definition. 

If you want to check if the exception is a UserLambdaValidationException you can do the following.

```dart
if (e is UserLambdaValidationException) {
  // handle exception 
}
```

Let me know if you have any other questions.

joaoantoniomartinsfilho:
Hey @Jordan-Nelson, how's it going?

Thanks for the reply.

[...]

PR #4804, opened 2024-04-30
fix: map Lambda exceptions correctly

Jordan-Nelson:
*Issue #, if available:* https://github.com/aws-amplify/amplify-flutter/issues/4693

*Description of changes:*
- Update LambdaException to use a generative constructor instead of a factory so that it can be extended
- Update `InvalidLambdaResponseException` `UnexpectedLambdaException`, and `UserLambdaValidationException` to extend `LambdaException`
- Map Lambda exceptions from the Cognito SDK to the equivalent Amplify exception
- Add tests for `transformSdkException`

All Lambda exceptions from the SDK are currently mapped to a generic LambdaException. The underlying exception is an exception from the SDK which is private. As a result, customers have no way to determine what type of Lambda exception occurred (without doing something hacky like string matching on the underlying exception).

This PR updates the three different types of lambda exceptions to extend `LambdaException`. This is useful because `LambdaException` has utilities for getting the lambda name/error. Additionally it makes this change non breaking as customers that were catching `LambdaException` can continue to do so if they would like. This PR also updates the mapping logic to map the SDK exceptions to specific Amplify exceptions rather than a generic `LambdaException`.

### DIFF
--- a/packages/auth/amplify_auth_cognito_dart/lib/src/sdk/sdk_exception.dart
+++ b/packages/auth/amplify_auth_cognito_dart/lib/src/sdk/sdk_exception.dart
@@ -27,31 +27,27 @@ sealed class CognitoServiceException extends core.AuthServiceException {
 /// {@endtemplate}
 final class LambdaException extends CognitoServiceException {
   /// {@macro amplify_auth_cognito_dart.sdk.lambda_exception}
-  factory LambdaException(
-    String message, {
-    String? recoverySuggestion,
-    Object? underlyingException,
-  }) {
-    final match = _errorRegex.firstMatch(message);
-    final lambdaName = match?.group(1);
-    final parsedMessage = match?.group(2);
-    if (parsedMessage != null) {
-      message = parsedMessage;
-    }
-    return LambdaException._(
-      message,
-      lambdaName: lambdaName,
-      recoverySuggestion: recoverySuggestion,
-      underlyingException: underlyingException,
-    );
-  }
-
-  const LambdaException._(
+  const LambdaException(
     super.message, {
-    this.lambdaName,
     super.recoverySuggestion,
     super.underlyingException,
-  });
+  }) : _message = message;
+
+  final String _message;
+
+  @override
+  String get message {
+    final match = _errorRegex.firstMatch(_message);
+    final parsedMessage = match?.group(2);
+    return parsedMessage ?? _message;
+  }
+
+  /// The name of the lambda which triggered this exception.
+  String? get lambdaName {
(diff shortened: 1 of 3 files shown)


khatruong2009 (approved):

packages/auth/amplify_auth_cognito_dart/lib/src/sdk/sdk_exception.dart
```diff
@@ -27,31 +27,27 @@ sealed class CognitoServiceException extends core.AuthServiceException {
...
     super.underlyingException,
-  });
+  }) : _message = message;
+
+  final String _message;
+
+  @override
+  String get message {
+    final match = _errorRegex.firstMatch(_message);
+    final parsedMessage = match?.group(2);
```
Equartey, line 41: Q: What happens if a match is not found, should we have a fallback value?

packages/auth/amplify_auth_cognito_dart/lib/src/sdk/sdk_exception.dart
```diff
@@ -27,31 +27,27 @@ sealed class CognitoServiceException extends core.AuthServiceException {
...
+    return parsedMessage ?? _message;
+  }
+
+  /// The name of the lambda which triggered this exception.
+  String? get lambdaName {
+    final match = _errorRegex.firstMatch(_message);
+    final lambdaName = match?.group(1);
```
Equartey, line 48: Same here

packages/auth/amplify_auth_cognito_dart/lib/src/sdk/sdk_exception.dart
Jordan-Nelson: The fallback is `_message` since the return is `parsedMessage ?? _message`. This is consistent with the existing behavior.

packages/auth/amplify_auth_cognito_dart/lib/src/sdk/sdk_exception.dart
Jordan-Nelson: No. There is no logical fallback. This is consistent with the existing behavior. `lambdaName` is nullable and has no fallback.

packages/auth/amplify_auth_cognito_dart/lib/src/sdk/sdk_exception.dart
Equartey: Somehow missed that, disregard!

Equartey (approved):

PR #4844, opened 2024-05-07
fix: wait for in progress requests to cancel before completing `.pause()` and `.cancel()`

Jordan-Nelson:
*Description of changes:*
- wait for in progress requests to cancel before completing `.pause()` and `.cancel()`. 
- add tests for multiple files sizes for pause/resume/cancel
- skip `.cancel()` tests for download. These are failing on flutter beta and it is not immediately obvious why. Further investigation is required.

The main change in this PR is to wait to complete the `.pause()` and `.cancel()` futures until in flight requests have been cancelled. There are a few motivations for this:
1. This resolved e2e test failures on Flutter beta. The cancelled networks requests which were completing after `.pause()` and `.cancel()` return are resulting in errors being thrown after tests complete. As of Flutter 3.22 these will result in test failures.
2. When a multi part upload is paused or canceled, any in progress requests are cancelled. When the request is cancelled, there is also a request to S3 to abort the multi part upload, which results in S3 discarding any previously uploaded chunks and preventing new chunks from being uploaded. S3 states that any requests that are currently in progress when the abort request is made **_may complete successfully_**. Since we currently we do not wait for the in progress requests to cancel before aborting the upload, those requests could still be uploaded to S3 resulting in additional cost.
3. Given  `.pause()` and `.cancel()` return futures, it is reasonable to expect that the operation is **_fully_** paused/cancelled when the future completes, but this isn't currently the case as there are network requests that are cancelled after the future returns.

### DIFF
--- a/packages/storage/amplify_storage_s3/example/integration_test/upload_file_test.dart
+++ b/packages/storage/amplify_storage_s3/example/integration_test/upload_file_test.dart
@@ -278,58 +278,110 @@ void main() {
       });
 
       group('pause, resume, cancel', () {
-        const size = 1024 * 1024 * 6;
-        const chars = 'qwertyuiopasdfghjklzxcvbnm';
-        final content = List.generate(size, (i) => chars[i % 25]).join();
-        testWidgets('can pause', (_) async {
-          final fileId = uuid();
-          final path = 'public/upload-file-pause-$fileId';
-          final filePath = await createFile(path: fileId, content: content);
-          addTearDownPath(StoragePath.fromString(path));
-          final operation = Amplify.Storage.uploadFile(
-            localFile: AWSFile.fromPath(filePath),
-            path: StoragePath.fromString(path),
+        /// file sizes in mb
+        const fileSizes = [
+          1, // non multi part upload
+          6, // small multi part upload (2 parts)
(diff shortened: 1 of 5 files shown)


packages/storage/amplify_storage_s3/example/integration_test/upload_file_test.dart
```diff
@@ -278,58 +278,110 @@ void main() {
...
-          final expectException = expectLater(
-            () => operation.result,
-            throwsA(isA<StorageOperationCanceledException>()),
-          );
-          await operation.cancel();
-          await expectException;
-        });
+          testWidgets('can cancel (file size: $fileSize mb)', (_) async {
+            final fileId = uuid();
+            final path = 'public/upload-file-cancel-$fileId';
```
khatruong2009, line 359: should this also be checking if the file size is 5mb like pause test since small file uploads can't be canceled??

Jordan-Nelson: Cancel is supported for both single part and multi part uploads. Only resume/pause are not supported for single part uploads

Equartey (approved):

PR #4858, opened 2024-05-08
chore(datastore): add query & mutate impl

Equartey:
*Description of changes:*
Adds Swift FlutterApiPlugin query and mutate implementations.  


packages/amplify_core/lib/src/types/datastore/hub/hub_event_element_with_metadata.dart
```diff
@@ -35,7 +35,7 @@ class HubEventElementWithMetadata<M extends Model> extends HubEventElement<M> {
   final int version;
 
   /// The last time the model was updated locally, in seconds since epoch.
-  final int lastChangedAt;
+  final int? lastChangedAt;
```
Jordan-Nelson, line 38: This is part of the public API so it is technically a breaking change. Was this intended to be part of the public API? Is this a class that customers use (is there a use case in which they might use it)?

Ideally we should try to find a way to make this non-breaking. If that is not feasible and this was never intended to be public was can discuss deprecating this or marking it as internal.

packages/amplify_datastore/example/ios/unit_tests/DataStoreHubEventStreamHandlerTests.swift
```diff
@@ -63,7 +63,7 @@ class DataStoreHubEventStreamHandlerTests: XCTestCase {
         waitForExpectations(timeout: 1.0)
         // cancellation needed to make sure that Hub token is invalidated to
         // prevent collisions between tests
-        hubHandler.onCancel(withArguments: nil)
+        let _ = hubHandler.onCancel(withArguments: nil)
```
Jordan-Nelson, line 66: Q: what is the purpose of this change? Is this the norm in swift or enforced via linting?

packages/amplify_datastore/example/ios/Podfile
```diff
@@ -32,9 +32,9 @@ target 'Runner' do
   use_modular_headers!
 
   # Point pods to internal source files
-  pod 'Amplify', path: '../../ios/internal/'
-  pod 'AWSPluginsCore', path: '../../ios/internal/'
-  pod 'AWSDataStorePlugin', path: '../../ios/internal/'
+  pod 'Amplify', path: '../../ios/internal/', :inhibit_warnings => true
```
Jordan-Nelson, line 35: Q: What is the purpose of this change?

packages/amplify_datastore/example/ios/unit_tests/DataStoreHubEventStreamHandlerTests.swift
```diff
@@ -63,7 +63,7 @@ class DataStoreHubEventStreamHandlerTests: XCTestCase {
         waitForExpectations(timeout: 1.0)
```
Jordan-Nelson, line 63: Not related to this PR, but we should remove these from generated files in `.gitattributes`. Everything under `**/example/ios/**` is treated as generated code since the majority of it is generated by flutter during `flutter create`. We should remove `unit_tests` from this.

Could you make that change in this feature branch (can be in a follow up PR) since there will probably be quite a few PRs that will updates native tests?

packages/amplify_datastore/lib/src/utils/native_api_helpers.dart
```diff
@@ -13,6 +14,17 @@ GraphQLRequest<String> nativeRequestToGraphQLRequest(
   );
 }
 
+/// Convert a [GraphQLResponse] to a [NativeGraphQLResponse]
+NativeGraphQLResponse GraphQLResponseToNativeResponse(
```
Jordan-Nelson, line 18:
```suggestion
NativeGraphQLResponse graphQLResponseToNativeResponse(
```

packages/amplify_datastore/example/ios/unit_tests/DataStoreHubEventStreamHandlerTests.swift
Equartey: Swift linter complains about the value was not bing used. This is a common pattern to satisfy the linter rule, while denoting the return value is not used. Relevant [SO](https://stackoverflow.com/a/32656172)

packages/amplify_datastore/example/ios/Podfile
Equartey: CI logs were showing tons of lint warnings for these files, but since we they are from Amplify Swift we don't care. We're not going to modify these files. Adding this line prevents warning messages in CI, making those logs more readable.

packages/amplify_datastore/example/ios/unit_tests/DataStoreHubEventStreamHandlerTests.swift
Equartey: Good call out, I added that to this branch.

Jordan-Nelson: Thank you

Jordan-Nelson (approved):
Change LGTM. We may have to make changes to for the breaking change but we can discuss offline and handle that in a follow up.

khatruong2009 (approved):

PR #4913, opened 2024-05-21
Chore(datastore): migrate cognito plugin

khatruong2009:
*Description of changes:*

- Migrate the Swift Cognito Plugin to use async/await pattern. 
- Other minor Auth related V2 migrations


packages/amplify_datastore/ios/Classes/CognitoPlugin.swift
```diff
@@ -24,45 +24,24 @@ public class CognitoPlugin: AuthCategoryPlugin {
     
     public func configure(using configuration: Any?) throws {}
     
-    public func reset() async {
-        // TODO: reset plugin -- used for testing
```
Jordan-Nelson, line 28: Q: should this TODO be left?

Jordan-Nelson (approved):

packages/amplify_datastore/ios/Classes/CognitoPlugin.swift
Equartey: Discussed offline, the [previous implementation](https://github.com/aws-amplify/amplify-flutter/blob/main/packages/amplify_datastore/ios/Classes/CognitoPlugin.swift#L48-L50) of this function was essentially a no-op and we're choosing to omit now.

tyllark (approved):

PR #4935, opened 2024-05-28
chore: map AmplifyConfig to AmplifyOutputs

Jordan-Nelson:
*Description of changes:*
- Add mapping from AmplifyConfig to AmplifyOutputs
- Add Gen 2 app to generate configs for testing, add tests


Equartey (approved):

packages/amplify_core/lib/src/config/api/api_config.dart
```diff
@@ -30,4 +31,25 @@ class ApiConfig extends AmplifyPluginConfigMap {
 
   @override
   Map<String, Object?> toJson() => _$ApiConfigToJson(this);
+
+  DataOutputs? toDataOutputs({AWSConfigMap<CognitoAppSyncConfig>? appSync}) {
+    final plugin = awsPlugin?.default$;
```
NikaHsn, line 36: I'm not sure if using `default$` is correct as it uses `authorizationType == APIAuthorizationType.apiKey` to get the default plugin. I think we want to get the plugin with `endpointType == EndpointType.graphQL` instead.

NikaHsn (approved):

PR #5026, opened 2024-06-14
chore(infra-gen2): Gen 2 infra

Equartey:
*Description of changes:*
This adds a new `infra-gen2` directory to hold Gen 2 backends for integration testing. It includes a script `tool/deploy_gen2.dart` that will: 

1. Run the amplify sandbox command in each folder under `/backends`
2. Combine multiple backend outputs for a given category into `amplifyEnvironments`
3. Copy and distribute the `amplify_outputs.dart` to example apps
4. Create and upload the config files to respective S3 buckets for CI use
Screenshot 2024-06-14 at 11 40 40 AM

Additionally, this includes a bash file, `pull_test_gen2_backend.sh`, that lives in example apps to allow GitHub Actions to pull the required config for a given App. 


infra-gen2/tool/deploy_gen2.dart
```diff
@@ -0,0 +1,391 @@
...
+  final List<AmplifyBackend> backends;
+
+  /// This is the default output path for the category
+  final String defaultOutput;
+
+  /// This is the list of shared output paths for the category
+  final List<String> sharedOutputs;
+}
+
+class AmplifyBackend {
```
Jordan-Nelson, line 32: Q: Could this info could be created by iterating over the infra-gen2/backends dir? Is there an advantage to listing it out?

infra-gen2/tool/deploy_gen2.dart
```diff
@@ -0,0 +1,391 @@
...
+    );
+  } else {
+    print(
+      '👍 ${category.name} ${backend.identifier} sandbox deployed',
+    );
+  }
+}
+
+/// Cache the config contents to create environments map
+void _cacheConfigContents(
```
Jordan-Nelson, line 232: *suggestion:* Make this a pure function. Instead of mutating `amplifyEnvironments`, have this fn return a new `amplifyEnvironments`.

infra-gen2/tool/deploy_gen2.dart
```diff
@@ -0,0 +1,391 @@
...
+/// 4. Run `dart tool/deploy_gen2.dart` to deploy the backend
+const List<AmplifyIntegCategory> infraConfig = [
+  AmplifyIntegCategory(
+    category: Category.analytics,
+    backends: [],
+    defaultOutput: '',
+  ),
+  AmplifyIntegCategory(
+    category: Category.api,
+    backends: [
```
Jordan-Nelson, line 67: Q: Which backend becomes the "default" (the one set to `const amplifyConfig = ... ` in the outputs file? I added a new backend and that one is now the "default". Maybe it would be beneficial to specify one as default?

infra-gen2/tool/deploy_gen2.dart
Jordan-Nelson: I am not sure if I am following the purpose of this fn. Why not overwrite the file each time? If I have my own config in there will this prevent it from being overwritten?

infra-gen2/tool/deploy_gen2.dart
```diff
@@ -0,0 +1,391 @@
...
+      /// Cache the config contents to create environments map
+      _cacheConfigContents(
+        backendName,
+        amplifyOutputs,
+        amplifyEnvironments,
+      );
+    }
+
+    /// Only append environments if there are multiple backends
+    if (category.backends.length > 1) {
```
Jordan-Nelson, line 151: Q: Do you need this since you have `if (category.backends.length <= 1) continue` above?

infra-gen2/tool/deploy_gen2.dart
```diff
@@ -0,0 +1,391 @@
...
+  final bucketNameWithoutUUID = bucketName.substring(0, bucketName.length - 8);
+  final uuidMatcher = '[a-f0-9]{8}\$';
+  final pattern = '($bucketNameWithoutUUID)$uuidMatcher';
+
+  List<String> bucketNames = output.split('\t').map((e) => e.trim()).toList();
+  RegExp regex = RegExp(pattern);
+  String matchingBuckets = bucketNames.firstWhere(
+    (name) => regex.hasMatch(name),
+    orElse: () => '',
+  );
```
Jordan-Nelson, line 338: Q: This will return an empty string if there is no bucket match. It will return the first bucket if there are multiple matches. Is that the desired behavior?

You could update to the following if you want to thrown an exception in either of those cases.

```
  final matchingBuckets = bucketNames.where((name) => regex.hasMatch(name));
  if (matchingBuckets.length != 1) {
    throw Exception(...);
  }
  return matchingBuckets.single;
```

infra-gen2/pubspec.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+environment:
+  sdk: ^3.3.0
+
+dependencies:
+  amplify_core: any
+  path: any
+
+dev_dependencies:
+  amplify_lints: any
```
Jordan-Nelson, line 12: *Suggestion:* I don't see this used. Can you enable linting in this project?

NikaHsn (approved):

packages/api/amplify_api/example/integration_test/main_test.dart
```diff
@@ -29,4 +29,20 @@ void main() async {
     graph_user_pools_test.main(useExistingTestUser: true);
     rest_test.main(useExistingTestUser: true);
   });
+
+  group('amplify_api gen 2', () {
```
Jordan-Nelson, line 33: *Note:* We should discuss when the tests with gen1/cdk are removed. They add some benefit as they test that we are still properly supporting Gen 1 config formats. Integration tests are not an efficient way to test that though. If we feel good about the mapping from Gen 1 -> Gen 2 we could probably remove them.

canaries/.gitignore
```diff
@@ -65,5 +65,7 @@ amplifytools.xcconfig
 **.sample
 #amplify-do-not-edit-end
 
+amplify_outputs.dart
```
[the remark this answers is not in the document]
Equartey: No reason not to, added 👍

infra-gen2/tool/deploy_gen2.dart
```diff
@@ -0,0 +1,391 @@
...
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:amplify_core/amplify_core.dart';
+import 'package:path/path.dart' as p;
+
+class AmplifyIntegCategory {
```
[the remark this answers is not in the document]
Equartey: Good call out, went with `AmplifyBackendGroup` for brevity.

infra-gen2/tool/deploy_gen2.dart
Equartey: If we performed this each time, we would need to at least add more logic to reason where the new string should appended in the `amplifyEnvironments` map string. This is primarily what I wanted to avoid. 

This approach has the `amplifyEnvironments` map only get created once, after all the back ends have been deployed. Also, I append because we want to preserve the first, historically the default, environment named the `amplifyConfig`. 

To be fair, this logic is easier to see when there are multiple back ends defined in a group.

infra-gen2/tool/deploy_gen2.dart
Equartey: It's currently implicit, as the first one in the list will be default. I'm not sure its work adding more logic to explicitly set the default since this script is typically used once-per team member.

infra-gen2/tool/deploy_gen2.dart
Equartey: Yeah, this is outside the for loop (scope of previous if statement), so single backend groups would still hit this. If it was removed, all output files would have an empty `amplifyEnvironments` definition, which might not be a bad thing?

infra-gen2/tool/deploy_gen2.dart
Equartey: Good suggestion! Cleaned this up using your approach. We only want to throw if there are more than 1 bucket. If there isn't a bucket we want to create one.

infra-gen2/tool/deploy_gen2.dart
Equartey: Yeah, that was my first approach. I went this route because there were small nuances like naming requirements differing between `bucketName` & `bucketIdentifier` and specifying them explicitly made it easier to reason about. 

Also, this lets us create/reuse back ends not bound to one particular plugin category.  

Now, it certainly could be different/improved, open to suggestions.

Jordan-Nelson (approved):

NikaHsn (approved):

Jordan-Nelson (approved):

PR #5043, opened 2024-06-20
chore(infra): Api migrate to Gen 2 E2E

Equartey:
*Description of changes:*

- Adds Gen 2 API category backend
- Enables Gen 2 API integration tests 


.gitignore
```diff
@@ -34,11 +34,11 @@ doc/api/
 node_modules/
 aws-exports.js
 awsconfiguration.json
-amplifyconfiguration.dart
+amplifyconfiguration*.dart
```
Jordan-Nelson, line 37: Q: Why this change?

infra-gen2/backends/api/apiMultiAuth/amplify/backend.ts
```diff
@@ -1,11 +1,23 @@
...
  */
-defineBackend({
+const backend = defineBackend({
   auth,
   data,
 });
+// extract L1 CfnUserPool resources
+const { cfnUserPool } = backend.auth.resources.cfnResources;
+// modify cfnUserPool policies directly
+cfnUserPool.policies = {
```
Jordan-Nelson, line 15: Q: What is the default? Why do we need to change it? Can that change not be made in Gen 2?

infra-gen2/backends/api/apiMultiAuth/package.json
```diff
@@ -1,5 +1,8 @@
 {
   "name": "api-multi-auth",
   "version": "1.0.0",
-  "main": "index.js"
+  "main": "index.js",
+  "dependencies": {
+    "infra-common": "1.0.0"
```
Jordan-Nelson, line 6: Q: I thought dependencies were not added to the individual package.json files?

packages/api/amplify_api/example/lib/models/gen2/Blog.dart
```diff
@@ -0,0 +1,340 @@
+/*
```
Jordan-Nelson, line 1: Can you update .gitattributes to treat these as generated?

Jordan-Nelson (commented):
Left a few questions. Will review the rest later

.gitignore
Equartey: To allow multiple configuration files in your local dev. For example, I've been using a pattern like `amplifyconfiguration.gen1.dart`, `amplifyconfiguration.gen2.dart`, etc.

infra-gen2/backends/api/apiMultiAuth/amplify/backend.ts
Equartey: > What is the default?

8 Chars, 1 lower, 1 upper, 1 symbol

> Why do we need to change it?

To match the [current pattern](https://github.com/aws-amplify/amplify-flutter/blob/main/infra/lib/common.ts#L419-L427) and password generator methods we have. 

> Can that change not be made in Gen 2?

This is the [recommended](https://docs.amplify.aws/flutter/build-a-backend/auth/modify-resources-with-cdk/#override-cognito-userpool-password-policies) way when using Gen 2.

infra-gen2/backends/api/apiMultiAuth/package.json
Equartey: Good catch, this can be hoisted to be dep for all sub packages.

.gitignore
Jordan-Nelson: *issue:* There are multiple files committed that match `amplifyconfiguration*.dart` or `amplify_outputs*.dart`. You will either need to update the gitignore (preferably in the relevant packages) to allow those files explicitly or update those file names. Below are the files that I am aware of that match these patterns:
- packages/amplify_core/lib/src/config/amplify_outputs/amplify_outputs.dart
- packages/amplify_core/test/config/amplify_outputs_mapping/data/amplifyconfiguration.g.dart
- packages/amplify_core/test/config/amplify_outputs_mapping/data/amplify_outputs.g.dart

.gitattributes
```diff
@@ -94,3 +94,7 @@ yarn.lock                                   linguist-generated
 **/dist/*.mjs                               linguist-generated
 **/dist/*.cjs                               linguist-generated
 **/dist/*.cjs.map                           linguist-generated
+
+## Codegen Models
+**/example/lib/models/**                          linguist-generated
+**/example/lib/models/gen2/Gen2ModelProvider.dart linguist-generated=false
```
Jordan-Nelson, line 100: Q: Is this file not generated? It appears to be generated. If it is manually edited can you leave a comment in the file noting what manual changes were made?

infra-gen2/backends/api/apiMultiAuth/amplify/auth/pre-sign-up-handler.ts
```diff
@@ -0,0 +1,4 @@
+import { PreSignUpTriggerHandler } from "aws-lambda";
```
Jordan-Nelson, line 1: *nit:* `infra-gen2/backends/api/apiMultiAuth/amplify/auth/pre-sign-up-handler.ts` should be `infra-gen2/backends/api/api-multi-auth/amplify/auth/pre-sign-up-handler.ts` if following the naming conventions we generally follow for JS/TS projects.

infra-gen2/backends/api/apiMultiAuth/amplify/backend.ts
Jordan-Nelson: The current generatePassword method used in integ tests inlcudes a UUID, an upper case character, and a symbol. That should always meet those requirements. Am I missing something?

https://github.com/aws-amplify/amplify-flutter/blob/7cca27606e9079a335ac7d6779f3438abff7cde5/packages/test/amplify_test/lib/src/mock_data.dart#L62-L65

packages/amplify_core/lib/amplify_core.dart
```diff
@@ -18,6 +18,7 @@ export 'src/category/amplify_categories.dart';
 
 /// Config
 export 'src/config/amplify_config.dart';
+export 'src/config/amplify_outputs/amplify_outputs.dart';
```
Jordan-Nelson, line 21: Note: I think merge conflicts are causing changes to show in this PR that are not part of the PR. I am going to ignore files that I don't think are intended to be part of the PR and we re-review once the PR is updated.

packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
```diff
@@ -10,14 +10,18 @@ import 'package:integration_test/integration_test.dart';
 
 import '../util.dart';
 
-void main({bool useExistingTestUser = false}) {
+void main({bool useExistingTestUser = false, bool useGen2 = false}) {
```
Jordan-Nelson, line 13: Q: It looks like the PR updates all tests to pass `true` for `useGen2`. What is the benefit of including a flag to run the tests with the old CI stack?

I could see some benefit is being able to use the old config format, but not sure how much benefit we are getting from running the tests on the old stack.

packages/api/amplify_api/example/integration_test/util.dart
```diff
@@ -35,14 +44,15 @@ class TestUser {
   final String _username;
   final String _password;
 
-  Future<void> signUp() async {
+  String get testEmail => '$_username@amazon.com';
```
Jordan-Nelson, line 47: *suggestion:* I would add a uuid even if it isn't required right now. If this user isn't deleted after the test run and a subsequent run does not handle the case where the user already exists that may be an issue.

.gitignore
Jordan-Nelson: I pushed a commit to add gitignore files to allow those files

infra-gen2/backends/api/apiMultiAuth/amplify/backend.ts
Equartey: Ah, this function wasn't being called. Previously it was just a uuid for the password. Updated.

packages/api/amplify_api/example/integration_test/graphql/api_key_test.dart
Jordan-Nelson: We discussed offline. IIRC the summary was that we are not running the tests on the old stack. We are generating the old config from the new stack. Long term we should look to increase the unit tests for the AmplifyConfig -> AmplifyOutputs mapping so that we do not have to rely on running integ tests for this.

packages/api/amplify_api/example/integration_test/util.dart
```diff
@@ -89,10 +87,36 @@ Future<void> configureAmplify() async {
         options: APIPluginOptions(modelProvider: ModelProvider.instance),
       ),
     ]);
-    await Amplify.configure(amplifyconfig);
+    await Amplify.configure(gen1.amplifyConfig);
+  }
+}
+
+Future<void> configureAmplifyGen2() async {
```
Jordan-Nelson, line 94: Q: Rather than having two functions, can you have one function with a boolean argument? I think it would make a lot of the tests a bit cleaner.

```dart
Future<void> configureAmplify({bool useGen1Config = false}) async {
  if (!Amplify.isConfigured) {
    await Amplify.addPlugins([
      AmplifyAuthCognito(
        secureStorageFactory: AmplifySecureStorage.factoryFrom(
          macOSOptions: MacOSSecureStorageOptions(useDataProtection: false),
        ),
      ),
      AmplifyAPI(
        options: APIPluginOptions(modelProvider: ModelProvider.instance),
      ),
    ]);
    if (useGen1Config) {
      await Amplify.configure(amplifyconfig);
    } else {
      await Amplify.configure(amplify_outputs.amplifyConfig);
    }
  }
}
```

Jordan-Nelson (dismissed):

packages/api/amplify_api/example/integration_test/graphql/iam_test.dart
```diff
@@ -417,7 +442,10 @@ void main({bool useExistingTestUser = false}) {
 
       testWidgets('should fetch model that allows guest access',
           (WidgetTester tester) async {
-        final req = ModelQueries.list<Blog>(Blog.classType);
+        final req = ModelQueries.list<Blog>(
+          Blog.classType,
+          authorizationMode: APIAuthorizationType.iam,
```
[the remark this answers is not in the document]
Equartey: It's unclear why this is necessary now, but it appears to be more accurate since these models have multiple authentication modes and an explicit mode should have been defined from the start.

Jordan-Nelson (dismissed):

infra-gen2/tool/deploy_gen2.dart
```diff
@@ -108,15 +124,18 @@ void main(List<String> arguments) async {
...
+    _copyConfigFile(
       backendGroup.sharedOutputs,
       amplifyOutputs,
     );
-
-    var bucketName = _createBucketName(categoryName);
+    _copyConfigFile(
+      backendGroup.sharedOutputs,
+      amplifyConfiguration,
+    );
```
NikaHsn, line 135: nit: instead of calling the  _copyConfigFile two times we can update the func to accept a list of file instead of one file and so call it once and pass both config files.

infra-gen2/tool/deploy_gen2.dart
```diff
@@ -126,10 +145,14 @@ void main(List<String> arguments) async {
...
+    // Upload config files to S3 bucket
+    _uploadConfigFileToS3(
       bucketName,
-      amplifyOutputs.path,
+      amplifyOutputs,
+    );
+    _uploadConfigFileToS3(
+      bucketName,
+      amplifyConfiguration,
     );
```
NikaHsn, line 156: same as above

NikaHsn (dismissed):

Jordan-Nelson (approved):

NikaHsn (approved):

PR #5075, opened 2024-06-25
Feat: fetchCurrentDevice API Implementation

hahnandrew:
*Description of changes:*

Implemented `fetchCurrentDevice` API

### DIFF
--- a/packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
+++ b/packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
@@ -31,6 +31,7 @@ import 'package:amplify_auth_cognito_dart/src/sdk/cognito_identity_provider.dart
         ForgotPasswordRequest,
         GetUserAttributeVerificationCodeRequest,
         GetUserRequest,
+        GetDeviceRequest,
         ListDevicesRequest,
         ResendConfirmationCodeRequest,
         UserContextDataType,
@@ -39,6 +40,7 @@ import 'package:amplify_auth_cognito_dart/src/sdk/cognito_identity_provider.dart
         VerifyUserAttributeRequest;
 import 'package:amplify_auth_cognito_dart/src/sdk/sdk_bridge.dart';
 import 'package:amplify_auth_cognito_dart/src/sdk/src/cognito_identity_provider/model/analytics_metadata_type.dart';
+import 'package:amplify_auth_cognito_dart/src/sdk/src/cognito_identity_provider/model/get_device_response.dart';
 import 'package:amplify_auth_cognito_dart/src/state/cognito_state_machine.dart';
 import 'package:amplify_auth_cognito_dart/src/state/state.dart';
 import 'package:amplify_auth_cognito_dart/src/util/cognito_iam_auth_provider.dart';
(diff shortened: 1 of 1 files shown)


khatruong2009 (approved):

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
```diff
@@ -993,6 +996,46 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface
...
+  @override
+  Future<CognitoDevice> fetchCurrentDevice() async {
+    final tokens = await stateMachine.getUserPoolTokens();
+    final deviceSecrets = await _deviceRepo.get(tokens.username);
+    final deviceKey = deviceSecrets?.deviceKey;
+    if (deviceSecrets == null || deviceKey == null) {
+      throw const DeviceNotTrackedException();
+    }
+
+    late GetDeviceResponse resp;
```
Equartey, line 1008: nit: IMO the abbreviation doesn't add much value here.
```suggestion
    late GetDeviceResponse response;
```

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
```diff
@@ -993,6 +996,46 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface
...
+    final device = resp.device;
+    final attributes =
+        device.deviceAttributes ?? const <cognito.AttributeType>[];
+
+    return CognitoDevice(
+      id: deviceKey,
+      attributes: {
+        for (final attribute in attributes)
+          attribute.name: attribute.value ?? '',
+      },
```
Equartey, line 1032: Q: Does this need a for loop? Does `attributes` have other properties? 
```suggestion
      attributes: attributes,
```

hahnandrew: > Q: Does this need a for loop? 

No, you can rewrite this (after checking attributes is [not empty](https://pub.dev/documentation/built_collection/latest/built_collection/BuiltList/isNotEmpty.html) like this:
```
    final test = attributes.elementAt(0);
    final name = test.name;
    final value = test.value ?? '';
    ...
    attributes = {name: value}
```
... among other methods in the [BuiltList class](https://pub.dev/documentation/built_collection/latest/built_collection/BuiltList-class.html). The loop is mostly stylistic and for parity, being that this is how the other devices API's extracts the name and values from the attributes map. 

Note that `attributes: attributes` is type invalid.  

> Does `attributes` have other properties?

As of now, no

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
hahnandrew: added

packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
Equartey: Makes sense, thanks for adding that context.

Equartey (approved):
LGTM

PR #5104, opened 2024-07-01
chore(infra): analytics integ test gen 2 backend

khatruong2009:
*Issue #, if available:*

*Description of changes:*


packages/analytics/amplify_analytics_pinpoint/example/tool/pull_test_backend.sh
```diff
@@ -4,4 +4,4 @@
 
 set -e
 
-aws s3 cp s3://$AFS_ANALYTICS_BUCKET_NAME/amplifyconfiguration.dart lib/amplifyconfiguration.dart
+aws s3 cp s3://$AFS_ANALYTICS_GEN2_BUCKET_NAME/amplify_outputs.dart lib/amplify_outputs.dart
```
Equartey, line 7: open question: do we want to run the test suite with both config versions?

Equartey (approved):
LGTM

packages/analytics/amplify_analytics_pinpoint/example/tool/pull_test_backend.sh
Jordan-Nelson: IMO we should not. The only benefit of running it with both configs is to test the config mapping. The analytics config only has 2 properties. We have unit tests to ensure those two properties are mapped correctly. No need to run the full e2e test suite for that.

I think the same argument hold true for other categories. While the mapping is a bit more complex in other categories (especially auth) I would rather invest some time into writing unit tests for the mapping rather than relying on the e2e suite.

infra-gen2/backends/analytics/main/amplify/backend.ts
```diff
@@ -0,0 +1,30 @@
+import { defineBackend } from "@aws-amplify/backend";
+import { addAnalyticsExtensions } from "infra-common";
+import { auth } from "./auth/resource";
+
+// const __dirname = path.resolve();
```
NikaHsn, line 5: clean up the commend code

infra-gen2/backends/analytics/no-unauth-access/amplify/backend.ts
```diff
@@ -0,0 +1,28 @@
+import { defineBackend } from "@aws-amplify/backend";
+import { addAnalyticsExtensions } from "infra-common";
+import { auth } from "./auth/resource";
+
+// const __dirname = path.resolve();
```
NikaHsn, line 5: clean up the commented code

infra-gen2/backends/analytics/no-unauth-identities/amplify/backend.ts
```diff
@@ -0,0 +1,33 @@
+import { defineBackend } from "@aws-amplify/backend";
+import { addAnalyticsExtensions } from "infra-common";
+import { auth } from "./auth/resource";
+
+// const __dirname = path.resolve();
```
NikaHsn, line 5: same as above

packages/analytics/amplify_analytics_pinpoint/example/tool/pull_test_backend.sh
NikaHsn: +1 ^^

Equartey (approved):

NikaHsn (approved):

PR #5134, opened 2024-07-10
chore(actions): xcodes ios 18 fix

Equartey:
*Description of changes:*

- Changes `ios-version: latest` to `ios-version: "17.5"` when trying to download the iOS simulator runtime. This is due the breaking change Apple introduced with iOS 18. The cli tool we use `xcodes` does not support iOS 18. This can revisited when the community finds a reliable way to download iOS 18 runtimes.
- Updates `pnpm` setup action version to 4.0.0
- Update `pnpm` version to 9.5


Jordan-Nelson (dismissed):

.github/composite_actions/launch_ios_simulator/action.yaml
```diff
@@ -3,7 +3,7 @@ description: Launches an iOS simulator and caches it for further action runs
 inputs:
   ios-version:
     description: The iOS version
-    default: "latest"
```
khatruong2009, line 6: do we have to change this back to latest once ios 18 releases? if so, should we put a todo in there?

Equartey: Not necessarily. It's likely that we will have to completely revise our entire process in order to download iOS18 runtimes. I can add a todo to reflect that.

khatruong2009 (approved):

tyllark (approved):

Issue #5165, opened 2024-07-17
Compile issues after upgrading to 2.2.0


iosephmagno:
### Description

After upgrading to latest flutter master channel and upgrading Amplify to 2.2.0, I get several compile issues that I cannto solve:

```
// Error
error: cannot find symbol
      flutterEngine.getPlugins().add(new com.amazonaws.amplify.amplify_secure_storage.AmplifySecureStoragePlugin());
```

```
// Warnings 
Package amplify_secure_storage:linux references amplify_secure_storage_dart:linux as the default plugin, but the package does not exist.
Ask the maintainers of amplify_secure_storage to either avoid referencing a default implementation via `platforms: linux: default_package: amplify_secure_storage_dart` or create a plugin named amplify_secure_storage_dart.

Package amplify_secure_storage:windows references amplify_secure_storage_dart:windows as the default plugin, but the package does not exist.
Ask the maintainers of amplify_secure_storage to either avoid referencing a default implementation via `platforms: windows: default_package: amplify_secure_storage_dart` or create a plugin named amplify_secure_storage_dart.

Package amplify_secure_storage:linux references amplify_secure_storage_dart:linux as the default plugin, but the package does not exist.
Ask the maintainers of amplify_secure_storage to either avoid referencing a default implementation via `platforms: linux: default_package: amplify_secure_storage_dart` or create a plugin named amplify_secure_storage_dart.

Package amplify_secure_storage:windows references amplify_secure_storage_dart:windows as the default plugin, but the package does not exist.
Ask the maintainers of amplify_secure_storage to either avoid referencing a default implementation via `platforms: windows: default_package: amplify_secure_storage_dart` or create a plugin named amplify_secure_storage_dart.

Package amplify_secure_storage:linux references amplify_secure_storage_dart:linux as the default plugin, but the package does not exist.
Ask the maintainers of amplify_secure_storage to either avoid referencing a default implementation via `platforms: linux: default_package: amplify_secure_storage_dart` or create a plugin named amplify_secure_storage_dart.

Package amplify_secure_storage:windows references amplify_secure_storage_dart:windows as the default plugin, but the package does not exist.
Ask the maintainers of amplify_secure_storage to either avoid referencing a default implementation via `platforms: windows: default_package: amplify_secure_storage_dart` or create a plugin named amplify_secure_storage_dart.

```

My pubspec is:
```
  amplify_auth_cognito: 2.2.0
  amplify_flutter: 2.2.0 
[...]

PR #5178, opened 2024-07-18
fix(secure_storage): Add plugin platform entry.

Equartey:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-flutter/issues/5165
*Description of changes:*
Followed [Flutter doc guidance](https://docs.flutter.dev/packages-and-plugins/developing-packages#dart-only-platform-implementations) to specify platform specific entry points for the Secure Storage plugin.  This resolves the build errors from Flutter Master channel: 
```
Package amplify_secure_storage:linux references amplify_secure_storage_dart:linux as the default plugin, but the package does not exist.
Ask the maintainers of amplify_secure_storage to either avoid referencing a default implementation via `platforms: linux: default_package: amplify_secure_storage_dart` or create a plugin named amplify_secure_storage_dart.

Package amplify_secure_storage:windows references amplify_secure_storage_dart:windows as the default plugin, but the package does not exist.
Ask the maintainers of amplify_secure_storage to either avoid referencing a default implementation via `platforms: windows: default_package: amplify_secure_storage_dart` or create a plugin named amplify_secure_storage_dart.
```

### DIFF
--- a/packages/secure_storage/amplify_secure_storage_dart/lib/src/amplify_secure_storage_dart.dart
+++ b/packages/secure_storage/amplify_secure_storage_dart/lib/src/amplify_secure_storage_dart.dart
@@ -64,6 +64,10 @@ class AmplifySecureStorageDart extends AmplifySecureStorageInterface
       );
     };
   }
+
+  // Required to register the plugin on windows and linux platforms.
+  // https://docs.flutter.dev/packages-and-plugins/developing-packages#plugin-platforms
+  static void registerWith() {}
 }
 
 /// {@template amplify_secure_storage_dart.amplify_secure_storage_worker}
--- a/packages/secure_storage/amplify_secure_storage_dart/lib/src/amplify_secure_storage_dart.dart
+++ b/packages/secure_storage/amplify_secure_storage_dart/lib/src/amplify_secure_storage_dart.dart
@@ -64,10 +64,6 @@ class AmplifySecureStorageDart extends AmplifySecureStorageInterface
       );
     };
   }
-
-  // Required to register the plugin on windows and linux platforms.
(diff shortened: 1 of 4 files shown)


packages/secure_storage/amplify_secure_storage_dart/lib/src/amplify_secure_storage_dart.dart
```diff
@@ -64,6 +64,10 @@ class AmplifySecureStorageDart extends AmplifySecureStorageInterface
       );
     };
   }
+
+  // Required to register the plugin on windows and linux platforms.
+  // https://docs.flutter.dev/packages-and-plugins/developing-packages#plugin-platforms
+  static void registerWith() {}
```
Equartey, line 70: Can this can be a no-op? There isn't a instance from the current impl to map here.

packages/secure_storage/amplify_secure_storage_dart/pubspec.yaml
```diff
@@ -48,3 +48,9 @@ flutter:
   assets:
     - lib/src/worker/workers.min.js
     - lib/src/worker/workers.min.js.map
+  plugin:
+    platforms:
+      linux:
+        default_package: amplify_secure_storage_dart
```
Jordan-Nelson, line 54: `default_package` would typically refer to another package. This references itself. Is that something flutter docs suggest doing?

What I am seeing in the docs is to use `dartPluginClass` 

```suggestion
        dartPluginClass: AmplifySecureStorageDart
```

You will also have to add a `registerWith()` method to AmplifySecureStorageDart I believe. This is based on https://docs.flutter.dev/packages-and-plugins/developing-packages#dart-only-platform-implementations

Jordan-Nelson (approved):

PR #5193, opened 2024-07-22
chore(api): update endpoint config to use ApiOutputs instead of AWSApiConfig type

NikaHsn:
*Issue #, if available:*

*Description of changes:*
- add a new ApiOutputs abstract interface
- DataOutputs and RestApiOutputs to implement the ApiOutputs
- Use ApiOutputs instead of AWSApiConfig type

### DIFF
--- a/packages/amplify_core/lib/src/config/amplify_outputs/data/data_outputs.dart
+++ b/packages/amplify_core/lib/src/config/amplify_outputs/data/data_outputs.dart
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import 'package:amplify_core/amplify_core.dart';
+import 'package:amplify_core/src/config/amplify_outputs/api_outputs.dart';
 
 part 'data_outputs.g.dart';
 
@@ -10,7 +11,8 @@ part 'data_outputs.g.dart';
 /// {@endtemplate}
 @zAmplifyOutputsSerializable
 class DataOutputs
-    with AWSEquatable<DataOutputs>, AWSSerializable, AWSDebuggable {
+    with AWSEquatable<DataOutputs>, AWSSerializable, AWSDebuggable
+    implements ApiOutputs {
   /// {@macro amplify_core.amplify_outputs.data_outputs}
   const DataOutputs({
     required this.awsRegion,
@@ -24,12 +26,15 @@ class DataOutputs
       _$DataOutputsFromJson(json);
 
   /// The AWS region of Amazon AppSync resources.
+  @override
   final String awsRegion;
 
   /// The AppSync endpoint URL.
+  @override
   final String url;
 
   /// The AppSync API Key.
+  @override
   final String? apiKey;
 
   /// The default authorization type for AWS AppSync.
@@ -37,6 +42,13 @@ class DataOutputs
 
   /// List of supported authorization types for AWS AppSync.
   final List<APIAuthorizationType> authorizationTypes;
+
+  @override
+  ApiType get apiType => ApiType.graphQL;
+
+  @override
+  APIAuthorizationType<AmplifyAuthProvider> get authorizationType =>
+      defaultAuthorizationType;
 
   @override
   List<Object?> get props => [
--- a/packages/amplify_core/lib/src/config/amplify_outputs/data/data_outputs.dart
+++ b/packages/amplify_core/lib/src/config/amplify_outputs/data/data_outputs.dart
@@ -43,9 +43,11 @@ class DataOutputs
   /// List of supported authorization types for AWS AppSync.
   final List<APIAuthorizationType> authorizationTypes;
 
+  /// The GraphQL Api type.
   @override
   ApiType get apiType => ApiType.graphQL;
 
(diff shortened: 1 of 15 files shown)


packages/api/amplify_api_dart/test/amplify_api_config_test.dart
```diff
@@ -58,20 +61,28 @@ void main() {
...
+      authorizationTypes: [authorizationType],
+    );
+  } else {
+    config = RestApiOutputs(
+      url: endpoint,
+      awsRegion: region,
+      authorizationType: authorizationType,
+      apiKey: apiKey,
+    );
+  }
```
Jordan-Nelson, line 85: *suggestion:* Use a switch statement. It will allow config to be final and will ensure that we are handling each case (if we added a new ApiType in the future, we would be alerted that this code needs to be updated with a switch statement)

```suggestion
  final config = switch(type) {
    ApiType.graphQL => DataOutputs(
      url: endpoint,
      awsRegion: region,
      defaultAuthorizationType: authorizationType,
      apiKey: apiKey,
      authorizationTypes: [authorizationType],
    ),
    ApiType.rest => RestApiOutputs(
      url: endpoint,
      awsRegion: region,
      authorizationType: authorizationType,
      apiKey: apiKey,
    )
  };
```

packages/amplify_core/lib/src/config/amplify_outputs/data/data_outputs.dart
```diff
@@ -38,6 +43,13 @@ class DataOutputs
   /// List of supported authorization types for AWS AppSync.
   final List<APIAuthorizationType> authorizationTypes;
 
+  @override
+  ApiType get apiType => ApiType.graphQL;
```
Jordan-Nelson, line 47: *Question:* Does this get included in the json serialization/de-serialization? I don't see the generated file changed, but that could just be caused by codegen not being run

packages/amplify_core/lib/src/config/amplify_outputs/rest_api/rest_api_outputs.dart
```diff
@@ -25,18 +27,25 @@ class RestApiOutputs
...
+  /// The Amazon API Gateway API Key.
+  @override
   final String? apiKey;
 
   /// The authorization type.
+  @override
   final APIAuthorizationType authorizationType;
 
+  @override
+  ApiType get apiType => ApiType.rest;
```
Equartey, line 47: nit: this is missing its docs comment. Same with the `DataOutputs.apiType`

packages/amplify_core/lib/src/config/amplify_outputs/data/data_outputs.dart
```diff
@@ -38,6 +43,13 @@ class DataOutputs
...
+
+  @override
+  APIAuthorizationType<AmplifyAuthProvider> get authorizationType =>
```
Equartey, line 50: nit: missing doc comment

packages/api/amplify_api_dart/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart
```diff
@@ -12,6 +12,8 @@ import 'package:amplify_api_dart/src/graphql/web_socket/types/connectivity_platf
 import 'package:amplify_api_dart/src/graphql/web_socket/types/subscriptions_event.dart';
 import 'package:amplify_api_dart/src/graphql/web_socket/types/web_socket_types.dart';
 import 'package:amplify_core/amplify_core.dart' hide SubscriptionEvent;
+// ignore: implementation_imports
+import 'package:amplify_core/src/config/amplify_outputs/api_outputs.dart';
```
Equartey, line 16: Q: Needing this ignore seems odd, are we sure this is the best way to import/export this type?

packages/api/amplify_api_dart/test/amplify_api_config_test.dart
Equartey: +1. Can also add a default case to throw with more meaningful messaging.

NikaHsn: type is an enum and all the valuse are covered in the switch. hence adding a default case (_) cuase analyzer error: This case is covered by the previous cases.

packages/amplify_core/lib/src/config/amplify_outputs/data/data_outputs.dart
NikaHsn: code gen does not change the generated codes, I think this is because they are getters.

packages/api/amplify_api_dart/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart
NikaHsn: I think this is the way we do in the monorepo when we do not want to export files/types from one package but still want to use them in other packages.

Equartey (approved):

Jordan-Nelson (approved):

Issue #5190, opened 2024-07-22
Amplify Fails to Configure - WorkerBeeExceptionImpl


CarlosEspinozaMaruri:
### Description

**This issue is a duplicate of [this](https://github.com/aws-amplify/amplify-flutter/issues/4883) that was closed due to lack of information. I have managed to obtain the requested JSON files and attach them below.**

---

When attempting to configure Amplify using Amplify.configure(), the following exception is being thrown:

Exception:

```
 WorkerBeeExceptionImpl {
  error=FormatException: Unexpected character (at character 27907)
...33,74,41,5,53,129,103,142,83,81,166,215,198,74,136,96,91,72,189,196]},196]}
                                                                        ^
,
  stackTrace=#0      _ChunkedJsonParser.fail (dart:convert-patch/convert_patch.dart:1380)
#1      _ChunkedJsonParser.parse (dart:convert-patch/convert_patch.dart:882)
#2      _parseJson (dart:convert-patch/convert_patch.dart:35)
#3      JsonDecoder.convert (dart:convert/json.dart:610)
#4      JsonCodec.decode (dart:convert/json.dart:216)
#5      FileKeyValueStore.readAll (package:amplify_secure_storage_dart/src/utils/file_key_value_store.dart:96)
<asynchronous suspension>
#6      FileKeyValueStore.readKey (package:amplify_secure_storage_dart/src/utils/file_key_value_store.dart:70)
<asynchronous suspension>
#7      AmplifySecureStorageWindows.read (package:amplify_secure_storage_dart/src/platforms/amplify_secure_storage_windows.dart:45)
<asynchronous suspension>
#8      SecureStorageWorker.run (package:amplify_secure_storage_dart/src/worker/secure_storage_worker.dart:53)
<asynchronous suspension>
#9      _run (package:amplify_secure_storage_dart/src/worker/secure_storage_worker.worker.vm.dart:14)
<asynchronous suspension>
```

Stacktrace:

```
dart:convert-patch/convert_patch.dart 1383                                                 _ChunkedJsonParser.fail
dart:convert-patch/convert_patch.dart 913                                                  _ChunkedJsonParser.parse
dart:convert-patch/convert_patch.dart 35                                                   _parseJson
dart:convert/json.dart 610                                                                 JsonDecoder.convert
dart:convert/json.dart 216                                                                 JsonCodec.decode
[...]

PR #5194, opened 2024-07-23
fix(secure_storage): handle parallel read/write/delete operations on Windows

Jordan-Nelson:
*Issue #, if available:* https://github.com/aws-amplify/amplify-flutter/issues/5190

*Description of changes:*
- Update FileKeyValueStore to process tasks one at a time in the order in which they are received

This PR accomplished two things:
1. Addresses a bug on Windows in which the file can become corrupted when there are many multiple writes/deletes occurring without awaiting them. On windows it appears that it is not safe to make multiple File system API calls in parallel (MacOS does not have this issue). 
2. Ensures that even if the caller does not await write/delete events, that they are processed in the order in which they are received.


Jordan-Nelson:
Closing in favor of https://github.com/aws-amplify/amplify-flutter/pull/5195

PR #5195, opened 2024-07-23
fix(secure_storage): process file system operations one at a time on Windows

Jordan-Nelson:
*Issue #, if available:* https://github.com/aws-amplify/amplify-flutter/issues/5190

*Description of changes:*
- Update FileKeyValueStore to process tasks one at a time in the order in which they are received to prevent multiple parallel file operations from corrupting the file
- Update readAll to log an error and clear the file contents if it is corrupted as it will be unreadable and unrecoverable
- Annotate readAll/writeAll as visible for testing
- add tests for in-memory fs and local fs

### DIFF
--- a/packages/secure_storage/amplify_secure_storage_dart/lib/src/utils/file_key_value_store.dart
+++ b/packages/secure_storage/amplify_secure_storage_dart/lib/src/utils/file_key_value_store.dart
@@ -1,6 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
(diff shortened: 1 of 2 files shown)


packages/secure_storage/amplify_secure_storage_dart/lib/src/utils/file_key_value_store.dart
```diff
@@ -47,9 +53,11 @@ class FileKeyValueStore {
...
-    data[key] = value;
-    return writeAll(data);
+    return _scheduler.schedule(() async {
+      final data = await readAll();
+      data[key] = value;
+      return writeAll(data);
+    });
   }
 
   /// Overwrites the existing data.
```
tyllark, line 63: Should writeAll be wrapped by the scheduler as well?

Jordan-Nelson: Discussed offline - I updated writeAll and readAll to be annotated with `@visibleForTesting` as these apis are really private and only exposed for testing.

Equartey (approved):

tyllark (approved):

CarlosEspinozaMaruri (approved):

Issue #5122, opened 2024-07-08
Pinpoint push notification campaign events are not being recorded. 


naveenthontepu:
### Description

We are unable to track pinpoint push notification campaign and journey events as they are not being recorded from the flutter SDK.

### Categories

- [X] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [ ] Auth
- [ ] Authenticator
- [ ] DataStore
- [X] Notifications (Push)
- [ ] Storage

### Steps to Reproduce

1. Create a push notification campaign in pinpoint.
2. Run the campaign.
3. Go to the campaign metrics section and only messages delivered is visible, the app open metric is not available.  

### Platforms

- [X] iOS
- [X] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Flutter Version

3.10.2

### Amplify Flutter Version

2.2.0

### Deployment Method

Custom Pipeline

khatruong2009:
Hi @naveenthontepu, thank you for submitting this issue. We will take a look and get back to you with any updates.

tyllark:
Hello @naveenthontepu, I was able to reproduce the issue and determined that the root cause is that the [Push Notification Campaign Events](https://docs.aws.amazon.com/pinpoint/latest/developerguide/event-streams-data-campaign.html#event-streams-data-campaign-attributes) are being cached in memory, but are never being sent to PinPoint. I'm marking this issue as a bug and we will get back to you when we have a fix.

NikaHsn:
I'm going to close this issue as the fix has been released in [2.4.0](https://github.com/aws-amplify/amplify-flutter/releases/tag/v2.4.0) version. If you see otherwise or have additional questions, please open a new issue.

PR #5215, opened 2024-07-29
fix: push notification flush events

tyllark:
[Github Issue 5122](https://github.com/aws-amplify/amplify-flutter/issues/5122)

*Description of changes:*
1. Moved StoppableTimer to aws_common so it can be reused to flush events for push notifications and analytics.
2. [Re-added logic](https://github.com/aws-amplify/amplify-flutter/commit/f0f572868e852b1da393507b37c07cb4110fc56b) for flushing events every 10 seconds.
3. Updated push notification event source names to match [Pinpoint documentation](https://docs.aws.amazon.com/pinpoint/latest/developerguide/event-streams-data-campaign.html#event-streams-data-campaign-attributes).


packages/amplify_core/lib/src/types/notifications/push/pinpoint_event_source.dart
```diff
@@ -11,8 +11,8 @@
 /// [Journeys](https://docs.aws.amazon.com/pinpoint/latest/userguide/journeys.html)
 /// {@endtemplate}
 enum PinpointEventSource {
```
NikaHsn, line 13: is this type public/exported? if so is this considered a breaking change?

tyllark: I marked the enum as deprecated and created a new private enum with the updated values. thanks for the callout!

packages/aws_common/lib/src/util/stoppable_timer.dart
```diff
@@ -0,0 +1,44 @@
...
+
+import 'dart:async';
+
+import 'package:meta/meta.dart';
+
+/// {@template amplify_analytics_pinpoint_dart.stoppable_timer}
+/// A Timer that can be stopped and started again.
+/// {@endtemplate}
+@protected
+class StoppableTimer {
```
Jordan-Nelson, line 12: Question: Can StoppableTimer be removed from amplify_analytics_pinpoint?

tyllark: Good catch. I initially moved StoppableTimer into amplify_core, but forgot to remove it after migrating it again to aws_common.

Jordan-Nelson (approved):

Equartey (approved):

NikaHsn (approved):

Issue #5217, opened 2024-07-29
Datastore V2+ not raising networkStatus event and not syncronizing data when offline on iOS


filipesbragio:
### Description

We've upgraded recently from Amplify V1 to V2
```
  amplify_api: ^1.8.0
  amplify_auth_cognito: ^1.8.0
  amplify_authenticator: ^1.0.0
  amplify_datastore: ^1.8.0
  amplify_flutter: ^1.8.0
```

to 

```
  amplify_api: ^2.3.0
  amplify_auth_cognito: ^2.3.0
  amplify_authenticator: ^2.1.1
  amplify_datastore: ^2.3.0
  amplify_flutter: ^2.3.0
```

Ever since this upgrade we haven't been receiving the networkStatus event when listening to datastore changes. Other events seem to be raised normally. Besides this, any changes we make while offline seem to be ignored and never updated on dynamoDB. Any changes made after becoming online again sync normally. I haven't created different issues for these as I believe they're related. I haven't been able to reproduce this on android, only on iOS.

We've also tried the 2.1.0 version and it seemed to have the same issue.

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [ ] Auth
- [ ] Authenticator
- [X] DataStore
- [ ] Notifications (Push)
- [ ] Storage

### Steps to Reproduce

on iOS, listen to data store events and turn on/off connection. No networkStatus events will be raised. 
Similarly, become offline and make changes to the model and then become online. The offline changes will not be synchronized. Subsequent changes while online will be synchronized normally.

```
StreamSubscription<DataStoreHubEvent> _createStreamSubscription() {
    return Amplify.Hub.listen(HubChannel.DataStore, (DataStoreHubEvent event) async {
      print(event.eventName);
    });
  }
```

### Platforms

- [X] iOS
- [ ] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Flutter Version

3.22.3

### Amplify Flutter Version

2.3.0

### Deployment Method

Amplify CLI

Equartey:
Hi @filipesbragio, thank you for the detailed write up. We will investigate the root cause and give you updates as we have them.

Equartey:
Hi @filipesbragio, quick update for you. 

I have opened https://github.com/aws-amplify/amplify-flutter/pull/5218 to resolve this issue, once it gets merged and released I'll notify you here. Thanks again for submitting this.

[...]

PR #5218, opened 2024-07-30
fix(datastore): Restart Sync Engine when network on/off

Equartey:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-flutter/issues/5217
*Description of changes:*
- Added Native Swift network monitor to listen to hardware level events
- Added callback for Swift to tell Flutter network was lost
- Logic to close Web socket on network loss
- Added an internal `WebSocketOptions` as a static way to turn off reconnection logic. This allows Swift DataStore remote sync engine to control when to reestablish the subscriptions.

**Note:** Currently there exists a delay in restarting the sync engine from a network loss event if the connection is not immediately reestablished. This is due to the retry/back off logic from the Swift remote sync engine. Once network is turned back on, DataStore will eventually restart the sync engine if it was previously running and `Amplify.Datastore.stop()` was never called. 
 
### DIFF
--- a/packages/amplify_core/lib/amplify_core.dart
+++ b/packages/amplify_core/lib/amplify_core.dart
@@ -62,11 +62,11 @@ export 'src/state_machine/transition.dart';
 
 /// Analytics
 export 'src/types/analytics/analytics_types.dart';
-// ignore: invalid_export_of_internal_element
-export 'src/types/api/api_types.dart' show WebSocketOptions;
 
 /// API
 export 'src/types/api/api_types.dart' hide WebSocketOptions;
+// ignore: invalid_export_of_internal_element
+export 'src/types/api/api_types.dart' show WebSocketOptions;
 
 /// App path provider
 export 'src/types/app_path_provider/app_path_provider.dart';
--- a/packages/amplify_core/lib/amplify_core.dart
+++ b/packages/amplify_core/lib/amplify_core.dart
@@ -62,9 +62,11 @@ export 'src/state_machine/transition.dart';
 
 /// Analytics
 export 'src/types/analytics/analytics_types.dart';
+// ignore: invalid_export_of_internal_element
+export 'src/types/api/api_types.dart' show WebSocketOptions;
 
 /// API
-export 'src/types/api/api_types.dart';
+export 'src/types/api/api_types.dart' hide WebSocketOptions;
 
 /// App path provider
 export 'src/types/app_path_provider/app_path_provider.dart';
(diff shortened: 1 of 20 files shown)


packages/amplify_datastore/lib/amplify_datastore.dart
```diff
@@ -350,7 +350,9 @@ class NativeAmplifyApi
   Future<NativeGraphQLSubscriptionResponse> subscribe(
       NativeGraphQLRequest request) async {
     final flutterRequest = nativeRequestToGraphQLRequest(request);
-
+    // Turn off then default reconnection behavior to allow native side to trigger reconnect
```
NikaHsn, line 353: Q: I don't see the native code to call WebSocketOptions.autoReconnect = true, is it expected?

Equartey: Yes this is expected. Swift has it's own reconnection logic that is used for DataStore. Having them both enabled created race conditions, so I disabled Flutter's in favor of Swift because it's also tied to restarting the sync engine.

packages/amplify_core/lib/amplify_core.dart
```diff
@@ -62,9 +62,11 @@ export 'src/state_machine/transition.dart';
 
 /// Analytics
 export 'src/types/analytics/analytics_types.dart';
+// ignore: invalid_export_of_internal_element
+export 'src/types/api/api_types.dart' show WebSocketOptions;
 
 /// API
-export 'src/types/api/api_types.dart';
+export 'src/types/api/api_types.dart' hide WebSocketOptions;
```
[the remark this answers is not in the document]
Equartey: This is broken into two lines to specify which import is getting the lint ignore. If it's hidden, it can't be used by the other categories.

NikaHsn (approved):

packages/amplify_datastore/example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -693,14 +693,15 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
 				CURRENT_PROJECT_VERSION = "$(FLUTTER_BUILD_NUMBER)";
```
tyllark, line 695: Are the DEVELOPMENT_TEAM and PRODUCT_BUNDLE_IDENTIFIER changes here intentional?

tyllark (approved):

packages/amplify_datastore/example/ios/Runner.xcodeproj/project.pbxproj
Equartey: Unintentional, thanks.

NikaHsn (approved):

tyllark (approved):

PR #5291, opened 2024-08-13
chore: migrate sms only MFA infra to Gen 2

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Migrate two backends that use SMS only for MFA (MFA required and MFA optional) to Gen 2
- Refactor e2e tests for new backends
- Resolve existing issue in auth lambda infra


infra-gen2/infra-common/src/auth-user-extensions/custom-sender-lambda.ts
```diff
@@ -44,9 +44,6 @@ export function addCustomSenderLambda({
         "custom-email-sender.js"
       ),
       runtime: Runtime.NODEJS_18_X,
-      bundling: {
-        nodeModules: ["@aws-crypto/client-node"],
-      },
```
Jordan-Nelson, line 49: Note: This was causing an issue deploying the infra, and is not needed. It am not sure why it was not causing an issue previously.

infra-gen2/infra-common/src/auth-user-extensions/auth-user-extensions.ts
```diff
@@ -23,6 +24,7 @@ export const addAuthUserExtensions = ({
   addCustomSenderLambda({ name, stack, cfnUserPool, graphQL });
   addCreateUserLambda({ name, stack, userPool, graphQL });
   addDeleteUserLambda({ name, stack, userPool, graphQL });
+  addEnableSmsMfaLambda({ name, stack, userPool, graphQL });
```
Equartey, line 27: Q: Will this add the lambda to every backend that calls this method? Can this be optional?

packages/test/amplify_auth_integration_test/lib/src/test_runner.dart
```diff
@@ -263,13 +275,17 @@ class AuthTestRunner {
   }
 
   /// Runs [body] in a [group] which configures [environment].
-  void withEnvironment(TestEnvironment environment, void Function() body) {
-    group(environment.environmentName, () {
+  void withEnvironment(
+      EnvironmentInfo environment, void Function(EnvironmentInfo env) body) {
```
Equartey, line 279: nit: trailing coma

packages/test/amplify_auth_integration_test/lib/src/environments.dart
Equartey, line 1: Q: General question on this file, can any of this info be obtained from the outputs files to reduce sources of truth?

infra-gen2/infra-common/src/auth-user-extensions/auth-user-extensions.ts
Jordan-Nelson: Yes it will add it to each env. It could be optional. So could the other lambdas. I think it is simpler to add all the lambdas to each env rather than try to remember which envs have which lambda. It doesn't really cost anything if it doesn't get invoked.

packages/test/amplify_auth_integration_test/lib/src/environments.dart
Jordan-Nelson: We could, yes. 

There are some pieces of info that I do not want to derive from the outputs file because that could result in false positives or false negatives if the outputs file is wrong. For example, we have tests that loop over all the environments where `mfaEnabled` is true. If that were derived from the outputs file, and a bug in the outputs file caused it to always be false, those test would never run.

There are _some_ pieces of info that would be fine to derive from the outputs file for _some_ tests. For example deriving the username alias would be fine for the MFA tests. However it would be problematic for the sign in / sign up tests that are supposed to test each alias.

packages/test/amplify_auth_integration_test/lib/src/test_runner.dart
Jordan-Nelson: Updated

packages/test/amplify_auth_integration_test/lib/src/environments.dart
Equartey: I see, this file helps strengthen the testing contract. I'm good with that.

infra-gen2/infra-common/src/auth-user-extensions/custom-sender-lambda.ts
NikaHsn: thanks for the clean up.

infra-gen2/infra-common/src/auth-user-extensions/auth-user-extensions.ts
Equartey: It doesn't need to come from this PR, but it would be a nice optimization to have these be pulled in as needed. Besides reducing the amount of resources added to the AWS account, it may speed up deployment of the entire infra stack. May also prevent maintenance confusion by having the bare minimum for each env.

Equartey (approved):

NikaHsn (approved):

PR #5436, opened 2024-09-10
chore: fix `aft version-bump`

Jordan-Nelson:
*Description of changes:*
1. Propagate version bumps n levels deep (instead of only 2 levels deep) and propagate version bumps when a component's config requires propagation (opposed to only for breaking changes)
    - ex: if amplify_auth_cognito receives a minor bump, this should result in a minor bump to amplify_core (since it is part of the same component) AND a patch bump to any package that depends on amplify_core (this was not happening)
 2. Do not update a dependency constraints in all dependent packages for non-breaking changes (unless the dependent is part of the same component and that component requires propagation)
    - ex: if amplify_auth_cognito_dart has a patch bump, the dependency constraints for amplify_auth_cognito_dart in amplify_auth_cognito do not need to be updated
3. Add a new test case for a multi package update with a breaking change in a common package to show how breaking and non breaking changes are treated differently.

### DIFF
--- a/packages/aft/lib/src/repo.dart
+++ b/packages/aft/lib/src/repo.dart
@@ -298,7 +298,6 @@ class Repo {
     required VersionBumpType type,
     required bool Function(PackageInfo) canBump,
     required bool includeInChangelog,
-    bool? propagateToComponent,
   }) {
     logger.verbose('bumpVersion ${package.name} $commit');
     final componentName = aftConfig.componentForPackage(package.name);
@@ -317,7 +316,7 @@ class Repo {
       ],
       (version) => version,
     )!;
-    propagateToComponent ??= component != null &&
+    final propagateToComponent = component != null &&
         component.propagate.propagateToComponent(
           currentVersion,
           newVersion,
@@ -359,11 +358,13 @@ class Repo {
             pkg.pubspecInfo.pubspec.dependencies.keys.contains(package.name) ||
             pkg.pubspecInfo.pubspec.devDependencies.keys.contains(package.name),
       );
-      if (commit.isBreakingChange) {
-        // Back-propagate to all dependent packages for breaking changes.
+      if (commit.isBreakingChange || propagateToComponent) {
+        // Back-propagate to all dependent packages for breaking changes or
+        // changes that need to propagate to a component.
         //
-        // Since we set semantic version constraints, only a breaking change
-        // in a direct dependency necessitates a version bump.
+        // Since we set semantic version constraints, only a breaking change in
+        // a direct dependency or a change that requires propagation
+        // necessitates a version bump.
         logger.verbose(
           'Breaking change. Performing dfs on dependent packages...',
         );
@@ -378,12 +379,6 @@ class Repo {
               includeInChangelog: false,
             );
           }
-          updateConstraint(package, dependent);
-        }
-      } else if (type == VersionBumpType.nonBreaking) {
-        // For non-breaking changes, we still need to update all constraints
-        // since we "pin" to minor versions.
-        for (final dependent in packageDependents) {
           updateConstraint(package, dependent);
         }
       }
@@ -408,7 +403,6 @@ class Repo {
               type: type,
               canBump: canBump,
               includeInChangelog: false,
-              propagateToComponent: false,
             );
           },
         );
--- a/packages/aft/test/version_bump/data/snapshots/multi_package_update.diff
+++ b/packages/aft/test/version_bump/data/snapshots/multi_package_update.diff
@@ -12,8 +12,6 @@ diff --git a/packages/amplify/amplify_flutter/pubspec.yaml b/packages/amplify/am
 +version: 2.4.0
 -  amplify_core: ">=2.3.0 <2.4.0"
 +  amplify_core: ">=2.4.0 <2.5.0"
--  aws_common: ">=0.7.1 <0.8.0"
-+  aws_common: ">=0.7.2 <0.8.0"
 diff --git a/packages/amplify_core/CHANGELOG.md b/packages/amplify_core/CHANGELOG.md
 +## 2.4.0
 +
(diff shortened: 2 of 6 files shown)


packages/aft/test/version_bump/data/snapshots/multi_package_update.diff
```diff
@@ -12,8 +12,6 @@ diff --git a/packages/amplify/amplify_flutter/pubspec.yaml b/packages/amplify/am
 +version: 2.4.0
 -  amplify_core: ">=2.3.0 <2.4.0"
 +  amplify_core: ">=2.4.0 <2.5.0"
--  aws_common: ">=0.7.1 <0.8.0"
-+  aws_common: ">=0.7.2 <0.8.0"
```
Jordan-Nelson, line 16: Note to reviewers: The constraints on aws_common do not need to be updated since this is not a breaking change. The existing version constraint of `">=0.7.1 <0.8.0"` is still valid (`">=0.7.0 <0.8.0"` would have also been valid). A customer can run pub upgrade to get the latest version of aws_common.

I have added a new test case (see .diff below) where aws_common receives a breaking change, bumping the range to `">=0.8.0 <0.9.0"`, which would result in an update to this file.

packages/aft/test/version_bump/data/snapshots/multi_package_update.diff
```diff
@@ -60,13 +56,16 @@ diff --git a/packages/analytics/amplify_analytics_pinpoint/pubspec.yaml b/packag
...
+diff --git a/packages/analytics/amplify_analytics_pinpoint_dart/CHANGELOG.md b/packages/analytics/amplify_analytics_pinpoint_dart/CHANGELOG.md
++## 0.4.4
++
++- Minor bug fixes and improvements
++
 diff --git a/packages/analytics/amplify_analytics_pinpoint_dart/pubspec.yaml b/packages/analytics/amplify_analytics_pinpoint_dart/pubspec.yaml
+-version: 0.4.3
++version: 0.4.4
```
Jordan-Nelson, line 66: Note to reviewers: This is an example of a bug being fixed where version bumps were not properly propagated. Since amplify_analytics_pinpoint_dart has a dependency constraint being updated, the package version needs to be updated as well.

packages/aft/test/version_bump/data/snapshots/single_dart_package_feat.diff
```diff
@@ -1,6 +1,3 @@
-diff --git a/packages/auth/amplify_auth_cognito/pubspec.yaml b/packages/auth/amplify_auth_cognito/pubspec.yaml
--  amplify_auth_cognito_dart: ">=0.11.3 <0.12.0"
-+  amplify_auth_cognito_dart: ">=0.11.4 <0.12.0"
```
Jordan-Nelson, line 3: Note to reviewers: This is another example of a bug being fixed where a dependency constraint is made when it does not need to. The existing version constraint (`">=0.11.3 <0.12.0"`) is valid. Prior to this PR, this diff would have been problematic since amplify_auth_cognito has a constraint update, but is not published.

packages/aft/lib/src/repo.dart
```diff
@@ -317,7 +316,7 @@ class Repo {
       ],
       (version) => version,
     )!;
-    propagateToComponent ??= component != null &&
+    final propagateToComponent = component != null &&
```
Equartey, line 319: Q: Can you provide context around the purpose of this flag? How is it set?

packages/aft/lib/src/repo.dart
```diff
@@ -380,12 +381,6 @@ class Repo {
           }
           updateConstraint(package, dependent);
         }
-      } else if (type == VersionBumpType.nonBreaking) {
-        // For non-breaking changes, we still need to update all constraints
-        // since we "pin" to minor versions.
-        for (final dependent in packageDependents) {
-          updateConstraint(package, dependent);
-        }
```
Equartey, line 388: Q: If I'm following this correctly, by omitting this we will no longer see constraints on the Dart packages updated from non-major bumps. Is that correct?

packages/aft/lib/src/repo.dart
Jordan-Nelson: aft defines components in the root pubspec.yaml. Below is an example of the "Amplify Flutter" component which includes all amplify packages intended for direct consumption in a Flutter app.

https://github.com/aws-amplify/amplify-flutter/blob/f825f86d68aba9d8e69eb934873cd212783bfde2/pubspec.yaml#L69-L82

By default, if a package has a minor/major bump, that bump should propagate to all other packages in the components. For example, if amplify_auth_cognito goes from 2.3.1 to 2.4.0, all packages in that component have to have their version bumped to 2.4.0.

Components can override the default behavior by specifying `propagate: "major" | "minor" | "all" | "none"` in the pubpsec.yaml file. The "Amplify Dart" component has this override. When amplify_auth_cognito_dart has a minor version bump, other packages in the group do not need to also be bumped.

https://github.com/aws-amplify/amplify-flutter/blob/f825f86d68aba9d8e69eb934873cd212783bfde2/pubspec.yaml#L83-L90

The `component.propagate.propagateToComponent()` fn determines if a change needs to propagate based on the component config in the root pubpsec file and the version update.

Prior to this PR, this fn accepted an optional `propagateToComponent` flag. I do not know what the intended purpose of this was, but it was preventing propagation from occurring more than 1 level deep. This is problematic. See the PR description for an example of an issue this caused.

packages/aft/lib/src/repo.dart
Jordan-Nelson: There is an example in the PR description that I think answers this question. Let me know if it doesn't.

packages/aft/lib/src/repo.dart
Equartey: Makes more sense now, thanks for providing those references!

Equartey (approved):

NikaHsn (approved):

PR #5453, opened 2024-09-13
chore(infra): Gen 1 Refresh

Equartey:
*Description of changes:*
- Updates Infra stack to achieve a clean deploy
- Removes old infra stacks (analytics & storage) that are now tested in Gen 2
- Fixed some type issues
- Lots of auto formatting changes.


khatruong2009 (approved):

infra/lib/auth/stack.ts
```diff
@@ -260,7 +260,6 @@ class AuthIntegrationTestStackEnvironment extends IntegrationTestStackEnvironmen
     const customSenderKmsKey = new kms.Key(this, "CustomSenderKey", {
       description: `Key for encrypting/decrypting SMS messages sent from ${this.name} user pool`,
       removalPolicy: RemovalPolicy.DESTROY,
-      alias: this.name,
```
NikaHsn, line 263: Q: don't we need the alias anymore?

infra/lib/common.ts
```diff
@@ -14,14 +14,20 @@ import * as secretsmanager from 'aws-cdk-lib/aws-secretsmanager';
...
   account: process.env.CDK_DEFAULT_ACCOUNT,
   region: 'us-west-2',
 };
 
+export enum StorageAccessLevel {
+  public = "guest",
+  protected = "protected",
+  private = "private",
+}
+
```
NikaHsn, line 30: I'm not sure if we need this anymore, we deleted the storage stack and hence we can remove storage config and also not adding it to the AmplifyConfig. we can also remove the analytic config and not add it to AmplifyConfig

infra/lib/auth/stack.ts
Equartey: This was causing my deployment to fail. Unsure what the cause was, but don't see the value in having defined. Also when this lifted to the gen 2 stack, it was dropped.

infra/lib/common.ts
Equartey: Good call out, I pushed a update removing the extra code.

khatruong2009 (approved):

NikaHsn (approved):

PR #5456, opened 2024-09-13
chore(auth): update auth plugin impl and config state machine to use AuthOutputs instead of Auth config

NikaHsn:
*Issue #, if available:*

*Description of changes:*
- Update auth plugin impl to remove the use of  CognitoUserPoolConfig  and CognitoIdentityCredentialsProvider instances from the dependency manager and instead use AuthOutputs instance from the dependency manager.
- Update the configuration state machine to remove the use of AuthConfiguration and not adding it and its properties to the dependency manager.
- Delete the auth_configuration.dart file 


packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
```diff
@@ -451,15 +449,16 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface
...
           ..analyticsMetadata = _analyticsMetadata?.toBuilder();
 
-        final clientSecret = _userPoolConfig.appClientSecret;
+        // ignore: invalid_use_of_internal_member
+        final clientSecret = _authOutputs.appClientSecret;
         if (clientSecret != null) {
           b.secretHash = computeSecretHash(
             username,
-            _userPoolConfig.appClientId,
+            _authOutputs.userPoolClientId!,
```
Equartey, line 461: Q: do we need a proper null check before force unwrapping `_authOutputs.userPoolClientId!`?

NikaHsn: we have a check when we get the _authOutputs from the dependency manager. it is line128

Equartey (approved):

khatruong2009 (approved):

PR #5476, opened 2024-09-17
feat(core): update storage outputs class for multi-bucket support

ekjotmultani:
added the list names buckets in storage_outputs.dart, built the new generator file for it, and updated the test to match the expected json output as described in the api review doc


packages/amplify_core/lib/src/config/amplify_outputs/storage/storage_outputs.dart
```diff
@@ -23,6 +23,9 @@ class StorageOutputs
   /// The Amazon S3 bucket name.
   final String bucketName;
 
+  /// The list of buckets if there are multiple buckets for the project
+  final List<Map<String,String>>? buckets;
```
NikaHsn, line 27: I suggest to define a class `Bucket` and use it instead of Map.

packages/amplify_core/lib/src/config/amplify_outputs/storage/storage_output_bucket.dart
```diff
@@ -0,0 +1,16 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+
+
+/// {@template amplify_core.amplify_outputs.amazon_pinpoint_channel}
+/// Supported channels for Amazon Pinpoint.
+/// {@endtemplate}
+class StorageOutputBucket {
```
NikaHsn, line 9:
```suggestion
class BucketOutputs {
```

packages/amplify_core/lib/src/config/amplify_outputs/storage/storage_output_bucket.dart
```diff
@@ -0,0 +1,16 @@
...
+  StorageOutputBucket(this.name, this.bucketName, this.awsRegion);
+  factory StorageOutputBucket.fromJson(Map<String, dynamic> json) => StorageOutputBucket(json['name'].toString(), json['bucket_name'].toString(), json['aws_region'].toString());
+  String name;
+  String bucketName;
+  String awsRegion;
+  Map<String, dynamic> toJson() => <String, dynamic>{'name':name, 'bucket_name':bucketName, 'aws_region':awsRegion};
+}
```
NikaHsn, line 16: instead of wrting the fromJson and toJson use generated code similar to other Amplify Outputs types.

packages/amplify_core/lib/src/config/amplify_outputs/storage/storage_output_bucket.dart
```diff
@@ -0,0 +1,16 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+
+
+/// {@template amplify_core.amplify_outputs.amazon_pinpoint_channel}
+/// Supported channels for Amazon Pinpoint.
+/// {@endtemplate}
+class StorageOutputBucket {
+  StorageOutputBucket(this.name, this.bucketName, this.awsRegion);
```
NikaHsn, line 10: the ctor can be const. I would suggest to use named parameter similat to other Amplify Outputs type

packages/amplify_core/doc/lib/auth.dart
ekjotmultani, line 1: I dont know how this got changed, I think I clicked a vscode 'optimization' by mistake, I will change this back right now

packages/amplify_core/lib/src/config/amplify_outputs/storage/bucket_output.dart
```diff
@@ -0,0 +1,37 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'package:amplify_core/amplify_core.dart';
+
+part 'bucket_output.g.dart';
+
+@zAmplifyOutputsSerializable
+class BucketOutput
```
NikaHsn, line 9:
```suggestion
class BucketOutputs
```

packages/amplify_core/lib/src/config/amplify_outputs/storage/bucket_output.dart
NikaHsn, line 1: missing doc comments https://dart.dev/effective-dart/documentation#doc-comments.

NikaHsn (approved):

packages/amplify_core/test/config/amplify_outputs/test_data.dart
Equartey, line 1: Question: I see the test data has been updated, but are the tests written in a way that validate the new functionality too? Do they need to be updated?

ekjotmultani: This is a good question, since multiple buckets are optional, I think we should validate that it works when only one bucket is supplied. I'll write a new test

ekjotmultani: I've verified that the test does work properly when only a single bucket is supplied and that the test itself actually validates the new function for multi-bucket support

Equartey (approved):

PR #5599, opened 2024-10-24
feat(storage): multi bucket download data

Equartey:
*Description of changes:*

- Updates `StorageDownloadDataOptions` to have a optional bucket param to specify target for operation
- Updates S3 impl to use bucket param
- Adds E2E tests

### DIFF
--- a/packages/storage/amplify_storage_s3/example/integration_test/download_data_test.dart
+++ b/packages/storage/amplify_storage_s3/example/integration_test/download_data_test.dart
@@ -137,6 +137,7 @@ void main() {
           final mainBucket =
               StorageBucket.fromOutputs('Storage Integ Test main bucket');
 
+          // TODO(equartey): Add download check for secondary bucket when upload supports multibucket
           final downloadResult = await Amplify.Storage.downloadData(
             path: StoragePath.fromIdentityId(
               (identityId) => 'private/$identityId/$identityName',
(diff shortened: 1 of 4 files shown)


packages/storage/amplify_storage_s3/example/integration_test/download_data_test.dart
```diff
@@ -132,6 +132,23 @@ void main() {
...
+
+        testWidgets('multi bucket', (_) async {
+          final mainBucket =
+              StorageBucket.fromOutputs('Storage Integ Test main bucket');
+
+          final downloadResult = await Amplify.Storage.downloadData(
+            path: StoragePath.fromIdentityId(
+              (identityId) => 'private/$identityId/$identityName',
+            ),
+            options: StorageDownloadDataOptions(bucket: mainBucket),
```
ekjotmultani, line 144: Q: should this be tested on the secondary bucket instead? otherwise this test targets the same bucket as the previous ones since they default to the main bucket with no bucket specified

Equartey: Good question, let me include a to-do to add the secondary bucket after upload file/data is merged. Right now there's no way to upload to the secondary bucket.

ekjotmultani (approved):

tyllark (approved):

PR #5656, opened 2024-11-06
feat(storage): multi bucket download file api

ekjotmultani:
*Description of changes:*

added integration tests
Updated download file options to accept bucket parameter


Equartey (approved):
LGTM. Are there any unit test that should be updated too?

packages/storage/amplify_storage_s3/example/integration_test/download_file_test.dart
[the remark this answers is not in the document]
ekjotmultani: This makes sense, I'll make a few separate ones under the standard config to test this then

packages/storage/amplify_storage_s3_dart/lib/src/amplify_storage_s3_dart_impl.dart
```diff
@@ -216,15 +216,15 @@ class AmplifyStorageS3Dart extends StoragePluginInterface
       defaultPluginOptions: const S3DownloadDataPluginOptions(),
     );
 
-    final s3Options = StorageDownloadDataOptions(
+    options = StorageDownloadDataOptions(
```
NikaHsn, line 219: Q: why to remove final?

ekjotmultani: it causes errors like `"Local variable 'options' can't be referenced before it is declared."` for using `options?.bucket` in the constructor, and another complaint of `The final variable 'options' can't be read because it's potentially unassigned at this point.`

ekjotmultani: it was final before since it was called s3Options, different variable name than the one that is passed into the function as a parameter as well

NikaHsn: I see the issue here, it is overriding the function input parameter and it is wrong to do so. Why to need to make this change? I would suggest to reveret it.

ekjotmultani: I did it to maintain consistency with the other functions options being called just `options` instead of `s3options.` I will revert it, I need to rebase with the most recent push to multi-bucket anyways

PR #5674, opened 2024-11-14
feat(storage): multi bucket copy

tyllark:
*Issue #, if available:*

*Description of changes:*

- Corrected toJson() for Storage Option classes
- Updated copy api to accept an optional buckets parameter
- Added tests

### DIFF
--- a/packages/amplify_core/lib/src/types/storage/copy_options.dart
+++ b/packages/amplify_core/lib/src/types/storage/copy_options.dart
@@ -1,7 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import 'package:aws_common/aws_common.dart';
+import 'package:amplify_core/amplify_core.dart';
 
 /// {@template amplify_core.storage.copy_options}
 /// Configurable options for `Amplify.Storage.copy`.
@@ -12,10 +12,16 @@ class StorageCopyOptions
         AWSSerializable<Map<String, Object?>>,
         AWSDebuggable {
   /// {@macro amplify_core.storage.copy_options}
-  const StorageCopyOptions({this.pluginOptions});
+  const StorageCopyOptions({
+    this.pluginOptions,
+    this.buckets,
+  });
 
   /// plugin specific options for `Amplify.Storage.copy`.
   final StorageCopyPluginOptions? pluginOptions;
+
+  /// Optionally specify which bucket to target
+  final CopyBuckets? buckets;
 
   @override
   List<Object?> get props => [pluginOptions];
@@ -26,6 +32,7 @@ class StorageCopyOptions
   @override
   Map<String, Object?> toJson() => {
         'pluginOptions': pluginOptions?.toJson(),
+        'buckets': buckets?.toJson(),
       };
 }
 
--- a/packages/amplify_core/lib/src/types/storage/copy_options.dart
+++ b/packages/amplify_core/lib/src/types/storage/copy_options.dart
@@ -20,11 +20,11 @@ class StorageCopyOptions
   /// plugin specific options for `Amplify.Storage.copy`.
   final StorageCopyPluginOptions? pluginOptions;
 
-  /// Optionally specify which bucket to target
+  /// Optionally specify which buckets to target
   final CopyBuckets? buckets;
 
   @override
-  List<Object?> get props => [pluginOptions];
+  List<Object?> get props => [pluginOptions, buckets];
 
   @override
   String get runtimeTypeName => 'StorageCopyOptions';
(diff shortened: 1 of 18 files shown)


packages/amplify_core/lib/src/types/storage/copy_options.dart
```diff
@@ -12,11 +12,17 @@ class StorageCopyOptions
...
-  const StorageCopyOptions({this.pluginOptions});
+  const StorageCopyOptions({
+    this.pluginOptions,
+    this.buckets,
+  });
 
   /// plugin specific options for `Amplify.Storage.copy`.
   final StorageCopyPluginOptions? pluginOptions;
 
+  /// Optionally specify which bucket to target
```
Equartey, line 23:
```suggestion
  /// Optionally specify which buckets to target
```

packages/amplify_core/lib/src/types/storage/copy_options.dart
```diff
@@ -12,11 +12,17 @@ class StorageCopyOptions
...
+  final CopyBuckets? buckets;
+
   @override
   List<Object?> get props => [pluginOptions];
```
Equartey, line 27:
```suggestion
  List<Object?> get props => [pluginOptions, buckets];
```

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
```diff
@@ -412,15 +412,19 @@ class StorageS3Service {
...
 
     final copyRequest = s3.CopyObjectRequest.build((builder) {
       builder
-        ..bucket = _storageOutputs.bucketName
-        ..copySource = '${_storageOutputs.bucketName}/$sourcePath'
+        ..bucket = s3ClientInfoDestination.bucketName
+        ..copySource = '${s3ClientInfoSource.bucketName}/$sourcePath'
         ..key = destinationPath
         ..metadataDirective = s3.MetadataDirective.copy;
     });
```
Equartey, line 430: Q: I think the code below this needs to be updated based on the destination bucket. For example the `_defaultS3Client` and `_storageOutputs.bucketName`. Can you double check this?

packages/amplify_core/lib/src/types/storage/copy_options.dart
tyllark: Good catch, I accidentally deleted this when updating the implementation to use CopyBuckets instead of 2 StorageBucket.

packages/amplify_core/lib/src/types/storage/copy_options.dart
```diff
@@ -12,11 +12,17 @@ class StorageCopyOptions
...
+  const StorageCopyOptions({
+    this.pluginOptions,
+    this.buckets,
+  });
 
   /// plugin specific options for `Amplify.Storage.copy`.
   final StorageCopyPluginOptions? pluginOptions;
 
+  /// Optionally specify which bucket to target
+  final CopyBuckets? buckets;
```
NikaHsn, line 24: Q: Does JS also use `CopyBuckets buckets` for copy API optinal param?

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/storage_s3_service_impl.dart
NikaHsn: +1

packages/amplify_core/lib/src/types/storage/copy_options.dart
tyllark: No they have a different [copy signature](https://docs.amplify.aws/nextjs/build-a-backend/storage/copy-files/#specify-a-bucket-or-copy-across-buckets--regions) that is more extensible:

- source
  - path
  - bucket
- destination
   - path
   - bucket

I prefer their signature over ours, but it would be a breaking change to wrap our source/destination parameters in an object, so I think we stick with the CopyBuckets approach.

Equartey (approved):

packages/amplify_core/lib/src/types/storage/storage_bucket.dart
```diff
@@ -16,4 +16,9 @@ class StorageBucket {
 
   @internal
   BucketInfo resolveBucketInfo(StorageOutputs? storageOutputs) => _info;
+
+  @override
+  Map<String, Object?> toJson() => {
+        '_info': _info.toJson(),
+      };
```
NikaHsn, line 23: how about to use `bucket?.resolveBucketInfo.toJson()` for the storage options instead of adding toJson() to `StorageBucket` and to `StorageBucketFromOutputs`?

PR #5741, opened 2024-12-02
chore(datastore,analytics,auth,notifications,storage): updated/regene…

tyllark:
*Issue #, if available:*

*Description of changes:*
1. Updated Pigeon to v22.6.2. Regenerated files and fixed breaking changes
-renamed setup to setUp in various locations
-added pigeonVar_binaryMessenger and pigeonVar_messageChannelSuffix to AmplifyNativeLegacyWrapper
-renamed PigeonLegacyDataProvider to PinpointLegacyDataProvider (API can’t start with Pigeon)
-updated signInWithUrl’s return value (this was updated 3 years ago)


packages/amplify/amplify_flutter/CHANGELOG.md
```diff
@@ -93,7 +93,7 @@ As always, you can find us on [GitHub](https://github.com/aws-amplify/amplify-fl
 ## 1.8.0
 
 NOTE: This version has been updated to:
-- Flutter: ">=3.19.0"
```
tyllark, line 96: Revert this change as we should not update old changelogs unless they are incorrect.

ekjotmultani (dismissed):

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/NativeAuthPluginWrapper.kt
```diff
@@ -85,45 +86,49 @@ class NativeAuthPluginWrapper(
...
+                    val authSession = FlutterFactory.createAWSCognitoAuthSession(
+                        session.isSignedIn,
+                        AuthSessionResult.success(session.identityId),
+                        awsCredentials,
+                        AuthSessionResult.success(session.userSub),
+                        userPoolTokens
+                    )
+                    onSuccess.accept(authSession)
+                } else {
+                    AuthSessionResult.failure<AWSCognitoUserPoolTokens>(UnknownException("Could not fetch"))
```
mattcreaser, line 130: Should this include the upstream error?

mattcreaser (commented):
I didn't look at the generated code but only had one question from a Kotlin perspective.

thisisabhash (approved):

packages/amplify_datastore/android/src/main/kotlin/com/amazonaws/amplify/amplify_datastore/NativeAuthPluginWrapper.kt
tyllark: Good callout. We are now passing the throwable message:

```
else {
    val error = UnknownException(result.exceptionOrNull()?.message ?: "Could not fetch")
    AuthSessionResult.failure<AWSCognitoUserPoolTokens>(error)
}
```

PR #5794, opened 2025-01-06
fix(authenticator): make phone number mandatory when "required" option is specified

ekjotmultani:
*Issue #, if available:* [#4079](https://github.com/aws-amplify/amplify-flutter/issues/4079)

*Description of changes:*

This is done in the exact same way as https://github.com/aws-amplify/amplify-flutter/pull/4106 from November last year. This fix was correct for the main issue in the original issue however it didn't address being able to re-input a dial code and still have it pass through validation

Truncate a re-inputted dial code if it exceeds the maximum length of a phone number from that given country
Using a map of dial codes to their countries' maximum length of phone number (generated from [this open source repo](https://github.com/google/libphonenumber/tree/master)), we will check if the inputted phone number begins with the selected dial code, if it does, then check if it surpasses the length of number in that country (since it could be the case that the number naturally begins with the dial code), if this is also true then ignore the repeat dial code and take the leftover number as the phone number

Credit: @Jordan-Nelson for most of the code


packages/authenticator/amplify_authenticator/lib/src/mixins/authenticator_username_field.dart
```diff
@@ -164,6 +165,58 @@ mixin AuthenticatorUsernameField<FieldType extends Enum,
     }
   }
 
+  String displayPhoneNumber(String? phoneNumber) {
```
godwinpang, line 168: why is this method defined 3 times in this file? I see there are small differences in the implementation, don't know if it is a Dart thing that I am not familiar with

godwinpang: also duplicated with https://github.com/aws-amplify/amplify-flutter/pull/5794/files#diff-3415b04b0444d866a70af8fcfe8cc0a4f4e5d47ee1e90c644c24916a84096c61R51, maybe worth moving to `lib/src/utils`?

packages/authenticator/amplify_authenticator/lib/src/mixins/authenticator_username_field.dart
```diff
@@ -183,7 +236,7 @@ mixin AuthenticatorUsernameField<FieldType extends Enum,
               isOptional: isOptional,
               context: context,
               inputResolver: stringResolver.inputs,
-            )(input?.username);
+            )(displayPhoneNumber(input?.username));
```
godwinpang, line 239: we are passing the username as a param into the phone number method? that doesn't seem right

packages/authenticator/amplify_authenticator/lib/src/mixins/authenticator_username_field.dart
AagamVadecha: Good to note that the comments seem copy pasted too

packages/authenticator/amplify_authenticator/lib/src/mixins/authenticator_username_field.dart
ekjotmultani: The username can be a phone number, and at that point in the code its been validated as such. We use AWS Cognito as our base provider for auth, [which accepts phone and email as usernames](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-attributes.html#user-pool-settings-usernames) if it has been configured that way.

packages/authenticator/amplify_authenticator/lib/src/mixins/authenticator_username_field.dart
ekjotmultani: this is because i accidentally copy pasted it too many times, it should be two different functions

packages/authenticator/amplify_authenticator/lib/src/mixins/authenticator_username_field.dart
godwinpang: TIL! thank you for explaining :)

PR #6159, opened 2025-06-03
chore(dev): static api analysis

ekjotmultani:
*Issue #, if available:*

*Description of changes:*

Introduced the `aft generate api-report` command to AFT. This command uses [dart_apitool](https://pub.dev/packages/dart_apitool) from @bmw-tech to create snapshots of our package's dart level API surfaces, and a change report for these. Developers must now always use this AFT command to generate these files for each package (just running the command as is once here works) on **every Pull Request** that is expected to merge into main.

This PR also includes a workflow for checking that these change reports are actually updated according to the changes on the PR branch, and a workflow for requiring additional approvals when an API change actually occurs (as shown by the lovely bot in this thread).

Overall this change to our development flow will enforce that developers are always aware of API changes to help us get the right eyes on it and to help with versioning.


packages/aft/lib/src/commands/generate/generate_api_report_command.dart
```diff
@@ -0,0 +1,208 @@
...
+      logger.info('Generating API JSON files for all packages...');
+      final failedPackages = <String>[];
+
+      // Filter packages to only include development packages (not examples or test packages)
+      var packagePaths = aftConfig.allPackages.values
+          .where((package) => 
+              package.isDevelopmentPackage && 
+              package.isPublishable &&
+              !package.isLintsPackage && 
+              !package.path.contains('_test'))
```
tyllark: Can we use the `isTestPackage` property for consistency ?

packages/aft/lib/src/commands/generate/generate_api_report_command.dart
```diff
@@ -0,0 +1,208 @@
...
+          .map((package) => package.path)
+          .toList();
+      
+      // Filter by specific packages if provided
+      final specificPackages = argResults?['packages'] as String?;
```
tyllark: (optional) This is logically correct, but I would remove the "if" check by assigning an empty string. I would also include this filter in the "where" clause above:
```
      final specificPackages = argResults?['packages'] as String? ?? '';
      final packageNamesToProcess = specificPackages
          .split(',')
          .map((s) => s.trim())
          .toSet();

      // Filter packages to only include specified development packages (not examples or test packages)
      final packagePaths = aftConfig.allPackages.values
          .where(
            (package) =>
                package.isDevelopmentPackage &&
                package.isPublishable &&
                !package.isLintsPackage &&
                !package.isTestPackage &&
                packageNamesToProcess.contains(package.name),
          )
          .map((package) => package.path)
          .toList();
```

packages/aft/lib/src/commands/generate/generate_api_report_command.dart
```diff
@@ -0,0 +1,208 @@
...
+    final pubspecFile = File(path.join(packagePath, 'pubspec.yaml'));
+    String packageName;
+    
+    if (pubspecFile.existsSync()) {
+      final pubspecContent = pubspecFile.readAsStringSync();
+      final nameMatch = RegExp(r'name:\s*([^\s]+)').firstMatch(pubspecContent);
+      packageName = nameMatch?.group(1) ?? path.basename(packagePath);
+    } else {
+      packageName = path.basename(packagePath);
+    }
```
tyllark, line 111: [pubspec_parse](https://pub.dev/packages/pubspec_parse) might be a better means of extracting the package name as the Regex could fail on a rogue comment or other data. Parsing all the YAML files would slow down the process and we do control the pubspec.yaml files. Let me know your thoughts.

packages/aft/lib/src/commands/generate/generate_api_report_command.dart
tyllark: Also we use "include" for specifying individual packages in other commands. See 'packages/aft/lib/src/options/glob_options.dart'

tyllark (changes requested):

packages/aft/lib/src/commands/generate/generate_api_report_command.dart
ekjotmultani: this should be taken care of by using the globbing mixin

.github/workflows/api-report-check.yaml
```diff
@@ -0,0 +1,118 @@
...
+                echo "The API report in your PR doesn't match the current code. Please run 'aft generate api-report' and commit the changes."
+                echo "Full diff for debugging:"
+                git diff "$file"
+                ACTUAL_API_CHANGES=true
+              else
+                echo "API report is up-to-date for: $file (only ordering differences)"
+              fi
+              
+              rm -f /tmp/pr_interfaces.txt /tmp/generated_interfaces.txt /tmp/pr_sorted.txt /tmp/generated_sorted.txt
+            done
```
tyllark, line 83: Can we remove the sorting here since we are now doing it within AFT.

tyllark (changes requested):
Can you update the AFT ReadMe to reference this new command.

Issue #6205, opened 2025-07-11
Amplify.Auth.signOut() Fails to Sign Out Locally with UnknownServiceException in Flutter


AntoniKwok:
### Description

When calling Amplify.Auth.signOut() in the Flutter app, the sign-out process intermittently fails with an UnknownServiceException and "signedOutLocally": false. The exception message is "An unknown error occurred" with an underlying exception of UnknownSmithyHttpException. This results in the user not being signed out locally, causing further authentication and authorization errors (e.g., 403 responses, malformed authorization headers) until the app is restarted.

``` Amplify.Auth.signOut().then((result) {
  if (result is CognitoSignOutResult) {
    if (result.signedOutLocally) {
      safePrint('Signed out locally.');
    } else {
      safePrint('Sign out failed: ${result.exception}');
    }
  }
  loading.value = false;
}).catchError((e) {
  safePrint('Error signing out: $e');
  AuthEvents.triggerForceLogout('Force Logout from signOut');
}); ```

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [ ] Auth
- [ ] Authenticator
- [ ] DataStore
- [ ] Notifications (Push)
- [ ] Storage

### Steps to Reproduce

I tried to implement force logout when session expired, and when force logout triggered the logout didn't work properly which can't re-login

### Screenshots

_No response_

### Platforms

- [x] iOS
- [x] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Flutter Version

3.24.5

### Amplify Flutter Version

1.7.0

### Deployment Method

Amplify CLI (Gen 1)

### Schema

```GraphQL

```

AntoniKwok:
Additional comment : The session became expired, it's because we enabled the refresh token setting in AWS dashboard. After the setting enabled, then the session is expired and failed to logout

harsh62:
@AntoniKwok Can you elaborate the following statement:
> it's because we enabled the refresh token setting in AWS dashboard.

Are you trying to use the new refresh token feature that Cognito has released? If yes, it is not supported in Amplify yet.

AntoniKwok:
> Are you trying to use the new refresh token feature that Cognito has released? If yes, it is not supported in Amplify yet.

Hi @harsh62 , thanks for the response!

[...]

PR #6226, opened 2025-07-15
fix(auth): handle fallthrough exceptions in sign out state

ekjotmultani:
*Issue #, if available:*
#6205 

*Description of changes:*

_onInitiate in our sign out state machine function catches only two exceptions, we add a catch all that also clears the credential store in this PR to handle other cases. Notably this allows users to exit a bad state where they would be unable to sign in or out properly if they have refresh token rotation enabled (linked issue), which causes the `REFRESH_TOKEN_AUTH` flow to be disabled.

### DIFF
--- a/packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_out_state_machine.dart
+++ b/packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_out_state_machine.dart
@@ -88,7 +88,7 @@ final class SignOutStateMachine
       await manager.clearCredentials();
       return emit(const SignOutState.success());
     } on Exception catch (e) {
-      // on any other exception, we should clear credentials and rethrow the exception
+      // unable to read tokens, clear the credentials to clear this invalid state.
       invalidTokenException = InvalidTokenException(underlyingException: e);
       await dispatchAndComplete(const CredentialStoreEvent.clearCredentials());
       return emit(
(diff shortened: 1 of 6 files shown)


packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_out_state_machine.dart
```diff
@@ -80,6 +80,10 @@ final class SignOutStateMachine
       // to clear the credentials associated with the non-existent user.
       await manager.clearCredentials();
       return emit(const SignOutState.success());
+    } on Exception catch (e) {
+      // on any other exception, we should clear credentials and rethrow the exception
+      await dispatchAndComplete(const CredentialStoreEvent.clearCredentials());
+      return emit(SignOutFailure(e));
```
tyllark: Since we are deleting the credential cache this we should return an `SignOutPartialFailure` rather than an `SignOutFailure`. We should also match the rest of the function and use the SignOutState factory constructors when we emit a state update `return emit(SignOutState.partialFailure());`.

none of the partialFailure parameter errors `HostedUiException`, `GlobalSignOutException`, and `RevokeTokenException` apply to this use case. We should add a new exception type here such as `InvalidTokenException` to help indicate there is a problem with the cached tokens.

tyllark (changes requested):

packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_out_state_machine.dart
```diff
@@ -80,13 +87,20 @@ final class SignOutStateMachine
       // to clear the credentials associated with the non-existent user.
       await manager.clearCredentials();
       return emit(const SignOutState.success());
+    } on Exception catch (e) {
+      // on any other exception, we should clear credentials and rethrow the exception
```
tyllark: We are no longer rethrowing the exception. I would write:

>// unable to read tokens, clear the credentials to clear this invalid state.

tyllark (approved):

tyllark (approved):

PR #6356, opened 2025-09-17
chore(aws_common): updated zIsWeb check to support WASM

tyllark:
*Issue #, if available:*

*Description of changes:*
updated zIsWeb check to support WASM


cadivus:
WASM support is already added

PR #6398, opened 2025-09-30
Feat/passwordless password

tyllark:
*Description of changes:*
Added Password and Password SRP first factor

### DIFF
--- a/packages/authenticator/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
+++ b/packages/authenticator/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
@@ -262,14 +262,16 @@ class StateMachineBloc
           }
           yield* _checkUserVerification();
         case AuthSignInStep.continueSignInWithFirstFactorSelection:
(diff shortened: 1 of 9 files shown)


packages/authenticator/amplify_authenticator/lib/src/blocs/auth/auth_bloc.dart
```diff
@@ -262,14 +262,16 @@ class StateMachineBloc
           }
           yield* _checkUserVerification();
         case AuthSignInStep.continueSignInWithFirstFactorSelection:
```
ekjotmultani, line 264: are these cases meant to just fall through eachother?

tyllark: Yes, while no user should be getting into this state we want these new steps to display an error if they are achieved. When we implement the Passwordless Authenticator changes they will have individual behavior. This aligns with the current Amplify Swift/Android behavior.

ekjotmultani (approved):

PR #6424, opened 2025-11-04
feat(authenticator): Add TextEditingController support to form fields

### LINKED ISSUE #6161: [Feature request] Enhancing Controller Support for Form Fields in Amplify Authenticator UI
### On which framework/platform would you like to see this feature implemented?

Flutter

### Which UI component is this feature-request for?

Authenticator

### Please describe your feature-request in detail.

Currently, the Amplify Authenticator UI for Flutter does not allow programmatic modification of form field values through a controller—unlike other conventional form fields. This limitation restricts use cases such as pre-populating a field with data retrieved from an API (for example, automatically setting a user’s address based on GPS coordinates) or auto-filling an SMS verification code.

To improve functionality and flexibility, it is recommended that the Authenticator UI Library, particularly the SignUpFormField, be extended to include controller support. This enhancement would enable developers to seamlessly manage form values, thereby expanding the component’s utility and improving overall user experience.

### Please describe a solution you'd like.

Create a new controller class for the fields used in the Auth lib, and expose it on the form field.

angelorodem:
This PR adds `TextEditingController` support to Amplify Authenticator form fields, enabling developers to programmatically control text input fields during authentication flows. It also introduces per-field visibility and enabled-state overrides for Sign Up fields so apps can keep legacy attributes off-screen or block user edits while still syncing values via controllers.

cc: @tyllark 

### What was changed
- Added new `AuthenticatorTextFieldController` class that wraps Flutter's `TextEditingController` with identical semantics.
- Added optional `controller` parameter to all text-based form fields across all authentication steps:
  - SignIn (username, password)
  - SignUp (username, email, password, phone number, address, and all custom attributes)
  - ConfirmSignIn (verification code, new password, custom auth)
  - ConfirmSignUp (username, verification code)
  - ResetPassword (username, verification code, new password)
  - VerifyUser (verification code)
  - EmailSetup (email)
- Added `visible` (default `true`) and `enabled` overrides to every `SignUpFormField` factory so applications can hide fields while keeping controller updates, or disable manual editing for attributes filled from background processes.
- Implemented bidirectional synchronization between controllers and internal authenticator state, including post-frame buffering to avoid setState-during-build issues.
- Added proper lifecycle management with post-frame callbacks to ensure correct sync timing and included diagnostics exposure for `controller`, `enabled`, and `visible` values.

### Why was this change made
Addresses the need for programmatic control over authenticator form fields to enable use cases such as:
- Pre-populating fields with data from external sources (GPS coordinates, API responses)
- Auto-filling verification codes from SMS or other sources
- Integration with third-party form validation libraries
- Dynamic form manipulation based on business logic
- Keeping legacy/system attributes hidden from end users while still providing values
- Disabling fields that should not accept manual input when values are provided automatically

### How was this change tested
- Added comprehensive unit/widget test suite with 13 test cases covering:
  - Bidirectional sync between controller and state
  - Multiple controllers on different fields
  - Rapid text updates and special key handling
  - Standard `TextEditingController` compatibility
  - Initial value population from controllers
  - Controller disposal and lifecycle management
  - Hidden and disabled Sign Up fields continuing to sync via controllers without user interaction
- All existing tests pass (301 tests)
- Created example app (`authenticator_with_controllers.dart`) demonstrating real-world usage patterns

## Related Issues
Fixes #6161

## Migration Guide
This is a non-breaking change. Existing code continues to work without modifications. To use the new controller functionality:

```dart
// Create a controller
final usernameController = AuthenticatorTextFieldController(text: 'initial_value');

// Pass it to a form field
Authenticator(
    fields: [
      SignUpFormField.username(
        controller: usernameController,
        enabled: false, // optional
        visible: false, // optional
      ),
      // ... other fields
    ],
  ),
  // ...
);

// Update programmatically
usernameController.text = 'new_value';

// Dispose when done
usernameController.dispose();
```

Standard `TextEditingController` is also supported for flexibility.

## Screenshots/Videos
https://github.com/user-attachments/assets/2d3d730f-9555-4f4c-8f80-cf38582ff8a3

---

**By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.**

### DIFF
--- a/packages/authenticator/amplify_authenticator/CHANGELOG.md
+++ b/packages/authenticator/amplify_authenticator/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.3.9
+
+- Minor bug fixes and improvements
+
 ## 2.3.8
 
 ### Chores
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -13,7 +13,7 @@
     "packages/smithy/goldens/lib",
     "packages/smithy/goldens/lib2"
   ],
-  "dart.onlyAnalyzeProjectsWithOpenFiles": true,
+  "dart.onlyAnalyzeProjectsWithOpenFiles": false,
   "files.insertFinalNewline": true,
   "editor.codeActionsOnSave": {
     "source.fixAll": "explicit",
--- a/packages/amplify/amplify_flutter/CHANGELOG.md
+++ b/packages/amplify/amplify_flutter/CHANGELOG.md
@@ -1,3 +1,22 @@
+## 2.8.0
+
+### Features
+- feat(auth): add support for cognito oidc parameters in managed login ([#6442](https://github.com/aws-amplify/amplify-flutter/pull/6442))
+
+### Chores
+- chore(all): bumped analyzer to ^9.0.0 ([#6457](https://github.com/aws-amplify/amplify-flutter/pull/6457))
+- chore(all): bumped built_collection to ^5.1.1 ([#6457](https://github.com/aws-amplify/amplify-flutter/pull/6457))
+- chore(all): bumped collection to ^1.18.0 ([#6457](https://github.com/aws-amplify/amplify-flutter/pull/6457))
+- chore(all): bumped crypto: ^3.0.7 ([#6457](https://github.com/aws-amplify/amplify-flutter/pull/6457))
+- chore(all): bumped grpc to ^5.0.0 ([#6457](https://github.com/aws-amplify/amplify-flutter/pull/6457))
+- chore(all): bumped json_serializable to ^6.11.0 ([#6457](https://github.com/aws-amplify/amplify-flutter/pull/6457))
+- chore(all): bumped meta to ^1.16.0 ([#6457](https://github.com/aws-amplify/amplify-flutter/pull/6457))
+- chore(all): bumped pigeon to ^26.0.0 ([#6457](https://github.com/aws-amplify/amplify-flutter/pull/6457))
+- chore(all): bumped protobuf to ^5.0.0 ([#6457](https://github.com/aws-amplify/amplify-flutter/pull/6457))
+- chore(all): bumped shelf to ^1.4.0 ([#6457](https://github.com/aws-amplify/amplify-flutter/pull/6457))
+- chore(all): bumped source_gen to ^4.0.0 ([#6457](https://github.com/aws-amplify/amplify-flutter/pull/6457))
+- chore(all): bumped stream_transform to ^2.1.0 ([#6457](https://github.com/aws-amplify/amplify-flutter/pull/6457))
+
 ## 2.7.0
 
 ### Features
--- a/packages/amplify/amplify_flutter/pubspec.yaml
+++ b/packages/amplify/amplify_flutter/pubspec.yaml
@@ -1,6 +1,6 @@
 name: amplify_flutter
 description: The top level Flutter package for the AWS Amplify libraries.
-version: 2.7.0
+version: 2.8.0
 homepage: https://docs.amplify.aws/lib/q/platform/flutter/
 repository: https://github.com/aws-amplify/amplify-flutter/tree/main/packages/amplify/amplify_flutter
 issue_tracker: https://github.com/aws-amplify/amplify-flutter/issues
@@ -19,9 +19,9 @@ platforms:
   web:
 
 dependencies:
-  amplify_core: ">=2.7.0 <2.8.0"
-  amplify_secure_storage: ">=0.5.13 <0.6.0"
-  aws_common: ">=0.7.11 <0.8.0"
+  amplify_core: ">=2.8.0 <2.9.0"
+  amplify_secure_storage: ">=0.5.14 <0.6.0"
+  aws_common: ">=0.7.12 <0.8.0"
   collection: ^1.18.0
   flutter:
     sdk: flutter
--- a/packages/amplify_core/CHANGELOG.md
+++ b/packages/amplify_core/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 2.8.0
+
+### Features
+- feat(auth): add support for cognito oidc parameters in managed login ([#6442](https://github.com/aws-amplify/amplify-flutter/pull/6442))
+
 ## 2.7.0
 
 ### Features
--- a/packages/amplify_core/lib/src/config/amplify_config.g.dart
+++ b/packages/amplify_core/lib/src/config/amplify_config.g.dart
@@ -57,10 +57,9 @@ Map<String, dynamic> _$AmplifyConfigToJson(AmplifyConfig instance) =>
     <String, dynamic>{
       'UserAgent': instance.userAgent,
       'Version': instance.version,
-      if (instance.api?.toJson() case final value?) 'api': value,
-      if (instance.analytics?.toJson() case final value?) 'analytics': value,
-      if (instance.auth?.toJson() case final value?) 'auth': value,
-      if (instance.notifications?.toJson() case final value?)
-        'notifications': value,
-      if (instance.storage?.toJson() case final value?) 'storage': value,
+      'api': ?instance.api?.toJson(),
+      'analytics': ?instance.analytics?.toJson(),
+      'auth': ?instance.auth?.toJson(),
+      'notifications': ?instance.notifications?.toJson(),
+      'storage': ?instance.storage?.toJson(),
     };
--- a/packages/amplify_core/lib/src/config/amplify_outputs/amplify_outputs.g.dart
+++ b/packages/amplify_core/lib/src/config/amplify_outputs/amplify_outputs.g.dart
@@ -53,13 +53,12 @@ AmplifyOutputs _$AmplifyOutputsFromJson(
 
 Map<String, dynamic> _$AmplifyOutputsToJson(AmplifyOutputs instance) =>
     <String, dynamic>{
-      if (instance.schema case final value?) 'schema': value,
+      'schema': ?instance.schema,
       'version': instance.version,
-      if (instance.analytics?.toJson() case final value?) 'analytics': value,
-      if (instance.auth?.toJson() case final value?) 'auth': value,
-      if (_dataToJson(instance.data) case final value?) 'data': value,
-      if (instance.notifications?.toJson() case final value?)
-        'notifications': value,
-      if (instance.storage?.toJson() case final value?) 'storage': value,
-      if (instance.custom case final value?) 'custom': value,
+      'analytics': ?instance.analytics?.toJson(),
+      'auth': ?instance.auth?.toJson(),
+      'data': ?_dataToJson(instance.data),
+      'notifications': ?instance.notifications?.toJson(),
+      'storage': ?instance.storage?.toJson(),
+      'custom': ?instance.custom,
     };
--- a/packages/amplify_core/lib/src/config/amplify_outputs/analytics/analytics_outputs.g.dart
+++ b/packages/amplify_core/lib/src/config/amplify_outputs/analytics/analytics_outputs.g.dart
@@ -22,7 +22,4 @@ AnalyticsOutputs _$AnalyticsOutputsFromJson(Map<String, dynamic> json) =>
     }, fieldKeyMap: const {'amazonPinpoint': 'amazon_pinpoint'});
 
 Map<String, dynamic> _$AnalyticsOutputsToJson(AnalyticsOutputs instance) =>
-    <String, dynamic>{
-      if (instance.amazonPinpoint?.toJson() case final value?)
-        'amazon_pinpoint': value,
-    };
+    <String, dynamic>{'amazon_pinpoint': ?instance.amazonPinpoint?.toJson()};
--- a/packages/amplify_core/lib/src/config/amplify_outputs/auth/auth_outputs.g.dart
+++ b/packages/amplify_core/lib/src/config/amplify_outputs/auth/auth_outputs.g.dart
@@ -98,40 +98,31 @@ AuthOutputs _$AuthOutputsFromJson(Map<String, dynamic> json) => $checkedCreate(
   },
 );
 
-Map<String, dynamic> _$AuthOutputsToJson(
-  AuthOutputs instance,
-) => <String, dynamic>{
-  'aws_region': instance.awsRegion,
-  if (instance.userPoolId case final value?) 'user_pool_id': value,
-  if (instance.userPoolClientId case final value?) 'user_pool_client_id': value,
-  if (instance.appClientSecret case final value?) 'app_client_secret': value,
-  if (instance.identityPoolId case final value?) 'identity_pool_id': value,
-  if (instance.passwordPolicy?.toJson() case final value?)
-    'password_policy': value,
-  if (instance.oauth?.toJson() case final value?) 'oauth': value,
-  if (instance.standardRequiredAttributes
+Map<String, dynamic> _$AuthOutputsToJson(AuthOutputs instance) =>
+    <String, dynamic>{
+      'aws_region': instance.awsRegion,
+      'user_pool_id': ?instance.userPoolId,
+      'user_pool_client_id': ?instance.userPoolClientId,
+      'app_client_secret': ?instance.appClientSecret,
+      'identity_pool_id': ?instance.identityPoolId,
+      'password_policy': ?instance.passwordPolicy?.toJson(),
+      'oauth': ?instance.oauth?.toJson(),
+      'standard_required_attributes': ?instance.standardRequiredAttributes
           ?.map(const CognitoUserAttributeKeyConverter().toJson)
-          .toList()
-      case final value?)
-    'standard_required_attributes': value,
-  if (instance.usernameAttributes
+          .toList(),
+      'username_attributes': ?instance.usernameAttributes
           ?.map(const CognitoUserAttributeKeyConverter().toJson)
-          .toList()
-      case final value?)
-    'username_attributes': value,
-  if (instance.userVerificationTypes
+          .toList(),
+      'user_verification_types': ?instance.userVerificationTypes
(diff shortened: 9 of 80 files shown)


angelorodem:
@tyllark could you please review?

angelorodem:
Here is a real world usage of this PR, it also uses the hide functionality to hide and automatically fill the Address field of aws cognito using the other address info available on the form.

https://github.com/user-attachments/assets/6c3e82e6-55fb-4bfa-bf92-fdf537b3c374

angelorodem:
@tyllark @ekjotmultani can you guys please review?

packages/authenticator/amplify_authenticator/CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
...
+### Features
+
+- feat(authenticator): Add TextEditingController support to form fields
+  - Added `AuthenticatorTextFieldController` class for programmatic control of form fields
+  - All text-based form fields now accept an optional `controller` parameter
+  - Enables pre-populating fields (e.g., from GPS/API data) and auto-filling verification codes
+  - Bidirectional sync between controller and internal state
+  - Compatible with standard `TextEditingController` for flexibility
+- feat(authenticator): Allow SignUpFormField inputs to be disabled or hidden so apps can prefill values programmatically or keep legacy attributes off-screen
+
```
tyllark: Please remove this change as we don't know the version number this will be released in. We have a package `aft` for preparing releases and it will generate the CHANGELOG updates from commit messages.

tyllark: Please remove this change as we don't know the version number this will be released in. We have a package `aft` for preparing releases and it will generate the CHANGELOG updates from commit messages.

packages/authenticator/amplify_authenticator/lib/src/mixins/authenticator_text_field.dart
```diff
@@ -4,33 +4,164 @@
...
+    super.dispose();
+  }
+
   @override
   Widget buildFormField(BuildContext context) {
     final inputResolver = stringResolver.inputs;
     final hintText = widget.hintText == null
         ? widget.hintTextKey?.resolve(context, inputResolver)
         : widget.hintText!;
+    _maybeUpdateEffectiveController();
```
tyllark: Ideally we shouldn't update the state inside the build function. I also think we can potentially reduce the overall complexity of the changes in this class, but I'll need to do some additional experimentation before I can say for certain.

packages/authenticator/amplify_authenticator/lib/src/widgets/form_field.dart
```diff
@@ -96,6 +98,20 @@ abstract class AuthenticatorFormField<
   /// Autocomplete hints to override the default value
   final Iterable<String>? autofillHints;
 
+  /// Optional text controller exposed by text-driven form fields.
+  TextEditingController? get controller => null;
```
tyllark: We like the `AuthenticatorTextFieldController` class you provided and would like to include that here in addition to all the form field constructors. This will minimize breaking changes if we need to make any changes to the underlying `TextEditingController` behavior.

packages/authenticator/amplify_authenticator/lib/src/widgets/form_field.dart
```diff
@@ -96,6 +98,20 @@ abstract class AuthenticatorFormField<
...
+
+  /// Whether the field can receive manual input.
+  ///
+  /// When `null`, the widget decides its enabled state.
+  final bool? enabledOverride;
```
tyllark: We think an enumeration called `AuthenticatorTextEnabledOverride` would work best here with the options `defaultSetting`, ~~`enabled`~~, and `disabled`. We are actually on the fence about including `enabled` as we have some fields that we intentionally disable to prevent Auth flows from breaking (such as disabling but displaying `username` throughout the signup flow). Developers should only be using `disabled` for fields that will be populated programmatically and `defaultSetting` for everything else.

packages/authenticator/amplify_authenticator/lib/src/widgets/form_field.dart
tyllark: Additionally we should name this property `authenticatorTextFieldController` instead of `controller`. I know `TextEditingController` uses the name `controller`, but we think it is better to have a more precise name.  This change should als be made for all the FormField named constructors as well.

packages/authenticator/amplify_authenticator/CHANGELOG.md
angelorodem: ok! removed

packages/authenticator/amplify_authenticator/lib/src/mixins/authenticator_text_field.dart
angelorodem: I tried to simplify and/or remove from the update cycle, but i did not see how else i could change it, maybe i could use a hand on how to do this.

packages/authenticator/amplify_authenticator/lib/src/widgets/form_field.dart
angelorodem: Changed, let me know if this is what you meant.

packages/authenticator/amplify_authenticator/lib/src/widgets/form_field.dart
angelorodem: Changed, please check if the new code helps,
though i still kept the enabled enum state, let me know if i should remove it, i think it is still important to have it.

angelorodem:
i'll be posting the changes shortly, double checking

angelorodem:
@tyllark  I've addressed most of your feedback, including the API naming changes and the controller class structure.

Regarding the update cycle and complexity: I explored alternative approaches to avoid state updates during the build phase, but the current implementation (leveraging post-frame callbacks) proved to be the most robust way to handle bidirectional synchronization without race conditions. I wasn't able to simplify it further without reintroducing issues.

packages/authenticator/amplify_authenticator/lib/src/mixins/authenticator_text_field.dart
```diff
@@ -19,7 +19,8 @@ mixin AuthenticatorTextField<
   bool _controllerUpdateScheduled = false;
 
   @protected
-  TextEditingController? get textController => null;
+  TextEditingController? get textController =>
```
tyllark: This type should also be `AuthenticatorTextFieldController`

packages/authenticator/amplify_authenticator/lib/src/mixins/authenticator_username_field.dart
```diff
@@ -24,7 +24,8 @@ mixin AuthenticatorUsernameField<
   bool _controllerUpdateScheduled = false;
 
   @protected
-  TextEditingController? get textController => null;
+  TextEditingController? get textController =>
```
tyllark: This type should also be `AuthenticatorTextFieldController`.

packages/authenticator/amplify_authenticator/lib/src/mixins/authenticator_username_field.dart
```diff
@@ -366,11 +398,14 @@ mixin AuthenticatorUsernameField<
...
-        enabled: enabled,
+        enabled: enabled
+            ? AuthenticatorTextEnabledOverride.enabled
+            : AuthenticatorTextEnabledOverride.disabled,
         errorMaxLines: errorMaxLines,
         initialValue: state.username,
         autofillHints: autofillHints,
-        controller: textController,
+        authenticatorTextFieldController:
+            textController as AuthenticatorTextFieldController?,
```
tyllark: We should be able to remove the casting once the type is updated.

packages/authenticator/amplify_authenticator/lib/src/enums/authenticator_text_enabled_override.dart
```diff
@@ -0,0 +1,14 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/// Enum for overriding the enabled state of a form field.
+enum AuthenticatorTextEnabledOverride {
+  /// Use the default enabled state.
+  defaultSetting,
+
+  /// Force the field to be enabled.
+  enabled,
```
tyllark: Sorry if I wasn't clear with my previous comment. I think we should remove `enabled`. If the default setting causes is disabling the field it is because the field is not allowed to be modified at that time otherwise the Auth flow could break. Let me know if you have a use case that requires `enabled`.

packages/authenticator/amplify_authenticator/lib/src/mixins/authenticator_username_field.dart
```diff
@@ -366,11 +398,14 @@ mixin AuthenticatorUsernameField<
         requiredOverride: true,
         onChanged: handleChanged,
         validator: validator,
-        enabled: enabled,
+        enabled: enabled
+            ? AuthenticatorTextEnabledOverride.enabled
+            : AuthenticatorTextEnabledOverride.disabled,
```
tyllark: We should retain the original `AuthenticatorTextEnabledOverride` value rather than converting it to a `bool` and back. You might need to add a getter to `AuthenticatorFormFieldState`

packages/authenticator/amplify_authenticator/lib/src/enums/authenticator_text_enabled_override.dart
tyllark: Currently `confirm_sign_up_form_field.dart` is the only example I'm aware of  where we disable form fields.

packages/authenticator/amplify_authenticator/lib/src/widgets/form_field.dart
```diff
@@ -187,7 +196,17 @@ abstract class AuthenticatorFormFieldState<
...
+    switch (widget.enabledOverride) {
+      case AuthenticatorTextEnabledOverride.enabled:
+        return true;
+      case AuthenticatorTextEnabledOverride.disabled:
+        return false;
+      case AuthenticatorTextEnabledOverride.defaultSetting:
+      case null:
+        return true;
+    }
+  }
```
tyllark: If a subclass overrides this property the logic will be lost unless. I think we should introduce a new property (`effectiveEnabled` or `isEnabled`) that combines these 2 properties. For example:
```
bool get effectiveEnabled {
    switch (widget.enabledOverride) {
      case AuthenticatorTextEnabledOverride.disabled:
        return false;
      case AuthenticatorTextEnabledOverride.defaultSetting:
      case null:
        return enabled;
    }
  }
```

packages/authenticator/amplify_authenticator/lib/src/widgets/form_fields/phone_number_field.dart
```diff
@@ -77,10 +82,17 @@ class _AuthenticatorPhoneFieldState<FieldType extends Enum>
...
+    switch (widget.enabled) {
+      case AuthenticatorTextEnabledOverride.enabled:
+        return true;
+      case AuthenticatorTextEnabledOverride.disabled:
+        return false;
+      case AuthenticatorTextEnabledOverride.defaultSetting:
+      case null:
+        return super.enabled;
+    }
+  }
```
tyllark: I believe we can revert this change with the `effectiveEnabled` property I suggested.

packages/authenticator/amplify_authenticator/lib/src/widgets/form_fields/confirm_sign_up_form_field.dart
```diff
@@ -94,9 +94,14 @@ abstract class _ConfirmSignUpFormFieldState<FieldValue extends Object>
...
-    if (override != null) {
-      return override;
+    switch (widget.enabledOverride) {
+      case AuthenticatorTextEnabledOverride.enabled:
+        return true;
+      case AuthenticatorTextEnabledOverride.disabled:
+        return false;
+      case AuthenticatorTextEnabledOverride.defaultSetting:
+      case null:
+        break;
```
tyllark: I believe we can revert this change with the `effectiveEnabled` property I suggested.

packages/authenticator/amplify_authenticator/lib/src/enums/authenticator_text_enabled_override.dart
angelorodem: changed.

packages/authenticator/amplify_authenticator/lib/src/mixins/authenticator_text_field.dart
angelorodem: fixed

packages/authenticator/amplify_authenticator/lib/src/mixins/authenticator_username_field.dart
angelorodem: fixed.

packages/authenticator/amplify_authenticator/lib/src/widgets/form_fields/confirm_sign_up_form_field.dart
angelorodem: reverted.

packages/authenticator/amplify_authenticator/lib/src/widgets/form_field.dart
angelorodem: changed.

packages/authenticator/amplify_authenticator/lib/src/widgets/form_fields/phone_number_field.dart
angelorodem: changed.

angelorodem:
@tyllark, i've applied the changes as you mentioned, please let me know if there is any other needed changes.

tyllark:
Hello @angelorodem, thank you for making the updates. I'm good with the changes and will approve once all the workflows pass. Currently the analyzer is giving the following errors:

>info • The referenced name isn't visible in scope • lib/src/widgets/form_field.dart:201:61 • comment_references
   info • Unnecessary duplication of receiver • lib/src/widgets/form_field.dart:375:5 • cascade_invocations
   info • Unnecessary duplication of receiver • test/form_field_controller_test.dart:467:7 • cascade_invocations

angelorodem:
@tyllark should be g2g, waiting on workflows

tyllark:
@angelorodem It looks like we are still getting the following error:
>info • The referenced name isn't visible in scope • lib/src/widgets/form_field.dart:201:61 • comment_references

I think we should change `[enabledOverride]` to `[AuthenticatorFormField.enabledOverride]`.

angelorodem:
> @angelorodem It looks like we are still getting the following error:
> 
> > info • The referenced name isn't visible in scope • lib/src/widgets/form_field.dart:201:61 • comment_references
> 
> I think we should change `[enabledOverride]` to `[AuthenticatorFormField.enabledOverride]`.

ok, will fix again, sorry i do not get these linting issues on my end, i'll be submiting the fix shortly

angelorodem:
@tyllark lmk if it's good

angelorodem:
@tyllark seems like it passed

tyllark:
Hi @angelorodem, could you please update the branch (otherwise the commit will be associated with me). I've asked @ekjotmultani to take a quick glance, but it looks like we should be good to go!

angelorodem:
@tyllark Branch updated, let me know if there is anything else

angelorodem:
@tyllark i think it is ready, any other steps? let me know, 
thanks.

packages/authenticator/amplify_authenticator/lib/src/widgets/form_fields/sign_up_form_field.dart
```diff
@@ -23,8 +23,10 @@ abstract class SignUpFormField<FieldValue extends Object>
     CognitoUserAttributeKey? customAttributeKey,
     bool? required,
     super.autofillHints,
+    AuthenticatorTextEnabledOverride? enabled,
```
tyllark: I believe this should be `super.enabledOverride`.  we will need to clean up the extra parameters from all the SignUpFormFields in this file. this is what the call should look like:

```
SignUpFormField.passwordConfirmation(
   visible: false,
   enabledOverride: AuthenticatorTextEnabledOverride.disabled,
   authenticatorTextFieldController: passwordController,
),
```

packages/authenticator/amplify_authenticator/lib/src/widgets/form_fields/sign_in_form_field.dart
```diff
@@ -33,24 +33,28 @@ abstract class SignInFormField<FieldValue extends Object>
     Key? key,
```
tyllark, line 35: We should include the following parameters in the constructor so they are available.
`super.enabledOverride`
`super.visible`

We will need need to do this for `ResetPasswordFormField`, `ConfirmSignUpFormField`, `ConfirmSignInFormField` as well.

angelorodem:
Hi @tyllark ,sorry for the bug, yes, feel free to chime in and finish, i would appreciate a lot your help, it is better to someone with better knowledge of the repo to finish this change.

Thank you!

angelorodem:
@tyllark i think i made all the changes needed, can you run the workflow and check it is working properly?

angelorodem:
@tyllark i think it is good?

tyllark:
@angelorodem I just merged in main into the branch (we did a release yesterday) and I'm waiting for the test to pass again. I'll squash and merge once all the checks pass and will push out another release. I'll ping you here and on the Github issue once that's all done!

angelorodem:
@tyllark Just saw that the e2e tests failed, though i could not find in the logs the reason it freezes my browser, is it related to this change?

tyllark:
@angelorodem I just restarted the failed tests. Some of our Github actions are a bit flakey and might fail due to Android emulators not starting up or network issues. These tests all passed before so I don't expect there to be any problems but if there are I'll push a fix.

PR #6429, opened 2025-11-11
fix(amplify_push_notifications): replace removeFirst() with removeAt(0) for Android < 11 compatibility

LucasNovaes:
### Summary

This pull request fixes a crash that occurs on Android devices running versions lower than Android 11 (API 30) due to the usage of the `removeFirst()` method from the Kotlin standard library.

The method `removeFirst()` is only available on Android 11+ (API 30+) or when running on a full Java 11+ runtime. When executed on older Android versions, it results in a `NoSuchMethodError`.

### Changes

- Replaced: 
  ```kotlin 
  eventQueue.removeFirst()
  ```

- With:
  ```kotlin
  eventQueue.removeAt(0)
  ```
### Technical background

`removeFirst()` is implemented as a default interface method, which is not supported by older Android runtimes.
`removeAt(0)` provides the same behavior and works correctly on all supported Android API levels (`minSdkVersion 21+`).

No functional behavior changes have been made — only improved runtime compatibility.

### Checklist
- [x] Verified that behavior is unchanged on Android 11+.
- [x] Verified that crash is fixed on Android 10 and below.
- [x] No new dependencies introduced.
- [x] Maintains API compatibility.

cc: @tyllark

### DIFF
--- a/packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationEventsStreamHandler.kt
+++ b/packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationEventsStreamHandler.kt
@@ -102,7 +102,7 @@ class PushNotificationEventsStreamHandler constructor(
         try {
             eventSink?.let {
                 while (eventQueue.isNotEmpty()) {
-                    val eventFromQueue = eventQueue.removeFirst()
+                    val eventFromQueue = eventQueue.removeAt(0)
                     // Check if it is an Error event and handle accordingly by using .error method
                     if (eventFromQueue.event.eventName == NativeEvent.ERROR.eventName) {
                         val exception = eventFromQueue.error


tylerjroach:
Yikes! Thanks for the report and PR. Would you happen to be compiling for SDK 35 here? 

http://youtrack.jetbrains.com/issue/KT-71375/Prevent-Kotlins-removeFirst-and-removeLast-from-causing-crashes-on-Android-14-and-below-after-upgrading-to-Android-API-Level-35

Looks like this is a new crash introduction if thats the case. 

@tyllark @ekjotmultani This change looks fine but there's no good mechanism at the moment to ensure we don't later use a method such as removeFirst elsewhere. We will want to look further into this.

tyllark:
Thank you for the PR @LucasNovaes! I will start to prepare a release for tomorrow.

LucasNovaes:
> Yikes! Thanks for the report and PR. Would you happen to be compiling for SDK 35 here?

Confirming that the issue was observed on a project built with compileSdkVersion 36, Kotlin 2.1.20, and AGP 8.13.0.
Even with SDK 36, the crash occurred on devices running Android 14 (API 34) and below, consistent with [KT-71375](https://youtrack.jetbrains.com/issue/KT-71375).
The merged change successfully resolves the compatibility issue across affected versions.
Thanks for the quick review and merge!

PR #5926, opened 2025-02-11
fix(auth,core): Added auth userpool endpoint into AmplifyOutputs mapping

tyllark:
*Issue #, if available:*
#5784 

*Description of changes:*
Auth user pool endpoint configuration was not included in the mapping from AmplifyConfig to AmplifyOutputs. This change adds a `userPoolEndpoint` field to AuthOutputs along with mapping logic. Additional any deleted references to this endpoint value were added back in.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #5784, opened 2024-12-23
Amplify Does Not Use Endpoint for Cognito Auth Calls


mozakassab:
### Description

When configuring Amplify in a Flutter application with existing resources and specifying an Endpoint for CognitoUserPool, the SDK does not route Cognito calls through the provided endpoint. This behavior is unexpected, and was working before with v1.
Here's an example of the configuration passed for existing resources: 

```dart
String amplifyconfig = ''' {
  "UserAgent": "aws-amplify-cli/2.0",
  "Version": "1.0",
  "auth": {
    "plugins": {
      "awsCognitoAuthPlugin": {
        "IdentityManager": {
          "Default": {}
        },
        "CognitoUserPool": {
          "Default": {
            "PoolId": "${EnvVariables.cognitoUserPoolID.value}",
            "AppClientId": "${EnvVariables.cognitoAppClientID.value}",
            "Region": "${EnvVariables.awsRegion.value}",
            "Endpoint": "${EnvVariables.awsCognitoEndpoint.value}"
          }
        },
        "Auth": {
          "Default": {
            "authenticationFlowType": "CUSTOM_AUTH"
          }
        }
      }
    }
  }
}''';
```

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [X] Auth
- [ ] Authenticator
- [ ] DataStore
- [ ] Notifications (Push)
- [ ] Storage

### Steps to Reproduce

1. Configure Amplify with existing resources, and specify an endpoint for CognitoUserPool:
```dart
String amplifyconfig = ''' {
  "UserAgent": "aws-amplify-cli/2.0",
  "Version": "1.0",
  "auth": {
    "plugins": {
      "awsCognitoAuthPlugin": {
        "IdentityManager": {
          "Default": {}
        },
        "CognitoUserPool": {
          "Default": {
            "PoolId": "${EnvVariables.cognitoUserPoolID.value}",
            "AppClientId": "${EnvVariables.cognitoAppClientID.value}",
            "Region": "${EnvVariables.awsRegion.value}",
            "Endpoint": "${EnvVariables.awsCognitoEndpoint.value}"
          }
        },
        "Auth": {
          "Default": {
            "authenticationFlowType": "CUSTOM_AUTH"
          }
        }
      }
    }
  }
}''';
```

2. Add Cognito Auth plugin:
```dart
Amplify.addPlugin(
        AmplifyAuthCognito(),
      );
```

[...]

PR #6475, opened 2025-12-11
Fix Cognito Auth endpoints mapping

nathan-dovire:
#5784 

## Completes the `userPoolEndpoint` Implementation

This PR completes completes the [implementation](https://github.com/aws-amplify/amplify-flutter/pull/5926) started in commit `f37cb7e69` which added `userPoolEndpoint` to `AuthOutputs` but was missing a critical mapping step.

### Problem
The original work added the `userPoolEndpoint` field to support custom Cognito endpoints (e.g., routing through a proxy), but the implementation was incomplete. The endpoint was being collected from Gen 1 configs but never actually passed to the Cognito User Pool configuration.

### Root Cause
The `CognitoUserPoolConfig.fromAuthOutputs()` factory method was missing the endpoint mapping, breaking the configuration flow:

```dart
// Missing in original PR:
endpoint: authOutputs.userPoolEndpoint
```

### DIFF
--- a/packages/amplify_core/lib/src/amplify_class.dart
+++ b/packages/amplify_core/lib/src/amplify_class.dart
@@ -180,6 +180,9 @@ abstract class AmplifyClass {
 
   /// Configures all plugins in topologically-sorted order.
   Future<void> _configurePlugins(AmplifyOutputs config) async {
+    safePrint(
+      '⚙️ [Amplify Core] Configuring plugins with AuthOutputs.userPoolEndpoint: ${config.auth?.userPoolEndpoint}',
+    );
     await Future.wait(_addPluginFutures);
     _addPluginFutures.clear();
     final categories = <Category, AmplifyCategory>{
--- a/packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
+++ b/packages/auth/amplify_auth_cognito_dart/lib/src/auth_plugin_impl.dart
@@ -515,6 +515,10 @@ class AmplifyAuthCognitoDart extends AuthPluginInterface
     String? password,
     SignInOptions? options,
   }) async {
+    safePrint(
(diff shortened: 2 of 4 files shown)


cadivus:
Hi @nathan-dovire ,

Thank you for your contribution to the AWS Amplify Flutter Library! Is there a reason this Pull Request is still a draft?

nathan-dovire:
Hi @cadivus,
I’ve been testing this and noticed an issue with authentication when using the endpoint. Even after configuring the `userPoolEndpoint`, authentication still fails.

For context, in my use case authentication is handled via a custom auth endpoint that wraps the Cognito client. This Cognito client is protected by a client secret that is only available at runtime. The flow is:
- A request is made to the custom auth endpoint.
- That endpoint injects the Cognito client secret (provided at runtime).
- The endpoint then redirects the request directly to the Cognito authentication endpoint.

With this setup, simply adding `userPoolEndpoint` to the configuration does not seem to be sufficient. The auth request directly hits the Cognito endpoint and fails to authenticate, which suggests the current implementation may not fully support Cognito flows where the authentication request is executed directly against Cognito rather than through the custom endpoint.

Importantly, this flow works as expected when testing against v1.8.*. However, after upgrading to v2.x and above, authentication breaks with the same configuration, which strongly suggests a regression introduced in v2+.

Please let me know if this behavior is expected, or if there’s additional configuration or a recommended integration pattern for this scenario. I’m happy to help investigate further or contribute a fix once the intended behavior is clear.

Thanks!

cadivus:
Hi @nathan-dovire ,
we just merged https://github.com/aws-amplify/amplify-flutter/pull/6477 which is similar to your PR. How exactly and where are you configuring your Cognito endpoint?

The merged PR worked for us by setting the `userPoolEndpoint` like this directly in `amplify_outputs.dart`:
Screenshot 2026-01-12 at 14 51 06

PR #6516, opened 2026-01-22
feat(kinesis): Kinesis Data Streams Library

ekjotmultani:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify_core/lib/src/category/amplify_categories.dart
+++ b/packages/amplify_core/lib/src/category/amplify_categories.dart
@@ -50,10 +50,7 @@ enum Category {
   storage,
 
   /// Push Notifications
-  pushNotifications,
-
-  /// Kinesis streaming (Data Streams + Firehose)
-  kinesis;
+  pushNotifications;
 
   String get name => switch (this) {
     Category.analytics => 'Analytics',
@@ -63,7 +60,6 @@ enum Category {
     Category.hub => 'Hub',
     Category.storage => 'Storage',
     Category.pushNotifications => 'PushNotifications',
-    Category.kinesis => 'Kinesis',
   };
 }
 
--- a/infra-gen2/tool/deploy_gen2.dart
+++ b/infra-gen2/tool/deploy_gen2.dart
@@ -7,7 +7,7 @@
 import 'dart:convert';
 import 'dart:io';
 
-import 'package:amplify_core/amplify_core.dart';
+import 'package:amplify_core/amplify_core.dart' show UUID;
 import 'package:args/args.dart';
 import 'package:path/path.dart' as p;
 
@@ -20,7 +20,7 @@ import 'package:path/path.dart' as p;
 /// 4. Run `dart tool/deploy_gen2.dart` to deploy the backend
 const List<AmplifyBackendGroup> infraConfig = [
   AmplifyBackendGroup(
-    category: Category.api,
+    category: 'API',
     defaultOutput: 'packages/api/amplify_api/example/lib',
     backends: [
       AmplifyBackend(
@@ -31,7 +31,7 @@ const List<AmplifyBackendGroup> infraConfig = [
     ],
   ),
   AmplifyBackendGroup(
-    category: Category.auth,
+    category: 'Auth',
     defaultOutput: 'packages/auth/amplify_auth_cognito/example/lib',
     sharedOutputs: [
       'packages/auth/amplify_auth_cognito_dart/example/lib',
@@ -96,7 +96,7 @@ const List<AmplifyBackendGroup> infraConfig = [
     ],
   ),
   AmplifyBackendGroup(
-    category: Category.storage,
+    category: 'Storage',
     defaultOutput: 'packages/storage/amplify_storage_s3/example/lib',
     backends: [
       AmplifyBackend(
@@ -112,7 +112,7 @@ const List<AmplifyBackendGroup> infraConfig = [
     ],
   ),
   AmplifyBackendGroup(
-    category: Category.analytics,
+    category: 'Analytics',
     defaultOutput: 'packages/analytics/amplify_analytics_pinpoint/example/lib',
     backends: [
       AmplifyBackend(
@@ -133,7 +133,7 @@ const List<AmplifyBackendGroup> infraConfig = [
     ],
   ),
   AmplifyBackendGroup(
-    category: Category.kinesis,
+    category: 'Kinesis',
     defaultOutput: 'packages/kinesis/amplify_kinesis_dart/lib',
     backends: [
       AmplifyBackend(
@@ -159,8 +159,7 @@ void main(List<String> arguments) async {
 
   print('🚀 Deploying Gen 2 backends!');
   for (final backendGroup in infraConfig) {
-    if (categoryToDeploy != null &&
-        backendGroup.category.name != categoryToDeploy) {
+    if (categoryToDeploy != null && backendGroup.category != categoryToDeploy) {
       continue;
     }
     // TODO(equartey): Could be removed when all backends are defined.
@@ -169,7 +168,7 @@ void main(List<String> arguments) async {
     }
     var gen2Environments = <String, String>{};
     var gen1Environments = <String, String>{};
-    final categoryName = backendGroup.category.name;
+    final categoryName = backendGroup.category;
     final outputPath = p.join(repoRoot.path, backendGroup.defaultOutput);
     final amplifyOutputs = File(p.join(outputPath, 'amplify_outputs.dart'));
     final amplifyConfiguration = File(
@@ -267,7 +266,7 @@ ArgResults _parseArgs(List<String> args) {
       'category',
       abbr: 'c',
       help: 'Specify the category to deploy.',
-      allowed: Category.values.map((e) => e.name).toList(),
+      allowed: infraConfig.map((e) => e.category).toList(),
       defaultsTo: null,
     );
 
@@ -276,14 +275,12 @@ ArgResults _parseArgs(List<String> args) {
 
 /// Deploy Sandbox for a given backend backend
 Future<String> _deployBackend(
-  Category category,
+  String category,
   AmplifyBackend backend,
   String outputPath,
   bool verbose,
 ) async {
-  print(
-    '🏖️  Deploying ${category.name} ${backend.name}, this may take a while...',
-  );
+  print('🏖️  Deploying $category ${backend.name}, this may take a while...');
 
   final outputFile = File(p.join(outputPath, 'amplify_outputs.dart'));
   if (outputFile.existsSync()) {
@@ -329,7 +326,6 @@ Future<String> _deployBackend(
       in process.stdout
           .transform(utf8.decoder)
           .transform(const LineSplitter())) {
-
     if (!line.startsWith('    at ') && line.trim().isNotEmpty) {
       // This line does not belong to a cdk build error
       postingBackendBuildError = false;
@@ -369,22 +365,24 @@ Future<String> _deployBackend(
 
   if (exitCode != 0) {
     throw Exception(
-      '❌ Error deploying ${category.name} ${backend.identifier} sandbox',
+      '❌ Error deploying $category ${backend.identifier} sandbox',
     );
-  } else if (postedDeploymentError && postedDeploymentErrorReason && postedDeploymentUpdateFailed) {
+  } else if (postedDeploymentError &&
+      postedDeploymentErrorReason &&
+      postedDeploymentUpdateFailed) {
     throw Exception(
-      '❌ Error deploying ${category.name} ${backend.identifier} sandbox: Update failed',
+      '❌ Error deploying $category ${backend.identifier} sandbox: Update failed',
     );
   } else if (postedBackendBuildError) {
     throw Exception(
-      '❌ Error deploying ${category.name} ${backend.identifier} sandbox - CDK build failed',
+      '❌ Error deploying $category ${backend.identifier} sandbox - CDK build failed',
     );
   } else if (!outputFile.existsSync()) {
     throw Exception(
-      '❌ Error deploying ${category.name} ${backend.identifier} sandbox - Output file not generated',
+      '❌ Error deploying $category ${backend.identifier} sandbox - Output file not generated',
     );
   } else {
-    print('👍 ${category.name} ${backend.identifier} sandbox deployed');
+    print('👍 $category ${backend.identifier} sandbox deployed');
     return stackID;
   }
 }
@@ -461,7 +459,7 @@ void _copyConfigFile(List<String> outputPaths, List<File> configFiles) {
 
 /// Create a unique bucket name
 String _createBucketName(String base) {
-  final uniqueShort = uuid().substring(0, 8);
+  final uniqueShort = UUID.getUUID().substring(0, 8);
   return '${base.toLowerCase()}-gen2-integ-$uniqueShort';
 }
 
@@ -560,14 +558,12 @@ void _uploadConfigFileToS3(String bucketName, List<File> configFiles) {
 
 /// Generates gen 1 amplifyconfiguration.dart file
 void _generateGen1Config(
-  Category category,
+  String category,
   AmplifyBackend backend,
   String outputPath,
   String stack,
 ) {
-  print(
-    '📁 Generating gen 1 config file for ${category.name} ${backend.name}...',
-  );
+  print('📁 Generating gen 1 config file for $category ${backend.name}...');
 
   final outputFile = File(p.join(outputPath, 'amplifyconfiguration.dart'));
   if (outputFile.existsSync()) {
@@ -594,16 +590,14 @@ void _generateGen1Config(
 
   if (process.exitCode != 0) {
     throw Exception(
-      '❌ Error generating gen 1 config file for ${category.name} ${backend.name}:: ${process.stdout}',
+      '❌ Error generating gen 1 config file for $category ${backend.name}:: ${process.stdout}',
     );
-  } else if(!outputFile.existsSync()) {
+  } else if (!outputFile.existsSync()) {
     throw Exception(
-      '❌ Error generating gen 1 config file for ${category.name} ${backend.identifier} - Output file not generated',
+      '❌ Error generating gen 1 config file for $category ${backend.identifier} - Output file not generated',
     );
   } else {
-    print(
-      '👍 Gen 1 config file for ${category.name} ${backend.name} generated',
-    );
+    print('👍 Gen 1 config file for $category ${backend.name} generated');
   }
 }
 
@@ -616,7 +610,7 @@ class AmplifyBackendGroup {
   });
 
   /// This is the category of the integration group
-  final Category category;
+  final String category;
 
   /// This is the list of backends for the group
   final List<AmplifyBackend> backends;
--- a/packages/amplify_core/lib/src/amplify_class_impl.dart
+++ b/packages/amplify_core/lib/src/amplify_class_impl.dart
@@ -46,8 +46,6 @@ class AmplifyClassImpl extends AmplifyClass {
           plugin.cast(),
           authProviderRepo: authProviderRepo,
         );
-      case Category.kinesis:
-        throw UnimplementedError();
     }
   }
 }


packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_storage.dart
```diff
@@ -0,0 +1,124 @@
...
+
+  /// Retrieves a batch of records sorted by stream_name, partition_key, id.
+  ///
+  /// Returns records up to [maxCount] records and [maxBytes] total size.
+  Future<List<StoredRecord>> getRecordsBatch({
+    int maxCount = 500,
+    int maxBytes = 5 * 1024 * 1024,
+  }) async {
+    // Query records sorted by stream_name, partition_key, id
+    final query = _db.select(_db.kinesisRecords)
```
jvh-aws: Can we optimize this query to avoid consuming high amounts of RAM by loading data we don't process? See also the [Android implementation](https://github.com/aws-amplify/amplify-android/pull/3188/changes#diff-09c7e6d34984c2de551ce40c2015f0020ba2197f728069306387c4d8b852b01bR114-R126).

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/kinesis_sender.dart
```diff
@@ -0,0 +1,153 @@
...
+      return _parseResponse(response, records.length);
+    } on ProvisionedThroughputExceededException {
+      // All records should be retried
+      return PutRecordsResult(
+        successfulRecordIndices: const [],
+        failedRecordIndices: const [],
+        retryableRecordIndices: List.generate(records.length, (i) => i),
+      );
+    } on SmithyHttpException catch (e) {
+      // Check if it's a retryable HTTP error (5xx)
```
jvh-aws: How about 429 errors?

packages/kinesis/aws_kinesis_datastreams/lib/src/exception/kinesis_exception.dart
```diff
@@ -0,0 +1,113 @@
...
+///
+/// This exception is thrown when the library cannot communicate with
+/// AWS Kinesis Data Streams due to network issues.
+///
+/// Records that fail due to network errors are retained in the cache
+/// and will be retried on the next flush operation.
+/// {@endtemplate}
+class NetworkException extends KinesisDataStreamsException {
+  /// {@macro aws_kinesis_datastreams.network_exception}
+  const NetworkException(super.message, {super.underlyingException})
```
jvh-aws: How are we using this exception actually?

packages/kinesis/aws_kinesis_datastreams_dart/lib/src/impl/auto_flush_scheduler.dart
```diff
@@ -0,0 +1,93 @@
...
+    required KinesisDataStreamsFlushStrategy strategy,
+    required Future<void> Function() onFlush,
+  })  : _strategy = strategy,
+        _onFlush = onFlush;
+
+  final KinesisDataStreamsFlushStrategy _strategy;
+  final Future<void> Function() _onFlush;
+  Timer? _timer;
+  bool _enabled = true;
+  bool _closed = false;
```
jvh-aws: What is the benefit of having both `_enabled` and `_closed`?

packages/kinesis/aws_kinesis_datastreams_dart/lib/src/impl/kinesis_sender.dart
```diff
@@ -0,0 +1,153 @@
...
+    final failedIndices = <int>[];
+    final retryableIndices = <int>[];
+
+    final resultEntries = response.records.toList();
+
+    for (var i = 0; i < resultEntries.length; i++) {
+      final entry = resultEntries[i];
+
+      if (entry.errorCode == null) {
+        // Success - record has sequenceNumber and shardId
```
jvh-aws: > record has sequenceNumber and shardId

Is this comment relevant? How are we using shardId?

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_storage.dart
```diff
@@ -0,0 +1,124 @@
...
+          KinesisRecordsCompanion.custom(
+            retryCount: _db.kinesisRecords.retryCount + const Constant(1),
+          ),
+        );
+    }
+  }
+
+  /// Deletes records that have exceeded the maximum retry count.
+  Future<void> deleteRecordsExceedingRetries(int maxRetries) async {
+    await (_db.delete(_db.kinesisRecords)
```
jvh-aws: Can we avoid an extra DB query here by memorizing the retry counts of records when we load and send them?

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_client.dart
```diff
@@ -0,0 +1,174 @@
...
+  /// Flushes all cached records to Kinesis.
+  ///
+  /// Records are batched by stream name and sent in batches respecting
+  /// Kinesis limits (500 records, 5MB per batch).
+  ///
+  /// Successful records are deleted from the cache.
+  /// Failed records with retryable errors have their retry count incremented.
+  /// Records exceeding max retries are removed from the cache.
+  Future<void> flush() async {
+    if (_closed) return;
```
jvh-aws: Wouldn't it be better to throw an error or at least warning when we try flushing while the client is closed?

packages/kinesis/aws_kinesis_datastreams/test/e2e/kinesis_e2e_test.dart
```diff
@@ -0,0 +1,290 @@
...
+      testSessionToken,
+    );
+
+    client = KinesisDataStreams(
+      region: testRegion,
+      credentialsProvider: const AWSCredentialsProvider(credentials),
+      storagePath: tempDir.path,
+      options: const KinesisDataStreamsOptions(
+        flushStrategy: KinesisDataStreamsInterval(
+          interval: Duration(minutes: 5), // Disable auto-flush for tests
```
jvh-aws: Shouldn't we be able to disable auto-flush using a call to `disable`?

packages/kinesis/aws_kinesis_datastreams/test/e2e/kinesis_e2e_test.dart
```diff
@@ -0,0 +1,290 @@
...
+      // Act
+      await client.record(
+        data: data,
+        partitionKey: 'e2e-test-partition',
+        streamName: testStreamName,
+      );
+      await client.flush();
+
+      // Assert - if no exception thrown, the record was sent successfully
+      // The flush would throw if the stream doesn't exist or credentials are invalid
```
jvh-aws: Can we have a more comprehensive assert here?

packages/kinesis/aws_kinesis_datastreams/test/e2e/kinesis_e2e_test.dart
```diff
@@ -0,0 +1,290 @@
...
+        final testData = {'event': 'batch_test', 'index': i};
+        await client.record(
+          data: Uint8List.fromList(utf8.encode(jsonEncode(testData))),
+          partitionKey: 'e2e-batch-partition',
+          streamName: testStreamName,
+        );
+      }
+      await client.flush();
+
+      // Assert - all records sent without error
```
jvh-aws: Same here

packages/kinesis/aws_kinesis_datastreams_dart/test/e2e/kinesis_e2e_test.dart
```diff
@@ -0,0 +1,290 @@
...
+      print('''
+╔══════════════════════════════════════════════════════════════════╗
+║  E2E TESTS SKIPPED                                               ║
+║                                                                  ║
+║  To run E2E tests:                                               ║
+║  1. Edit test/e2e/test_config.dart                               ║
+║  2. Fill in your AWS credentials and stream name                 ║
+║  3. Set isConfigured = true                                      ║
+║  4. Run: dart test test/e2e/ --tags=e2e                          ║
+╚══════════════════════════════════════════════════════════════════╝
```
jvh-aws: Can we actually implement the e2e test before merging?

packages/kinesis/aws_kinesis_datastreams/test/e2e/kinesis_e2e_test.dart
```diff
@@ -0,0 +1,290 @@
...
+        data: Uint8List.fromList(utf8.encode('{"ignored": true}')),
+        partitionKey: 'disabled-test',
+        streamName: testStreamName,
+      );
+
+      // Re-enable and flush - should have nothing to send
+      client.enable();
+      await client.flush();
+
+      // Assert - no exception means success (nothing was queued)
```
jvh-aws: Same here

packages/kinesis/aws_kinesis_datastreams_dart/test/record_storage_test.dart
```diff
@@ -0,0 +1,262 @@
...
+
+        final after = await storage.getRecordsBatch();
+        expect(after, isEmpty);
+
+        final size = await storage.getCurrentCacheSize();
+        expect(size, equals(0));
+      });
+    });
+  });
+}
```
jvh-aws: Can we ensure concurrent access to the record storage works via a test?

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/kinesis_sender.dart
ekjotmultani: Good catch, addressing

packages/kinesis/aws_kinesis_datastreams/lib/src/exception/kinesis_exception.dart
ekjotmultani: For now it serves as being here while vNext common error classes are coming, should be thrown when network errors arise in send operations via the kinesis client, will address in commit

packages/kinesis/aws_kinesis_datastreams_dart/lib/src/impl/auto_flush_scheduler.dart
ekjotmultani: _closed can serve as a flag for other code to know whether or not the client itself is shut down or if just flushing is enabled or not. In our implementation they do similar things but it is useful for getters to know the internal state of the client I feel. What do you think?

packages/kinesis/aws_kinesis_datastreams_dart/lib/src/impl/kinesis_sender.dart
ekjotmultani: we arent using them, the comment just documents that we must have been returned these from the sdk call

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_storage.dart
ekjotmultani: do you mean that we shouldn't have the retry counts as a column in the db and store retry counts to records in memory?

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_client.dart
ekjotmultani: yes good point, addressed to throw an error when attempting to operate on a closed client

packages/kinesis/aws_kinesis_datastreams_dart/test/e2e/kinesis_e2e_test.dart
ekjotmultani: yes before merge these tests will be live

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_storage.dart
jvh-aws: I agree with you, that we should keep them in the DB to persist them. But, I think we can delete the records exceeding retry counts without an additional DB lookup. Here's how I think this could be implemented in 

`record_cient`'s `flush` ([link](https://github.com/aws-amplify/amplify-flutter/pull/6516/changes#diff-a85b964840a1238c085213d02a91214362e416db58f92c2a2070e04193e68465R91-R110)): 

 * We're already keeping the results of `getRecordsBatch` in memory
 * `_sendStreamBatch` can return a list of record ids that failed and exceed retries (we're passing the records into it already)
 * `deleteRecordsExceedingRetries` can accept this list and delete those right away.

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/kinesis_sender.dart
```diff
@@ -0,0 +1,174 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:io';
```
cadivus: This package is intended for web too, right? If so, I don't think we can import `dart:io` like this.

packages/kinesis/aws_kinesis_datastreams/lib/src/amplify_kinesis_client.dart
```diff
@@ -0,0 +1,205 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:typed_data';
+
+import 'package:amplify_foundation_dart/amplify_foundation_dart.dart';
+import 'package:aws_kinesis_datastreams/aws_kinesis_datastreams.dart' show KinesisLimitExceededException;
+import 'package:aws_kinesis_datastreams/src/db/kinesis_record_database.dart';
+import 'package:aws_kinesis_datastreams/src/exception/amplify_kinesis_exception.dart' show KinesisLimitExceededException;
```
cadivus: This is redundant to
```Dart
import 'package:aws_kinesis_datastreams/aws_kinesis_datastreams.dart' show KinesisLimitExceededException;
```

packages/kinesis/aws_kinesis_datastreams_dart/lib/src/kinesis_data_streams_options.dart
```diff
@@ -0,0 +1,100 @@
...
+          _minFlushIntervalMs,
+          _maxFlushIntervalMs,
+          'flushStrategy.interval',
+          'Flush interval must be between 1 second and 1 hour',
+        );
+      }
+    }
+  }
+
+  static const int _maxCacheBytes = 4 * 1024 * 1024 * 1024; // 4GB
```
cadivus: Not sure, can 4GB be a problem on 32Bit systems? @jvh-aws How do they other platforms treat that?
However, probably a nit considering how uncommon 32Bit got.

packages/kinesis/aws_kinesis_datastreams_dart/lib/src/kinesis_data_streams_options.dart
```diff
@@ -0,0 +1,100 @@
...
+
+/// {@template aws_kinesis_datastreams.amplify_kinesis_client_options}
+/// Configuration options for [AmplifyKinesisClient].
+///
+/// All numeric properties are validated to be positive and within
+/// service-imposed upper bounds.
+/// {@endtemplate}
+class AmplifyKinesisClientOptions {
+  /// {@macro aws_kinesis_datastreams.amplify_kinesis_client_options}
+  AmplifyKinesisClientOptions({
```
cadivus: Can't this be `const`?

packages/kinesis/aws_kinesis_datastreams/lib/src/models/flush_data.dart
cadivus, line 1: Why do we have this empty file (https://github.com/aws-amplify/amplify-flutter/blob/feat/kinesis-data-streams-library/packages/kinesis/aws_kinesis_datastreams/lib/src/models/flush_data.dart / https://github.com/aws-amplify/amplify-flutter/blob/d9ec0471dc40b3cbcead0133e3be08558f39a347/packages/kinesis/aws_kinesis_datastreams/lib/src/models/flush_data.dart)?

packages/kinesis/aws_kinesis_datastreams/lib/src/model/clear_cache_data.dart
```diff
@@ -0,0 +1,16 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/// {@template aws_kinesis_datastreams.clear_cache_data}
+/// Data returned from a clearCache operation.
+/// {@endtemplate}
+class ClearCacheData {
```
cadivus: Should this be extended? If not, I think we should set it to final.

packages/kinesis/aws_kinesis_datastreams/lib/src/model/flush_data.dart
```diff
@@ -0,0 +1,20 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/// {@template aws_kinesis_datastreams.flush_data}
+/// Data returned from a flush operation.
+/// {@endtemplate}
+class FlushData {
```
cadivus: Should this be extended? If not, I think we should set it to final.

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/kinesis_sender.dart
```diff
@@ -0,0 +1,174 @@
...
+import 'package:amplify_foundation_dart/amplify_foundation_dart.dart'
+    as foundation;
+import 'package:aws_common/aws_common.dart';
+import 'package:aws_kinesis_datastreams/src/exception/amplify_kinesis_exception.dart';
+import 'package:aws_kinesis_datastreams/src/sdk/kinesis.dart';
+import 'package:aws_kinesis_datastreams/src/sdk/sdk_bridge.dart';
+import 'package:smithy/smithy.dart';
+
+/// Result of a PutRecords operation.
+class PutRecordsResult {
```
cadivus: Is this a simple non-extended data class? If so, I'd set it to `final`.

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/kinesis_sender.dart
```diff
@@ -0,0 +1,174 @@
...
+
+  /// Indices of records that failed with non-retryable errors.
+  final List<int> failedRecordIndices;
+
+  /// Indices of records that failed with retryable errors.
+  final List<int> retryableRecordIndices;
+}
+
+/// A record to be sent to Kinesis.
+class KinesisSenderRecord {
```
cadivus: Is this a simple non-extended data class? If so, I'd set it to `final`.

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/kinesis_record.dart
```diff
@@ -0,0 +1,42 @@
...
+  }) : dataSize = data.length + utf8.encode(partitionKey).length,
+       createdAt = DateTime.now();
+
+  /// Creates a Kinesis record with a specific creation timestamp (for testing).
+  KinesisRecord.withTimestamp({
+    required this.data,
+    required this.partitionKey,
+    required this.streamName,
+    required this.createdAt,
+  }) : dataSize = data.length + utf8.encode(partitionKey).length;
```
cadivus: Shouldn't the default constructor require a timestamp and the factory sets a value for this?
if we then also move
```Dart
dataSize = data.length + utf8.encode(partitionKey).length
```
to a getter, we could have a const constructor here.

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/kinesis_record.dart
```diff
@@ -0,0 +1,42 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:convert';
+import 'dart:typed_data';
+
+/// Internal representation of a record to be sent to Kinesis.
+class KinesisRecord {
+  /// Creates a new Kinesis record.
+  KinesisRecord({
```
cadivus: Should this be a final class?

packages/kinesis/aws_kinesis_datastreams/lib/src/credentials/credentials.dart
```diff
@@ -0,0 +1,32 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/// AWS credentials for authenticating API requests.
+class Credentials {
```
cadivus: Do we want to allow subclassing here?

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/auto_flush_scheduler.dart
```diff
@@ -0,0 +1,55 @@
...
+  Future<void> close() async {
+    _closed = true;
+    _cancelTimer();
+  }
+
+  void _cancelTimer() { _timer?.cancel(); _timer = null; }
+
+  void _handleTimerTick() {
+    if (!_enabled || _closed) return;
+    _onFlush();
```
cadivus: Should we add error handling here? Since it's a non-awaited `Future`, there is no handling yet.

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_client.dart
```diff
@@ -0,0 +1,153 @@
...
+    } on Exception {
+      await _storage.incrementRetryCount(records.map((r) => r.id));
+      return 0;
+    }
+  }
+
+  /// Clears all cached records.
+  Future<ClearCacheData> clearCache() async {
+    final batch = await _storage.getRecordsBatch();
+    final count = batch.length;
```
cadivus: We get all the data just to count it. Shouldn't we use a count query.

@jvh-aws What do you think?

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_storage.dart
```diff
@@ -0,0 +1,132 @@
...
+
+    for (final id in ids) {
+      await (_db.update(_db.kinesisRecords)
+        ..where((t) => t.id.equals(id)))
+        .write(
+          KinesisRecordsCompanion.custom(
+            retryCount: _db.kinesisRecords.retryCount + const Constant(1),
+          ),
+        );
+    }
```
cadivus: I am not so familiar with this but don't we query more updates than needed here? Shouldn't we use something like
```SQL
UPDATE ... WHERE id IN (...) 
```
Maybe, @jvh-aws knows more here.

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_client.dart
```diff
@@ -0,0 +1,153 @@
...
+    await _storage.clear();
+    return ClearCacheData(recordsCleared: count);
+  }
+
+  void enable() { _enabled = true; _scheduler.enable(); }
+  void disable() { _enabled = false; _scheduler.disable(); }
+  void enableAutoFlush() { _scheduler.enable(); }
+  void disableAutoFlush() { _scheduler.disable(); }
```
cadivus: Nit: I don't think this matches our codestyle.

packages/kinesis/aws_kinesis_datastreams/lib/src/exception/amplify_kinesis_exception.dart
```diff
@@ -0,0 +1,120 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'package:amplify_foundation_dart/amplify_foundation_dart.dart';
+
+/// {@template aws_kinesis_datastreams.amplify_kinesis_exception}
+/// Base exception for Amplify Kinesis Data Streams errors.
+/// {@endtemplate}
+sealed class AmplifyKinesisException extends AmplifyException {
```
cadivus: Since this is sealed, shouldn't the classes below extending this be `final` to keep the hierarchy sealed?

packages/kinesis/aws_kinesis_datastreams/pubspec.yaml
```diff
@@ -0,0 +1,32 @@
...
+version: 0.1.0
+homepage: https://docs.amplify.aws/lib/q/platform/flutter/
+repository: https://github.com/aws-amplify/amplify-flutter/tree/main/packages/kinesis/aws_kinesis_datastreams
+issue_tracker: https://github.com/aws-amplify/amplify-flutter/issues
+
+environment:
+  sdk: ^3.9.0
+
+dependencies:
+  amplify_core: ">=2.7.0 <2.8.0"
```
cadivus: This is not compatible with main, we are on a newer version.

packages/kinesis/aws_kinesis_datastreams/example/.flutter-plugins-dependencies
```diff
@@ -0,0 +1 @@
...e_info_plus-12.3.0/","dependencies":[],"dev_dependency":false},{"name":"package_info_plus","path":"/Users/ekjotm/.pub-cache/hosted/pub.dev/package_info_plus-9.0.0/","dependencies":[],"dev_dependency":false}]},"dependencyGraph":[{"name":"amplify_analytics_pinpoint","dependencies":["amplify_db_common","amplify_secure_storage","device_info_plus","package_info_plus","path_provider"]},{"name":"amplify_auth_cognito","dependencies":["amplify_analytics_pinpoint","amplify_secure_storage"]},{"name":"amplify_db_common","dependencies":["path_provider"]},{"name":"amplify_secure_storage","dependencies":["path_provider","path_provider_windows"]},{"name":"device_info_plus","dependencies":[]},{"name":"package_info_plus","dependencies":[]},{"name":"path_provider","dependencies":["path_provider_android","path_provider_foundation","path_provider_linux","path_provider_windows"]},{"name":"path_provider_android","dependencies":[]},{"name":"path_provider_foundation","dependencies":[]},{"name":"path_provider_linux","dependencies":[]},{"name":"path_provider_windows","dependencies":[]}],"date_created":"2026-01-26 14:26:14.510543","version":"3.35.4","swift_package_manager_enabled":{"ios":false,"macos":false}}
```
cadivus: Please remove this.

packages/kinesis/aws_kinesis_datastreams_dart/lib/src/kinesis_data_streams_options.dart
ekjotmultani: Can't be const because of runtime validation (RangeError.checkValueInInterval). A factory+private const constructor wouldn't help since the factory itself can't be const.

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_storage.dart
ekjotmultani: agreed, this could be one query, good catch

packages/kinesis/aws_kinesis_datastreams/example/.flutter-plugins-dependencies
ekjotmultani: thank you!!!! where not commented to your other comments in the review , consider them addressed

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_client.dart
jvh-aws: Agreed. We can probably use the ORM to count.

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_storage.dart
jvh-aws: This is what we use in Android/Swift: 
```sql
UPDATE records SET retry_count = retry_count + 1 WHERE id IN ($placeholders)
```

packages/amplify_core/lib/src/category/amplify_categories.dart
```diff
@@ -60,6 +63,7 @@ enum Category {
     Category.hub => 'Hub',
     Category.storage => 'Storage',
     Category.pushNotifications => 'PushNotifications',
+    Category.kinesis => 'kinesis',
```
jvh-aws: Why are we adding Kinesis Category but below ([here](https://github.com/aws-amplify/amplify-flutter/pull/6516/changes#diff-3f912949c3c591421a55e7142c203d9cacf8bcb5be68dffdfb216b5366de96d2R49-R50)) marking it as not implemented? Shouldn't we not have to add any category/plugin for the Kinesis Client, since they are independent of the plugins?

packages/kinesis/aws_kinesis_datastreams_dart/lib/src/db/kinesis_record_database.dart
```diff
@@ -0,0 +1,72 @@
...
+
+  @override
+  MigrationStrategy get migration {
+    return MigrationStrategy(
+      onCreate: (Migrator m) async {
+        await m.createAll();
+        // Create index for efficient batching queries
+        await customStatement(
+          'CREATE INDEX IF NOT EXISTS idx_stream_partition '
+          'ON kinesis_records(stream_name, partition_key, id)',
```
jvh-aws: Does an index over data size make sense for faster aggregations? 

```sql
CREATE INDEX IF NOT EXISTS idx_data_size ON records (data_size)
```

packages/kinesis/aws_kinesis_datastreams/lib/src/exception/amplify_kinesis_exception.dart
```diff
@@ -0,0 +1,120 @@
...
+      : super(
+          message: message,
+          recoverySuggestion: 'Try clearing the cache or reinitializing.',
+        );
+}
+
+/// {@template aws_kinesis_datastreams.kinesis_service_exception}
+/// Thrown when a Kinesis SDK/API error occurs. Inspect [sdkException] for details.
+/// {@endtemplate}
+final class KinesisServiceException extends AmplifyKinesisException {
```
jvh-aws: I think we discussed that those errors are silently swallowed and not exposed.

packages/kinesis/aws_kinesis_datastreams/lib/src/exception/amplify_kinesis_exception.dart
```diff
@@ -0,0 +1,120 @@
...
+          message: message,
+          recoverySuggestion: 'Check network connectivity and try again.',
+        );
+}
+
+/// {@template aws_kinesis_datastreams.kinesis_record_too_large_exception}
+/// Thrown when a single record exceeds the Kinesis per-record size limit
+/// (10 MiB, partition key + data blob combined).
+/// {@endtemplate}
+final class KinesisRecordTooLargeException extends AmplifyKinesisException {
```
jvh-aws: If we have a specific type for records being too large, we should also have one for the partition key being too long.

packages/kinesis/aws_kinesis_datastreams_dart/lib/src/impl/auto_flush_scheduler.dart
jvh-aws: I think both approaches are fine, but personally, I recommend merging them for the auto flush scheduler, since `_closed` is not really used.

packages/kinesis/aws_kinesis_datastreams_dart/lib/src/impl/kinesis_sender.dart
jvh-aws: Gotcha. I wonder if it makes sense to remove the comment for cleaner code (nit)

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_client.dart
```diff
@@ -0,0 +1,165 @@
...
+  ///
+  /// Throws [ClientClosedException] if the client has been closed.
+  /// Throws [KinesisRecordTooLargeException] if the record exceeds the
+  /// per-record size limit (10 MiB, partition key + data blob).
+  /// Throws [KinesisLimitExceededException] if the cache is full.
+  Future<void> record(KinesisRecord record) async {
+    if (_closed) throw ClientClosedException();
+    if (!_enabled) return;
+
+    if (record.dataSize > kKinesisMaxRecordBytes) {
```
jvh-aws: We should also check the partition key length here. It should be in (0, 256]

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_client.dart
```diff
@@ -0,0 +1,165 @@
...
+    try {
+      final result = await _sender.putRecords(streamName: streamName, records: senderRecords);
+
+      await _storage.deleteRecords(result.successfulRecordIndices.map((i) => records[i].id));
+      await _storage.incrementRetryCount(result.retryableRecordIndices.map((i) => records[i].id));
+      await _storage.deleteRecords(result.failedRecordIndices.map((i) => records[i].id));
+
+      return result.successfulRecordIndices.length;
+    } on Exception {
+      await _storage.incrementRetryCount(records.map((r) => r.id));
```
jvh-aws: I think, we wanted to propagate non-SDK errors up to the caller of `flush`

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_storage.dart
```diff
@@ -0,0 +1,141 @@
...
+          ROW_NUMBER() OVER (ORDER BY stream_name, partition_key, id) as rn,
+          SUM(data_size) OVER (ORDER BY stream_name, partition_key, id) as running_size
+        FROM kinesis_records
+      )
+      WHERE rn <= ?1 AND (running_size <= ?2 OR rn = 1)
+      ORDER BY stream_name, partition_key, id
+      ''',
+      variables: [Variable.withInt(maxCount), Variable.withInt(maxBytes)],
+      readsFrom: {_db.kinesisRecords},
+    ).get();
```
jvh-aws: What happens if this query fails? In my understanding, currently, we'd swallow the error in `flush`'s try-catch block. However, I think we would like to expose such errors as storage errors. Can we add logic to expose such errors as `Kinesis*` errors? Or am I missing something?

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_storage.dart
```diff
@@ -0,0 +1,141 @@
...
+  /// Deletes records that have exceeded the maximum retry count.
+  Future<void> deleteRecordsExceedingRetries(int maxRetries) async {
+    await (_db.delete(_db.kinesisRecords)
+      ..where((t) => t.retryCount.isBiggerThanValue(maxRetries)))
+      .go();
+  }
+
+  /// Returns the current total size of cached data in bytes.
+  Future<int> getCurrentCacheSize() async {
+    final query = _db.selectOnly(_db.kinesisRecords)
```
jvh-aws: I think, I commented this before: We can keep an integer for the cache size and update / check against that, when we add new records. We will only need to run a DB query after removing records. This allows us to avoid the extra query when recording, which is most likely the most frequent operation.

packages/kinesis/aws_kinesis_datastreams/lib/src/amplify_kinesis_client.dart
```diff
@@ -0,0 +1,205 @@
...
+  }
+}
+
+// ---------------------------------------------------------------------------
+// Deprecated alias for backward compatibility
+// ---------------------------------------------------------------------------
+
+/// Deprecated: Use [AmplifyKinesisClient] instead.
+@Deprecated('Use AmplifyKinesisClient instead')
+typedef KinesisDataStreams = AmplifyKinesisClient;
```
cadivus: Why do we need this in a brand new package?

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_client.dart
```diff
@@ -0,0 +1,165 @@
...
+        maxBytes: kKinesisMaxRecordBytes,
+      );
+    }
+
+    final currentSize = await _storage.getCurrentCacheSize();
+    if (currentSize + record.dataSize > _storage.maxCacheBytes) {
+      throw KinesisLimitExceededException();
+    }
+
+    await _storage.saveRecord(record);
```
cadivus: Don't we need a mutex/lock here? What happens if we call `record()` async two times?

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_client.dart
```diff
@@ -0,0 +1,165 @@
...
+          recordsByStream.putIfAbsent(record.streamName, () => []).add(record);
+        }
+
+        for (final entry in recordsByStream.entries) {
+          final flushed = await _sendStreamBatch(entry.key, entry.value);
+          totalFlushed += flushed;
+        }
+
+        await _storage.deleteRecordsExceedingRetries(_maxRetries);
+      }
```
cadivus: Can't we get into an endless loop here? Is there any detection of a situation where `batch` does not get smaller?

packages/kinesis/aws_kinesis_datastreams/lib/src/flush_strategy/flush_strategy.dart
```diff
@@ -0,0 +1,35 @@
...
+/// {@endtemplate}
+sealed class KinesisDataStreamsFlushStrategy {
+  /// {@macro aws_kinesis_datastreams.flush_strategy}
+  const KinesisDataStreamsFlushStrategy();
+}
+
+/// {@template aws_kinesis_datastreams.interval_flush_strategy}
+/// A flush strategy that triggers automatic flushes at a fixed interval.
+/// {@endtemplate}
+class KinesisDataStreamsInterval extends KinesisDataStreamsFlushStrategy {
```
cadivus: I think this should be final. We don't want to extend it, right?

packages/kinesis/aws_kinesis_datastreams/lib/src/flush_strategy/flush_strategy.dart
```diff
@@ -0,0 +1,35 @@
...
+  });
+
+  /// The interval between automatic flush operations.
+  final Duration interval;
+}
+
+/// {@template aws_kinesis_datastreams.none_flush_strategy}
+/// A flush strategy that disables automatic flushing.
+/// {@endtemplate}
+class KinesisDataStreamsNone extends KinesisDataStreamsFlushStrategy {
```
cadivus: I think this should be final. We don't want to extend it, right?

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_client.dart
```diff
@@ -0,0 +1,165 @@
...
+import 'package:aws_kinesis_datastreams/src/impl/record_storage.dart';
+import 'package:aws_kinesis_datastreams/src/kinesis_data_streams_options.dart';
+import 'package:aws_kinesis_datastreams/src/model/clear_cache_data.dart';
+import 'package:aws_kinesis_datastreams/src/model/flush_data.dart';
+
+/// {@template aws_kinesis_datastreams.record_client}
+/// Orchestrates record operations, managing the flow between storage,
+/// scheduling, and sending.
+/// {@endtemplate}
+class RecordClient {
```
cadivus: Should this be final?

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_storage.dart
```diff
@@ -0,0 +1,141 @@
...
+import 'dart:async';
+
+import 'package:aws_kinesis_datastreams/src/db/kinesis_record_database.dart';
+import 'package:aws_kinesis_datastreams/src/impl/kinesis_record.dart';
+import 'package:drift/drift.dart';
+
+/// {@template aws_kinesis_datastreams.record_storage}
+/// Manages SQLite database operations for record persistence.
+/// {@endtemplate}
+class RecordStorage {
```
cadivus: Should this be final?

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/auto_flush_scheduler.dart
```diff
@@ -0,0 +1,55 @@
...
+  bool get isEnabled => _enabled;
+  bool get isClosed => _closed;
+
+  void start() {
+    if (_closed) return;
+    _cancelTimer();
+
+    switch (_strategy) {
+      case KinesisDataStreamsInterval(:final interval):
+        _timer = Timer.periodic(interval, (_) { _handleTimerTick(); });
```
cadivus: Nit (style): I think this should be like:

```suggestion
        _timer = Timer.periodic(interval, (_) => _handleTimerTick());
```

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_storage.dart
```diff
@@ -0,0 +1,141 @@
...
+    );
+  }
+
+  /// Retrieves a batch of records sorted by stream_name, partition_key, id.
+  ///
+  /// Returns records up to [maxCount] records and [maxBytes] total size.
+  /// Uses window functions to efficiently limit at the database level.
+  Future<List<StoredRecord>> getRecordsBatch({
+    int maxCount = 500,
+    int maxBytes = 10 * 1024 * 1024, // 10 MiB for Kinesis PutRecords
```
cadivus: Don't we already have a constant for this somewhere? Maybe, we should move it to a constant file and use it everywhere.

packages/amplify_core/lib/src/category/amplify_categories.dart
ekjotmultani: good call, removed

packages/kinesis/aws_kinesis_datastreams_dart/lib/src/db/kinesis_record_database.dart
ekjotmultani: getCurrentCacheSize() does SUM(data_size) over all rows, a full table scan regardless of index. An index on data_size wouldn't help aggregations. However, this is now moot because we added in-memory cache tracking from a later comment, so the DB query only runs on init and after deletes.

packages/kinesis/aws_kinesis_datastreams/lib/src/exception/amplify_kinesis_exception.dart
ekjotmultani: agreed, added a corresponding type

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_storage.dart
ekjotmultani: no, good call, wrapped the failures in error object

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_storage.dart
ekjotmultani: yes, done

packages/kinesis/aws_kinesis_datastreams/lib/src/amplify_kinesis_client.dart
ekjotmultani: leftover from a previous generation, thanks

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_client.dart
ekjotmultani: good call, added a Completer for the method

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_client.dart
ekjotmultani: can't make it final because MockRecordClient extends Mock implements RecordClient in tests. Added doc comment explaining why.

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_storage.dart
ekjotmultani: yes you are right

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_client.dart
ekjotmultani: yes, I've added a hard limit and a counter for retries without the batch size decreasing, `if maxRetries`+2 iterations go without the batch getting smaller we break

packages/kinesis/aws_kinesis_datastreams_dart/lib/src/db/kinesis_record_database.dart
jvh-aws: Nice!

packages/kinesis/aws_kinesis_datastreams/lib/src/credentials/credentials.dart
```diff
@@ -0,0 +1,32 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
```
cadivus: isn't this dead code/part of the foundation?

packages/kinesis/aws_kinesis_datastreams/lib/src/credentials/credentials_provider.dart
```diff
@@ -0,0 +1,20 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
```
cadivus: isn't this dead code/part of the foundation?

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_client.dart
```diff
@@ -0,0 +1,276 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:async';
+
+import 'package:aws_kinesis_datastreams/aws_kinesis_datastreams.dart' show AmplifyKinesisClient;
+import 'package:aws_kinesis_datastreams/src/amplify_kinesis_client.dart' show AmplifyKinesisClient;
```
cadivus: Isn't this a duplicate import?

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/auto_flush_scheduler.dart
```diff
@@ -0,0 +1,55 @@
...
+  Future<void> close() async {
+    _closed = true;
+    _cancelTimer();
+  }
+
+  void _cancelTimer() { _timer?.cancel(); _timer = null; }
+
+  void _handleTimerTick() {
+    if (!_enabled || _closed) return;
+    unawaited(_onFlush().catchError((_) {}));
```
cadivus: We don't even log the errors, is this intended?

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/auto_flush_scheduler.dart
```diff
@@ -0,0 +1,55 @@
...
+      case KinesisDataStreamsInterval(:final interval):
+        _timer = Timer.periodic(interval, (_) => _handleTimerTick());
+      case KinesisDataStreamsNone():
+        break;
+    }
+  }
+
+  void stop() { _cancelTimer(); }
+  void enable() { _enabled = true; }
+  void disable() { _enabled = false; }
```
cadivus: These are customer facing, right? If so, shouldn't we add docstring here?

cadivus: Does this fit our codestyle?

cadivus (changes requested):

packages/kinesis/aws_kinesis_datastreams/lib/src/amplify_kinesis_client.dart
```diff
@@ -0,0 +1,197 @@
...
+    recordClient = RecordClient(
+      storage: storage,
+      sender: _kinesisSender,
+      scheduler: scheduler = AutoFlushScheduler(
+        strategy: options.flushStrategy,
+        onFlush: () => recordClient.flush(),
+      ),
+      maxRetries: options.maxRetries,
+      maxRecords: options.maxRecords,
+    );
```
cadivus: Isn't this better since we can't avoid `late`?

```suggestion
   final scheduler =  AutoFlushScheduler(
      strategy: options.flushStrategy,
      onFlush: () => recordClient.flush(),
    );
    
    final recordClient = RecordClient(
      storage: storage,
      sender: _kinesisSender,
      scheduler: scheduler,
      maxRetries: options.maxRetries,
      maxRecords: options.maxRecords,
    );
```

packages/kinesis/aws_kinesis_datastreams/lib/src/credentials/credentials.dart
ekjotmultani: yes these are old, will remove

packages/kinesis/aws_kinesis_datastreams/lib/src/credentials/credentials_provider.dart
ekjotmultani: yes these are old, will remove

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/auto_flush_scheduler.dart
jvh-aws: It's not customer facing.

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/auto_flush_scheduler.dart
jvh-aws: Yes. Those need to be logged here.

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_client.dart
ekjotmultani: yes, removed

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/auto_flush_scheduler.dart
ekjotmultani: added logging

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/auto_flush_scheduler.dart
ekjotmultani: it is idiomatic in Dart 3, do you see a problem?

packages/kinesis/aws_kinesis_datastreams/lib/src/amplify_kinesis_client.dart
ekjotmultani: yeah, separated it out

packages/kinesis/aws_kinesis_datastreams/test/e2e/kinesis_e2e_test.dart
jvh-aws: We can use flush strategy `none` now

packages/kinesis/aws_kinesis_datastreams_dart/lib/src/kinesis_data_streams_options.dart
jvh-aws: Yes, let's change it to 5MB to align with the other implementations.

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_client.dart
```diff
@@ -0,0 +1,276 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:async';
+
+import 'package:aws_kinesis_datastreams/aws_kinesis_datastreams.dart'
+    show AmplifyKinesisClient;
+import 'package:aws_kinesis_datastreams/src/amplify_kinesis_client.dart'
+    show AmplifyKinesisClient;
```
cadivus: Duplicate/redundant imports of AmplifyKinesisClient


```suggestion
import 'package:aws_kinesis_datastreams/src/amplify_kinesis_client.dart'
    show AmplifyKinesisClient;
```

packages/kinesis/aws_kinesis_datastreams/lib/src/sdk/sdk_bridge.dart
```diff
@@ -0,0 +1,113 @@
...
+  final KinesisClient _base;
+  final AWSHttpClient? _httpClient;
+
+  // ignore: invalid_use_of_protected_member, invalid_use_of_visible_for_testing_member
+  amplify_core.DependencyManager? _dependencyManager;
+
+  amplify_core.DependencyManager get _deps {
+    // ignore: invalid_use_of_protected_member, invalid_use_of_visible_for_testing_member
+    return _dependencyManager ??= amplify_core.Amplify.dependencies;
+  }
```
cadivus: Are you sure that this is a good idea if we need to ignore even two rules?

cadivus: And are you sure you want to assign that here? The entire construction here looks very broken, please fix that.

packages/kinesis/aws_kinesis_datastreams/lib/src/amplify_kinesis_client.dart
cadivus: No, the old problem is still here. Did you forget to push?

packages/kinesis/aws_kinesis_datastreams_dart/lib/src/amplify_kinesis_client.dart
```diff
@@ -0,0 +1,197 @@
...
+  late final KinesisSender _kinesisSender;
+
+  /// The AWS region for this client.
+  String get region => _region;
+
+  /// The configuration options for this client.
+  AmplifyKinesisClientOptions get options => _options;
+
+  /// Whether the client is currently enabled.
+  bool get isEnabled => _recordClient.isEnabled;
```
cadivus: `_recordClient` is `late final`. We should either guard access or get away from `late final`.
  
Getting away would work by making the record client a parameter and eventually adding a factory.

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_client.dart
```diff
@@ -0,0 +1,276 @@
...
+        recordBytes: record.dataSize,
+        maxBytes: kKinesisMaxRecordBytes,
+      );
+    }
+
+    // Acquire async lock to prevent concurrent cache size races.
+    while (_recordLock != null) {
+      await _recordLock!.future;
+    }
+    _recordLock = Completer<void>();
```
cadivus: I think the log implementation is broken. I suggest either using a package like https://pub.dev/packages/synchronized or getting some inspiration from there.

packages/kinesis/aws_kinesis_datastreams/lib/src/kinesis_data_streams_options.dart
```diff
@@ -0,0 +1,108 @@
...
+/// Maximum length of a Kinesis partition key (256 characters).
+const int kKinesisMaxPartitionKeyLength = 256;
+
+/// {@template aws_kinesis_datastreams.amplify_kinesis_client_options}
+/// Configuration options for [AmplifyKinesisClient].
+///
+/// All numeric properties are validated to be positive and within
+/// service-imposed upper bounds.
+/// {@endtemplate}
+class AmplifyKinesisClientOptions {
```
cadivus: Can't the class be final? Or do we want customers to extend it?

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/auto_flush_scheduler.dart
```diff
@@ -0,0 +1,74 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:async';
+
+import 'package:aws_kinesis_datastreams/src/flush_strategy/flush_strategy.dart';
+
+/// {@template aws_kinesis_datastreams.auto_flush_scheduler}
+/// Manages automatic flush scheduling based on the configured strategy.
+/// {@endtemplate}
+class AutoFlushScheduler {
```
cadivus: Do we want/expect subclassing here?

packages/kinesis/aws_kinesis_datastreams_dart/test/e2e/cognito_auth_helper.dart
```diff
@@ -0,0 +1,190 @@
...
+  Future<AWSCredentials> resolve() async {
+    final cached = _cached;
+    if (cached != null && cached is TemporaryCredentials) {
+      if (DateTime.now().isBefore(cached.expiration)) {
+        return cached;
+      }
+    }
+    _cached = await _helper.authenticate();
+    return _cached!;
+  }
```
cadivus: Shouldn't we have a semaphore here? I don't think this should be running two times at the same time.

packages/kinesis/aws_kinesis_datastreams/pubspec.yaml
```diff
@@ -0,0 +1,32 @@
...
+issue_tracker: https://github.com/aws-amplify/amplify-flutter/issues
+
+environment:
+  sdk: ^3.9.0
+
+dependencies:
+  amplify_core: ">=2.9.0 <2.10.0"
+  amplify_db_common_dart: ">=0.4.13 <0.5.0"
+  amplify_foundation_dart: ">=1.0.0 <2.0.0"
+  amplify_foundation_dart_bridge: ">=1.0.0 <2.0.0"
```
cadivus:
```suggestion
  amplify_foundation_dart_bridge: ">=1.0.0 <1.1.0"
```

We need to narrow this down for compatibility guarantees.

cadivus (changes requested):

packages/kinesis/aws_kinesis_datastreams/lib/src/amplify_kinesis_client.dart
ekjotmultani: ah yes I did, fixed now

packages/kinesis/aws_kinesis_datastreams/lib/src/sdk/sdk_bridge.dart
ekjotmultani: yeah you're right this came out really ugly, changed it to just straightforwardly accept an awshttpclient

ekjotmultani: I think it was mimicking https://github.com/aws-amplify/amplify-flutter/blob/feat/kinesis-data-streams-library/packages/analytics/amplify_analytics_pinpoint_dart/lib/src/sdk/sdk_bridge.dart#L46, but we aren't tied to Amplify so we can accept a client directly

packages/kinesis/aws_kinesis_datastreams_dart/lib/src/amplify_kinesis_client.dart
ekjotmultani: yeah, good call

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/record_client.dart
ekjotmultani: Do you mean lock? In that case then what would be broken here?

packages/kinesis/aws_kinesis_datastreams/lib/src/kinesis_data_streams_options.dart
ekjotmultani: made final

packages/kinesis/aws_kinesis_datastreams/lib/src/impl/auto_flush_scheduler.dart
ekjotmultani: made final

packages/kinesis/aws_kinesis_datastreams_dart/test/e2e/cognito_auth_helper.dart
ekjotmultani: This is a test utility which runs on single threaded kinesis clients, concurrent calls shouldn't be a problem here I think, even in the worst case two concurrent authenticate calls just overwrite the cached auth token

cadivus: Sounds reasonable

packages/kinesis/aws_kinesis_datastreams_dart/lib/src/impl/record_client.dart
```diff
@@ -0,0 +1,319 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:async';
+
+import 'package:amplify_foundation_dart/amplify_foundation_dart.dart';
+import 'package:aws_kinesis_datastreams_dart/aws_kinesis_datastreams_dart.dart'
```
ekjotmultani: unneeded import

packages/kinesis/aws_kinesis_datastreams_dart/lib/src/impl/storage/record_storage_indexeddb.dart
```diff
@@ -0,0 +1,247 @@
...
+
+  @override
+  Future<void> deleteRecords(Iterable<int> ids) async {
+    if (ids.isEmpty) return;
+    await _openEvent;
+    final store = _getStore();
+    for (final id in ids) {
+      await store.delete(id.toJS).future;
+    }
+  }
```
cadivus: @jvh-aws Shouldn't this be batches in some way?

cadivus (dismissed):

PR #6549, opened 2026-01-29
feat(ci): Windows - Support multiple attempts with timeouts

cadivus:
This allows us to configure multiple attempts for Windows runners, each with a timeout to increase test stability.  
For now, we set 5 attempts with 60min timeout each.

### DIFF
--- a/build-support/linux_like_timeout.ps1
+++ b/build-support/linux_like_timeout.ps1
@@ -1,0 +1,186 @@
+# This copies the behavior of the Linux timeout command
+
+[CmdletBinding()]
+param(
+    [Parameter(Mandatory=$false)]
+    [Alias('k')]
+    [string]$KillAfter = "",
+
+    [Parameter(Mandatory=$false)]
+    [Alias('s')]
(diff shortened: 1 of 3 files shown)


build-support/linux_like_timeout.ps1
```diff
@@ -0,0 +1,184 @@
+[CmdletBinding()]
```
jvh-aws, line 3: Shouldn't this rather be called according to what it does and not what it's like?

cadivus: It does the same. So if you know the Linux timeout, you know what this PS script does.

jvh-aws:
Also, can you please provide a PR description?

build-support/linux_like_timeout.ps1
jvh-aws: Gotcha, can we note this in the script as comment?

cadivus: Done

Issue #6548, opened 2026-01-29
[amplify_authenticator] Password autofill not working in Sign In form


Poloten:
### Description

When I create a new user in SignUpForm, Android's Google Password Manager offers to save the login and password. (When I open Google Password manager on device I see the password and user exist with my appId). However, when I try to log in to the account again, it only suggests entering the previous email address, the password is never auto-filled under any circumstances, and I have to enter it manually. 

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [ ] Auth
- [x] Authenticator
- [ ] DataStore
- [ ] Notifications (Push)
- [ ] Storage

### Steps to Reproduce

```Dart
 AuthenticatorForm(
    child: Padding(
     padding: const EdgeInsets.symmetric(horizontal: 16),
     child: Column(
      children: [
       SignUpFormField.email(required: true),
       SignUpFormField.password(),
       SignUpFormField.passwordConfirmation(),
       const SizedBox(height: 20),
       const SignUpButton(),
     ]))
```
```yaml
amplify_auth_cognito: ^2.9.0
amplify_flutter: ^2.9.0
amplify_authenticator: ^2.4.1
amplify_storage_s3: ^2.9.0
aws_common: ^0.7.12
```

### Platforms

- [ ] iOS
- [x] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Flutter Version

3.38.7

### Amplify Flutter Version

2.9.0

### Deployment Method

Amplify Gen 2

### Schema

```GraphQL

```

cadivus:
Hi @Poloten ,
 
Thank you for creating this issue. You are running on Android, right? Is it an Android app or are you using a web app on Android?

Poloten:
> Hi [@Poloten](https://github.com/Poloten) ,
> 
> Thank you for creating this issue. You are running on Android, right? Is it an Android app or are you using a web app on Android?

Hi [cadivus](https://github.com/cadivus), I use Android App

cadivus:
Hi @Poloten,
 
Thanks for the clarification!

cadivus:
Hi @Poloten,

The issue has been resolved and is now available in version 2.10.0. Please upgrade to this version to get the fix. 
You can find the full release notes [here](https://github.com/aws-amplify/amplify-flutter/releases/tag/v2.10.0). I am closing this issue now, but please do not hesitate to reopen 
[...]

PR #6554, opened 2026-01-30
fix(auth): Fix login autofill on login

cadivus:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-flutter/issues/6548

With the bug:  
Screenshot 2026-01-30 at 14 30 28

With the fix:  
Screenshot 2026-01-30 at 14 28 43


jvh-aws:
Does it make sense to test this, e.g. by using a widget test that verifies the `AutofillGroup` exists?

PR #6559, opened 2026-02-02
chore: fix null-aware-elements warning in form field

jvh-aws:
*Issue #, if available:*

*Description of changes:* It appears that recent updates might have introduced [`use_null_aware_elements`](https://dart.dev/tools/linter-rules/use_null_aware_elements) checks that affect existing code and hence causing linter issues in PRs. This ignores the check for existing code. 

### DIFF
--- a/packages/authenticator/amplify_authenticator/lib/src/widgets/form_field.dart
+++ b/packages/authenticator/amplify_authenticator/lib/src/widgets/form_field.dart
@@ -317,12 +317,10 @@ abstract class AuthenticatorFormFieldState<
           Column(
             crossAxisAlignment: CrossAxisAlignment.start,
             children: <Widget>[
-              // ignore: use_null_aware_elements
-              if (surlabel != null) surlabel!,
+              ?surlabel,
               SizedBox(height: labelGap),
               buildFormField(context),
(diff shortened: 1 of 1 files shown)


packages/authenticator/amplify_authenticator/lib/src/widgets/form_field.dart
```diff
@@ -317,9 +317,11 @@ abstract class AuthenticatorFormFieldState<
           Column(
             crossAxisAlignment: CrossAxisAlignment.start,
             children: <Widget>[
+              // ignore: use_null_aware_elements
               if (surlabel != null) surlabel!,
```
cadivus: Sure this is the best way to do this?
Not rather something like

```suggestion
              surlabel ?? SizedBox.shrink(),
```
instead of adding an ignore?

cadivus: (and remove the ignore)

jvh-aws: We can use null-aware elements to avoid the warning and avoid adding new elements to the list

PR #6570, opened 2026-02-03
chore: Apply format fixes

cadivus:
Use null-aware elements in various places


jvh-aws:
Can you add a description? E.g. "Use null-aware elements in various places"

cadivus:
> Can you add a description? E.g. "Use null-aware elements in various places"

Done

jvh-aws (approved):

Issue #6568, opened 2026-02-03
[Authenticator][A11Y] Password visibility toggle button missing accessible name (WCAG 4.1.2)


timstarkk:
### Description

The password visibility toggle button (eye icon) in `SignInForm`, `SignUpForm`, and other password fields lacks a semantic label/tooltip. Screen readers (VoiceOver, TalkBack) announce verbose, incorrect text like "Enter your Password menu pop-up, secure text field with autofill menu" instead of the expected "Show password" / "Hide password".

This violates **WCAG 4.1.2 (Name, Role, Value - Level A)**: User interface components must have an accessible name that describes their purpose.

**Root cause:**

In `lib/src/widgets/form.dart` lines 144-158, the `obscureTextToggle` getter creates an `IconButton` without a `tooltip` parameter:

```dart
Widget get obscureTextToggle {
  return ValueListenableBuilder<bool>(
    valueListenable: obscureTextToggleValue,
    builder: (BuildContext context, bool toggleObscureText, Widget? _) {
      return IconButton(
        onPressed: () {
          obscureTextToggleValue.value = !toggleObscureText;
        },
        icon: Icon(
          toggleObscureText ? Icons.visibility : Icons.visibility_off,
        ),
        // Missing: tooltip
      );
    },
  );
}
```

**Suggested fix:**

Add a `tooltip` parameter that changes based on the visibility state:

```dart
IconButton(
  onPressed: () {
    obscureTextToggleValue.value = !toggleObscureText;
  },
  icon: Icon(
    toggleObscureText ? Icons.visibility : Icons.visibility_off,
  ),
  tooltip: toggleObscureText ? 'Show password' : 'Hide password',
);
```

The tooltip text should ideally come from the existing localization system (`InputResolver`) for i18n support.

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [ ] Auth
- [x] Authenticator
- [ ] DataStore
- [ ] Notifications (Push)
- [ ] Storage

### Steps to Reproduce

1. Use `amplify_authenticator` package with `SignInForm()` or any form with a password field
2. Enable a screen reader (VoiceOver on iOS/macOS, TalkBack on Android, or browser screen reader on web)
3. Navigate to the password visibility toggle button (eye icon)
4. Listen to the screen reader announcement - it does not announce "Show password" or "Hide password"

### Platforms

- [x] iOS
- [x] Android
- [x] Web
- [x] macOS
[...]

PR #6577, opened 2026-02-04
fix(authenticator): add tooltip text to obscure widget and wrap materialbanner content in semantics

ekjotmultani:
*Issue #, if available:*
#6568 
#6581

*Description of changes:*
Adding text on the obscure widget on password fields in forms on Authenticator, making it screen readable, wrapped authenticatorform in semantics so errors are screen readable as ewll (credit @timstarkk for finding these misses)


packages/authenticator/amplify_authenticator/lib/src/l10n/generated/input_localizations_en.dart
```diff
@@ -153,4 +153,10 @@ class AuthenticatorInputLocalizationsEn
   @override
   String get totpCodePrompt =>
       'Please enter the code from your registered Authenticator app';
+
+  @override
+  String get showPassword => 'Show password';
```
jvh-aws, line 158: This should be generated from the ARB file `inputs_en.arb`, which should be checked in

jvh-aws:
Can we try to update our references in `goldens/` for the tests to pass? According to our [contribution guide](https://github.com/aws-amplify/amplify-flutter/blob/main/CONTRIBUTING.md), you'll need to run: 

```bash
$ cd packages/authenticator/amplify_authenticator
$ flutter test --update-goldens
```

We should also verify the snapshots to avoid any unnecessary UI changes.

PR #6567, opened 2026-02-02
TMP test `fix(infra_gen2): Fix backends and deployment#6582`

cadivus:
Writes Amplify envs the gen1 config files too.

PR #6582, opened 2026-02-04
fix(infra_gen2): Fix backends and deployment

cadivus:
The backends in `infra_gen2` were not in sync with the backends used for integration tests. Also, the deployment tool didn't print errors. Now, build and deployment errors will be displayed.


cadivus:
Workflows were triggered here: https://github.com/aws-amplify/amplify-flutter/pull/6567. 
(after deploying the changes in this PR)

Screenshot 2026-02-05 at 00 10 26

PR #6666, opened 2026-02-23
fix(ci): Fix post-action logging on Windows

cadivus:
Fixes this:  
Screenshot 2026-02-23 at 14 33 44

Skip git clean/reset to avoid failures from Windows file locks after long-running tests. Safe here because this checkout only needs the composite action files and nothing runs after it.


cadivus:
Tested with a TMP commit
Screenshot 2026-02-24 at 09 48 59

PR #6693, opened 2026-02-27
Feat/kinesis e2e infra

ekjotmultani:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/kinesis_e2e.yaml
+++ b/.github/workflows/kinesis_e2e.yaml
@@ -69,7 +69,7 @@ jobs:
 
       - name: Run E2E tests
         working-directory: packages/kinesis/aws_kinesis_datastreams
-        run: dart test test/e2e/ --tags=e2e
+        run: dart test test/e2e/ --tags=e2e --run-skipped
 
       - name: Log success/failure
         if: always()
@@ -130,7 +130,7 @@ jobs:
 
       - name: Run E2E tests
         working-directory: packages/kinesis/aws_amazon_firehose
-        run: dart test test/e2e/ --tags=e2e
+        run: dart test test/e2e/ --tags=e2e --run-skipped
 
       - name: Log success/failure
         if: always()


.github/workflows/kinesis_e2e.yaml
```diff
@@ -0,0 +1,144 @@
+name: Kinesis E2E Tests
+on:
+  pull_request:
+    paths:
+      - "packages/kinesis/**"
+      - ".github/workflows/kinesis_e2e.yaml"
+  workflow_dispatch:
```
cadivus, line 14: How does this differ so much from all the other e2e tests that we need to implement everything custom?

.github/workflows/kinesis_e2e.yaml
```diff
@@ -0,0 +1,144 @@
...
+
+      - name: Fetch Kinesis E2E secrets
+        uses: aws-actions/aws-secretsmanager-get-secrets@022e8919774ecb75e8e375656d7b1898936ab878 # 1.0.4
+        with:
+          secret-ids: |
+            KINESIS_E2E,${{ secrets.KINESIS_E2E_SECRET_ARN }}
+          parse-json-secrets: true
+
+      - name: Install native sqlite3
+        run: sudo apt-get update && sudo apt-get install -y libsqlite3-dev
```
cadivus, line 52: Why is this needed? For building on Android?

.github/workflows/kinesis_e2e.yaml
```diff
@@ -0,0 +1,144 @@
...
+            aws_common:
+              path: ../../aws_common
+            aws_signature_v4:
+              path: ../../aws_signature_v4
+            smithy:
+              path: ../../smithy/smithy
+            smithy_aws:
+              path: ../../smithy/smithy_aws
+          EOF
+
```
cadivus: Why don't we use the aft bootstrapping? Why this custom solution?

infra/lib/stack.ts
```diff
@@ -5,8 +5,8 @@ import * as cdk from "aws-cdk-lib";
 import * as cognito from "aws-cdk-lib/aws-cognito";
 import { Construct } from "constructs";
 import {
-  AuthIntegrationTestStack,
-  AuthIntegrationTestStackEnvironmentProps,
+    AuthIntegrationTestStack,
+    AuthIntegrationTestStackEnvironmentProps,
```
cadivus: Why these `infra` changes? Isn't `infra` unrelated to this PR?

infra-gen2/backends/kinesis/main/amplify/backend.ts
```diff
@@ -0,0 +1,107 @@
...
+  description: "Kinesis E2E test credentials and resource names",
+  secretObjectValue: {
+    ACCESS_KEY_ID: cdk.SecretValue.unsafePlainText(accessKey.accessKeyId),
+    SECRET_ACCESS_KEY: accessKey.secretAccessKey,
+    STREAM_NAME: cdk.SecretValue.unsafePlainText(stream.streamName),
+    DELIVERY_STREAM_NAME: cdk.SecretValue.unsafePlainText(
+      deliveryStream.deliveryStreamName!,
+    ),
+  },
+});
```
cadivus: What are these secrets needed for? If they are for fetching the e2e backends, I don't think this is the right place.

ekjotmultani: not for fetching but for creating

ekjotmultani: but I am pushing out the updated branch, this whole commit was outdated and mistakenly pushed

.github/workflows/kinesis_e2e.yaml
ekjotmultani: this custom code is outdated sorry, but still kinesis E2E is fundamentally different, it needs OIDC + Secrets Manager for credentials rather than fetch_backends + amplify pull

cadivus: > it needs OIDC + Secrets Manager for credentials rather than fetch_backends + amplify pull


But fetch_backends [already uses OIDC + Secrets Manager](https://github.com/aws-amplify/amplify-flutter/blob/main/.github/composite_actions/fetch_backends/action.yaml#L25-L41). The only thing it does on top is run tool/pull_test_backend.sh to download `amplify_outputs.dart` from S3. 

For the credentials, the [Android PR](https://github.com/aws-amplify/amplify-android/pull/3194) shows the better approach:
Include auth as a category and grant Kinesis permissions to the authenticated  IAM role:

```typescript
// Android: aws-kinesis/infra/amplify/backend.ts
const backend = defineBackend({ auth });

backend.auth.resources.authenticatedUserIamRole.addToPrincipalPolicy(
    new PolicyStatement({
        actions: ['kinesis:PutRecord', 'kinesis:PutRecords', 'kinesis:DescribeStream'],
        resources: [stream.streamArn],
    })
);
```
([link](https://github.com/aws-amplify/amplify-android/pull/3194/changes#diff-06ecbb40c8dc9e763d676cd20be2f28b22019c3fa2059016aaf4938f84cb6e05R19-R29))

This is exactly the same pattern the Flutter repo already uses for [Analytics/Pinpoint](https://github.com/aws-amplify/amplify-flutter/blob/main/infra-gen2/backends/analytics/main/amplify/backend.ts), which also isn't a "native" Amplify  category but piggybacks on auth + custom CDK resources + backend.addOutput().  
  
Also, using some tokens in the app is not a real e2e test. This is not what customers most likely do. Using Cognito instead of some "injected" credentials is way closer to what a customer would to.
And this way, we need the `amplify_outputs.dart`.

ekjotmultani: I would say that these dart only next version packages could be decoupled from the cognito pattern, but for consistency I have updated them to match the other platforms and our own packages

infra-gen2/backends/kinesis/main/amplify/backend.ts
```diff
@@ -0,0 +1,78 @@
...
+
+// --- S3 bucket (Firehose destination) ---
+const bucket = new s3.Bucket(kinesisStack, "FirehoseDestBucket", {
+  removalPolicy: RemovalPolicy.DESTROY,
+  autoDeleteObjects: true,
+  enforceSSL: true,
+});
+
+// --- IAM role for Firehose → S3 ---
```
jvh-aws, line 32: I think this is unrelated to Kinesis. Can we handle this separately and remove from this PR?

cadivus:
Done in
https://github.com/aws-amplify/amplify-flutter/pull/6786

PR #6752, opened 2026-03-09
Kinesis: Separate out dart package and various fixes

jvh-aws:
Addition to Kinesis Client Feature

 * Add dart package 
 * Add record storage error and wrap all errors in kinesis exception (add consistent `from` method for kinesis exception type)
 * Do not expose logger as part of public constructor
 * Avoid reexporting foundation types
 * Fix error message when sqlite3.wasm file not found in dev mode
 * Enable full flush but attempt batches only once per flush
 * Fix partition key length verification
 * Minor fixes


packages/kinesis/aws_kinesis_datastreams/lib/src/flutter_path_provider/flutter_path_provider.dart
```diff
@@ -0,0 +1,6 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export 'flutter_path_provider_stub.dart'
+    if (dart.library.js_interop) 'flutter_path_provider_html.dart'
+    if (dart.library.io) 'flutter_path_provider_io.dart';
```
cadivus, line 6: Is there a reason we have our own implementation here?

packages/kinesis/aws_kinesis_datastreams_dart/lib/src/impl/record_client.dart
```diff
@@ -0,0 +1,319 @@
...
+  /// Flushes all cached records to Kinesis.
+  ///
+  /// Each flush processes all pending records in batches per stream. Records
+  /// that fail or are retryable within a flush cycle are skipped via an
+  /// exclusion set and will be picked up in the next flush cycle.
+  ///
+  /// SDK Kinesis errors (throttling, invalid stream, etc.) are logged and
+  /// skipped so other streams can still flush. Non-SDK errors (e.g. network,
+  /// storage) abort the flush and propagate to the caller.
+  Future<FlushData> flush() async {
```
cadivus, line 146: Why don't we use the result types anymor?

packages/kinesis/aws_kinesis_datastreams/lib/src/flutter_path_provider/flutter_path_provider.dart
jvh-aws: This is idiomatic in flutter and aids tree shaking. We use [`path_provider`](https://pub.dev/packages/path_provider) under when `io` is available.

packages/kinesis/aws_kinesis_datastreams_dart/lib/src/impl/record_client.dart
jvh-aws: Due to merge conflicts, We've temporarily moved the result boundary up and kept try-catch in the internal code. We will adapt this in a follow up.

PR #6765, opened 2026-03-12
fix: enable manual flushing when client is disabled

ekjotmultani:
*Issue #, if available:*

*Description of changes:*


jvh-aws:
Good catch!

PR #6795, opened 2026-03-16
(fix): remove kinesis from Category enum and refactor deploy script

ekjotmultani:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/infra-gen2/tool/deploy_gen2.dart
+++ b/infra-gen2/tool/deploy_gen2.dart
@@ -20,7 +20,7 @@ import 'package:path/path.dart' as p;
 /// 4. Run `dart tool/deploy_gen2.dart` to deploy the backend
 const List<AmplifyBackendGroup> infraConfig = [
   AmplifyBackendGroup(
-    category: 'Api',
+    category: 'API',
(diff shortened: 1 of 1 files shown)


infra-gen2/tool/deploy_gen2.dart
```diff
@@ -133,7 +133,7 @@ const List<AmplifyBackendGroup> infraConfig = [
     ],
   ),
   AmplifyBackendGroup(
-    category: Category.kinesis,
+    category: 'Kinesis',
```
cadivus, line 136: @ekjotmultani Are you sure it's upper case? It needs to align what we had. Otherwise, we might deploy it all new instead of running updates.

jvh-aws: I think, `category.name` resolves to uppercase, see [here](https://github.com/aws-amplify/amplify-flutter/blob/3b3cdafc05ef6a90ea41c97d3ccc8f55e5fd1ed0/packages/amplify_core/lib/src/category/amplify_categories.dart#L55)

ekjotmultani: yeah the old enum was uppercase

PR #6828, opened 2026-03-24
feat(firehose): add exceptions and record storage types

ekjotmultani:
added the exceptions and record storage objects for use in future prs and the web in-memory impl

*Issue #, if available:*

*Description of changes:*


packages/kinesis/amplify_firehose_dart/lib/src/impl/storage/record_storage.dart
```diff
@@ -0,0 +1,156 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jvh-aws, line 1: Can we avoid re-implementing those for the firehose client?

ekjotmultani:
Closing, this landed on main through #6845. The exceptions are in amplify_firehose_dart and the record storage types moved into amplify_record_cache_dart during the consolidation.

PR #6845, opened 2026-04-02
feat(firehose): amplify firehose library

ekjotmultani:
*Issue #, if available:*

*Description of changes:*

Amplify Firehose Library, dart and flutter impl


packages/kinesis/amplify_firehose_dart/lib/src/impl/firehose_sender.dart
```diff
@@ -0,0 +1,58 @@
...
+    final requestRecords = records
+        .map((record) => sdk.Record(data: record.data))
+        .toList();
+
+    final request = sdk.PutRecordBatchInput(
+      deliveryStreamName: streamName,
+      records: requestRecords,
+    );
+
+    final response = await _firehoseClient.putRecordBatch(request).result;
```
jvh-aws, line 47: How are errors handled when putRecordBatch fails? Can we have a comment that explains how this is handled above? in the record client?

packages/kinesis/amplify_firehose_dart/lib/src/amplify_firehose_client.dart
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  /// Closes the client and releases all resources.
+  Future<void> close() async {
+    _closed = true;
+    _scheduler?.stop();
+    await _recordClient.close();
+  }
+
+  Future<Result<T>> _wrapError<T>(Future<T> Function() operation) async {
```
jvh-aws, line 210: Can we re-use the wrapError from Kinesis and avoid duplicating the code here?

packages/kinesis/amplify_kinesis_dart/lib/src/impl/kinesis_sender.dart
```diff
@@ -68,7 +37,7 @@ class KinesisSender {
         .map(
           (record) => PutRecordsRequestEntry(
             data: record.data,
-            partitionKey: record.partitionKey,
+            partitionKey: record.partitionKey!,
```
jvh-aws, line 40: Can we throw a state error with appropriate message here?

packages/kinesis/amplify_kinesis_dart/test/auto_flush_scheduler_test.dart
```diff
@@ -1,9 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import 'package:amplify_kinesis_dart/src/impl/auto_flush_scheduler.dart';
```
jvh-aws, line 4: Doesn't this test record cache functionality? Can we move the test there?

packages/kinesis/amplify_kinesis_dart/test/in_memory_record_storage_test.dart
```diff
@@ -4,16 +4,20 @@
 import 'dart:typed_data';
```
jvh-aws, line 4: Doesn't this test record cache functionality? Can we move the test there? (Same for the other recordcache classes)

packages/kinesis/amplify_kinesis_dart/test/in_memory_record_storage_test.dart
```diff
@@ -112,21 +116,21 @@ void main() {
...
-          RecordInput.now(
+          createKinesisRecordInputNow(
             data: Uint8List.fromList([2]),
             partitionKey: 'pk',
             streamName: 'stream-b',
           ),
         );
         await storage.addRecord(
-          RecordInput.now(
+          createKinesisRecordInputNow(
```
jvh-aws, line 133: nit: Can this be a factory method which is more idiomatic?

packages/kinesis/amplify_kinesis_dart/test/kinesis_sender_test.dart
```diff
@@ -79,7 +79,7 @@ void main() {
 
       final sender = _DirectMockSender(mockClient, maxRetries: maxRetries);
 
-      await sender.putRecords(
+      await sender.sendBatch(
```
jvh-aws, line 82: nit: Why are we renaming that function?

packages/kinesis/amplify_kinesis_dart/test/record_validation_test.dart
```diff
@@ -18,13 +18,13 @@ library;
 import 'dart:convert';
 import 'dart:typed_data';
 
-import 'package:amplify_kinesis_dart/src/exception/record_cache_exception.dart';
```
jvh-aws, line 21: Doesn't part of this test record cache's sqlite storage functionality? Can we move the test to the record cache package? (Same for the other record cache classes)

packages/kinesis/amplify_kinesis_dart/test/sqlite_record_storage_cache_accuracy_test.dart
```diff
@@ -11,7 +11,7 @@ library;
 import 'dart:typed_data';
```
jvh-aws, line 11: Doesn't this test record cache functionality? Can we move the test there? (Same for the other recordcache classes)

packages/kinesis/amplify_record_cache_dart/lib/src/client/record_client.dart
```diff
@@ -21,22 +20,25 @@ import 'package:smithy/smithy.dart'
...
   final RecordStorage _storage;
-  final KinesisSender _sender;
+  final Sender _sender;
   final int _maxRetries;
   final Logger _logger = AmplifyLogging.logger('RecordClient');
 
   bool _flushing = false;
 
+  /// Provides access to the underlying storage (for testing).
+  RecordStorage get storage => _storage;
```
jvh-aws, line 40: annotate with @visibleForTesting from package:meta

jvh-aws: (same in other locations where we expose members for testing only)

packages/kinesis/amplify_record_cache_dart/pubspec.yaml
```diff
@@ -0,0 +1,24 @@
+name: amplify_record_cache_dart
+description: Shared record caching infrastructure for Amplify streaming clients (Kinesis Data Streams, Firehose). Provides offline storage, batching, retry logic, and auto-flush scheduling.
```
jvh-aws, line 2: This should be marked as internal

packages/kinesis/amplify_record_cache_dart/pubspec.yaml
```diff
@@ -0,0 +1,24 @@
...
+version: 0.1.0
```
jvh-aws, line 3: Also, I think the package is missing the license and readme.

packages/kinesis/amplify_firehose/example/integration_test/firehose_integration_test.dart
```diff
@@ -0,0 +1,612 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jvh-aws, line 1: This seems to contain a lot of duplicated code from the kinesis client's e2e tests. Can we avoid duplicating it and instead extract common tests into a module in recordcache? (As done in Android/Swift)

packages/kinesis/amplify_record_cache_dart/lib/src/storage/record_storage_sqlite.dart
```diff
@@ -70,7 +75,7 @@ final class SqliteRecordStorage extends RecordStorage {
         .insert(
           KinesisRecordsCompanion.insert(
             streamName: record.streamName,
-            partitionKey: record.partitionKey,
+            partitionKey: Value(record.partitionKey ?? ''),
```
jvh-aws, line 78: Didn't we discuss to have a separate schema? If that's not feasible, can we at least have a nullable column?

jvh-aws (dismissed):

packages/kinesis/amplify_kinesis_dart/lib/src/impl/kinesis_sender.dart
ekjotmultani: yes

packages/kinesis/amplify_record_cache_dart/lib/src/client/record_client.dart
ekjotmultani: good call

packages/kinesis/amplify_record_cache_dart/pubspec.yaml
ekjotmultani: yes

packages/kinesis/amplify_record_cache_dart/lib/src/storage/record_storage_sqlite.dart
ekjotmultani: we discussed and after investigation that it would be more effort and code than its worth to separate it out, we can make it nullable though

packages/kinesis/amplify_kinesis_dart/test/kinesis_sender_test.dart
ekjotmultani: it was renamed because the shared Sender interface uses sendBatch

packages/kinesis/amplify_kinesis_dart/test/in_memory_record_storage_test.dart
ekjotmultani: its a kinesis specific function (it does a calculation with the size of the record that includes the partition key), making this a factory would have kinesis specific concepts in the shared package

packages/kinesis/amplify_firehose_dart/lib/src/impl/firehose_sender.dart
ekjotmultani: done

packages/kinesis/amplify_firehose_dart/lib/src/amplify_firehose_client.dart
ekjotmultani: the exception mapping is different per client (`AmplifyFirehoseException.from` vs `AmplifyKinesisException.from`), so sharing would require a generic callback parameter which adds complexity for minimal gain

packages/kinesis/amplify_record_cache_dart/pubspec.yaml
ekjotmultani: Tried @internal but it blocks imports from amplify_firehose_dart and amplify_kinesis_dart since they're separate packages in the monorepo. Went with publish_to: none + a doc comment instead — same effect for customers (can't install it from pub) without breaking our internal consumers.

thisisabhash (approved):

packages/kinesis/amplify_record_cache_dart/README.md
```diff
@@ -0,0 +1,5 @@
+# amplify_record_cache_dart
+
+Shared record caching infrastructure for Amplify streaming clients (Kinesis Data Streams, Firehose).
+
+This package is internal to the Amplify Flutter monorepo and is not published to pub.dev.
```
jvh-aws, line 5: Typo here. It'll be published to pub.dev

jvh-aws:
Fixes https://github.com/aws-amplify/amplify-flutter/issues/6288

PR #6876, opened 2026-04-14
feat(ci): Add workflow for bumping versions

cadivus:
Introduces a workflow to create version bump PRs. They look like this:  

Screenshot 2026-04-14 at 17 57 06

### DIFF
--- a/.github/workflows/version_bump.yaml
+++ b/.github/workflows/version_bump.yaml
@@ -1,0 +1,145 @@
+name: Version Bump
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pull-requests: write
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  version-bump:
+    name: Version Bump
+    runs-on: ubuntu-latest
+    steps:
+      - name: Git Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+        with:
+          submodules: true
+          fetch-depth: 0
+
+      - name: Git Config
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          USER_NAME=$(gh api "/users/${{ github.actor }}" --jq '.name // "${{ github.actor }}"')
+          USER_EMAIL="${{ github.actor_id }}+${{ github.actor }}@users.noreply.github.com"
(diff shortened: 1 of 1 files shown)


.github/workflows/version_bump.yaml
```diff
@@ -0,0 +1,141 @@
...
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+        with:
+          submodules: true
+          fetch-depth: 0
+
+      - name: Git Config
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          USER_NAME=$(gh api "/users/${{ github.actor }}" --jq '.name // "${{ github.actor }}"')
```
soberm, line 32: Can you check if there is a risk of potential command injection through user input?

.github/workflows/version_bump.yaml
```diff
@@ -0,0 +1,141 @@
+name: Version Bump
+on:
+  workflow_dispatch:
```
soberm, line 3: Maybe you want to add concurrency control to ensure there are no conflicts.

.github/workflows/version_bump.yaml
```diff
@@ -0,0 +1,141 @@
...
+          unset GITHUB_HEAD_REF
+          aft version-bump
+
+      - name: Collect version changes and prepare PR body
+        run: |
+          SUMMARY=""
+          # Find all pubspec.yaml files that were modified
+          for pubspec in $(git diff --name-only -- '**/pubspec.yaml'); do
+            # Get the package name from the pubspec
+            PKG_NAME=$(grep -m1 '^name:' "$pubspec" | sed 's/name: *//')
```
soberm, line 91: What if grep or sed fails?

.github/workflows/version_bump.yaml
cadivus: These variables are set by Github: https://docs.github.com/en/actions/reference/workflows-and-actions/contexts

.github/workflows/version_bump.yaml
cadivus: In the worst case, we get a weird PR description. But getting there is extremely unlikely since this means that the files are broken and not even the analyzers pass (so the PR and release is blocked).

soberm (dismissed):

.github/workflows/version_bump.yaml
cadivus: Done

soberm (approved):

PR #6970, opened 2026-05-08
feat(ci): Add release workflow

cadivus:
Adds a release workflow for publishing to pub.dev.

Todos (that's why it's still a draft):
 - Set up Codebuild
 - Configure packages on pub.dev

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -52,11 +52,8 @@ jobs:
         with:
           channel: stable
 
-      - name: Checkout
-        uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-          fetch-tags: true
+      - name: Git Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
 
       - name: Create publishing token (pub.dev)
         run: |
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -50,9 +50,7 @@ jobs:
       - name: Setup Flutter
         uses: subosito/flutter-action@1a449444c387b1966244ae4d4f8c696479add0b2 # 2.23.0
         with:
-          cache: true
-          channel: ${{ matrix.channel }}
-          flutter-version: ${{ matrix.flutter-version }}
+          channel: stable
 
       - name: Checkout
         uses: actions/checkout@v4
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -48,9 +48,11 @@ jobs:
           git config --global --add safe.directory '*'
 
       - name: Setup Flutter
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,121 @@
...
+          apt-get install -y git libicu-dev curl xz-utils clang cmake ninja-build pkg-config libgtk-3-dev jq
+          
+          # Install Github CLI
+          curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | sudo dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg
+          echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | sudo tee /etc/apt/sources.list.d/github-cli.list > /dev/null
+          sudo apt-get update -y
+          sudo apt-get install -y gh
+
+          # Fix Git ownership fix
+          git config --global --add safe.directory '*'
```
soberm, line 48: Can we scope this to the workspace instead of disabling it completely ?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,121 @@
...
+          channel: stable
+
+      - name: Git Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+
+      - name: Create publishing token (pub.dev)
+        run: |
+          set -eo pipefail
+          PUB_TOKEN=$(curl --retry 5 --retry-connrefused -sLS "${ACTIONS_ID_TOKEN_REQUEST_URL}&audience=https://pub.dev" -H "User-Agent: actions/oidc-client" -H "Authorization: Bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN" | jq -r '.value')
+          echo "PUB_TOKEN=${PUB_TOKEN}" >> $GITHUB_ENV
```
soberm, line 62: Can we mask it ?

`echo "::add-mask::${PUB_TOKEN}"`

cadivus: https://stackoverflow.com/questions/70558558/how-to-mask-environment-variables-created-in-github-when-running-a-workflow

.github/workflows/release.yaml
```diff
@@ -0,0 +1,121 @@
+name: Release
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - '**/pubspec.yaml'
```
soberm, line 7: This will run on every version bump. Do you want to run release verifications (run E2E) before doing the actual release as part of this workflow?

cadivus: No. They ran after the version bump was merged. They get merged after the workflows passed.

soberm (approved):

PR #6985, opened 2026-05-19
feat: add amplify_push package

ekjotmultani:
## Summary

New Flutter plugin package `amplify_push` for standalone on-device push notification handling, decoupled from Pinpoint.

> This PR contains the **package implementation only**. To keep reviews scoped, the rest is split into stacked PRs based on this branch:
> - **Unit tests** (Dart + Android): #7078
> - **Example app**: #7079

### What is this?

A backend-agnostic push notifications client that handles:
- Device token lifecycle (FCM/APNs)
- Foreground notification routing
- Background notification processing (Android isolate, iOS callback)
- Tap/open notification handling + launch notification detection
- Permission request/status APIs
- Badge count (iOS)

No `Amplify.configure()` required. No AWS service dependency. Works standalone or with an optional `PushServiceProvider` for backend wiring (Connect client integration coming in Phase 2).

### Design

- **Standalone**: `AmplifyPushClient.create()` static factory — no Amplify plugin lifecycle
- **Result types**: Uses `Result<T>` from `amplify_foundation_dart`
- **Sealed exceptions**: `AmplifyPushException` hierarchy extending foundation `AmplifyException`
- **Optional provider**: `PushServiceProvider` interface with `onTokenReceived` and `onPushEvent` callbacks
- **Native code**: Ported from `amplify_push_notifications` (same platform channel bridge, backend-agnostic)
- **Android lifecycle**: depends on `flutter_plugin_android_lifecycle` (the plugin resolves the activity lifecycle via `FlutterLifecycleAdapter`)

### Usage

```dart
// Standalone — no backend
final result = await AmplifyPushClient.create(
  onBackgroundMessage: myHandler,
);

// With a provider (e.g., Connect)
final result = await AmplifyPushClient.create(
  provider: connectClient.asPushProvider(),
  onBackgroundMessage: myHandler,
);
```

### Verification

- `dart analyze`: 0 errors, 0 warnings (runs in this PR's CI)
- iOS umbrella header named to match the podspec module (`amplify_push.h`)
- Unit tests live in #7078 (31 Dart + 33 Android, all passing); the example app for on-device testing is in #7079

---

## Known follow-ups / not-yet-verified

- **Android background-isolate path**: needs on-device testing to confirm the background processor signals completion correctly when no Amplify re-init is performed
- **iOS AmplifyUtilsNotifications**: pod resolves and builds (verified via the example app), runtime behavior on device not yet tested
- **No Connect provider wired yet** (Phase 2 — depends on Connect client shipping)
- **Native code package namespaces**: still use `amplify_push_notifications` internally (cosmetic, channel names must match)


packages/amplify_push/lib/src/amplify_push_client.dart
```diff
@@ -0,0 +1,392 @@
...
+  void setBadgeCount(int count) {
+    if (_closed) return;
+    _hostApi.setBadgeCount(count).ignore();
+  }
+
+  /// Release resources held by this client.
+  ///
+  /// Cancels all provider subscriptions, clears background callbacks, and
+  /// disconnects the provider. The client cannot be reused after closing.
+  Future<void> close() async {
```
cadivus, line 375: The broadcast controller behind `onTokenReceived` never gets closed here. `onDone: tokenController.close` only fires if the source stream completes — cancelling the subscription won't trigger it. Worth holding a ref to the controller and calling `await _tokenController.close()` in here.

packages/amplify_push/lib/src/amplify_push_client.dart
```diff
@@ -0,0 +1,392 @@
...
+  Stream<PushNotificationMessage> get onNotificationOpened {
+    if (_closed) throw const PushClientClosedException();
+    return _openedStream;
+  }
+
+  /// The notification that launched the app, if any.
+  ///
+  /// Returns `null` if the app was launched by other means.
+  /// Consumed on first read (subsequent reads return `null`).
+  PushNotificationMessage? get launchNotification {
```
cadivus: This getter nullifies the field as a side effect of reading it. It's documented, but a getter that mutates state is easy to trip over. Something like `consumeLaunchNotification()` would read more honestly.

cadivus (commented):
One general note: some of the doc comments (the class template and the `create()` docs especially) run pretty long. Trimming them to a line or two would make them easier to scan — not blocking.

packages/amplify_push/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
```diff
@@ -0,0 +1,274 @@
...
+        applicationContext = flutterPluginBinding.applicationContext
+        sharedPreferences = applicationContext.getSharedPreferences(
+            PushNotificationPluginConstants.SHARED_PREFERENCES_KEY, Context.MODE_PRIVATE
+        )
+        refreshToken()
+    }
+
+    override fun onDetachedFromEngine(binding: FlutterPlugin.FlutterPluginBinding) {
+        mainBinaryMessenger = null
+        _flutterEngineCache.clear()
```
cadivus: `FlutterEngineCache.getInstance()` is a process-wide singleton, so `clear()` drops every cached engine in the app — including this plugin's background engine and anything other plugins or an add-to-app host put there. Safer to remove just our own key: `_flutterEngineCache.remove(FLUTTER_ENGINE_ID)`.

packages/amplify_push/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
```diff
@@ -0,0 +1,274 @@
...
+    override fun requestPermissions(
+        withPermissionOptions: PushNotificationsHostApiBindings.PermissionsOptions,
+        result: PushNotificationsHostApiBindings.Result<Boolean>
+    ) {
+        scope.launch {
+            val res = PushNotificationPermission(applicationContext).requestPermission()
+
+            if (res is PermissionRequestResult.Granted) {
+                result.success(true)
+            } else if (res is PermissionRequestResult.NotGranted){
```
cadivus: This handles `Granted` and `NotGranted` but has no terminal `else`. If `requestPermission()` ever returns another result (or null), `result.success(...)` never fires and the Dart future hangs forever. Worth a fallback `result.success(false)` to be safe.

cadivus (commented):
Went over the native side too. Generated files (Pigeon `.m`/`.h`/`.java`/`.g.dart`) aside, a couple of Android things inline.

One minor iOS nit: in `AmplifyPushNotificationsPlugin.swift` the badge getter/setter use `applicationIconBadgeNumber` (lines ~95 and ~99), which is deprecated as of iOS 17 in favor of `UNUserNotificationCenter.setBadgeCount(_:)`. Not urgent, just expect deprecation warnings.

packages/amplify_push/lib/src/amplify_push_client.dart
ekjotmultani: I've reworked close() to hold the token StreamController and close it directly, on top of cancelling the three provider subscriptions (token, foreground, opened) so nothing keeps listening after close() now.

packages/amplify_push/lib/src/amplify_push_client.dart
ekjotmultani: Agreed, renamed it to `consumeLaunchNotification() `

packages/amplify_push/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
ekjotmultani: Good point, switched it to remove only our own keys (`FLUTTER_ENGINE_ID` and `BACKGROUND_ENGINE_ID`)

packages/amplify_push/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/AmplifyPushNotificationsPlugin.kt
ekjotmultani: Nice catch, converted it to a `when` with a terminal `else` that returns `success(false)`, and wrapped the whole thing in a try/catch that surfaces failures through result.error, so the future always completes now.

packages/amplify_push/lib/src/amplify_push_client.dart
```diff
@@ -0,0 +1,395 @@
...
+@visibleForTesting
+const foregroundNotificationEventChannel = EventChannel(
+  'com.amazonaws.amplify/push_notification/event/FOREGROUND_MESSAGE_RECEIVED',
+);
+
+/// Background method channel for signaling completion.
+@visibleForTesting
+const backgroundChannel = MethodChannel(
+  'plugins.flutter.io/amplify_push_notification_plugin_background',
+);
```
cadivus, line 41: I think this channel is never invoked. Don't you need to run  
```Dart
invokeMethod('amplifyBackgroundProcessorFinished')
```
somewhere?

cadivus (changes requested):
It looks to me like something about the Android background code is broken.

Issue #6998, opened 2026-05-27
PushNotificationBackgroundService still uses deprecated JobIntentService, crashes on Android 16 / low-end devices


houzx2581396:
## Description
  `PushNotificationBackgroundService` in `amplify_push_notifications` still extends the deprecated
  `androidx.core.app.JobIntentService`. On Android 16 (SDK 36) and on low-end Samsung / Oppo / Redmi devices, this triggers a
  race in `JobParameters.completeWork()` that crashes the app with:

  ```
  java.lang.RuntimeException: An error occurred while executing doInBackground()
  Caused by: java.lang.IllegalArgumentException: Given work is not active: JobWorkItem{...
    cmp=<pkg>/com.google.firebase.iid.FirebaseInstanceIdReceiver (has extras) } dcount=1}
    at android.app.job.JobParameters.completeWork(JobParameters.java:503)
    at androidx.core.app.JobIntentService$JobServiceEngineImpl$WrapperWorkItem.complete(JobIntentService.java:273)
    at androidx.core.app.JobIntentService$CommandProcessor.doInBackground(JobIntentService.java:399)
  ```

  The TODO has been in the code for several minor releases:
  https://github.com/aws-amplify/amplify-flutter/blob/main/packages/notifications/push/amplify_push_notifications/android/src/
  main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt#L24

  ```kotlin
  //TODO(Samaritan1011001): Replace deprecated JobIntentService
  class PushNotificationBackgroundService : JobIntentService(), ...
  ```

  ## Reproduction
  - amplify_push_notifications: 2.10.2
  - amplify_push_notifications_pinpoint: 2.10.0
  - Flutter: 3.38.10
  - firebase-messaging (transitive): 25.0.1
  - Devices: Samsung A16/A53/A54, Oppo (MediaTek MT6769/MT6873/MT6897), Redmi (Mediatek)
  - OS distribution we see in Play Vitals: Android 16 Beta 52.5%, Android 14 16.9%, Android 13 10.2%
  - Crash frequency increases when the app is held in foreground/idle (e.g. force-update modal) while FCM pushes arrive.

  ## Expected
  `PushNotificationBackgroundService` should use `WorkManager` (or handle the FCM message synchronously inside
  `PushNotificationFirebaseMessagingService.onMessageReceived`, which already gets ~10s of execution time) instead of the
[...]

PR #7000, opened 2026-05-30
fix(push_notifications): Migrate from deprecated JobIntentService to WorkManager

jvh-aws:
*Issue #, if available:* #6998

*Description of changes:* 

Migrates the Android background push handler from the deprecated JobIntentService to WorkManager. The public `enqueueWork` API is unchanged. Notifications are processed on a serialized work chain, and the payload reaching the Dart background callback is identical.

### DIFF
--- a/packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
+++ b/packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
@@ -240,18 +240,33 @@ class PushNotificationBackgroundWorker(
             return Result.success()
         }
 
+        // Never return Result.failure() from this worker: failure cancels all
+        // work APPENDed behind this one on the unique chain, so one bad
+        // notification would silently kill every queued notification after it.
+        // Log and return success instead so the chain keeps processing.
+
         // WorkManager has no built-in retry cap; bail out so a permanent failure
         // (e.g. a misconfigured app) can't retry indefinitely.
         if (runAttemptCount >= MAX_RETRY_ATTEMPTS) {
-            Log.w(TAG, "Giving up after $runAttemptCount attempts to process push notification.")
-            return Result.failure()
+            Log.e(
+                TAG,
+                "Giving up after $runAttemptCount attempts to process push notification; " +
+                    "dropping it so queued notifications still process."
+            )
+            return Result.success()
         }
 
         Log.i(TAG, "Handling work in PushNotificationBackgroundWorker")
 
         when (startPushNotificationService(applicationContext)) {
             // No Dart callback registered — retrying can't fix a missing handle.
-            EngineStartResult.CALLBACK_NOT_REGISTERED -> return Result.failure()
+            EngineStartResult.CALLBACK_NOT_REGISTERED -> {
+                Log.e(
+                    TAG,
+                    "No Dart background callback registered; dropping notification."
+                )
+                return Result.success()
+            }
 
             // Transient startup failure; let WorkManager retry with backoff.
             EngineStartResult.FAILED -> return Result.retry()
--- a/packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
+++ b/packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
@@ -5,9 +5,11 @@ package com.amazonaws.amplify.amplify_push_notifications
 
 import android.content.Context
 import android.content.Intent
+import android.os.Bundle
 import android.os.Handler
 import android.util.Log
 import androidx.work.Data
+import androidx.work.ExistingWorkPolicy
 import androidx.work.OneTimeWorkRequestBuilder
 import androidx.work.WorkManager
 import androidx.work.Worker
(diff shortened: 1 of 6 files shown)


packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
```diff
@@ -141,30 +204,150 @@ class PushNotificationBackgroundService : JobIntentService(), MethodChannel.Meth
...
+                TAG,
+                "WorkManager is not initialized, dropping background notification. " +
+                    "If the host app uses on-demand WorkManager initialization, ensure " +
+                    "its Application implements androidx.work.Configuration.Provider.",
+                exception
+            )
+            return
+        }
+        val workRequest = OneTimeWorkRequestBuilder<PushNotificationBackgroundWorker>()
+            .setInputData(extras.toWorkData())
```
cadivus: Do we introduce a new payload ceiling here?
https://developer.android.com/reference/androidx/work/Data#MAX_DATA_BYTES()  
  
https://github.com/androidx/androidx/blob/0bd71ea9f0a84b6c67d6cc068dc81cae7febbec2/work/work-runtime/src/main/java/androidx/work/Data_.kt#L564

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
```diff
@@ -141,30 +204,150 @@ class PushNotificationBackgroundService : JobIntentService(), MethodChannel.Meth
...
+    for (key in keySet()) {
+        when (val value = get(key)) {
+            is String -> builder.putString(key, value)
+            is Int -> builder.putInt(key, value)
+            is Long -> builder.putLong(key, value)
+            is Boolean -> builder.putBoolean(key, value)
+            is Float -> builder.putFloat(key, value)
+            is Double -> builder.putDouble(key, value)
+            is ByteArray -> builder.putByteArray(key, value)
+            else -> if (value != null) builder.putString(key, value.toString())
```
cadivus: This is potentially lossy. Did you cover all the cases? Do we need to treat arrays?

houzx2581396:
Thank you for putting this PR together! This fix is highly anticipated and would be a huge help for our Flutter project currently in development.

Just wondering if there is an estimated timeline for reviewing and merging this? It would really help us with our roadmap planning. Appreciate all your hard work on this SDK! 🙏

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
cadivus: Data.MAX_DATA_BYTES is 10KB. In practice FCM caps payloads at 4KB so this shouldn't trigger, but I added a guard: if `build()` throws we log + drop the notification instead of crashing the messaging service. See f12d251.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
cadivus: Fixed — now handles all array types Data supports (String[], Boolean[], Byte[], Int[], Long[], Float[], Double[]) and drops unsupported types with a warning instead of stringifying them. Added round-trip tests for arrays.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
```diff
@@ -141,30 +204,214 @@ class PushNotificationBackgroundService : JobIntentService(), MethodChannel.Meth
...
         result.success(null)
     }
 
-    override fun onHandleWork(intent: Intent) {
-        Log.i(TAG, "Handling work in PushNotificationBackgroundService")
-        synchronized(serviceStarted) {
-            if (!serviceStarted.get()) {
-                // Queue up notification events while background processor is being executed
-                queue.add(intent)
+    override fun doWork(): Result {
```
mattcreaser, line 237: From this function I think we need to replace the `Result.failure()` with `Result.success()`, see reasoning in other comment.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
```diff
@@ -141,30 +204,214 @@ class PushNotificationBackgroundService : JobIntentService(), MethodChannel.Meth
...
+            return
+        }
+        val workRequest = OneTimeWorkRequestBuilder<PushNotificationBackgroundWorker>()
+            .setInputData(inputData)
+            .build()
+        // Serialize work onto a single chain so concurrent workers don't race on the
+        // shared background Flutter engine.
+        workManager.enqueueUniqueWork(
+            UNIQUE_WORK_NAME,
+            ExistingWorkPolicy.APPEND_OR_REPLACE,
```
mattcreaser, line 340: There's a subtle issue here: If the work for **notification A** is still pending when **notification B** arrives, we'll _append_ the work here, as per design. The issue is that if the work for **notification A** subsequently _fails_ then this will cause the work for **notification B** to be cancelled.

The fix is to make sure the work items don't returned a failure result. We can emit logging of the error but just return a success result so that the next queued item will be processed.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
cadivus: Done in 55075c5 - all paths now return `Result.success()` with error logging so the chain keeps going.

packages/notifications/push/amplify_push_notifications/android/src/main/kotlin/com/amazonaws/amplify/amplify_push_notifications/PushNotificationBackgroundService.kt
cadivus: Fixed in 55075c5. Worker never returns `failure()` anymore - logs the error and returns `success()` so queued notifications behind it still process.

mattcreaser (approved):

cadivus:
Hi @houzx2581396,
the fix is part of our latest release (https://github.com/aws-amplify/amplify-flutter/releases/tag/v2.13.0).

PR #7003, opened 2026-06-01
feat(native_legacy_wrapper): add Swift Package Manager support

ekjotmultani:
*Issue #, if available:*

*Description of changes:*


packages/amplify_native_legacy_wrapper/darwin/amplify_native_legacy_wrapper/Package.swift
```diff
@@ -0,0 +1,32 @@
...
+        .package(url: "https://github.com/aws-amplify/amplify-ios.git", exact: "1.29.2")
+    ],
+    targets: [
+        .target(
+            name: "amplify_native_legacy_wrapper",
+            dependencies: [
+                .product(name: "FlutterFramework", package: "FlutterFramework"),
+                .product(name: "Amplify", package: "amplify-ios"),
+                .product(name: "AWSPluginsCore", package: "amplify-ios"),
+                .product(name: "AWSCognitoAuthPlugin", package: "amplify-ios")
```
cadivus, line 28: You name the product `AWSCognitoAuthPlugin`, but import `AmplifyPlugins` in https://github.com/aws-amplify/amplify-flutter/blob/feat/spm-amplify-native-legacy-wrapper/packages/amplify_native_legacy_wrapper/darwin/amplify_native_legacy_wrapper/Sources/amplify_native_legacy_wrapper/AmplifyNativeLegacyWrapperPlugin.swift  
  
You see this issue here too:  
https://github.com/aws-amplify/amplify-flutter/actions/runs/26767926955/job/78899276702?pr=7003  
  
```
Swift Compiler Error (Xcode): Unable to resolve module dependency: 'AmplifyPlugins'
import AmplifyPlugins
       ^
/Users/runner/work/amplify-flutter/amplify-flutter/packages/amplify_native_legacy_wrapper/darwin/amplify_native_legacy_wrapper/Sources/amplify_native_legacy_wrapper/AmplifyNativeLegacyWrapperPlugin.swift:4:7
```
  
You need to change the import to:  
```
import AWSCognitoAuthPlugin
```

cadivus (dismissed):

packages/amplify_native_legacy_wrapper/darwin/amplify_native_legacy_wrapper/Package.swift
ekjotmultani: Thanks for catching this. Fixed in the latest push.

Rather than swapping import AmplifyPlugins → import AWSCognitoAuthPlugin outright, I used a conditional import:

```

#if canImport(AmplifyPlugins)
import AmplifyPlugins
#else
import AWSCognitoAuthPlugin
#endif
```

The reason: the two build systems expose the Cognito plugin under different module names. CocoaPods bundles it in the AmplifyPlugins umbrella module, while SPM (amplify-swift) ships it as the standalone AWSCognitoAuthPlugin product. A bare import AWSCognitoAuthPlugin would fix SPM but break the CocoaPods build. The canImport guard resolves correctly under both.

harsh62 (approved):

PR #7004, opened 2026-06-01
feat(datastore): add Swift Package Manager support

ekjotmultani:
Migrates amplify_datastore to SPM with 5 targets mirroring the existing podspec subspecs:
- AmplifyInternal (vendored Amplify framework source)
- AWSPluginsCoreInternal (vendored AWSPluginsCore source)
- AWSDataStorePluginInternal (vendored DataStore plugin, depends on SQLite.swift 0.13.2)
- amplify_datastore_objc (ObjC plugin shim)
- amplify_datastore (Swift plugin code, main target)

No native source code was modified — purely structural reorganization.

*Issue #, if available:*

*Description of changes:*


packages/amplify_datastore/darwin/amplify_datastore/Package.swift
```diff
@@ -0,0 +1,65 @@
...
+                .process("Resources")
+            ]
+        ),
+        .target(
+            name: "amplify_datastore_objc",
+            dependencies: [
+                .product(name: "FlutterFramework", package: "FlutterFramework")
+            ],
+            publicHeadersPath: "include"
+        ),
```
cadivus: `AmplifyDataStorePlugin.m` imports `amplify_datastore-Swift.h`, which SPM doesn't generate. This workflow shows it:  
https://github.com/aws-amplify/amplify-flutter/actions/runs/26775304466/job/78925197409?pr=7004  
  
I think you need to do more to add SPM support.

cadivus (dismissed):

packages/amplify_datastore/darwin/amplify_datastore/Package.swift
ekjotmultani: Good catch. Fixed in the latest push.

1. Removed the `AmplifyDataStorePlugin` ObjC shim. It only forwarded `register(with:)` to the Swift class and imported the generated `amplify_datastore-Swift.h`, which SPM doesn't produce. iOS registers `SwiftAmplifyDataStorePlugin` directly via `pluginClass`, so the shim was already dead code on iOS. This matches what `amplify_auth_cognito` does.
2. Collapsed the vendored sources into a single SPM target. The CocoaPods podspec compiled its subspecs (`Amplify`, `AWSPluginsCore`, `AWSDataStorePlugin`) into one module, so the vendored files reference each other without imports.

The SPM iOS build and example build pass now. I also confirmed locally that the example still builds under CocoaPods, so existing consumers aren't affected.

harsh62 (approved):

PR #7006, opened 2026-06-04
fix(ci): Use newer iOS version

cadivus:
Old one is not compatible with the runner


cadivus:
Workflows run:

Screenshot 2026-06-04 at 15 12 05

mattcreaser (approved):

PR #7024, opened 2026-06-12
fix(core): hide internal exports from core

jvh-aws:
*Issue #, if available:*

*Description of changes:* hide internal exports from core


ekjotmultani (approved):

cadivus (changes requested):
Should be fixed with https://github.com/aws-amplify/amplify-flutter/pull/7033 so I think you can close this. Besides this, your way creates a breaking change.

PR #7029, opened 2026-06-17
feat: add event enrichment client package

ekjotmultani:
## Summary

Event Enrichment client for analytics — the on-device companion that stamps device/app/session/identity context onto events and emits a Pinpoint-compatible JSON envelope, so events sent via the Kinesis/Firehose clients keep the analytics context Pinpoint used to provide automatically.

> This PR contains **core package implementation only** (lib files, no example, no tests). The example app and unit tests are in stacked PRs based on this branch.

### PR Stack (3-way split)

| PR | Branch | Content |
|---|---|---|
| **This PR (#7029)** | `feat/event-enrichment-client` | Core package (lib only — 22 files) |
| #7076 | `feat/event-enrichment-client-example` | Example app (92 files) |
| #7068 | `feat/event-enrichment-client-tests` | Unit tests (5 files) |

### Structure

Dart/Flutter split, mirroring the Kinesis packages:
- **`amplify_event_enrichment_dart`** — pure-Dart core (client, `EnrichedEvent`, `SessionManager`, global fields, metadata, `EventSink` interface). No Flutter dependency.
- **`amplify_event_enrichment`** — Flutter wrapper: lifecycle observer, `SharedPreferencesClientIdProvider`, injectable `DeviceMetadataProvider`, re-exports the core.

Built on v3 Amplify Foundation (`amplify_foundation_dart`, `Result<T>`, sealed exceptions, `AmplifyLogging`).

### Key decisions

- **Shared device id**: `shared_preferences` under `com.amplifyframework.device_id`, read-or-create — a cross-package contract shared with the Connect client (whichever inits first generates the UUID; the other reads it). Maps to native SharedPreferences/NSUserDefaults so a device gets one id everywhere.
- **Zero default dependencies**: `device_info_plus` is not a core dependency; device metadata comes from an injectable `DeviceMetadataProvider` defaulting to `dart:io Platform`. The `EventSink` ships as an interface only (no default sinks) — keeps enrichment transport-agnostic.

### Verification

- Example app: see #7076
- Tests: see #7068 (41 Dart + 6 Flutter, all passing)


cadivus:
Please make sure to generate and commit workflows using `aft` for these new packages.

packages/amplify_event_enrichment/amplify_event_enrichment/pubspec.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+topics:
+  - aws
+  - analytics
+  - event-enrichment
+  - aws-amplify
+
+environment:
+  sdk: ^3.10.0
+  flutter: ">=3.38.1"
```
cadivus: We just bumped both, please adapt to that (https://github.com/aws-amplify/amplify-flutter/pull/7098)

packages/amplify_event_enrichment/amplify_event_enrichment/pubspec.yaml
```diff
@@ -0,0 +1,39 @@
...
+  windows:
+  linux:
+  web:
+
+dependencies:
+  amplify_event_enrichment_dart: ">=0.1.0 <0.2.0"
+  amplify_foundation_dart: ">=2.11.0 <2.12.0"
+  flutter:
+    sdk: flutter
+  shared_preferences: ^2.3.0
```
cadivus: Please align to our other packages. They use `^2.0.15`

cadivus (commented):
Please add our LICENSE file. Also, a README is missing.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/metadata/device_metadata_provider.dart
```diff
@@ -0,0 +1,38 @@
...
+/// wire a custom provider (e.g. using `device_info_plus`) if those are needed.
+/// {@endtemplate}
+class IODeviceMetadataProvider implements DeviceMetadataProvider {
+  /// {@macro amplify_event_enrichment.io_device_metadata_provider}
+  const IODeviceMetadataProvider();
+
+  @override
+  Future<DeviceMetadata> getDeviceMetadata() async {
+    return DeviceMetadata(
+      platform: Platform.operatingSystem,
```
cadivus: This breaks on web.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/metadata/device_metadata_provider.dart
```diff
@@ -0,0 +1,38 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:io' show Platform;
```
cadivus: We cannot import `dart:io` if we want to be web compatible.

cadivus:
Please have a look at the pana package score (e.g. here: https://github.com/aws-amplify/amplify-flutter/actions/runs/28815090606/job/85452353135?pr=7029). It's a ogood indicator for misses.

packages/amplify_event_enrichment/amplify_event_enrichment/lib/src/event_enrichment_client_flutter.dart
```diff
@@ -0,0 +1,118 @@
...
+      '"$appId". When both are provided they must be the same value.',
+    );
+    final metadataProvider =
+        deviceMetadataProvider ?? const IODeviceMetadataProvider();
+    const clientIdProvider = SharedPreferencesClientIdProvider();
+
+    final results = await (
+      metadataProvider.getDeviceMetadata(),
+      clientIdProvider.getClientId(),
+    ).wait;
```
cadivus: Shouldn't we add error handling here?

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/event_enrichment_client.dart
```diff
@@ -0,0 +1,176 @@
...
+      attributes: mergedAttributes,
+      metrics: mergedMetrics,
+      device: _deviceMetadata,
+      app: _appMetadata,
+      sdk: _sdkMetadata,
+      clientId: _clientId,
+      userId: _userId,
+    );
+
+    _sink?.send(event);
```
cadivus: If this throws, we won't use Results for indicating the error => We need error handling

packages/amplify_event_enrichment/amplify_event_enrichment/pubspec.yaml
ekjotmultani: Done. Rebased onto main and bumped both to `^3.11.0` / `>=3.41.0`.

packages/amplify_event_enrichment/amplify_event_enrichment/pubspec.yaml
ekjotmultani: Done, aligned to `^2.0.15`.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/metadata/device_metadata_provider.dart
ekjotmultani: Good catch. The native path is behind a conditional import now with a web-safe stub, so web falls back cleanly.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/metadata/device_metadata_provider.dart
ekjotmultani: Fixed. `dart:io` is isolated in `device_metadata_io.dart` behind a conditional import, web gets the stub.

packages/amplify_event_enrichment/amplify_event_enrichment/lib/src/event_enrichment_client_flutter.dart
ekjotmultani: Added. Client id resolution falls back to a generated UUID now instead of propagating the error.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/event_enrichment_client.dart
ekjotmultani: Done, this goes through the `Result` path now.

ekjotmultani:
> Please make sure to generate and commit workflows using `aft`

Done. Generated workflows are committed and match current `aft` output.

ekjotmultani:
> Please have a look at the pana package score

Went through the pana report and addressed what it flagged (license, README, platform/web support).

ekjotmultani:
> Please add our LICENSE file. Also, a README is missing.

Added both, to each package.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/enriched_event.dart
```diff
@@ -0,0 +1,120 @@
...
+  ///
+  /// "3.1" identifies the layout of the envelope (the event, application,
+  /// client, device, and session sections and their field names) so
+  /// downstream consumers can tell which schema they are parsing. Bump this
+  /// only when the envelope structure changes in a way consumers must handle;
+  /// it is independent of the package version.
+  static const _eventVersion = '3.1';
+
+  /// Serializes to the analytics event envelope as a JSON string.
+  String toJson() {
```
cadivus: Do we wanna call it `toJSON`? Everywhere else in the repo, `toJSON` returns `Map<String, dynamic>`. One option could be to call it `toJsonString. The maybe better options would be to serialize it where we need the JSON string.  
  
Eventually, we should also have a look at https://github.com/aws-amplify/amplify-flutter/blob/main/AGENTS.md#8-serialization.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/event_enrichment_client_options.dart
```diff
@@ -0,0 +1,30 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+/// {@template amplify_event_enrichment.event_enrichment_client_options}
+/// Configuration options for `EventEnrichmentClient`.
+/// {@endtemplate}
+final class EventEnrichmentClientOptions {
+  /// {@macro amplify_event_enrichment.event_enrichment_client_options}
+  const EventEnrichmentClientOptions({
+    this.autoSessionTracking = true,
+    this.sessionTimeout = const Duration(seconds: 5),
```
cadivus, line 14: 5s feels low for a timeout. Is this really what we want here?

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/session/session.dart
```diff
@@ -0,0 +1,30 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'package:meta/meta.dart';
+
+/// {@template amplify_event_enrichment.session}
+/// Represents an app session with start/stop timestamps and duration.
+/// {@endtemplate}
+@immutable
+final class Session {
```
cadivus: Shouldn't we implement `== / hashCode` here to make sure same values equal? And is `AWSEquatable` still a thing here (not sure)? If so, https://github.com/aws-amplify/amplify-flutter/blob/main/AGENTS.md#6a-mixins-as-core-building-blocks might be interesting.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/enriched_event.dart
ekjotmultani: Good call. Went with your last option, `toJson()` returns the map now and the string serialization happens where it's sent.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/event_enrichment_client_options.dart
ekjotmultani: This is the backgrounded duration before a new session starts. It matches the Android enrichment client's default and the legacy Pinpoint auto session behavior, but we can change if it seems too low

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/session/session.dart
ekjotmultani: Yes `AWSEquatable` should still be the pattern. Added it with the four fields as props (pulled in `aws_common` for the mixin, same constraint the kinesis packages use).

packages/amplify_event_enrichment/amplify_event_enrichment/lib/src/shared_preferences_client_id_provider.dart
```diff
@@ -0,0 +1,36 @@
...
+import 'package:shared_preferences/shared_preferences.dart';
+import 'package:uuid/uuid.dart';
+
+/// SharedPreferences key for the persistent client/device ID.
+///
+/// This is a cross-package integration contract shared with the Connect client.
+/// Both packages use the same native store (SharedPreferences on Android,
+/// NSUserDefaults on iOS) so that a device gets ONE id everywhere — including
+/// Phase 2 native packages.
+const _storageKey = 'com.amplifyframework.device_id';
```
cadivus: NIT: shouldn't we have a shared constant to make sure this stays in sync?

packages/amplify_event_enrichment/amplify_event_enrichment/lib/src/event_enrichment_client_flutter.dart
```diff
@@ -0,0 +1,140 @@
...
+    EventSink? sink,
+  }) async {
+    assert(
+      appMetadata == null || appMetadata.appId == appId,
+      'appMetadata.appId ("${appMetadata.appId}") does not match the appId '
+      '"$appId". When both are provided they must be the same value.',
+    );
+    final metadataProvider =
+        deviceMetadataProvider ?? const PlatformDeviceMetadataProvider();
+    const clientIdProvider = SharedPreferencesClientIdProvider();
```
cadivus: Why is `metadataProvider` injectable but `clientIdProvider` isn't?

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/event_sink.dart
```diff
@@ -0,0 +1,15 @@
...
+
+/// {@template amplify_event_enrichment.event_sink}
+/// Interface for transporting enriched events to a destination.
+///
+/// Implement this to pipe events to Kinesis, Firehose, or any custom
+/// transport.
+/// {@endtemplate}
+abstract interface class EventSink {
+  /// Sends an enriched event to the configured destination.
+  void send(EnrichedEvent event);
```
cadivus: Should this really be sync? Isn't sending events async?

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/event_enrichment_client.dart
```diff
@@ -0,0 +1,182 @@
...
+        attributes: mergedAttributes,
+        metrics: mergedMetrics,
+        device: _deviceMetadata,
+        app: _appMetadata,
+        sdk: _sdkMetadata,
+        clientId: _clientId,
+        userId: _userId,
+      );
+
+      _sink?.send(event);
```
cadivus: This sink is a sync function as I found in the previous comment. In this case, it also means that try/catch is not gonna work here.

packages/amplify_event_enrichment/amplify_event_enrichment/lib/src/event_enrichment_client_flutter.dart
```diff
@@ -0,0 +1,140 @@
...
+import 'package:amplify_event_enrichment/src/flutter_lifecycle_observer.dart';
+import 'package:amplify_event_enrichment/src/shared_preferences_client_id_provider.dart';
+import 'package:amplify_event_enrichment_dart/amplify_event_enrichment.dart';
+import 'package:amplify_foundation_dart/amplify_foundation_dart.dart'
+    show AmplifyLogging, Result;
+import 'package:uuid/uuid.dart';
+
+/// {@template amplify_event_enrichment.event_enrichment_client_flutter}
+/// Flutter convenience wrapper that creates an [EventEnrichmentClient] with
+/// platform-specific providers wired automatically.
```
cadivus: Shouldn't we also wrap/expose session controls here? Won't most customers need that? If `autoSessionTracking == false`, can we even log anything?

packages/amplify_event_enrichment/amplify_event_enrichment/lib/src/shared_preferences_client_id_provider.dart
ekjotmultani: Pulled it into a named constant (`zClientIdStorageKey`) with a comment on why it can't drift: if the key changes, every existing install mints a new id and re-registers as a second device on the backend. Kept it local to this package for now. The Connect client has its own copy of the same literal, so making it genuinely shared means picking a home for it and touching a released package, and there's the same discussion open on the Android side. I'd rather do both packages in one dedicated PR than half of it here.

packages/amplify_event_enrichment/amplify_event_enrichment/lib/src/event_enrichment_client_flutter.dart
ekjotmultani: No good reason, just an oversight. It's injectable now, same shape as `deviceMetadataProvider`, defaulting to the shared preferences one. Added a `ClientIdProvider` interface next to its default impl so there's something to substitute. Tests cover the injected case and the failure fallback.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/event_sink.dart
ekjotmultani: You're right, it shouldn't. `send` returns `Future<void>` now. I followed `Sender` in `amplify_record_cache_dart` since that's the same kind of transport abstraction and it's what kinesis and firehose sit behind. `FutureOr` shows up in the repo for callback parameters but not for interface methods, so I stayed with plain `Future`. Sinks that do no async work can still just use an `async` body.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/event_enrichment_client.dart
ekjotmultani: Yeah, that was a real hole. `record` awaits the sink inside the try now, so a sink that suspends and then fails gets caught instead of escaping as an unhandled async error. That makes `record` return `Future<Result<EnrichedEvent>>`. Failure behavior is otherwise unchanged: logged and returned as an error `Result`, never thrown. Added a test with a sink that awaits before throwing, and confirmed it fails if the await is removed.

packages/amplify_event_enrichment/amplify_event_enrichment/lib/src/event_enrichment_client_flutter.dart
ekjotmultani: Yes, recording still works with it off. The first `record()` lazily starts a session, so you never get an event without one. That was only in the `autoSessionTracking` doc, so I've put it on the wrapper class doc too. Exposed the controls as well: `startSession`, `stopSession`, `handleAppPaused`, `handleAppResumed`, all delegating. The last two only matter when auto tracking is off, and they're guarded in `SessionManager` so calling them anyway can't double-count.

cadivus:
Do we need
```yaml
    - name: Event Enrichment
      summary: amplify_event_enrichment
      propagate: none
      packages:
        - amplify_event_enrichment
        - amplify_event_enrichment_dart

```
in https://github.com/aws-amplify/amplify-flutter/blob/main/pubspec.yaml#L69?

packages/amplify_event_enrichment/amplify_event_enrichment_dart/pubspec.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - aws
+  - analytics
+  - event-enrichment
+  - aws-amplify
+
+environment:
+  sdk: ^3.11.0
+
+dependencies:
+  amplify_foundation_dart: ">=2.11.0 <2.12.0"
```
cadivus: Please update this

packages/amplify_event_enrichment/amplify_event_enrichment_dart/pubspec.yaml
```diff
@@ -0,0 +1,26 @@
...
+  aws_common: ">=0.7.15 <0.8.0"
+  meta: ^1.16.0
+  uuid: ^4.5.0
```
cadivus: Please align with https://github.com/aws-amplify/amplify-flutter/blob/main/pubspec.yaml#L45

packages/amplify_event_enrichment/amplify_event_enrichment/pubspec.yaml
```diff
@@ -0,0 +1,39 @@
...
+  linux:
+  web:
+
+dependencies:
+  amplify_event_enrichment_dart: ">=0.1.0 <0.2.0"
+  amplify_foundation_dart: ">=2.11.0 <2.12.0"
+  flutter:
+    sdk: flutter
+  shared_preferences: ^2.0.15
+  uuid: ^4.5.0
```
cadivus: Please update the amplify libraries to the latest and align the external libs with https://github.com/aws-amplify/amplify-flutter/blob/main/pubspec.yaml#L16

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/event_sink.dart
```diff
@@ -0,0 +1,25 @@
...
+///
+/// Implement this to pipe events to Kinesis, Firehose, or any custom
+/// transport.
+///
+/// Sending is asynchronous: [send] returns a `Future` so the client can await
+/// delivery and observe failures. Implementations that do no async work can
+/// still return an already-completed future (an `async` method body with no
+/// `await` is enough).
+/// {@endtemplate}
+abstract interface class EventSink {
```
cadivus: Should we use this naming? It will collide with `dart:async` (https://api.dart.dev/dart-async/EventSink-class.html)

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/enriched_event.dart
```diff
@@ -0,0 +1,120 @@
...
+import 'package:amplify_event_enrichment_dart/src/metadata/sdk_metadata.dart';
+import 'package:amplify_event_enrichment_dart/src/session/session.dart';
+import 'package:meta/meta.dart';
+
+/// {@template amplify_event_enrichment.enriched_event}
+/// An analytics event enriched with device, app, session, and SDK metadata.
+///
+/// Use [toJson] to produce the structured analytics envelope as a
+/// JSON-compatible map.
+/// {@endtemplate}
```
cadivus, line 19: Shouldn't we act as described here: https://github.com/aws-amplify/amplify-flutter/blob/main/AGENTS.md#8-serialization
Instead, we implement everything from scratch here.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+## 0.1.0
+
+ - Initial release ([#7029](https://github.com/aws-amplify/amplify-flutter/pull/7029))
```
cadivus: I think the leading space is wrong here. All of our other CHANGELOGs don't have it

packages/amplify_event_enrichment/amplify_event_enrichment/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+## 0.1.0
+
+ - Initial release ([#7029](https://github.com/aws-amplify/amplify-flutter/pull/7029))
```
cadivus: Same as with the other CHANGELOG

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/session/session_manager.dart
```diff
@@ -0,0 +1,144 @@
...
+    switch (_state) {
+      case SessionState.paused:
+        _cancelTimer();
+        _state = SessionState.active;
+      case SessionState.stopped:
+        startSession();
+      case SessionState.active:
+        break;
+    }
+  }
```
cadivus, line 150: Din't we have a bug here if the customer themselves stopped the session? We would continue that too.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/event_enrichment_client.dart
```diff
@@ -0,0 +1,191 @@
...
+  /// The client cannot be reused after closing.
+  void close() {
+    _closed = true;
+    // Stop the session to record its end, then drop it so no stale session is
+    // readable after close.
+    _sessionManager
+      ..stopSession()
+      ..clearSession();
+    _logger.info('Client closed');
+  }
```
cadivus: Do we do something with the recorded data? Or will it be thrown away?

cadivus:
The package scores are not good. See:
https://github.com/aws-amplify/amplify-flutter/actions/runs/32055962365/job/95466058849?pr=7029
and
https://github.com/aws-amplify/amplify-flutter/actions/runs/32055962360/job/95466058764?pr=7029

Especially the first is bad. You can download the full reports from the artifacts and have a closer look.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/pubspec.yaml
ekjotmultani: Bumped to `>=2.12.1 <2.13.0`.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/pubspec.yaml
ekjotmultani: Done.

packages/amplify_event_enrichment/amplify_event_enrichment/pubspec.yaml
ekjotmultani: Updated `amplify_foundation_dart` to `>=2.12.1 <2.13.0` and `amplify_lints` to `>=3.1.6 <3.2.0` in both packages, and `uuid` to `^4.5.1`. Left `shared_preferences` at `^2.0.15` per your earlier comment. The new codegen deps (`build_runner`, `json_serializable`, `json_annotation`) match the root list too.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/event_sink.dart
ekjotmultani: Renamed to `EnrichedEventSender`.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/enriched_event.dart
ekjotmultani: Converted. The envelope isn't a field for field projection of `EnrichedEvent`, it regroups platform name/version, renames `manufacturer` to `make`, wraps the locale code, drops empty sections, and doesn't emit `eventId` at all. So annotating `EnrichedEvent` itself wouldn't produce the right shape. The envelope is modelled as its own set of private `json_serializable` classes and `toJson()` delegates to them.

I didn't use `zAmplifySerializable` since it lives in `amplify_core`, which this package doesn't depend on, and it doesn't set the snake_case renaming the envelope needs so we'd be overriding it everywhere anyway. Added a `zEventEnrichmentSerializable` in the package instead, same as how `amplify_foundation_dart` declares its own.

Output is unchanged, the existing envelope tests pass as written.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/CHANGELOG.md
ekjotmultani: Removed.

packages/amplify_event_enrichment/amplify_event_enrichment/CHANGELOG.md
ekjotmultani: Removed.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/session/session_manager.dart
ekjotmultani: Yeah, that's a real bug, thank you. The manager tracks whether a stop was explicit now: resume after an explicit stop stays stopped, resume after a timeout still starts fresh, and `record()` still lazily starts a session after any stop. Tests cover all three, including through `FlutterLifecycleObserver`. Android has the same behavior in its SessionManager, following up there separately.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/event_enrichment_client.dart
ekjotmultani: `record()` hands the event to the `EnrichedEventSender` and awaits it, and a failure comes back as an error `Result`.

Session end data was being thrown away though. `close()` computed the stop timestamp and duration and then dropped them one line later, and a public `stopSession()` did the same. Fixed now: session boundaries now emit `_session.start` and `_session.stop` events (same names legacy Analytics used, exposed as public constants) through full enrichment, with the stop carrying the ended session's id, timestamps and duration.

ekjotmultani:
> The package scores are not good.

Dug into both reports. The dart package's three deductions: the foundation constraint (fixed in the latest push, it's `>=2.12.1 <2.13.0` now), the missing example (added one, and the pubspec description was over pana's length cap so trimmed that too), and the repository URL check, which fails because the URL points at main and the package isn't merged yet. That last one resolves itself on merge.

The wrapper's 30 looks alarming but almost all of it is one cascading cause, pana resolves from pub.dev and `amplify_event_enrichment_dart` isn't published yet, so version solving fails and takes dartdoc, platform detection, static analysis and the dependency checks down with it. That's the normal pre-publish state for our stacked packages and heals when the pair publishes together. Its example comes with the example app PR in this stack, and the repository check heals on merge same as the dart one.

So post-merge and post-publish these should sit at or near full score. The real gaps in the reports were the dart example and the description length, both in now.

ekjotmultani:
> Do we need [the components entry] in pubspec.yaml?

Yes, thanks. Added, with `propagate: none` like kinesis and firehose since the Flutter and Dart packages are on independent version lines.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/session/session_manager.dart
cadivus: Logic fix looks right, thanks.

One thing left: the resume path trusts the pause timer. iOS suspends the Dart event loop while backgrounded, so the timer fires late and we don't control whether it or `didChangeAppLifecycleState(resumed)` wins. If resume wins, `_cancelTimer()` runs and a 20-minute background resumes as the same session, inflating duration and dropping a `_session.start`. Pinpoint compared timestamps instead of relying on a timer. Can we stash the pause time and decide on resume?

packages/amplify_event_enrichment/amplify_event_enrichment/lib/src/constants.dart
```diff
@@ -0,0 +1,12 @@
...
+
+/// SharedPreferences key for the persistent client/device ID.
+///
+/// This value is a cross-package integration contract: the Connect client
+/// reads and writes the same key in the same native store (SharedPreferences
+/// on Android, NSUserDefaults on iOS) so a device resolves to ONE id
+/// everywhere, including the Phase 2 native packages. Changing it re-mints an
+/// id for every existing install, which registers those devices a second time
+/// on the backend, so it must stay in lockstep across packages and platforms.
+const zClientIdStorageKey = 'com.amplifyframework.device_id';
```
cadivus, line 13: `SharedPreferences.getInstance()` prefixes keys with `flutter.`, so this lands at `flutter.com.amplifyframework.device_id`. A native SDK reading the raw platform store won't find it, so "including the Phase 2 native packages" doesn't hold. Connect uses the same key and API and documents the opposite in `shared_preferences_device_id_store.dart:17-21`.

Flutter-to-Flutter sharing with Connect is fine, so please just match Connect's wording, or switch to `SharedPreferencesAsync` / `setPrefix('')` if native parity is a real Phase 2 requirement.

Separately: `constants.dart` isn't exported from the barrel, so `[zClientIdStorageKey]` in the public dartdoc on `SharedPreferencesClientIdProvider` is a dangling link. Connect exports its copy as `connectDeviceIdKey`.

cadivus (commented):
Thanks for the last round, all of it checks out and I resolved those threads.

**Behaviour**

1. Session timeout relies on a `Timer` running while backgrounded, which iOS suspends. See the `session_manager.dart` thread.
2. The eager start in the constructor (`event_enrichment_client.dart:76-90`) hands `_session.start` to the sender before the caller can call `setUserId` or `addGlobalAttribute` - your example says as much at `amplify_event_enrichment_dart_example.dart:46-48`. So the first event of every launch is missing the user id and all globals, which is the one event downstream is most likely to segment on. Pinpoint started its session after `configure`. Either defer the eager start to the first `record()`, or let `create()` take an initial userId / globals.

**Hygiene**

3. `zClientIdStorageKey` comment is inaccurate, see inline.
4. Web is advertised but never compiled for web in CI, see inline on the dart pubspec.

**Nits**

- `Session` got `AWSEquatable`; `EnrichedEvent`, `DeviceMetadata`, `AppMetadata`, `SdkMetadata` and `EventEnrichmentClientOptions` didn't, and none use `AWSDebuggable`.
- `GlobalFieldsManager` is exported from the barrel but the client already fronts it.
- `propagate: none` copies the kinesis/firehose rationale, but both packages here are `0.1.0` so it doesn't apply. Drop it or say why.
- `eventId` is generated per event and never emitted. Worth a note on the field, not only on `toJson`.

`record()`, the explicit-stop flag and the `json_serializable` conversion all look good.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/pubspec.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+dependencies:
+  amplify_foundation_dart: ">=2.12.1 <2.13.0"
+  aws_common: ">=0.7.15 <0.8.0"
+  json_annotation: ^4.9.0
+  meta: ^1.16.0
+  uuid: ^4.5.1
+
+dev_dependencies:
+  amplify_lints: ">=3.1.6 <3.2.0"
```
cadivus, line 26: Could we add `build_test: ^3.1.1`? `aft` gates the browser jobs on exactly this dev dep (`generate_workflows_command.dart:283`), so the generated workflow is `dart_vm` only and the web stub is never compiled for a browser, despite the package advertising web. `amplify_kinesis_dart` and `amplify_record_cache_dart` both carry it and get `ddc_test` / `dart2js_test` / `dart2wasm_test`. Without it the earlier `dart:io` thread stays unverified by CI.

(`dart_native` is missing because there's no `test/` dir yet, which #7068 should fix.)

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/session/session_manager.dart
```diff
@@ -0,0 +1,315 @@
...
+  /// Resumes a paused session, or starts a new one if the session timeout
+  /// expired while backgrounded. Does nothing if tracking was stopped
+  /// explicitly via [stopSession] or [clearSession] — a session the customer
+  /// ended is not resurrected by a lifecycle transition.
+  /// Called when the app returns to foreground.
+  ///
+  /// Resumes a paused session, or starts a new one if the session timeout
+  /// expired while backgrounded. Does nothing if tracking was stopped
+  /// explicitly via [stopSession] or [clearSession] — a session the customer
+  /// ended is not resurrected by a lifecycle transition.
```
cadivus: Isn't this two times the same comment?

cadivus: I think the comments in general should be condensed.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/session/session_manager.dart
ekjotmultani: Done. Pause stashes a timestamp and both resume and record compare elapsed time against the timeout, so a suspended timer cannot keep a session alive. The stop is attributed to the pause time on every path, matching the Swift client.

packages/amplify_event_enrichment/amplify_event_enrichment/lib/src/constants.dart
ekjotmultani: Done. Reworded to match Connect and exported the constant so the link resolves.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/pubspec.yaml
ekjotmultani: Done. Added build_test ^3.1.1 and regenerated the workflow. One catch: aft only emits the browser jobs when the package has a test/ dir, so they appear in the stacked tests PR's workflow; this branch gets the path triggers.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/session/session_manager.dart
ekjotmultani: Done, deduped and condensed the doc comments generally.

ekjotmultani:
> Either defer the eager start to the first `record()`, or let `create()` take an initial userId / globals.

Went with create() params: initialUserId, initialGlobalAttributes and initialGlobalMetrics, applied before the eager start so the launch _session.start carries them. Kept the start at construction so duration measures from launch.

ekjotmultani:
Added AWSEquatable and AWSDebuggable across the models and options, dropped GlobalFieldsManager from the barrel, moved the eventId note onto the field, and kept propagate: none with a comment: the version lines split once released, like kinesis.

Also renamed the dart barrel to amplify_event_enrichment_dart.dart to match the kinesis/connect convention, fixed record() during an expired pause to close the dead session out first, and made a session started while backgrounded start paused so it can still time out.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/session/session_manager.dart
cadivus: The duplicate paragraph is gone, thanks. The second half of the ask isn't done for this file though: it grew rather than shrank.

| commit | lines | comment lines |
|---|---|---|
| 2b32738 | 315 | 146 |
| HEAD | 391 | 185 |

`ee9b09aa3` ("tighten the docs") doesn't touch `session_manager.dart`. It's now 47% comment. `startSession()` is a one-line delegation with a 19-line doc comment (L123-141), `sessionForRecording()` has 16 (L144-159), and most of it restates what the callee's own doc already says. Can we cut those down to the invariant plus the non-obvious bit?
This long comments make it look very AI sloppy...

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/session/session_manager.dart
```diff
@@ -0,0 +1,391 @@
...
+  /// timeout.
+  ///
+  /// Deliberately does not touch the explicit-stop flag, so the timeout path
+  /// and [startSession]'s implicit stop stay restartable by
+  /// [handleAppResumed].
+  Future<void>? _endCurrent({DateTime? at}) {
+    _cancelTimer();
+    _pausedAt = null;
+    if (_state == SessionState.stopped || _session == null) return null;
+    final stoppedAt = at ?? clock.now();
```
cadivus:
```suggestion
    final stoppedAt = at ?? _pausedAt ?? clock.now();
```

Don't we need to look at `_pausedAt` for when we paused the session before stopping it?

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/event_enrichment_client.dart
```diff
@@ -0,0 +1,331 @@
...
+  /// is enriched exactly like any other event.
+  EnrichedEvent _buildEvent(
+    String eventType,
+    Session session, {
+    Map<String, String>? attributes,
+    Map<String, double>? metrics,
+  }) => EnrichedEvent(
+    eventId: const Uuid().v4(),
+    eventType: eventType,
+    eventTimestamp: DateTime.now().millisecondsSinceEpoch,
```
cadivus: Why `DateTime` and not `clock` although you pulled in `clock` (https://github.com/aws-amplify/amplify-flutter/blob/d13bdcbf920122e14566bcf12764e912380e64ec/packages/amplify_event_enrichment/amplify_event_enrichment_dart/pubspec.yaml#L20)? Shouldn't we stick to one to stay consistent?

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/session/session_manager.dart
ekjotmultani: Done. Reworked the comments across both packages to match the auth and kinesis style, not just this file. session_manager.dart is 244 lines now, a third of what it was in comments, below the 2b32738 baseline. Dropped the private-member docs and kept the startSession and sessionForRecording docs to the invariant plus the non-obvious part.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/session/session_manager.dart
ekjotmultani: Yes. The suggestion as written would not work since _endCurrent nulls _pausedAt two lines earlier, so the pause capture moved into _endCurrent itself and the at parameter is gone. Every end path makes the same attribution from one line now. Added tests for a stop while paused, past the timeout, and after a resume.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/event_enrichment_client.dart
ekjotmultani: Done. Switched to clock.now().

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/session/session_manager.dart
```diff
@@ -0,0 +1,244 @@
...
+  Future<Session> sessionForRecording() {
+    if (_state == SessionState.paused) {
+      if (!_pauseExpired) return Future.value(_session!);
+      return _restart().then((_) => _session!);
+    }
+    if (_state == SessionState.stopped || _session == null) {
+      return _restart().then((_) => _session!);
+    }
+    return Future.value(_session!);
+  }
```
cadivus, line 94: What happens if we call it twice without awaiting it? Isn't this undefined behavior? I think we should have a lock/semaphore here. Not sure if we should copy it from other parts of the repo (I think s3 has one in `s3_upload_task.dart` and `amplify_core` has the `EventCompleter`) or if we should pull in a dependency.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/event_enrichment_client.dart
```diff
@@ -0,0 +1,232 @@
...
+  void addGlobalMetric(String key, double value) =>
+      _globalFields.addMetric(key, value);
+
+  /// Removes a global metric.
+  void removeGlobalMetric(String key) => _globalFields.removeMetric(key);
+
+  /// Ends any running session, emits its [zSessionStopEventType] event, and
+  /// releases resources. The client cannot be reused after closing.
+  Future<void> close() async {
+    if (_closed) return;
```
cadivus, line 225: Should we have locks for session management (not only for the `close` function)?

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/session/session_manager.dart
ekjotmultani: Looked into this. It's actually well defined without a lock because all the session state changes happen synchronously, there's no await between checking state and mutating it, so two unawaited calls just run one after the other on the event loop. Calling sessionForRecording twice like that gives you the same session back and one start event. Added tests for this and the double startSession case, and left a comment in the code explaining the invariant so we don't break it later.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/event_enrichment_client.dart
ekjotmultani: You were right to poke at close(). The session state was fine but a second unawaited close() would return before the first one actually finished tearing down, so its future completed before the final stop event went out. Changed it so repeated calls all get the same teardown future, and added a test that fails against the old version.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/session/session_manager.dart
ekjotmultani: Coming back to this one. The state changes are synchronous like I said, but there was still a problem on the return side. `sessionForRecording` was reading the session back after the restart's future resolved, so if `close()` landed in that timeframe it would null the session out from under it. It never threw, `record()` caught it and returned an error result, but it was there. Changed `_restart` to hand back the session it created so nothing reads state after an await anymore. Added a test for the race.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/event_enrichment_client.dart
```diff
@@ -0,0 +1,237 @@
...
+      await sender.send(_buildEvent(eventType, session));
+      _logger.verbose('Recorded event: $eventType');
+    } on Object catch (e, st) {
+      _logger.error('Failed to record event: $eventType', e, st);
+    }
+  }
+
+  /// Starts a new session and emits a [zSessionStartEventType] event, ending
+  /// any session already running.
+  Future<void> startSession() => _sessionManager.startSession();
```
cadivus, line 187: Prevent startSession from reviving a closed client

startSession bypasses _closed, so a caller can invoke it after close() and emit a new _session.start. Because later close() calls return the cached teardown, the revived session is never stopped.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/event_enrichment_client.dart
```diff
@@ -0,0 +1,237 @@
...
+
+      final event = _buildEvent(
+        eventType,
+        session,
+        attributes: attributes,
+        metrics: metrics,
+      );
+
+      // Awaited inside the try so async sender errors are caught here.
+      await _sender?.send(event);
```
cadivus, line 138: Serialize close with in-flight records

record checks _closed only before awaiting sessionForRecording(). A concurrent close() can stop and clear the session while this future yields; record then sends the captured event after the final stop and even after close() completes.

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/session/session_manager.dart
```diff
@@ -0,0 +1,257 @@
...
+      reported.onError<Object>((e, st) => _logger.error(failureMessage, e, st)),
+    );
+  }
+
+  void _onTimeoutExpired() {
+    // A timer that outlived its pause must not end the session after it.
+    if (_state != SessionState.paused) return;
+    final ended = _endCurrent();
+    if (ended == null) return;
+    _fireAndLog(ended, 'Failed to report the end of a session that timed out');
```
cadivus, line 232: Serialize a timeout stop before the next session start

The timeout path marks the state stopped and fires the asynchronous stop report without retaining it. If the app resumes while that sender future is pending, stopped-state handling reports the next session concurrently, so _session.start can reach the transport before the prior _session.stop.

packages/amplify_event_enrichment/amplify_event_enrichment/lib/src/shared_preferences_client_id_provider.dart
```diff
@@ -0,0 +1,39 @@
...
+  /// Returns the persistent client ID, creating one if it doesn't exist.
+  Future<String> getClientId();
+}
+
+/// {@template amplify_event_enrichment.shared_preferences_client_id_provider}
+/// Reads or creates a persistent client ID at [zClientIdStorageKey] in
+/// SharedPreferences.
+///
+/// The Connect client uses the same key, so whichever package initializes
+/// first generates the UUID and the other reads it.
```
cadivus, line 24: Qualify the shared client ID guarantee

Parallel initialization can let both providers read no value, generate different UUIDs, and return different IDs. Sharing a key guarantees one ID only when initialization is serialized.

```suggestion
/// The Connect client uses the same key. When initialization is serialized,
/// the first package creates the UUID and the other reads it.
```

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/exception/event_enrichment_exception.dart
```diff
@@ -0,0 +1,44 @@
...
+  /// {@macro amplify_event_enrichment.event_enrichment_exception}
+  const EventEnrichmentException({
+    required super.message,
+    required super.recoverySuggestion,
+    super.cause,
+  });
+}
+
+/// {@template amplify_event_enrichment.client_closed_exception}
+/// Thrown when an operation is attempted on a closed client.
```
cadivus, line 19: Describe the closed error as returned

The client returns this exception inside Result.error when record is called after close; it does not throw it. The comment states the wrong error-delivery contract.

```suggestion
/// Returned as [Result.error] when recording is attempted on a closed client.
```

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/session/session_manager.dart
```diff
@@ -0,0 +1,257 @@
...
+
+  void _pauseCurrent() {
+    _state = SessionState.paused;
+    _pausedAt = clock.now();
+    _pauseTimer = timerFactory(_sessionTimeout, _onTimeoutExpired);
+  }
+
+  // Returns the session it started along with the future that reports both
+  // boundaries, so a caller awaiting the report does not have to read
+  // _session back once it resolves.
```
cadivus, line 174: Do not claim every restart reports two boundaries

When no session is running, _endCurrent returns null and the future reports only the new start. The comment incorrectly says every restart reports both boundaries.

```suggestion
  // Returns the new session with a future that reports its start after any
  // prior end, avoiding a read of _session across the await.
```

packages/amplify_event_enrichment/amplify_event_enrichment_dart/lib/src/session/session_manager.dart
```diff
@@ -0,0 +1,257 @@
...
+    final ended = _endCurrent();
+    final started = _startFresh();
+    final reported = ended == null
+        ? _reportStart(started)
+        : ended.then((_) => _reportStart(started));
+    return (session: started, reported: reported);
+  }
+
+  // A paused session's stop is stamped at the pause, so no end path counts
+  // background time. Returns null when there was nothing to end.
```
cadivus, line 185: Limit the background-time claim to paused endings

A pause that resumes before timeout leaves the original session start unchanged, so a later active stop includes that background interval. Only a session ended while paused excludes it.

```suggestion
  // A session ended while paused is stamped at the pause. Returns null when
  // there was nothing to end.
```

packages/amplify_event_enrichment/amplify_event_enrichment_dart/example/amplify_event_enrichment_dart_example.dart
```diff
@@ -0,0 +1,64 @@
...
+
+import 'package:amplify_event_enrichment_dart/amplify_event_enrichment_dart.dart';
+import 'package:amplify_foundation_dart/amplify_foundation_dart.dart';
+
+/// A [EnrichedEventSender] that prints each enriched envelope, standing in for a real
+/// transport such as the Kinesis or Firehose client.
+///
+/// Sending is asynchronous, so a transport that suspends can be awaited and its
+/// failures observed. This one does no async work and still satisfies the
+/// contract with an `async` body.
```
cadivus, line 16: Suggestion: Condense the example sender documentation

The interface and async method already show the transport contract. The example needs only to say what this implementation does.

```suggestion
/// Prints each enriched envelope as JSON.
```

packages/amplify_event_enrichment/amplify_event_enrichment/lib/src/flutter_lifecycle_observer.dart
```diff
@@ -0,0 +1,45 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import 'package:amplify_event_enrichment_dart/amplify_event_enrichment_dart.dart';
+import 'package:flutter/widgets.dart';
+
+/// {@template amplify_event_enrichment.flutter_lifecycle_observer}
+/// Flutter implementation of [LifecycleObserver] using [WidgetsBindingObserver].
+///
+/// Listens to [AppLifecycleState] changes and calls the session manager's
+/// pause/resume methods accordingly.
```
cadivus, line 11: Suggestion: Condense the lifecycle observer summary

The implements clause, mixin, and switch already expose these mechanics. A shorter summary preserves the useful purpose without narrating the declaration and method body.

```suggestion
/// Bridges Flutter lifecycle changes to [SessionManager].
```

packages/amplify_event_enrichment/amplify_event_enrichment/lib/amplify_event_enrichment.dart
```diff
@@ -0,0 +1,14 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/// Amplify Event Enrichment client for Flutter.
+library;
+
+// Re-export everything from the Dart package.
+export 'package:amplify_event_enrichment_dart/amplify_event_enrichment_dart.dart';
+
+// Export Flutter-specific implementations.
```
cadivus, line 10: Suggestion: Remove narration above barrel exports

The export targets already distinguish the Dart barrel from the Flutter source files. Both comments repeat the adjacent syntax without adding a constraint or rationale.

cadivus (changes requested):
THanks for the updates! I think some comments need an update. Besides that, I have some small concerns about async behavior.

PR #7032, opened 2026-06-22
feat: Add wasm support to `amplify_core`, `amplify_foundation` and `aws_common`

cadivus:
This add wasm support to the base layer of amplify-flutter.

### DIFF
--- a/packages/amplify_core/build.yaml
+++ b/packages/amplify_core/build.yaml
@@ -9,6 +9,9 @@ targets:
           dart2js_args:
             - --define=dart.vm.product=true
             - --enable-asserts
+          dart2wasm_args:
+            - --define=dart.vm.product=true
+            - --enable-asserts
       source_gen:combining_builder:
         options:
           # TODO(equartey): Remove when customAuthWithSrp & customAuthWithoutSrp are supported by CLI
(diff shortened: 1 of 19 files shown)


cadivus:
While adding branch coverage for the `UnderlyingSource` factory I hit a pre-existing bug in `PropsReadableStreamReader.cancel` that's unrelated to this PR, so noting it here for follow-up rather than expanding scope.

`readable_stream.dart`:

```dart
Future<void> cancel([String? reason]) =>
    getProperty<JSPromise>('cancel'.toJS).toDart;
```

This reads the `cancel` *property* (the function object) and awaits it as if it were the returned Promise, so it never actually calls `cancel()` and the `reason` argument is silently dropped. It should invoke the method, roughly:

```dart
Future<void> cancel([String? reason]) =>
    callMethod<JSPromise>('cancel'.toJS, reason?.toJS).toDart;
```

`closed` on the same extension is a real property, so the getter there is fine.

The current tests don't exercise `reader.cancel()` directly, which is why it's gone unnoticed. It surfaced under dart2wasm (`TypeError: p.then is not a function`) but the underlying mistake is compiler-agnostic. Happy to pick this up in a separate change once this lands.

packages/amplify_core/build.yaml
ekjotmultani, line 1: should `amplify_core` get its own `--headless` browser tests on firefox and chrome?

ekjotmultani (approved):

packages/amplify_core/build.yaml
cadivus: We have browser tests for `amplify_core`

PR #7046, opened 2026-06-24
test(aft): Rename version_bump snapshot pubspecs to avoid pana 'multiple matching pubspec.yaml'

cadivus:
## Problem

pub.dev / [pana](https://github.com/dart-lang/pana) repository verification clones the **full git tree** (`git ls-tree -r`) and scans for every file whose basename is exactly `pubspec.yaml`. A pubspec is treated as a match for the package being scored when its `name:` matches, and if **two or more** `pubspec.yaml` files declare the same package name, verification fails with:

```
Repository has multiple matching `pubspec.yaml` with `name: X`
```

The `aft version-bump` test fixtures under `packages/aft/test/version_bump/data/repo_snapshot/` contain ~26 `pubspec.yaml` files copied from real published packages (same `name:`, no `publish_to`). When any of those packages is scored by pana, these fixtures collide with the real package's own `pubspec.yaml`, breaking repository verification on the pub.dev score.

Note: `.pubignore` and `publish_to: none` do **not** fix this — verification runs against the **git tree** (not the published archive), and the duplicate-name match is decided before `publish_to` is considered.

## Fix

Rename the 30 fixture `pubspec.yaml` files (29 packages + the snapshot root) to `pubspec.yaml.snapshot` so their basename is no longer `pubspec.yaml`, and strip the suffix where the snapshot is consumed/generated so behaviour is unchanged:

- **`version_bump_test.dart`** — `copyDirectory` strips a trailing `.snapshot` from `pubspec.yaml.snapshot` when materializing the temp repo, so the test input is byte-identical to before.
- **`tool/repo_snapshot.sh`** — writes `pubspec.yaml.snapshot` when (re)generating the snapshot.
- **`README.md`** — documents the suffix.

Renames were done with `git mv` (100% similarity, history preserved).

## Verification

`cd packages/aft && dart test ./test/version_bump/version_bump_test.dart` → **all 11 tests pass** with **no changes** to the stored `.diff` snapshots, confirming the materialized temp repo is identical to before.


cadivus:
Fixes this: 
Screenshot 2026-06-24 at 15 37 01

soberm (approved):

PR #7047, opened 2026-06-25
fix(aws_common): Invoke `ReadableStreamReader.cancel` instead of reading the property

cadivus:


jvh-aws (approved):
Can we write a short description (nit)

PR #7063, opened 2026-06-29
fix(authenticator): Make web e2e resilient to focus-traversal regression

cadivus:
Root cause is an unfixed Flutter framework regression
(https://github.com/flutter/flutter/issues/178109, P1, triaged-web), bisected by the Flutter team
to https://github.com/flutter/flutter/pull/166804 ("[web] close input connection when
window/iframe loses focus", shipped in 3.35.0). On web the browser can
dispatch a didChangeViewFocus notification after the previous widget
tree has been deactivated; focus traversal then walks a stale focus node
(FocusTraversalPolicy.findFirstFocus -> FocusNode.rect ->
Element.findRenderObject()) and the debug-only assertion throws "Cannot
get renderObject of inactive element", failing the integration tests.
It is web-only (Linux/iOS/Android e2e pass) and has no documented
migration -- it is a bug, not an intended behavior change.


cadivus:
Works: https://github.com/aws-amplify/amplify-flutter/actions/runs/28392146855/job/84122460055?pr=7063

ekjotmultani (approved):

PR #7076, opened 2026-06-30
feat(amplify_event_enrichment): add example app

ekjotmultani:
## Description

Adds the example Flutter app for the `amplify_event_enrichment` package, demonstrating event enrichment client usage with session tracking, global fields, and lifecycle observation.

**This PR is stacked on #7029** (core package implementation). Review #7029 first.

### Structure (3-way split)
| PR | Branch | Content |
|---|---|---|
| #7029 | `feat/event-enrichment-client` | Core package (lib only) |
| **This PR** | `feat/event-enrichment-client-example` | Example app (92 files) |
| #7068 | `feat/event-enrichment-client-tests` | Unit tests (5 files) |

### What's included
- Full Flutter example app with Android, iOS, and macOS targets
- Demonstrates `EventEnrichmentClient` initialization and usage
- Shows session management, global fields, and lifecycle observation


packages/amplify_event_enrichment/amplify_event_enrichment/example/pubspec.yaml
```diff
@@ -0,0 +1,25 @@
...
+publish_to: "none"
+
+environment:
+  sdk: ^3.10.0
+  flutter: ">=3.38.1"
+
+dependencies:
+  amplify_event_enrichment:
+    path: ../
+  amplify_foundation_dart: ">=2.11.0 <2.12.0"
```
cadivus: In example apps for our libs in this repo, I'd use `any` as a version so that we don't need to care about version bumps.

ekjotmultani: Done, switched the example deps to `any`.

PR #7081, opened 2026-07-01
feat(aft): Locate package by name or tag

cadivus:
This feature allows to locate a package by name or tag. It's mainly for being used in CI/CD.

### DIFF
--- a/packages/aft/lib/src/commands/locate_package_command.dart
+++ b/packages/aft/lib/src/commands/locate_package_command.dart
@@ -25,10 +25,12 @@ class LocatePackageCommand extends AmplifyCommand {
     if (!verbose) {
       AWSLogger().logLevel = LogLevel.warn;
     }
+    // Called only to satisfy `@mustCallSuper`; its Repo.open / GitDir.fromExisting
+    // git setup is unused here and can fail in shallow or sparse checkouts.
     await super.run();
 
     final rest = argResults!.rest;
-    if (rest.length != 1 || rest.single.isEmpty) {
+    if (rest.length != 1 || rest.single.trim().isEmpty) {
       usageException('Expected exactly one argument: <name-or-tag>');
     }
     final nameOrTag = rest.single;
--- a/packages/aft/lib/src/commands/locate_package_command.dart
+++ b/packages/aft/lib/src/commands/locate_package_command.dart
@@ -1,0 +1,43 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:io';
+
+import 'package:aft/aft.dart';
+import 'package:aws_common/aws_common.dart';
+import 'package:path/path.dart' as p;
+
+/// Prints the repo-relative path of a package identified by name or publish
+/// tag, e.g. `aft locate-package amplify_core-v2.10.1` -> `packages/amplify_core`.
+class LocatePackageCommand extends AmplifyCommand {
+  @override
+  String get description =>
+      'Prints the repo-relative path of a package by name or publish tag.';
+
+  @override
+  String get name => 'locate-package';
+
+  @override
+  Future<void> run() async {
+    // Reserve stdout for the package path so it can be captured directly,
+    // e.g. `PKG=$(aft locate-package amplify_core-v2.10.1)`. `--verbose`
+    // re-enables the default logging.
+    if (!verbose) {
+      AWSLogger().logLevel = LogLevel.warn;
+    }
+    await super.run();
+
+    final rest = argResults!.rest;
+    if (rest.length != 1 || rest.single.isEmpty) {
+      usageException('Expected exactly one argument: <name-or-tag>');
+    }
+    final nameOrTag = rest.single;
+
+    final package = aftConfig.locatePackage(nameOrTag);
+    if (package == null) {
+      exitError("Error: package '$nameOrTag' not found in workspace");
+    }
+
+    stdout.writeln(p.relative(package.path, from: rootDir.path));
+  }
+}
--- a/packages/aft/lib/src/config/config.dart
+++ b/packages/aft/lib/src/config/config.dart
@@ -96,6 +96,21 @@ abstract class AftConfig
         packageName;
   }
 
+  /// Locates the package identified by [nameOrTag] in [allPackages].
+  ///
+  /// [nameOrTag] may be a package name (`amplify_core`) or a publish tag of
+  /// the form `$name-v$version` as emitted by `aft publish --tags`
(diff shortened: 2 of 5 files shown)


packages/aft/lib/src/commands/locate_package_command.dart
```diff
@@ -0,0 +1,43 @@
...
+  String get name => 'locate-package';
+
+  @override
+  Future<void> run() async {
+    // Reserve stdout for the package path so it can be captured directly,
+    // e.g. `PKG=$(aft locate-package amplify_core-v2.10.1)`. `--verbose`
+    // re-enables the default logging.
+    if (!verbose) {
+      AWSLogger().logLevel = LogLevel.warn;
+    }
```
soberm, line 27: `super.run()` calls `Repo.open(aftConfig, logger: logger)`, which builds the full package graph and does git work — none of which this command uses. Since `locate-package` is intended specifically for CI scripting and only ever accesses `aftConfig`, that initialization is pure overhead and a potential failure point in shallow-checkout or sparse environments where git state is non-standard. Since `@mustCallSuper` makes this hard to avoid without a base-class change, worth at least calling it out in a comment here — or investigating whether `AmplifyCommand` could offer a lighter init path (e.g. make the `repo` initialization lazy or opt-in) for read-only commands.

packages/aft/lib/src/commands/locate_package_command.dart
```diff
@@ -0,0 +1,43 @@
...
+    await super.run();
+
+    final rest = argResults!.rest;
```
soberm, line 32: `rest.single.isEmpty` won't reject a whitespace-only argument like `' '` — the caller would get a "not found" error rather than a helpful usage message. `rest.single.trim().isEmpty` is more defensive.

packages/aft/lib/src/config/config.dart
```diff
@@ -96,6 +96,20 @@ abstract class AftConfig
...
 
+  /// Locates the package identified by [nameOrTag] in [allPackages], or `null`
+  /// if no matching package exists.
+  ///
+  /// [nameOrTag] may be a package name (`amplify_core`) or a publish tag of
+  /// the form `$name-v$version` as emitted by `aft publish --tags`
+  /// (`amplify_core-v2.10.1`); the trailing `-v$version` is stripped to recover
+  /// the name. Anchoring on `-v` + a digit (never `_v`) leaves names like
+  /// `aws_signature_v4` intact.
+  PackageInfo? locatePackage(String nameOrTag) {
```
soberm, line 107: Anchoring the separator on `-v\d` rather than just `-v` is exactly the right call — it handles `aws_signature_v4` correctly where a naive split on `-v` would break. Clean solution.

packages/aft/lib/src/config/config.dart
```diff
@@ -485,3 +499,8 @@ final Version activeDartSdkVersion = () {
 }();
 
 final _versionRegex = RegExp(r'\d+\.\d+\.\d+(-[a-zA-Z\d]+)?');
+
+// Captures the package name from a `$name-v$version` publish tag. The `\d`
```
soberm: This 3-line block comment largely repeats what the `locatePackage()` doc comment above already explains (the tag format, the `_v` exception, the anchoring). A one-liner like `// Strips the -v<semver> suffix; anchoring on a digit avoids splitting _v in names like aws_signature_v4.` keeps the key signal without the repetition.

packages/aft/test/config/config_loader_test.dart
```diff
@@ -115,4 +116,74 @@ aft:
         );
     });
   });
+
+  group('locatePackage', () {
+    late AftConfig config;
+
+    setUp(() async {
+      d.Descriptor package(String name, String version) => d.dir(name, [
```
soberm: Defining the `package()` helper inside `setUp` means it's re-created on every call. Hoisting it to the `group` scope (before `setUp`) is slightly cleaner and makes it clear it has no per-invocation state.

packages/aft/test/config/config_loader_test.dart
```diff
@@ -115,4 +116,74 @@ aft:
...
+        ..has((pkg) => pkg.name, 'name').equals('aws_signature_v4')
+        ..has((pkg) => p.normalize(pkg.path), 'path').equals(expectedPath);
+
+      // The tag form strips only the trailing `-v0.7.1`, not `_v4`.
+      check(config.locatePackage('aws_signature_v4-v0.7.1')).isNotNull()
+        ..has((pkg) => pkg.name, 'name').equals('aws_signature_v4')
+        ..has((pkg) => p.normalize(pkg.path), 'path').equals(expectedPath);
+    });
+
+    test('returns null for an unknown package', () {
```
soberm, line 185: The four existing tests cover plain name, tag, `_v`-in-name, and not-found — but there's no test for a malformed tag where `-v` is present but not followed by a digit, e.g. `locatePackage('amplify_core-vbeta')`. That's the boundary where the regex falls back to treating the whole string as a package name (and returns null). A short test there would confirm the fallback path and guard against future regex changes silently breaking it.

packages/aft/lib/src/config/config.dart
cadivus: Thanks!

packages/aft/lib/src/commands/locate_package_command.dart
cadivus: Done — switched to `rest.single.trim().isEmpty` so a whitespace-only argument now produces the usage error instead of a not-found.

packages/aft/lib/src/commands/locate_package_command.dart
cadivus: Good catch. `super.run()` calls `Repo.open` → `GitDir.fromExisting`, which this command never uses. One nuance: the package graph is `late final`, so it is not actually built unless something accesses it — the real cost here is just the git discovery. I have added a comment noting we only call `super.run()` for `@mustCallSuper` and that its git work is unused for this read-only command. I have left the base-class lazy-init refactor out of this PR to keep the scope tight.

packages/aft/lib/src/config/config.dart
cadivus: Collapsed to a one-liner as suggested — the `locatePackage()` doc comment already covers the tag format and the `_v` exception.

packages/aft/test/config/config_loader_test.dart
cadivus: Hoisted the `package()` helper up to the `group` scope so it is defined once.

packages/aft/test/config/config_loader_test.dart
cadivus: Added a test for the `-vbeta` case (no `-v<digit>`) that confirms the fallback to a plain-name lookup returns null.

soberm (approved):

PR #7082, opened 2026-07-01
fix(worker_bee): Coerce int-backed types for wasm

cadivus:
Under dart2wasm, `int`/`Duration`/`DateTime` come back from a worker `postMessage` as `double`, so built_value's `as int` cast throws. (dart2js never hit this — it treats int/double as the same JS number.)

Registers an `IntCoercingSerializer` that coerces `num`→`int` before delegating to built_value's default. Generalizes the existing `DateTime`-only fix (#7059) to `int` and `Duration`.

Tests round-trip all three (incl. the `double` wire case) on VM, dart2js, and dart2wasm. Also verified end-to-end in a browser worker.


packages/worker_bee/worker_bee/lib/src/serializers/int_coercing_serializer.dart
```diff
@@ -0,0 +1,65 @@
...
+  }) => _delegate.serialize(serializers, object, specifiedType: specifiedType);
+
+  @override
+  T deserialize(
+    Serializers serializers,
+    Object serialized, {
+    FullType specifiedType = FullType.unspecified,
+  }) => _delegate.deserialize(
+    serializers,
+    serialized is num ? serialized.toInt() : serialized,
```
soberm, line 65: This coerces *any* `num` down to `int`, so the wrapper is only safe for types whose wire form is a whole number. That's true for int/Duration/DateTime, but since the class is generic and `forType()`'s guard only checks that a `PrimitiveSerializer` exists (not that `T` is int-backed), someone could register `IntCoercingSerializer<double>` and silently truncate real fractional values. Worth a line in the doc — or an `assert` — stating this is for int-backed primitives only.

packages/worker_bee/worker_bee/lib/src/serializers/int_coercing_serializer.dart
```diff
@@ -0,0 +1,65 @@
...
+/// Serializer for an int-backed built_value type which tolerates a `double`
+/// wire value.
+///
+/// built_value's defaults for `int`, `Duration`, and `DateTime` cast the wire
+/// value straight to `int`, which fails on the `double` a `postMessage`
+/// round-trip yields under dart2wasm. This coerces to `int` first, then
+/// delegates to built_value's default so everything else is unchanged.
+/// {@endtemplate}
+@internal
+class IntCoercingSerializer<T extends Object>
```
soberm, line 20: Minor: the class is fully generic, but the doc comment above pins it to `int`/`Duration`/`DateTime`, and the name reads a little oddly at the non-int call sites (`IntCoercingSerializer<DateTime>`). Could describe it in terms of "any int-backed primitive" and leave the concrete list to `serializers.dart` where they're actually registered. Not blocking.

packages/worker_bee/worker_bee/test/int_coercing_serializer_test.dart
```diff
@@ -0,0 +1,108 @@
...
+          );
+          expect(deserialized, c.value);
+        });
+
+        test('accepts a double, as dart2wasm produces over postMessage', () {
+          final serialized = workerBeeSerializers.serialize(
+            c.value,
+            specifiedType: fullType,
+          );
+          final asDouble = (serialized as int).toDouble();
```
soberm, line 58: The deleted DateTime test had a one-liner explaining that `(serialized as int).toDouble()` is standing in for what dart2wasm hands you back over `postMessage`. That rationale got dropped in the consolidated test — worth re-adding a short comment here so the synthetic widening to `double` doesn't look arbitrary to the next reader.

packages/worker_bee/worker_bee/lib/src/serializers/serializers.dart
```diff
@@ -5,20 +5,23 @@ import 'package:built_value/serializer.dart';
...
+/// [IntCoercingSerializer]s registered here override built_value's defaults for
+/// `int`, `Duration`, and `DateTime` in worker payloads.
 @internal
 @SerializersFor([WorkerBeeExceptionImpl])
 final Serializers workerBeeSerializers =
     (_$workerBeeSerializers.toBuilder()
           ..add(const LogEntrySerializer())
           ..add(const StackTraceSerializer())
-          ..add(const DateTimeSerializer()))
+          ..add(IntCoercingSerializer<int>.forType())
```
soberm, line 24: These three cover the int-backed primitives that `dartify()` to `double` under wasm. Did you confirm nothing in the worker payloads travels as some other int-backed type (e.g. a `fixnum` `Int64`) that would hit the same `as int` cast? `built_value` serializes `Int64`/`BigInt` to strings, so they *should* be unaffected — just want to make sure the scope is genuinely these three and there's no remaining gap.

packages/worker_bee/worker_bee/lib/src/serializers/int_coercing_serializer.dart
cadivus: Good catch. Added an `assert` in `forType()` that `T` is int/Duration/DateTime, so `IntCoercingSerializer<double>` now throws instead of silently truncating. Test added for it.

packages/worker_bee/worker_bee/lib/src/serializers/int_coercing_serializer.dart
cadivus: Doc now leads with "an int-backed built_value type" and only mentions the concrete three as the built_value defaults being overridden. Left the name as-is to avoid churn.

packages/worker_bee/worker_bee/test/int_coercing_serializer_test.dart
cadivus: Re-added: `// Stand in for the double dart2wasm hands back over postMessage.`

packages/worker_bee/worker_bee/lib/src/serializers/serializers.dart
cadivus: Confirmed. Checked the worker payload models: no `double`/`Int32`/`Int64`/`fixnum` fields. There are `BigInt` fields (SrpInitResult.privateA/publicA), but built_value serializes `BigInt` and `Int64` to strings, so they cross postMessage as strings and never hit the `as int` cast. Scope is genuinely these three.

soberm (approved):

PR #7093, opened 2026-07-02
feat(aft): Await published version on pub.dev before awaiting analysis

cadivus:
With this change, we first await the package and then the analysis. Useful when the pending state does not start directly after kicking off publishing.

### DIFF
--- a/packages/aft/lib/src/commands/amplify_command.dart
+++ b/packages/aft/lib/src/commands/amplify_command.dart
@@ -226,6 +226,24 @@ abstract class AmplifyCommand extends Command<void>
     }
   }
 
+  /// Waits until [version] of [package] is live on `pub.dev`, so its analysis
+  /// has started before we await it.
+  Future<void> awaitPublishedVersion(String package, Version version) async {
+    final qaDurations = Platform.environment.containsKey('QA_DURATIONS');
+
+    final pollInterval = qaDurations
+        ? const Duration(seconds: 1)
+        : const Duration(seconds: 15);
+
+    while (true) {
+      final versionInfo = await resolveVersionInfo(package);
+      if (versionInfo?.allVersions.contains(version) ?? false) {
+        return;
+      }
+      await Future<void>.delayed(pollInterval);
+    }
+  }
+
   @override
   @mustCallSuper
   Future<void> run() async {
--- a/packages/aft/lib/src/publish_scheduler.dart
+++ b/packages/aft/lib/src/publish_scheduler.dart
@@ -224,7 +224,9 @@ class PublishScheduler {
   /// order and call [_publishPackage] for each one.
   Future<void> _runDryRun() async {
     for (final package in _packages) {
-      stdout.writeln('Publishing ${package.name} (${package.version}) [dry run]');
+      stdout.writeln(
+        'Publishing ${package.name} (${package.version}) [dry run]',
+      );
       await _publishPackage(package);
       stdout.writeln('');
     }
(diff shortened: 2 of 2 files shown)


packages/aft/lib/src/commands/amplify_command.dart
```diff
@@ -226,6 +226,24 @@ abstract class AmplifyCommand extends Command<void>
...
+  /// Waits until [version] of [package] is live on `pub.dev`, so its analysis
+  /// has started before we await it.
+  Future<void> awaitPublishedVersion(String package, Version version) async {
+    final qaDurations = Platform.environment.containsKey('QA_DURATIONS');
+
+    final pollInterval = qaDurations
+        ? const Duration(seconds: 1)
+        : const Duration(seconds: 15);
+
+    while (true) {
```
osama-rizk, line 238: `awaitPublishedVersion` loops `while (true)` with no timeout — the only exit is `allVersions.contains(version)` going true. If the version never lands (partial publish) or equality never matches, this hangs forever. Worth noting: pub_semver's `Version.==` includes build metadata (`1.2.3+1` != `1.2.3`), so any build-metadata drift between the local version and pub.dev's listing means `.contains` is false forever. Since dependents are only enqueued by `_markAnalysisComplete` after this future completes, a hang here deadlocks the whole run. Suggest a max wall-clock/attempt cap that throws, and/or comparing on canonicalized versions.

packages/aft/lib/src/commands/amplify_command.dart
```diff
@@ -226,6 +226,24 @@ abstract class AmplifyCommand extends Command<void>
...
+      final versionInfo = await resolveVersionInfo(package);
```
osama-rizk, line 239: `resolveVersionInfo` throws `AWSHttpException` on any non-200/404. This new loop polls pub.dev every 15s with no `try/catch`, so a single transient 5xx errors the `analysisFuture`. That future is only consumed via `whenComplete`, which doesn't swallow the error — so `_markAnalysisComplete` never runs (deadlock) and the error surfaces unhandled. Recommend catching transient HTTP errors here and treating them as "not ready, retry."

packages/aft/lib/src/commands/amplify_command.dart
```diff
@@ -226,6 +226,24 @@ abstract class AmplifyCommand extends Command<void>
     }
   }
 
+  /// Waits until [version] of [package] is live on `pub.dev`, so its analysis
+  /// has started before we await it.
+  Future<void> awaitPublishedVersion(String package, Version version) async {
+    final qaDurations = Platform.environment.containsKey('QA_DURATIONS');
```
osama-rizk, line 232: Minor: the `QA_DURATIONS` / poll-interval setup is duplicated verbatim from `awaitPendingAnalysis` above. Consider a shared helper so the two timing policies don't drift.

packages/aft/lib/src/commands/amplify_command.dart
cadivus: I think we can take this risk here. If we use this in e.g. a workflow, we have a timeout there. 
If we use it locally, the user will watch it. 
Definitely good enough for a dev tool.

packages/aft/lib/src/commands/amplify_command.dart
cadivus: The way it's written, it should crash the entire app which is fine.

packages/aft/lib/src/commands/amplify_command.dart
cadivus: I think that's fine at this stage.

osama-rizk (approved):

PR #7094, opened 2026-07-03
Feat/cognito/wasm

cadivus:
This introduces wasm support for auth (Cognito). It's hard to stack these PRs (test steup...), so I will merge with rebase and have the PR referenced in every of it.


packages/worker_bee/worker_bee/example/lib/workers.dart2js.js
thisisabhash, line 1: The example under packages/worker_bee/worker_bee/example/lib/ has compiled outputs checked in. These are large diffs of generated code. Is this intentional for the example app to work without a build step, or should these be .gitignored and built in CI?

thisisabhash (dismissed):

packages/worker_bee/worker_bee/example/lib/workers.dart2js.js
cadivus: Yes, it's intentional in this case.

thisisabhash (approved):

ekjotmultani (approved):

asaarnak:
@cadivus 
Tested that `flutter run --release` works but `flutter run --debug` gives me this error and the app crashes:
```
Refused to execute script from 'http://localhost:56488/packages/amplify_auth_cognito_dart/src/workers/workers.debug.dart.js' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.
```

cadivus:
@asaarnak Thanks for the feedback! I will have a look!

asaarnak:
Also got compilation errors because some `build_runner` files are in `.gitignore`
```
Compiling lib/main.dart for the Web...                          
Error detected in pubspec.yaml:
No file or variants found for asset: packages/amplify_secure_storage_dart/lib/src/worker/workers.min.js.

This asset was included from package amplify_secure_storage_dart.
Target web_release_bundle failed: Exception: Failed to bundle asset files.
#0      copyAssets 
```
As a workaround i built them locally and added path git dependency like this:
```
  amplify_secure_storage_dart:
    path: packages/amplify_secure_storage_dart
  
  amplify_auth_cognito_dart:
    path: packages/amplify_auth_cognito_dart
```

cadivus:
@asaarnak These files should be autogenerated when you bootstrap the repo with `aft bootstrap` (see https://github.com/aws-amplify/amplify-flutter/blob/main/CONTRIBUTING.md#setting-up-for-local-development). Did you do that? This also creates the dependency overrides you created manually.

Before publishing to pub.dev, the build_runners run and generate these files so that they get published.

asaarnak:
yep,  but the fork didn't have those files

asaarnak:
@cadivus any news about the issue?
https://github.com/aws-amplify/amplify-flutter/pull/7094#issuecomment-4956407174

cadivus:
@asaarnak Yes, we have this PR open that should solve it: https://github.com/aws-amplify/amplify-flutter/pull/7111
I case you want to test it (what would be very nice!), remember to bootstrap again after checking out the branch.

asaarnak:
@cadivus can you create an `dev` release where i don't need to use `aft bootstrap` ?

asaarnak:
Building latest(`0154e97d57f5dbe8f95fafc4c1d402524979f2a1`) master with `dependency_overrides` gives this error:
`No file or variants found for asset: packages/amplify_secure_storage_dart/lib/src/worker/workers.min.js.`

cadivus:
@asaarnak This looks very likely like a bootstrap problem (we can't reproduce this bug and our runners look fine too).  
We will get this released soon so that you can use it directly from pub.dev.

cadivus:
Hi @asaarnak,
we just released a new version to pub.dev that adds wasm support to Cognito.

PR #6687, opened 2026-02-27
feat(foundation): Add foundation

ekjotmultani:
### feat: Add `amplify_foundation_dart` and `amplify_foundation_dart_bridge` packages

Introduces two new packages under `packages/amplify_foundation/` that provide core V3 foundation types for the Amplify Flutter SDK.

#### `amplify_foundation_dart`

A pure-Dart package containing shared types and utilities used across Amplify packages:

- __Credentials__ — `AWSCredentials` (sealed class with `StaticCredentials` and `TemporaryCredentials` subtypes), `AWSCredentialsProvider` interface
- __Logging__ — A structured logging system with `Logger`, `LogSink`, `BroadcastLogger`, configurable `LogLevel`, `LogMessage`, and a default `AmplifySimplePrinterLogSink`
- __Result__ — A sealed `Result<S, E>` type (`ResultSuccess` / `ResultError`) for typed error handling
- __Exceptions__ — `AmplifyException` base class
- __Utilities__ — `Globals` (zone-scoped dependency container) and a `safePrint` helper

Includes unit tests for credentials, logging, result types, and exceptions, plus an example demonstrating usage.

#### `amplify_foundation_dart_bridge`

A bridge package (`publish_to: none`) that adapts V2 credential types (`aws_common` / `amplify_core`) to the new V3 `AWSCredentialsProvider` interface via `V2CredentialsProviderBridge`, enabling incremental migration. Includes unit tests.

PR #7096, opened 2026-07-06
ci: add event_enrichment and connect_client to log_cw_metric category allowlist

ekjotmultani:
Adds the `event_enrichment` and `connect_client` categories to the `log_cw_metric` composite action's allowlist so CI runs for the new `amplify_event_enrichment` and `amplify_connect_client` packages can log their success/failure metrics. Without these entries the action throws `WorkingDirectory input ... must contain a valid category` on every workflow run for those packages.

The `dist` output is rebuilt from source via `dart tool/build.dart` per the generated-code workflow.


actions/bin/log_cw_metric.dart
```diff
@@ -94,6 +94,8 @@ Future<void> logMetric() async {
     'amplify_lints',
     'amplify_native_legacy_wrapper',
     'kinesis',
+    'event_enrichment',
+    'connect_client',
```
cadivus, line 98: Are you sure about these two? The last new categories were introduced by https://github.com/aws-amplify/amplify-flutter/pull/6687 and https://github.com/aws-amplify/amplify-flutter/pull/6516. They matched exactly the package names.  
  
If I see https://github.com/aws-amplify/amplify-flutter/pull/7109 right, the new name will be `amplify_connect_client`. The tests are probably for the dart package only, right? This would be `amplify_connect_client_dart`. But I am not completely sure here since we barely touch these categories. I'd probably open the PR after I got an error message that exactly shows the missing category.

cadivus: At least the tests introduced by this PR run without adding a new category: https://github.com/aws-amplify/amplify-flutter/pull/7110

PR #7109, opened 2026-07-13
feat(connect): add Amplify Connect client for Customer Profiles

ekjotmultani:
## Description

Adds the Amplify Connect client for Amazon Connect Customer Profiles (the Pinpoint identifyUser replacement). This PR contains the **implementation only** — the two packages (`amplify_connect_client_dart` and `amplify_connect_client`) with their lib, pubspec, README, CHANGELOG, and LICENSE.

The example app and all tests (unit + integration) live in the stacked tests PR #7110.

Public API (Option C, per the finalized JS design): `identifyUser({userProfile})`, `registerDevice({token})`, `removeDevice()`. All three routes are SigV4-signed to the backend construct's endpoint; config is read from `notifications.amazon_connect` in amplify_outputs. Includes https-only endpoint validation and 255-char attribute bounds mirroring the backend.

Coordinated with the backend construct (aws-amplify/amplify-backend#3265) and the JS provider (aws-amplify/amplify-js#14866).

## Tests
See #7110 (example app + unit + integration tests).


cadivus:
Do we need to add the new client here?
https://github.com/aws-amplify/amplify-flutter/blob/b3d85ac08ef75bf632d72181c069f754927b70e5/pubspec.yaml#L69

cadivus (changes requested):
Please add `README.md`, `CHANGELOG.md`, and `LICENSE`

packages/amplify_connect_client/amplify_connect_client/pubspec.yaml
```diff
@@ -0,0 +1,40 @@
...
+  flutter: ">=3.38.1"
+
+# Helps `pana` since we do not use Flutter plugins for most platforms
+platforms:
+  ios:
+  android:
+  macos:
+  windows:
+  linux:
+  web:
```
cadivus, line 25: If we officially want to support web, we need to get rid of this unconditional import: 
https://github.com/aws-amplify/amplify-flutter/blob/a784c20d0b52b40b9b9b562d7c3dd7eee5aa6b01/packages/amplify_connect_client/amplify_connect_client/lib/src/connect_client_flutter.dart#L4

packages/amplify_connect_client/amplify_connect_client_dart/lib/src/connect_client.dart
```diff
@@ -0,0 +1,131 @@
...
+      address: options.address,
+      deviceId: deviceId,
+      channelType: options.channelType,
+      platform: options.platform ?? _platform,
+      appVersion: options.appVersion ?? _appVersion,
+      optOut: options.optOut,
+      previousGuestIdentityId: options.previousGuestIdentityId,
+    );
+  }
+}
```
cadivus: Every field is copied by hand. If a field is later added to `IdentifyUserOptions` and someone forgets to add it here, it will be silently dropped.

A `copyWith({String? deviceId, String? platform, String? appVersion})` on `IdentifyUserOptions` would make this robust and is the more conventional pattern.

packages/amplify_connect_client/amplify_connect_client_dart/lib/src/services/identify_user_service.dart
```diff
@@ -0,0 +1,118 @@
...
+      try {
+        request = await signer.sign(
+          unsigned,
+          credentialScope: AWSCredentialScope(
+            region: _region,
+            service: _service,
+          ),
+        );
+      } on Exception catch (e) {
+        throw ConnectNetworkException(cause: e);
```
cadivus: Should we label signing exceptions as network exceptions? I think this is misleading.

packages/amplify_connect_client/amplify_connect_client_dart/lib/src/models/identify_user_options.dart
```diff
@@ -0,0 +1,90 @@
...
+    'deviceId': ?deviceId,
+    'channelType': ?channelType?.value,
+    'platform': ?platform,
+    'appVersion': ?appVersion,
+    'optOut': ?optOut?.value,
+    'previousGuestIdentityId': ?previousGuestIdentityId,
+  };
+
+  /// Whether every field is null/empty (nothing to send).
+  bool get isEmpty => toJson().isEmpty;
```
cadivus: Every field is final, so this thing does not mutate, right? if so, can we make this class const?
And wouldn't it make sense to precalculate the `toJson` getter and/or `isEmpty`?
In calls like this: 
https://github.com/aws-amplify/amplify-flutter/blob/a784c20d0b52b40b9b9b562d7c3dd7eee5aa6b01/packages/amplify_connect_client/amplify_connect_client_dart/lib/src/connect_client.dart#L98
we run `toJson` to look if it's empty and then again for the actual value.

packages/amplify_connect_client/amplify_connect_client_dart/lib/src/connect_client_configuration.dart
```diff
@@ -0,0 +1,89 @@
...
+  });
+
+  /// Parses the configuration from the `custom.CustomerProfiles` section of a
+  /// decoded `amplify_outputs.json` map (the canonical output written by the
+  /// backend construct).
+  ///
+  /// Expects a shape of:
+  /// ```json
+  /// {
+  ///   "custom": {
```
thisisabhash: newer canary output is under
```
notifications.amazon_connect_customer_profile with aws_region
```

packages/amplify_connect_client/amplify_connect_client_dart/lib/src/models/identify_user_options.dart
```diff
@@ -0,0 +1,90 @@
...
+  final String? appVersion;
+
+  /// Opt-out preference. The backend currently accepts this field but does not
+  /// act on it yet, so setting it has no effect for now.
+  final OptOut? optOut;
+
+  /// On an authenticated call, the caller's prior guest Identity Pool
+  /// `identityId`. When present, the backend folds the guest profile (and its
+  /// devices) into the authenticated profile (merge-on-sign-in).
+  final String? previousGuestIdentityId;
```
thisisabhash: renamed to `guestIdentityId`

cadivus:
@ekjotmultani The new packages should also be added here: https://github.com/aws-amplify/amplify-flutter/blob/main/pubspec.yaml  
  
Otherwise, version bumps may fail

packages/amplify_connect_client/amplify_connect_client_dart/lib/src/connect_client_configuration.dart
ekjotmultani: updated

packages/amplify_connect_client/amplify_connect_client/pubspec.yaml
ekjotmultani: Good catch. The dart:io import is gone

packages/amplify_connect_client/amplify_connect_client_dart/lib/src/connect_client.dart
ekjotmultani: This one got resolved from a different direction. The API design was finalized after your review and IdentifyUserOptions was removed entirely. Device registration is now its own registerDevice({token}) call and the library fills in deviceId, platform, appVersion, and channelType internally, so there is no hand-copied options object left to drift.

packages/amplify_connect_client/amplify_connect_client_dart/lib/src/services/identify_user_service.dart
ekjotmultani: Agreed, that was misleading. Signing failures are now classified as service/configuration errors rather than network errors, and the classifier was reworked alongside the new three-route service. There are unit tests covering the error classification paths

packages/amplify_connect_client/amplify_connect_client_dart/lib/src/models/identify_user_options.dart
ekjotmultani: Also resolved by the API finalization. IdentifyUserOptions no longer exists, and the request serialization in the new write service builds each body once, so the double toJson evaluation is gone too.

packages/amplify_connect_client/amplify_connect_client/lib/src/connect_client_flutter.dart
```diff
@@ -0,0 +1,121 @@
...
+      case TargetPlatform.iOS:
+        // Debug builds register with the APNs sandbox.
+        return kDebugMode ? ChannelType.apnsSandbox : ChannelType.apns;
+      case TargetPlatform.macOS:
+      case TargetPlatform.windows:
+      case TargetPlatform.linux:
+      case TargetPlatform.fuchsia:
+        return null;
+    }
+  }
```
cadivus, line 118: Web is actually not handled here. `TargetPlatform` returns `android`/`iOS` if the app runs in a browser on Android/iOS.
Guess, this would lead to an exception.

```suggestion
  static ChannelType? _resolveChannelType() {
    if (kIsWeb) {
      return null;
    }
    switch (defaultTargetPlatform) {
      case TargetPlatform.android:
        return ChannelType.gcm;
      case TargetPlatform.iOS:
        // Debug builds register with the APNs sandbox.
        return kDebugMode ? ChannelType.apnsSandbox : ChannelType.apns;
      case TargetPlatform.macOS:
      case TargetPlatform.windows:
      case TargetPlatform.linux:
      case TargetPlatform.fuchsia:
        return null;
    }
  }
```

cadivus (dismissed):

packages/amplify_connect_client/amplify_connect_client/lib/src/connect_client_flutter.dart
ekjotmultani: `_resolveChannelType` now checks `kIsWeb` before `defaultTargetPlatform`, so web resolves no channel and `registerDevice` throws the unsupported operation exception instead of picking up the browser's host OS. Also added the same guard to `_platformName`

packages/amplify_connect_client/amplify_connect_client_dart/lib/src/connect_client.dart
```diff
@@ -0,0 +1,162 @@
...
+    );
+    _logger.info('registerDevice sent for channel ${channelType.value}');
+  }
+
+  /// Removes this device from the caller's profile.
+  ///
+  /// Resolves the stable device id from storage (the same id used to register).
+  Future<void> removeDevice() async {
+    final credentials = await _credentialsProvider.fetchCredentials();
+    final deviceId = await _deviceIdStore.getOrCreateDeviceId();
```
thisisabhash: connect_client.dart calls _deviceIdStore.getOrCreateDeviceId(). If nothing is persisted (fresh install, or registerDevice never called), this mints and persists a new UUID and then sends POST /remove-device with an id the backend has never seen. Should be readDeviceId() with an early return or a typed throw when null

packages/amplify_connect_client/amplify_connect_client/lib/src/shared_preferences_device_id_store.dart
```diff
@@ -0,0 +1,47 @@
...
+
+  final Uuid _uuid;
+
+  @override
+  Future<String> getOrCreateDeviceId() async {
+    final prefs = await SharedPreferences.getInstance();
+    final existing = prefs.getString(connectDeviceIdKey);
+    if (existing != null && existing.isNotEmpty) return existing;
+    final id = _uuid.v4();
+    await prefs.setString(connectDeviceIdKey, id);
```
thisisabhash, line 38: AI notes: 
shared_preferences_device_id_store.dart says the key, store, and semantics are "the same as the event enrichment client's client-id provider — so both packages resolve the identical id for a device." But shared_preferences prefixes every key with flutter. by default (_prefix = 'flutter.' in the plugin). The real stored key is flutter.com.amplifyframework.device_id, which does not match the raw key com.amplifyframework.device_id. Either call SharedPreferences.setPrefix('') (has its own migration hazards), read via a platform channel, or correct the docstring to scope the claim to Flutter-only.

packages/amplify_connect_client/amplify_connect_client/lib/src/connect_client_flutter.dart
```diff
@@ -0,0 +1,129 @@
...
+    // iOS in mobile browsers), so check kIsWeb first.
+    if (kIsWeb) {
+      return null;
+    }
+    switch (defaultTargetPlatform) {
+      case TargetPlatform.android:
+        return ChannelType.gcm;
+      case TargetPlatform.iOS:
+        // Debug builds register with the APNs sandbox.
+        return kDebugMode ? ChannelType.apnsSandbox : ChannelType.apns;
```
thisisabhash, line 111: AI note:
kDebugMode is false in profile mode, so a profile build — signed with a development cert, producing a sandbox token — registers as ChannelType.apns, and pushes silently fail. AmplifyConnectClientFlutter.create() exposes credentialsProvider, deviceIdStore, and appVersion but not channelType, so a developer who knows better (TestFlight, enterprise distribution) cannot override it. Expose ChannelType? on create()

packages/amplify_connect_client/amplify_connect_client_dart/lib/src/connect_client_configuration.dart
```diff
@@ -0,0 +1,110 @@
...
+    return ConnectClientConfiguration(
+      endpoint: _trimTrailingSlash(endpoint),
+      region: region,
+    );
+  }
+
+  static String _trimTrailingSlash(String value) =>
+      value.endsWith('/') ? value.substring(0, value.length - 1) : value;
+
+  /// The base identify endpoint URL. Clients call `POST {endpoint}/identify-user`
```
thisisabhash: nit for doc comment:
connect_client_configuration.dart documents endpoint as "Clients call POST {endpoint}/identify-user and POST {endpoint}/identify-user-guest", and the class docstring mentions "the region used to SigV4-sign guest requests." ConnectWriteService has only /identify-user, /register-device, /remove-device, and signs one path for everyone

packages/amplify_connect_client/amplify_connect_client_dart/lib/src/connect_client.dart
```diff
@@ -0,0 +1,162 @@
...
+    if (attributes != null) {
+      for (final entry in attributes.entries) {
+        _checkLength("customAttributes key '${entry.key}'", entry.key);
+        _checkLength("customAttributes value for '${entry.key}'", entry.value);
+      }
+    }
+  }
+
+  void _checkLength(String field, String? value) {
+    if (value != null && value.length > _maxAttributeLength) {
```
thisisabhash, line 161: nit: we can add check for empty strings as well

packages/amplify_connect_client/amplify_connect_client_dart/lib/src/services/connect_write_service.dart
```diff
@@ -0,0 +1,117 @@
...
+/// {@template amplify_connect_client.connect_write_service}
+/// Sends the three write requests to the Customer Profiles endpoint.
+///
+/// Every route is `POST`-ed to the endpoint and SigV4-signed for the
+/// `execute-api` service with the caller's AWS credentials (authenticated or
+/// guest). The backend derives the principal identity from the signature.
+///
+/// The HTTP client is injectable so tests can supply a `MockAWSHttpClient`.
+/// {@endtemplate}
+class ConnectWriteService {
```
thisisabhash, line 19: nit: we can mark this as internal class if not being exposed to public.

packages/amplify_connect_client/amplify_connect_client_dart/lib/src/storage/device_id_store.dart
```diff
@@ -0,0 +1,54 @@
...
+
+/// {@template amplify_connect_client.in_memory_device_id_store}
+/// An in-memory [DeviceIdStore] with no persistence.
+///
+/// Useful for tests and for the pure-Dart core. Flutter apps should use the
+/// shared-preferences-backed implementation from the `amplify_connect_client`
+/// package so the id survives restarts and is shared with other Amplify
+/// packages.
+/// {@endtemplate}
+class InMemoryDeviceIdStore implements DeviceIdStore {
```
thisisabhash, line 39: nit: we can mark this as internal class if not being exposed to public.

packages/amplify_connect_client/amplify_connect_client_dart/lib/src/connect_client.dart
ekjotmultani: Good catch, `removeDevice` now uses `readDeviceId` and returns early with a log line when nothing is persisted, so a fresh install no longer mints and stores an id just to remove it. Went with the no-op over a throw so logout flows can call it unconditionally. Added a unit test covering the no-op path. Fixed in latest commit

packages/amplify_connect_client/amplify_connect_client/lib/src/shared_preferences_device_id_store.dart
ekjotmultani: You're right about the prefix, the stored entry is `flutter.com.amplifyframework.device_id.` The claim itself still holds for the Flutter enrichment client since it goes through the same plugin and gets the same prefix, so both Flutter packages do resolve the same id. But it doesn't extend to native SDKs reading the raw platform store, and setPrefix('') isn't worth the migration hazard. Updated the docstring to scope the claim to Flutter packages

packages/amplify_connect_client/amplify_connect_client/lib/src/connect_client_flutter.dart
ekjotmultani: Agreed, `create()` and `createFromAmplifyOutputs` now take an optional `channelType` that overrides the automatic resolution, so distributions where build mode doesn't imply the APNs environment can set it explicitly. The default behavior is unchanged and the doc comment explains when to override.

packages/amplify_connect_client/amplify_connect_client_dart/lib/src/connect_client_configuration.dart
ekjotmultani: Fixed, the docs now list the three actual routes and drop the guest signing framing since all routes sign the same way

packages/amplify_connect_client/amplify_connect_client_dart/lib/src/connect_client.dart
ekjotmultani: We're intentionally holding the client-side validation to exactly what the backend enforces (255 max length, nothing else) so the two never drift. That came out of a security review, where the bounds were specified as length-only mirrors of the backend's `MAX_ATTRIBUTE_LENGTH`. If the backend adds a non-empty rule we'd mirror it, but adding it client-side first would reject payloads the backend accepts.

packages/amplify_connect_client/amplify_connect_client_dart/lib/src/services/connect_write_service.dart
ekjotmultani: Done, thank you

packages/amplify_connect_client/amplify_connect_client_dart/lib/src/storage/device_id_store.dart
ekjotmultani: This one we'd keep public. That line is `InMemoryDeviceIdStore`, which is there as a testing convenience so consumers can unit test code that talks to `ConnectClient` without hitting real storage. The` DeviceIdStore` interface next to it is the injection point that `create() `accepts and the Flutter package implements, so it needs to stay exported too.

thisisabhash (dismissed):

thisisabhash (approved):

PR #7111, opened 2026-07-14
fix(worker_bee): Point `fallbackUrls` at real filenames after `copy_builder` use

cadivus:
We use worker_bee in combination with copy_builder that renames `*.debug.*` and `*.release.*` files. Because of this, the fallback paths are broken. We add the new names while keeping the old one in the list of fallback urls since one of them will probably work. This is done because of https://github.com/aws-amplify/amplify-flutter/pull/7094#issuecomment-4956407174.

### DIFF
--- a/packages/worker_bee/worker_bee_builder/lib/src/impl/js.dart
+++ b/packages/worker_bee/worker_bee_builder/lib/src/impl/js.dart
@@ -78,35 +78,35 @@ return ${allocate(DartTypes.awsCommon.zDebugMode)} ? '$_workersJsPath' : '$_mini
   }
 
   Code get _fallbackUrls {
-    return Block.of([
-      const Code('''
-    // When running in a test, we need to find the `packages` directory which
-    // is symlinked in the root `test/` directory.
-    final baseUri = Uri.base;
-    final basePath = baseUri.pathSegments
-        .takeWhile((segment) => segment != 'test')
-        .map(Uri.encodeComponent)
-        .join('/');'''),
-      declareConst('relativePath')
-          .assign(
-            DartTypes.awsCommon.zDebugMode.conditional(
-              literalString(_debugWorkersJsPath),
-              literalString(_releaseWorkersJsPath),
-            ),
-          )
-          .statement,
-      const Code(r'''
-    final testRelativePath = Uri(
+    // Offer both the copy-builder output names (`workers.js` / `workers.min.js`)
+    // and the raw build_web_compilers names (`*.debug.dart.js` /
+    // `*.release.dart.js`), each also under the `test/` symlink used by tests.
+    return Code.scope(
+      (allocate) =>
+          '''
+// When running in a test, we need to find the `packages` directory which
+// is symlinked in the root `test/` directory.
+final baseUri = Uri.base;
+final basePath = baseUri.pathSegments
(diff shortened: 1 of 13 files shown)


packages/worker_bee/worker_bee_builder/lib/src/impl/js.dart
```diff
@@ -78,35 +78,35 @@ return ${allocate(DartTypes.awsCommon.zDebugMode)} ? '$_workersJsPath' : '$_mini
...
+          '''
+// When running in a test, we need to find the `packages` directory which
+// is symlinked in the root `test/` directory.
+final baseUri = Uri.base;
+final basePath = baseUri.pathSegments
+    .takeWhile((segment) => segment != 'test')
+    .map(Uri.encodeComponent)
+    .join('/');
+const relativePaths = ${allocate(DartTypes.awsCommon.zDebugMode)}
+    ? ['$_workersJsPath', '$_debugWorkersJsPath']
```
soberm, line 95: [question] `_workersJsPath` (debug) and `_minifiedWorkersJsPath` (release) are the same paths already returned by `jsEntrypoint`, so when the primary load fails and the runtime walks `fallbackUrls`, the very first entry is a retry of something that just failed. Does worker_bee deduplicate or skip the primary URL before iterating fallbacks? If it does, no problem. If it doesn't, this first entry is a guaranteed no-op on every worker load failure and could just be dropped, leaving only the `*.debug.dart.js` / `*.release.dart.js` names that actually differ.

packages/worker_bee/worker_bee_builder/lib/src/impl/js.dart
```diff
@@ -78,35 +78,35 @@ return ${allocate(DartTypes.awsCommon.zDebugMode)} ? '$_workersJsPath' : '$_mini
...
+    : ['$_minifiedWorkersJsPath', '$_releaseWorkersJsPath'];
+return [
+  for (final relativePath in relativePaths) ...[
```
soberm, line 98: [minor] The list interleaves base and test-path variants per filename: `[base1, test/base1, base2, test/base2]`. In a non-test environment every odd-indexed attempt (`test/base1`, `test/base2`) hits a guaranteed 404 before trying the next real candidate. Grouping base paths first — `[base1, base2, test/base1, test/base2]` — halves the failed fetches in the common case. Not a blocker since the runtime stops at the first success, but worth considering.

packages/worker_bee/worker_bee_builder/lib/src/impl/js.dart
cadivus: No, not exactly. We have `if (zIsFlutter)` in `jsEntrypoint` because we assume it will be in the `assets` dir. In case it isn't we add the `!jsEntrypoint` here.

> Does worker_bee deduplicate or skip

It skips invalid urls if there is nothing to attempt left. So no, we don't create a problem by adding invalid fallback options.

packages/worker_bee/worker_bee_builder/lib/src/impl/js.dart
cadivus: I think that's fine here. My thought was that we actually always use the copy_builder, so for our usecase `base2*` is unlikely (but should be supported).

soberm (approved):

PR #7125, opened 2026-07-17
fix(aft): Cascade constraints for `foundation_dart` +`foundation_dart_bridge` minor bumps

cadivus:
`amplify_foundation_dart` and `amplify_foundation_dart_bridge` are post-1.0 but in no aft component, so a minor bump doesn't widen dependents' `<nextMinor` constraints (the 1b4e201b7 failure fixed by hand in #7080).

Add solo, default-minor-propagating components so the constraint cascade happens automatically.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -127,6 +127,12 @@ aft:
       packages:
         - worker_bee
         - worker_bee_builder
+    - name: Foundation Dart
+      packages:
+        - amplify_foundation_dart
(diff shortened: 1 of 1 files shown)


cadivus:
Added `amplify_record_cache_dart` as a solo component too — pre-1.0 so it's inert today, but self-activates the dependent cascade the moment it bumps to 1.0.0 (same landmine `foundation_dart` hit).

soberm (approved):

PR #7129, opened 2026-07-19
fix(foundation): Autogenerate `version.dart`

cadivus:
`packageVersion` in `amplify_foundation_dart` was stuck at `2.10.0` (used in kinesis/firehose user agents).

- Add `build_version` so `aft` version bumps regenerate `version.dart`
- Regenerate: `2.10.0` → `2.12.0`

## Verification
- `dart analyze`: no issues
- `dart test -p vm`: all pass

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #7130, opened 2026-07-19
ci: Verify `version.dart` is autogenerated

cadivus:
Follow up of https://github.com/aws-amplify/amplify-flutter/pull/7129

---

Catches stale `version.dart` files (like amplify_foundation_dart being stuck at 2.10.0, fixed in #7129).

New CI step in `dart_vm` and `flutter_vm` (stable only):
1. If `lib/src/version.dart` exists: save it, delete it, regenerate via `build_runner`
2. Fail if it's not regenerated or differs, with a hint to check `build_version` setup
3. No `version.dart` → skip

## Verification
Ran the script locally on three cases:
- `aws_signature_v4` (correct setup) → pass
- `aws_common` (no version.dart) → skip
- `amplify_foundation_dart` on main (stale) → fail with hint

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.


.github/composite_actions/check_version_dart/action.yaml
```diff
@@ -0,0 +1,42 @@
...
+        if [ ! -f lib/src/version.dart ]; then
+          echo "No lib/src/version.dart — skipping"
+          exit 0
+        fi
+        cp lib/src/version.dart "$RUNNER_TEMP/version.dart.orig"
+        rm lib/src/version.dart
+        # Full build, no --build-filter: build_runner deletes committed
+        # generated files as conflicting outputs, so a filtered build would
+        # delete sibling *.g.dart files without regenerating them.
+        dart run build_runner build --delete-conflicting-outputs || true
```
soberm, line 29: [minor] The `|| true` swallows the build_runner exit code entirely. When build_runner fails for a reason unrelated to `version.dart` — a compilation error elsewhere in the package, a transient pub resolution failure, OOM — the script falls through to the `if [ ! -f lib/src/version.dart ]` branch and emits a misleading hint about `build_version` configuration when that's not actually the problem.

The `|| true` seems to exist to guarantee `git checkout -- .` runs for cleanup regardless of build_runner's outcome. You can get both correct diagnostics and guaranteed cleanup by capturing the exit code separately and moving the restore outside the conditional:

```bash
dart run build_runner build --delete-conflicting-outputs
br_exit=$?
result=0
if [ $br_exit -ne 0 ]; then
  echo "::error::build_runner failed (exit $br_exit) — see above for details"
  result=1
elif [ ! -f lib/src/version.dart ]; then
  echo "::error::lib/src/version.dart was not regenerated..."
  result=1
elif ! diff "$RUNNER_TEMP/version.dart.orig" lib/src/version.dart; then
  echo "::error::The committed lib/src/version.dart does not match..."
  result=1
else
  echo "lib/src/version.dart matches the generated output"
fi
# Always restore — unconditional, not nested inside the if chain.
git checkout -- .
exit $result
```

.github/composite_actions/check_version_dart/action.yaml
```diff
@@ -0,0 +1,42 @@
...
+          echo "::error::lib/src/version.dart was not regenerated by build_runner. Is 'build_version' configured correctly in dev_dependencies?"
+          result=1
+        elif ! diff "$RUNNER_TEMP/version.dart.orig" lib/src/version.dart; then
+          echo "::error::The committed lib/src/version.dart does not match the generated output. Is 'build_version' used the right way? Run 'dart run build_runner build' and commit the result."
+          result=1
+        else
+          echo "lib/src/version.dart matches the generated output"
+        fi
+        # Leave the tree exactly as committed for the following steps.
+        git checkout -- .
```
soberm, line 41: [nit] `git checkout -- .` restores modified/deleted tracked files but leaves any newly created untracked files from the build_runner run. If build_runner generates a file that wasn't previously tracked, it'll linger for subsequent steps. Since this step runs last (only the CloudWatch logger follows), it won't cause a real problem in practice — but `git clean -fd -- .` after the checkout would make the restore fully hermetic.

.github/composite_actions/check_version_dart/action.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+runs:
+  using: "composite"
+  steps:
+    - name: Check version.dart is generated
+      shell: bash
+      working-directory: ${{ inputs.working-directory }}
+      run: |
+        if [ ! -f lib/src/version.dart ]; then
+          echo "No lib/src/version.dart — skipping"
```
soberm: [nit] Unicode em dash (—) in a shell string. Modern runners handle UTF-8 fine so this won't break anything, and adjacent CI scripts in this repo tend to stick to ASCII in shell output strings. Optional.

soberm (commented):
CI guard for stale `version.dart` — solid design. Composite action is the right abstraction, the ordering rationale is well-commented, and the `always() && steps.bootstrap.conclusion == 'success'` guard is exactly right. One minor issue with `|| true` on the `build_runner` invocation worth fixing before merge; everything else is nits.

.github/composite_actions/check_version_dart/action.yaml
cadivus: This should be fine here since we ran all the bootstrap and build runners as before.

.github/composite_actions/check_version_dart/action.yaml
cadivus: Wouldn't this also undo `aft bootstrap`?

.github/composite_actions/check_version_dart/action.yaml
cadivus: Fixed

soberm (approved):

PR #7168, opened 2026-07-24
feat(firehose): Add wasm support for Firehose

cadivus:
Adds dart2wasm support for the Firehose family: `amplify_firehose`, `amplify_firehose_dart`, and the shared `amplify_record_cache_dart`.

Smoke tests opt each package into the dart2wasm CI leg and exercise the web IndexedDB storage path. Verified a live record + flush round trip against a Firehose sandbox under both dart2js and dart2wasm.


osama-rizk (commented):
Reviewed the wasm wiring + smoke tests. Overall clean, well-scoped, and it matches the pattern already used by the ~7 packages that support dart2wasm. Nice touch on the `record_cache` round-trip test asserting a real add→read→delete cycle so the `JSNumber.toDartInt` conversions are actually exercised, not just compiled — that's exactly the wasm-specific risk.

A few non-blocking observations inline. The main one worth a decision is the `native_test` leg on `amplify_record_cache_dart`.

.github/workflows/amplify_record_cache_dart.yaml
```diff
@@ -73,3 +83,31 @@ jobs:
     with:
       package-name: amplify_record_cache_dart
       working-directory: packages/kinesis/amplify_record_cache_dart
+  native_test:
```
osama-rizk, line 86: The new `native_test` (VM) leg runs almost no real coverage for this package. The entire test suite here is `wasm_smoke_test.dart`, and only its first test (`RecordInput carries its fields`, a plain constructor check) is VM-runnable — the round-trip is `testOn: 'browser'`. So the VM sqlite storage path (`record_storage_sqlite.dart`) gets no coverage, while this leg spins up a macOS + Windows matrix.

Is VM/sqlite coverage intentionally out of scope for a wasm PR? If so, does this leg earn its CI minutes, or is it here only for parity with `amplify_firehose_dart`? Either add a VM storage round-trip, or consider dropping the native leg.

.github/workflows/amplify_record_cache_dart.yaml
```diff
@@ -73,3 +83,31 @@ jobs:
...
+    needs: test
+    uses: ./.github/workflows/dart_native.yaml
+    secrets: inherit
```
osama-rizk, line 89: Nit: trailing whitespace after `inherit`.

packages/kinesis/amplify_firehose/example/test/wasm_smoke_test.dart
```diff
@@ -0,0 +1,23 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+library;
+
+import 'package:amplify_firehose/amplify_firehose.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+/// Presence of this file opts the example into the `dart2wasm` E2E leg
```
osama-rizk, line 9: Small doc-accuracy nit: the comment says this file's presence "opts the example into the `dart2wasm` E2E leg (`run-wasm: true`)", but the actual trigger is `run-wasm: true` set by hand in `amplify_firehose_example.yaml`. The two are kept in sync manually. If `aft` auto-generates that input from this file's presence, the comment is correct and ignore this — otherwise it reads as more automatic than it is.

packages/kinesis/amplify_firehose_dart/test/wasm_smoke_test.dart
```diff
@@ -0,0 +1,45 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+@TestOn('browser')
```
osama-rizk, line 4: Worth a one-line note here explaining why this file is fully `@TestOn('browser')`, whereas the `amplify_record_cache_dart` twin deliberately keeps one platform-neutral test "so the VM test job has something to run." This file is fine — the package has other VM tests — but the differing rationale between the two twins could trip up the next person into "fixing" one to match the other.

.github/workflows/amplify_record_cache_dart.yaml
cadivus: Good catch. Added `record_storage_sqlite_test.dart` — a VM round-trip (add→read→delete) over the sqlite path, mirroring the IndexedDB test. Native leg now earns its minutes. (183bd6144)

packages/kinesis/amplify_firehose_dart/test/wasm_smoke_test.dart
cadivus: Made the twins consistent: with the VM sqlite test now covering that path, `record_cache`'s `wasm_smoke_test.dart` is also fully `@TestOn('browser')`. No more divergent rationale. (183bd6144)

.github/workflows/amplify_record_cache_dart.yaml
cadivus: This is emitted by the `aft generate workflows` template (all 35 generated workflows have it). Not editable by hand here — worth a one-line fix in the generator as a separate cleanup.

packages/kinesis/amplify_firehose/example/test/wasm_smoke_test.dart
cadivus: The comment is accurate — it's auto-generated, not hand-synced. `aft`'s `_e2eNeedsWasm` maps the example to its `_dart` sibling and emits `run-wasm: true` when this file is present (generate_workflows_command.dart). No manual step.

osama-rizk (approved):

PR #7228, opened 2026-08-03
fix(core): support Connect-only notifications shape in AmplifyOutputs

ekjotmultani:
## Description

`Amplify.configure` throws a parse error when `amplify_outputs.json` contains the notifications shape emitted by the `@aws-amplify/backend-notifications` construct:

```json
{
  "notifications": {
    "amazon_connect": {
      "aws_region": "us-east-1",
      "endpoint": "https://<api-id>.execute-api.<region>.amazonaws.com"
    }
  }
}
```

`NotificationsOutputs` declares the legacy Pinpoint fields (`aws_region`, `amazon_pinpoint_app_id`, `channels`) as required, so the section fails to parse when only `amazon_connect` is present. Until this change every app consuming the construct had to strip the `notifications` key before calling `Amplify.configure`, which is what the `amplify_connect_client` example app was doing.

### Changes

- Make `awsRegion` and `amazonPinpointAppId` nullable on `NotificationsOutputs`, and default `channels` to an empty list, with doc comments noting that `notifications` may be present without Pinpoint being configured — for example when only Amazon Connect is provisioned.
- Add `AmazonConnectOutputs` (`aws_region` + `endpoint`, required within the sub-object) and an `amazonConnect` field, matching how amplify-android (`AmplifyOutputsData.Notifications.AmazonConnect`) and amplify-swift (optional `Notifications` fields) already handle this shape on their mains.
- Make the same fields optional on the darwin `AmplifyOutputsData.Notifications` struct in `amplify_datastore`, so a `notifications` key holding only `amazon_connect` no longer fails the native `Codable` decode.
- Drop the strip-`notifications` workaround from the `amplify_connect_client` example app and its integration test setup. Both now pass the full outputs to `Amplify.configure`, making the example an end-to-end check of the Connect-only shape.
- `pinpoint_provider.dart` is the only consumer reading the now-nullable fields; it throws its existing `ConfigurationError` when the Pinpoint fields are absent, with behavior unchanged when they are present.
- `.g.dart` files regenerated with build_runner.

### Tests

New parse tests covering: Connect-only shape, legacy Pinpoint-only shape, both sections together, and full outputs with `auth` + `notifications.amazon_connect` round-tripping through `AmplifyOutputs.fromJson`.

- `amplify_core`: `dart test -p vm` — 160 tests passing (4 new), `dart analyze` clean
- `amplify_push_notifications_pinpoint`: 15 tests passing, analyze clean
- `amplify_push_notifications`: 28 tests passing, analyze clean
- `amplify_connect_client` example: `flutter analyze` clean
- darwin `AmplifyOutputsData.swift`: `xcrun swiftc -parse` clean

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

### DIFF
--- a/packages/amplify_core/lib/src/config/amplify_outputs/notifications/notifications_outputs.dart
+++ b/packages/amplify_core/lib/src/config/amplify_outputs/notifications/notifications_outputs.dart
@@ -2,37 +2,59 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import 'package:amplify_core/amplify_core.dart';
+import 'package:amplify_core/src/config/amplify_outputs/notifications/amazon_connect_outputs.dart';
 import 'package:amplify_core/src/config/amplify_outputs/notifications/amazon_pinpoint_channel.dart';
 
 part 'notifications_outputs.g.dart';
 
 /// {@template amplify_core.amplify_outputs.notifications_outputs}
 /// The Amplify Gen 2 outputs for Notifications category.
+///
+/// Notifications may be present without Pinpoint being configured — for
+/// example when only Amazon Connect is provisioned.
 /// {@endtemplate}
 @zAmplifyOutputsSerializable
 class NotificationsOutputs
     with AWSEquatable<NotificationsOutputs>, AWSSerializable, AWSDebuggable {
   /// {@macro amplify_core.amplify_outputs.notifications_outputs}
   const NotificationsOutputs({
-    required this.awsRegion,
-    required this.amazonPinpointAppId,
-    required this.channels,
+    this.awsRegion,
+    this.amazonPinpointAppId,
+    this.channels,
+    this.amazonConnect,
   });
 
   factory NotificationsOutputs.fromJson(Map<String, Object?> json) =>
       _$NotificationsOutputsFromJson(json);
 
   /// The AWS region of Amazon Pinpoint resources.
-  final String awsRegion;
+  ///
+  /// May be `null` when Pinpoint is not configured — for example when only
+  /// Amazon Connect is provisioned.
+  final String? awsRegion;
 
   /// Amazon Pinpoint App ID.
-  final String amazonPinpointAppId;
+  ///
+  /// May be `null` when Pinpoint is not configured — for example when only
+  /// Amazon Connect is provisioned.
+  final String? amazonPinpointAppId;
 
   /// {@macro amplify_core.amplify_outputs.amazon_pinpoint_channel}
-  final List<AmazonPinpointChannel> channels;
+  ///
+  /// May be `null` when Pinpoint is not configured — for example when only
+  /// Amazon Connect is provisioned.
+  final List<AmazonPinpointChannel>? channels;
+
+  /// {@macro amplify_core.amplify_outputs.amazon_connect_outputs}
+  final AmazonConnectOutputs? amazonConnect;
 
   @override
-  List<Object?> get props => [awsRegion, amazonPinpointAppId, channels];
+  List<Object?> get props => [
+    awsRegion,
+    amazonPinpointAppId,
+    channels,
+    amazonConnect,
+  ];
 
   @override
   String get runtimeTypeName => 'NotificationsOutputs';
--- a/packages/notifications/push/amplify_push_notifications_pinpoint/lib/src/pinpoint_provider.dart
+++ b/packages/notifications/push/amplify_push_notifications_pinpoint/lib/src/pinpoint_provider.dart
@@ -90,6 +90,17 @@ class PinpointProvider implements ServiceProviderClient {
         final region = config.awsRegion;
         final appId = config.amazonPinpointAppId;
 
+        if (region == null || appId == null) {
+          throw ConfigurationError(
+            'No Pinpoint configuration found for the Notifications category. '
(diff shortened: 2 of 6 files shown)


cadivus:
Do we need to adapt https://github.com/aws-amplify/amplify-flutter/blob/fix/notifications-outputs-connect-schema/packages/amplify_datastore/darwin/amplify_datastore/Sources/amplify_datastore/AmplifyInternal/Core/Configuration/AmplifyOutputsData.swift#L168L173 too to make it optional here? So eventually like
```Dart
      @_spi(InternalAmplifyConfiguration)
      public struct Notifications: Codable {
          public let awsRegion: String?
          public let amazonPinpointAppId: String?
          public let channels: [AmazonPinpointChannelType]?
          public let amazonConnect: AmazonConnect?

          @_spi(InternalAmplifyConfiguration)
          public struct AmazonConnect: Codable {
              public let awsRegion: String
              public let endpoint: String
          }
      }
```

packages/amplify_core/lib/src/config/amplify_outputs/notifications/notifications_outputs.dart
```diff
@@ -2,37 +2,59 @@
...
+  /// May be `null` when Pinpoint is not configured — for example when only
+  /// Amazon Connect is provisioned.
+  final String? amazonPinpointAppId;
 
   /// {@macro amplify_core.amplify_outputs.amazon_pinpoint_channel}
-  final List<AmazonPinpointChannel> channels;
+  ///
+  /// May be `null` when Pinpoint is not configured — for example when only
+  /// Amazon Connect is provisioned.
+  final List<AmazonPinpointChannel>? channels;
```
cadivus: Now, this is nullable instead of sett ing an empty list as a default.  The Android native code has it as an empty list on default: https://github.com/aws-amplify/amplify-android/blob/a91e94e39315593a3214becfe82afeaec432e172/core/src/main/java/com/amplifyframework/core/configuration/AmplifyOutputsData.kt#L173 Same in the darwin native code above (what also might change my suggestion for the needed code change).

packages/notifications/push/amplify_push_notifications_pinpoint/lib/src/pinpoint_provider.dart
```diff
@@ -90,6 +90,17 @@ class PinpointProvider implements ServiceProviderClient {
...
+          throw ConfigurationError(
+            'No Pinpoint configuration found for the Notifications category. '
+            'Notifications may be present without Pinpoint being configured — '
+            'for example when only Amazon Connect is provisioned.',
+            recoverySuggestion:
+                'Make sure Amazon Pinpoint is configured in the Notifications '
+                'category of your Amplify configuration.',
+          );
+        }
+
```
cadivus, line 103: Shouldn't we add a test for this new branch?

packages/amplify_core/lib/src/config/amplify_outputs/notifications/notifications_outputs.dart
ekjotmultani: Good catch, switched it to `const []` so it matches. `channels` is non-nullable again and just comes back empty when there's no Pinpoint config, same as the Kotlin side. Also went and checked the Swift one and it's optional there rather than defaulted, so the three aren't perfectly aligned, but empty-list is the better default for us here since it keeps every existing `channels` call site working untouched.

packages/notifications/push/amplify_push_notifications_pinpoint/lib/src/pinpoint_provider.dart
ekjotmultani: Yep, added one

packages/amplify_core/lib/src/config/amplify_outputs/notifications/notifications_outputs.dart
```diff
@@ -2,37 +2,59 @@
...
     with AWSEquatable<NotificationsOutputs>, AWSSerializable, AWSDebuggable {
   /// {@macro amplify_core.amplify_outputs.notifications_outputs}
   const NotificationsOutputs({
-    required this.awsRegion,
-    required this.amazonPinpointAppId,
-    required this.channels,
+    this.awsRegion,
+    this.amazonPinpointAppId,
+    this.channels = const [],
+    this.amazonConnect,
```
cadivus, line 24: You changed many fields to optional here. Don't we need to adapt native code like this one: https://github.com/aws-amplify/amplify-flutter/blob/9d21da7bccfe86619700b7e11a1fe8096cd1c266/packages/amplify_datastore/darwin/amplify_datastore/Sources/amplify_datastore/AmplifyInternal/Core/Configuration/AmplifyOutputsData.swift#L169-L173 ?

ekjotmultani: Yeah it does. Made the same fields optional in the darwin copy, and it turns out upstream amplify-swift already shipped this exact change on that struct, so we're just matching them. Android is a different story, this repo doesn't parse the outputs at all, it hands them to com.amplifyframework, which had the same bug and fixed it in 2.39.0. So the Android side would be a dependency bump rather than a code change here, if we want it.

PR #7234, opened 2026-08-04
Ios27 support (Xcode 27 runners)

cadivus:


cadivus:
https://github.com/aws-amplify/amplify-flutter/pull/7239 is merged. At one point, we need to upgrade the runners too (after it being beta)

PR #7237, opened 2026-08-05
fix(api): serialize TemporalDateTime in identifier for ModelQueries.get

VarshithaPamisetty:
`ModelQueries.get()` throws `JsonUnsupportedObjectError` when an identifier has a
`TemporalDateTime` in it. The get path doesn't run identifier values through
`_getSerializedValue` like the list/filter path does, so the date never gets
stringified before JSON-encoding.

Re: #6298.

- Serialize `get` variables via `buildVariablesForGetRequest`.
- Serialize the composite-key `get` document values too (date comes out quoted).
- Add `CpkTemporalPrimaryKey` test model + regression test.
- Add matching `CpkTemporalPrimaryKey` model to the `api-multi-auth` gen2 backend.

`dart analyze` clean, tests red to green. Draft: live AppSync e2e still pending backend deploy + codegen.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #6298, opened 2025-08-13
Get Query Using TemporalDateTime Objects


Maanasaaaaa:
### Environment information

```plain text
System:
  OS: macOS 15.5
  CPU: (10) arm64 Apple M4
  Memory: 174.03 MB / 16.00 GB
  Shell: /bin/zsh
Binaries:
  Node: 20.19.3 - ~/.nvm/versions/node/v20.19.3/bin/node
  Yarn: undefined - undefined
  npm: 10.8.2 - ~/.nvm/versions/node/v20.19.3/bin/npm
  pnpm: undefined - undefined
NPM Packages:
  @aws-amplify/auth-construct: 1.8.1
  @aws-amplify/backend: 1.16.1
  @aws-amplify/backend-ai: Not Found
  @aws-amplify/backend-auth: 1.7.1
  @aws-amplify/backend-cli: 1.8.0
  @aws-amplify/backend-data: 1.6.1
  @aws-amplify/backend-deployer: 2.1.3
  @aws-amplify/backend-function: 1.14.1
  @aws-amplify/backend-output-schemas: 1.7.0
  @aws-amplify/backend-output-storage: 1.3.1
  @aws-amplify/backend-secret: 1.4.0
  @aws-amplify/backend-storage: 1.4.1
  @aws-amplify/cli-core: 2.2.1
  @aws-amplify/client-config: 1.8.0
  @aws-amplify/data-construct: 1.16.3
  @aws-amplify/data-schema: 1.21.1
  @aws-amplify/deployed-backend-client: 1.8.0
  @aws-amplify/form-generator: 1.2.4
  @aws-amplify/model-generator: 1.2.0
  @aws-amplify/platform-core: 1.10.0
  @aws-amplify/plugin-types: 1.11.0
  @aws-amplify/sandbox: 2.1.2
  @aws-amplify/schema-generator: 1.4.0
  @aws-cdk/toolkit-lib: 1.1.1
  aws-amplify: 6.15.5
  aws-cdk-lib: 2.204.0
  typescript: 5.9.2
npm warn exec The following package was not found and will be installed: cdk@2.1024.0
No AWS environment variables
No CDK environment variables
```

### Describe the bug

1. When using Amplify's ModelQueries.get() in a Flutter application with a schema that includes a TemporalDateTime field, the query fails with a JsonUnsupportedObjectError. This occurs when attempting to serialize a TemporalDateTime object for a GraphQL request.

[...]

PR #7255, opened 2026-08-11
fix(api): Serialize TemporalDateTime in identifier for ModelQueries.get

VarshithaPamisetty:
## Description

`ModelQueries.get()` on a model whose custom identifier includes a `TemporalDateTime` threw `JsonUnsupportedObjectError`. The get-by-identifier path passed identifier values straight to JSON encoding without running them through `_getSerializedValue` (which the list/filter path already uses), so the `TemporalDateTime` was never stringified.

This serializes identifier values for both the get **variables** and the composite-key **document**. Adds a Gen2-generated `CpkTemporalPrimaryKey` test model + regression test.

## Testing

- New regression test in `graphql_test.dart` (asserts the get document and JSON-encodable, correctly-serialized variables).
- Verified end-to-end against a live AppSync API: create + `ModelQueries.get()` by the composite key with a `TemporalDateTime`, across multiple date formats (incl. timezone offset), all returning cleanly.

Supersedes #7237.

Issue: #6298


packages/api/amplify_api_dart/lib/src/graphql/factories/model_queries_factory.dart
```diff
@@ -22,7 +22,8 @@ class ModelQueriesFactory {
     APIAuthorizationType? authorizationMode,
     Map<String, String>? headers,
   }) {
-    final variables = modelIdentifier.serializeAsMap();
+    final variables = GraphQLRequestFactory.instance
+        .buildVariablesForGetRequest(modelIdentifier: modelIdentifier);
```
cadivus: I think this need to be added to `delete` operations too.

packages/api/amplify_api_dart/test/test_models/CpkTemporalPrimaryKey.dart
```diff
@@ -0,0 +1,329 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// NOTE: This file is generated and may not follow lint rules defined in your app
+// Generated files can be excluded from analysis in analysis_options.yaml
+// For more info, see: https://dart.dev/guides/language/analysis-options#excluding-code-from-analysis
```
cadivus, line 18: Who/what generated it? The Amazon header is usually not used for autogenerated files.

packages/api/amplify_api_dart/lib/src/graphql/factories/model_queries_factory.dart
cadivus: And do we need to adapt test models (e.g. `test/test_models/schema.graphql`)? Do we need to add something like
```
type CpkTemporalPrimaryKey @model {
  hwid: String! @primaryKey(sortKeyFields: ["sessionStart"])
  sessionStart: AWSDateTime!
}
```

infra-gen2/backends/api/api-multi-auth/amplify/data/resource.ts
```diff
@@ -106,6 +106,14 @@ const schema = a.schema({
...
     .authorization((allow) => [allow.owner()]),
 
+  CpkTemporalPrimaryKey: a
+    .model({
+      hwid: a.string().required(),
+      sessionStart: a.datetime().required(),
+    })
+    .identifier(["hwid", "sessionStart"])
+    .authorization((allow) => [allow.authenticated()]),
+
```
cadivus: Where is this used in the modified tests? If we want to extend e2e, we should also modify the test itself (`example/integration_test/graphql/user_pools_test.dart`).

packages/api/amplify_api_dart/test/test_models/CpkTemporalPrimaryKey.dart
VarshithaPamisetty: I added the header manually to match our license convention, it's not from modelgen. Yes, it doesn't belong on a generated file. I'll remove it and restore the generator's default.

packages/api/amplify_api_dart/lib/src/graphql/factories/model_queries_factory.dart
VarshithaPamisetty: Added the delete operation and the `CpkTemporalPrimaryKey` type to the test schema (as suggested).

infra-gen2/backends/api/api-multi-auth/amplify/data/resource.ts
VarshithaPamisetty: Added an e2e test that exercises it, along with the generated example model it needs.

PR #7256, opened 2026-08-11
feat(aft): CI mode for `publish` command

cadivus:
This enables us to automatically push new package versions to pub.dev after a version bump. Because of the requirements by https://dart.dev/tools/pub/automated-publishing#configuring-a-github-action-workflow-for-publishing-to-pub-dev, we need to create a new tag for each bumped package that triggers its own workflow. In order to be able to do this, we introduce a Github app that has the needed permissions. 

The `--ci` mode of `aft publish` will push tags instead of running `dart pub publish`. Everything else stays the same and has already proven its functionality. The release-tag workflows runs a few checks to make sure the tag matches the package version.
The version bump is triggered with another workflow that has also shown its functionality a few times - it stays untouched. Since version bumping is its own PR, releases are triggered intentionally. 

Tested in a private repo by mocking a few things.


ernest24 (approved):

.github/workflows/release.yaml
```diff
@@ -81,44 +75,90 @@ jobs:
...
+          #!/usr/bin/env bash
+          set -euo pipefail
+          b64url() { openssl base64 -A | tr '+/' '-_' | tr -d '='; }
+          now=$(date +%s)
+          header=$(printf '{"alg":"RS256","typ":"JWT"}' | b64url)
+          payload=$(printf '{"iat":%d,"exp":%d,"iss":"%s"}' "$((now-60))" "$((now+300))" "$APP_ID" | b64url)
+          sig=$(printf '%s.%s' "$header" "$payload" \
+            | openssl dgst -sha256 -sign "$APP_PRIVATE_KEY_FILE" | b64url)
+          jwt="$header.$payload.$sig"
+          inst_id=$(curl -sf -H "Authorization: Bearer $jwt" \
```
soberm: [minor] The curl calls here (JWT exchange, then the installation-token exchange two lines down) don't have retry flags, but the comment above says git mints a fresh token per operation, so this can run several times per release, and one transient blip to api.github.com fails the whole job. publish-tag.yaml's pub.dev OIDC step already uses `--retry 5 --retry-connrefused`, might as well match it here.

.github/workflows/release.yaml
```diff
@@ -81,44 +75,90 @@ jobs:
...
+            "https://api.github.com/repos/$GITHUB_REPOSITORY/installation" | jq -r .id)
+          curl -sf -X POST -H "Authorization: Bearer $jwt" \
+            "https://api.github.com/app/installations/$inst_id/access_tokens" | jq -r .token
```
soberm, line 158: [nit] This request doesn't send a body, so the token comes back scoped to whatever the App is configured with rather than an explicit narrowed `{"permissions":..., "repositories":...}`. Can't tell from here if the App's already minimally scoped at the settings level, but sending an explicit narrow scope in the request itself would be a cheap belt-and-suspenders addition.

.github/workflows/release.yaml
```diff
@@ -81,44 +75,90 @@ jobs:
...
+          SCRIPT
+          chmod +x "$MINT"
+
+          git config --global user.name "flutter-release[bot]"
+          git config --global user.email "flutter-release[bot]@users.noreply.github.com"
+
+          git config --global credential.https://github.com.helper \
+            "!f() { test \"\$1\" = get || return 0; echo username=x-access-token; echo password=\$(APP_ID=$APP_ID APP_PRIVATE_KEY_FILE=$KEY_FILE GITHUB_REPOSITORY=$GITHUB_REPOSITORY $MINT); }; f"
```
soberm, line 166: [minor] This token never gets `::add-mask::`'d before it's used here, unlike `PUB_TOKEN` in publish-tag.yaml. Probably harmless today since it only flows through the credential helper's stdout, but this is a public repo (logs are public too) and masking is already the pattern one file over, worth adding for consistency.

packages/aft/lib/src/commands/publish_command.dart
```diff
@@ -235,6 +241,63 @@ mixin PublishHelpers on AmplifyCommand {
...
+    ], throwOnError: false);
+    return result.exitCode == 0;
+  }
+
+  /// Creates and pushes the release tag for [package] in lieu of publishing it
+  /// with `pub`.
+  ///
+  /// An existing tag on the current commit is skipped so a partially completed
+  /// release can be re-run; one on any other commit is a conflict.
+  Future<void> publishTag(PackageInfo package) async {
```
soberm, line 265: [major] This is new tag create/skip/conflict/rollback logic driving the real release pipeline, and there's no test covering the skip-if-same-commit vs. conflict-if-different-commit branching. Worth adding, a regression here would silently mis-tag a release.

packages/aft/lib/src/commands/publish_command.dart
```diff
@@ -293,6 +386,19 @@ class PublishCommand extends AmplifyCommand with GlobOptions, PublishHelpers {
 
   @override
   Future<void> run() async {
+    if (ci && dryRun) {
+      exitError('--ci cannot be combined with --dry-run');
```
soberm, line 390: [minor] Same gap here, the `--ci`/`--prepare-only`/`--skip-pre-publish` combination validation has no test exercising the invalid-combination paths.

.github/workflows/release.yaml
```diff
@@ -81,44 +75,90 @@ jobs:
...
+  release:
+    name: Release packages
+    needs: verify
+    if: needs.verify.outputs.has_tags == 'true'
+    runs-on:
+      - codebuild-amplify-flutter-release-${{ github.run_id }}-${{ github.run_attempt }}
+      - image:ubuntu-8.0
+      - instance-size:medium
+    steps:
+      - name: Set up runner
```
soberm, line 97: [nit] "Set up runner" and "Install aft" here are duplicated verbatim from the `verify` job above (line 34 onward). Could pull into a composite action so they can't drift out of sync later.

.github/workflows/publish-tag.yaml
```diff
@@ -0,0 +1,176 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
+permissions: {}
+
+defaults:
+  run:
+    shell: bash
```
soberm, line 17: [nit] This sets an explicit bash default; release.yaml doesn't (relies on the runner default). Not a bug, just inconsistent now that these two workflows are tightly coupled.

.github/workflows/release.yaml
cadivus: I’m leaving this unchanged. The helper’s stdout is the Git credential protocol (`username=...` / `password=...`), so emitting `::add-mask::` there would be consumed by Git and can break authentication. Sending the minted token to the runner log just to register a mask would also widen its exposure; today it only travels from the helper directly to Git. Keeping it out of workflow output is the safer boundary.

.github/workflows/release.yaml
cadivus: I’m keeping these steps separate here. Runner setup intentionally happens before checkout so Git ownership is configured first, but a repository-local composite action cannot be loaded until checkout has completed. Moving checkout earlier just to enable the action would change that bootstrap order. The post-checkout `Install aft` duplication is only three commands, so I’d rather avoid the extra coupling in this release path.

packages/aft/lib/src/commands/publish_command.dart
cadivus: This should be fine

.github/workflows/publish-tag.yaml
cadivus: Added

.github/workflows/release.yaml
cadivus: Changed

.github/workflows/release.yaml
cadivus: added

soberm (approved):

PR #7263, opened 2026-08-11
fix(aws_common): Retry transient transport failures (AWSHttpException)

VarshithaPamisetty:
## Description
`AWSRetryer` only retried based on the HTTP response, so a connection that died before any response arrived was never retried. This showed up as flaky Windows S3 e2e runs failing on `HandshakeException: Connection terminated during handshake`.

The IO HTTP client now flags `SocketException` and `HandshakeException` as retryable, and the retryer retries only those. The web client has no equivalent signal.

## Tests
- `aws_common`: a terminated TLS handshake and a refused connection are flagged retryable; other errors are not.
- `smithy_aws`: retryer retries a flagged `AWSHttpException` and recovers; does not retry a non-flagged one.
- `dart analyze --fatal-infos --fatal-warnings` and `dart format` clean.


packages/smithy/smithy_aws/lib/src/http/retry/aws_retryer.dart
```diff
@@ -96,6 +96,10 @@ class AWSRetryer implements Retryer {
     if (exception is TimeoutException) {
       return true;
     }
+    // Transport-level failures never reached the server, so are safe to retry.
+    if (exception is AWSHttpException) {
+      return true;
+    }
```
cadivus: Are you sure that this is always transport-level? I see these two exceptions: 
 - https://github.com/aws-amplify/amplify-flutter/blob/df9817b091e42323a4050fd00b7702d3323126b9/packages/common/amplify_db_common_dart/lib/src/connect_html.dart#L74L80
 - https://github.com/aws-amplify/amplify-flutter/blob/df9817b091e42323a4050fd00b7702d3323126b9/packages/common/amplify_db_common_dart/lib/src/connect_html.dart#L62L66

Should we e.g. check the exception? See if we see specific HTTP codes? We could have a look at the actual exception we want to retry.

VarshithaPamisetty: Good point. `AWSHttpException` has no status code so we can't check codes, and those two are response errors, not transport. Instead of trusting the type, the HTTP client now marks the real transport failures with `AWSHttpException.retryable(...)` and the retryer only retries the flagged ones, so those two stay non-retryable.

cadivus (dismissed):

Issue #6395, opened 2025-09-29
signInWithWebUI throws UrlLauncherException instead of UserCancelledException when running on release mode


nikorehnback:
### Description

When launching SSO login with Android and returning back UrlLauncherException gets thrown by Amplify Auth instead of UserCancelledException.

This happens only when running on **release** mode with `--release` flag.

Issue does not reproduce when running with **debug** or **profile** mode.

### Categories

- [ ] Analytics
- [ ] API (REST)
- [ ] API (GraphQL)
- [x] Auth
- [ ] Authenticator
- [ ] DataStore
- [ ] Notifications (Push)
- [ ] Storage

### Steps to Reproduce

Below is minimal sample Flutter code to reproduce the issue.

_AndroidManifest.xml_ 
---
    <queries>
        <intent>
            <action android:name="android.intent.action.SEND" />
            <data android:mimeType="text/plain" />
        </intent>
        <intent>
            <action android:name="android.intent.action.VIEW" />
            <data android:scheme="https" />
        </intent>
        <intent>
            <action android:name="android.support.customtabs.action.CustomTabsService" />
        </intent>
    </queries>

_main.dart_
---

    import 'dart:convert';
    import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
    import 'package:amplify_core/amplify_core.dart';
    import 'package:flutter/material.dart';
    
    void main() {
      runApp(const MyApp());
    }
    
    class MyApp extends StatelessWidget {
      const MyApp({super.key});
    
      @override
      Widget build(BuildContext context) {
        return MaterialApp(
          title: 'Flutter Demo',
          theme: ThemeData(
            colorScheme: ColorScheme.fromSeed(seedColor: Colors.deepPurple),
          ),
          home: const MyHomePage(title: 'Flutter Demo Home Page'),
        );
      }
    }
    
    class MyHomePage extends StatefulWidget {
      const MyHomePage({super.key, required this.title});
    
      final String title;
    
      @override
      State<MyHomePage> createState() => _MyHomePageState();
    }
    
    class _MyHomePageState extends State<MyHomePage> {
      bool _initialized = false;
    
      @override
      void initState() {
        super.initState();
    
        _init();
      }
    
      void _init() async {
        final amplifyConfig = {
          "auth": {
            "plugins": {
              "awsCognitoAuthPlugin": {
                "IdentityManager": {"Default": {}},
                "CredentialsProvider": {
                  "CognitoIdentity": {
[...]

PR #7264, opened 2026-08-11
fix(auth): Preserve HostedUiException names under R8 in release builds

VarshithaPamisetty:
**Description:**
On Android (release only), cancelling `signInWithWebUI` throws `UrlLauncherException` instead of `UserCancelledException`. R8 obfuscates `HostedUiException$CANCELLED`, so its `simpleName` no longer matches the `"CANCELLED"` code Dart checks. Fixed by shipping a consumer ProGuard rule (`consumerProguardFiles`) that keeps `HostedUiException` under R8.

**Tests:**
Verified e2e on a real Cognito Hosted UI (release): cancel → `UrlLauncherException` without the fix, `UserCancelledException` with it.

| Before (no fix) | After (fix) |
|---|---|
| before | after |

Issue: #6395


packages/auth/amplify_auth_cognito/android/consumer-rules.pro
ekjotmultani: I think it should have `-keepattributes InnerClasses,EnclosingMethod`, we might miss nested classes otherwise

cadivus (approved):

cadivus:
Please make sure the workflows pass

VarshithaPamisetty:
Merging the PR since the e2e fails are known flakes and failed the same way on multiple reruns.

PR #7282, opened 2026-08-17
fix(api): Don't block subscription events during WebSocket reconnect

### LINKED ISSUE #7001: Subscription events are blocked while reconnect retry is in progress
### Description

When network connectivity becomes unstable, the WebSocket reconnect flow can enter a state where incoming subscription messages are still being received from AppSync, but they are never delivered to subscription listeners.

After investigating the source code and debugging the internal WebSocket implementation of Amplify, it appears that the reconnect workflow blocks the entire WebSocket event processing pipeline.

Observed Behavior

1. App starts with active GraphQL subscriptions
2. Network becomes slow/intermittent (I used Network link conditioner to test it)
3. Connectivity reports a disconnect and NetworkLossEvent is emitted
4. ReconnectEvent is added and `_reconnect()` begins executing
5. While `_reconnect()` is running, AppSync continues sending and ARE received by the WebSocketService:
    * `KeepAliveEvent`
    * `SubscriptionDataEvent`
6. `AmplifyWebSocketService._onData()` receives these events successfully.
7. `WebSocketBloc._safeAdd()` is called successfully.
8. `_wsEventController.add(event)` is executed successfully.
9. However, `WebSocketBloc._eventTransformer()` is never entered for these events.

As a result:

* Subscription updates stop reaching application code.
* Incoming realtime events are effectively frozen.
* The socket appears healthy enough to continue receiving data, but the events are not processed

After investigating the whole behavior, when a ReconnectEvent is emitted, and taking too long to complete, all healthy, incoming events are queued behind the reconnect operation.

The issue lies here:
[`final blocStream = _wsEventStream.asyncExpand(_eventTransformer);`](https://github.com/aws-amplify/amplify-flutter/blob/5d896cacb4139faef0aa1a807555ccfd88045e59/packages/api/amplify_api_dart/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart#L53)

`asyncExpand` processes events sequentially, the running ReconnectEvent appears to block all subsequent events from being processed.

The expected behavior here is that incoming WebSocket messages should continue to be processed while reconnect logic is running, especially if:

* The WebSocket connection is still receiving keep-alive messages.
* Subscription data messages are still arriving from AppSync.

A reconnect attempt should not block processing of valid incoming subscription traffic like SubscriptionDataEvent, KeepAliveEvent, ConnectionAckMessageEvent and others

Isn't better for reconnect operations be performed outside the main event processing queue?

### Categories

- [ ] Analytics
- [ ] API (REST)
- [x] API (GraphQL)
- [ ] Auth
- [ ] Authenticator
- [ ] DataStore
- [ ] Notifications (Push)
- [ ] Storage

### Steps to Reproduce

1. Create a subscription
2. Simulate a very slow internet connection with high RetryOptions max attempts so that a Reconnect operation keeps running
3. Execute a mutation for that subscription
4. Observe under verbose logging that the WebSocketService received the Data event and keeps receiving KeepAliveEvents but the WebSocketBloc does not.

### Screenshots

_No response_

### Platforms

- [x] iOS
- [x] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Flutter Version

3.44

### Amplify Flutter Version

2.11.0

### Deployment Method

Amplify CLI (Gen 1)

### Schema

```GraphQL

```

VarshithaPamisetty:
## Description

The WebSocket reconnect ran on the serial event queue (`asyncExpand`), so a slow reconnect blocked every event behind it: subscription data and keep-alives were received but never processed, freezing active subscriptions on a healthy socket. This moves reconnect off the queue so events keep flowing, and guards against shutdown mid-reconnect.

## Testing

- Regression test: data flows during an in-progress reconnect (fails without the change); plus a clean-shutdown-mid-reconnect test.
- WebSocket suite 35/35; `analyze`/`format` clean (Dart 3.11 and 3.13).
- Verified e2e against live AppSync: delivery, reconnect recovery, new subscription mid-reconnect.

Fixes #7001


packages/api/amplify_api_dart/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart
```diff
@@ -108,6 +108,9 @@ class WebSocketBloc with AWSDebuggable, AmplifyLoggerMixin {
 
   late WebSocketState _currentState;
 
+  /// The in-flight reconnection, if any. Ensures at most one runs at a time.
+  Future<void>? _reconnectOperation;
```
cadivus, line 112: Don't we need to cancel or await this? E.g. in `_close`?

packages/api/amplify_api_dart/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart
```diff
@@ -378,12 +381,25 @@ class WebSocketBloc with AWSDebuggable, AmplifyLoggerMixin {
...
+  /// events (e.g. subscription data, keep alives) from being processed.
   Stream<WebSocketState> _reconnect() async* {
     assert(
       _currentState is ReconnectingState,
       'Bloc should be set to connecting before starting reconnection.',
     );
-    final state = _currentState;
+    _reconnectOperation ??= _performReconnect(
+      _currentState,
+    ).whenComplete(() => _reconnectOperation = null);
```
cadivus, line 394: Should we handle errors that aren't `Exception` (https://api.dart.dev/dart-core/#exceptions)? `_performReconnect` catches `Exception` only (`} on Exception catch (e, st) `).

packages/api/amplify_api_dart/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart
VarshithaPamisetty: Tested awaiting it in `_close()` and it hangs shutdown (breaks the shutdown-during-reconnect test), and a Future can't be cancelled. Instead, `_isShuttingDown` makes the reconnect a no-op once the bloc closes.

packages/api/amplify_api_dart/lib/src/graphql/web_socket/blocs/web_socket_bloc.dart
VarshithaPamisetty: `on Exception` is intentional here. Per Effective Dart we shouldn't swallow Errors, and the one that actually surfaced in testing, `add-after-close`, is already prevented by the `_isShuttingDown/_safeAdd` guards. So leaving it as-is would be better.

packages/authenticator/amplify_authenticator/example/test_driver/integration_test.dart
```diff
@@ -5,4 +5,4 @@ import 'package:integration_test/integration_test_driver.dart';
 
 // Required for running integration tests in the browser:
 // https://docs.flutter.dev/cookbook/testing/integration/introduction#5b-web
-Future<void> main() => integrationDriver();
+Future<void> main() => integrationDriver(timeout: const Duration(minutes: 60));
```
cadivus, line 8: Minor: Is this change wanted? It looks unrelated.

cadivus (approved):

packages/authenticator/amplify_authenticator/example/test_driver/integration_test.dart
VarshithaPamisetty: The authenticator web-e2e suite runs past the 20-min default driver timeout, so I bumped it to 60 (auth_cognito is already 90). The latest passing run took about ~36min.

PR #7285, opened 2026-08-18
chore: Regenerate files for Flutter 3.47

cadivus:
Flutter 3.47.0 added `AnalysisOptionsMigration` (flutter/flutter#187940), which runs on `flutter pub get` and appends `build/**` plus the platform directories to each project's `analysis_options.yaml`. The same `pub get` also regenerates the macOS plugin registrants, which pick up `path_provider_foundation` again in the examples whose graph resolves it to 2.5.1 — 2.6.0 dropped the native `pluginClass` in favor of Dart-only registration.

Applying this repo-wide keeps the churn out of unrelated commits, e.g. the `aft version-bump` workflow, which runs `flutter pub get` in `amplify_authenticator` and commits with `git add -A`.

Generated with `aft bootstrap --no-upgrade --no-build` on Flutter 3.47.0.

That regeneration is needed was found here: https://github.com/aws-amplify/amplify-flutter/pull/7283

### DIFF
--- a/packages/analytics/amplify_analytics_pinpoint/pubspec.yaml
+++ b/packages/analytics/amplify_analytics_pinpoint/pubspec.yaml
@@ -20,9 +20,9 @@ platforms:
 
 dependencies:
   amplify_analytics_pinpoint_dart: ">=0.4.20 <0.5.0"
-  amplify_core: ">=2.15.0 <2.16.0"
-  amplify_db_common: ">=0.4.21 <0.5.0"
-  amplify_secure_storage: ">=0.5.21 <0.6.0"
+  amplify_core: ">=2.14.0 <2.15.0"
+  amplify_db_common: ">=0.4.20 <0.5.0"
+  amplify_secure_storage: ">=0.5.20 <0.6.0"
   aws_common: ">=0.7.15 <0.8.0"
   device_info_plus: ^13.1.0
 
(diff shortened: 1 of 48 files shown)


packages/analytics/amplify_analytics_pinpoint/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: amplify_analytics_pinpoint
 description: The Amplify Flutter Analytics category plugin using the AWS Pinpoint provider.
-version: 2.14.0
+version: 2.15.0
```
VarshithaPamisetty: Is this version bump needed here? seems unrelated to the regen.

cadivus: Ohh, good catch!

packages/analytics/amplify_analytics_pinpoint/pubspec.yaml
```diff
@@ -20,9 +20,9 @@ platforms:
 
 dependencies:
   amplify_analytics_pinpoint_dart: ">=0.4.20 <0.5.0"
-  amplify_core: ">=2.14.0 <2.15.0"
-  amplify_db_common: ">=0.4.20 <0.5.0"
-  amplify_secure_storage: ">=0.5.20 <0.6.0"
+  amplify_core: ">=2.15.0 <2.16.0"
+  amplify_db_common: ">=0.4.21 <0.5.0"
+  amplify_secure_storage: ">=0.5.21 <0.6.0"
```
VarshithaPamisetty: Should these dependency constraint bumps be reverted too? they go with the version bump you reverted. The package is still 2.14.0 but now requires amplify_core ≥2.15.0.

cadivus: Thanks for this one as well!

VarshithaPamisetty (approved):

PR #7318, opened 2026-08-26
test(authenticator): Retry user creation when the generated phone number already exists

VarshithaPamisetty:
## Description

The phone-based authenticator sign-in tests use a generated `555-01XX` number as the Cognito username. On the shared user pool it can collide with a leftover user, so `adminCreateUser` throws `UsernameExistsException` in `setUp` and the test fails before running.

Add `retryOnUsernameExists` (with an `adminCreateUserWithGeneratedPhoneNumber` wrapper) that regenerates the number and retries, and use it in the affected tests.

The same shared-pool collision also affects the "Sign in with unknown credentials" tests. They sign in with a generated number expecting `UserNotFoundException`, but if that number matches a leftover user, Cognito returns `NotAuthorizedException` and the "User does not exist" banner never shows.

Add `generateUnusedUSPhoneNumber` (on a generic `retryUntilUnused` helper) that tries signing in and regenerates until the number returns `UserNotFoundException`, then use it in those tests.

## Tests

Verified on #7317, which stacks this fix on the Android e2e branch and runs the full authenticator e2e. The previously-failing `sign-in-sms-mfa` tests pass on Android, iOS, and Windows.

- `retryUntilUnused` is unit-tested (immediate success, regenerate-on-collision, max-attempts, error propagation).
- Verified against a real Cognito pool: an unknown number returns `UserNotFoundException`; an existing user with a wrong password returns `NotAuthorizedException`.
- Ran the "Sign in with unknown credentials" test on macOS against a live pool. The helper returned a free number and the sign-in flow proceeded.


cadivus (commented):
Overall, I like the idea. However, the wasm tests show:
```
Expected: exactly one matching candidate containing User does not exist descending from widgets with key ['authenticatorBanner']: []
Which: means none were found but one was expected
```

I am not sure if the idea is complete seeing that multiple workflows still fails with known patterns.

VarshithaPamisetty:
I dug into the wasm failure and found one more bug that caused the failure: in `Sign in with unknown credentials`, if the generated number happens to match a leftover user in the shared pool, sign-in returns `NotAuthorizedException` instead of `User does not exist`, so the test fails. 
Committed a fix that regenerates the number until it's actually unused.

VarshithaPamisetty:
The 2 failing web e2e checks are the pre-existing browser-attach hang (all attempts time out before any test runs, on both js and wasm), unrelated to the change.

Issue #6401, opened 2025-10-03
AWS S3: Native crash when calling cancel() on Storage operations


dkliss:
### Description

**Description:**

Calling cancel() on active storage operations causes the application to crash at the native level with no Dart exception or stack trace. The crash affects both download and upload operations and occurs specifically when:
1. A storage operation is actively transferring data
2. Progress callbacks are being fired
3. cancel() is called on the operation

**Environment:**

- Platform: macOS, iOS, Android (all native platforms)
- Amplify Flutter version: 2.6.5
  - amplify_flutter: ^2.6.5
  - amplify_auth_cognito: ^2.6.5
  - amplify_storage_s3: ^2.6.5
  - amplify_api: ^2.6.5
- Flutter version: 3.35.3 (stable channel)
- Dart version: 3.9.2

**Expected Behavior:**

Calling cancel() should gracefully stop the operation and allow the result future to complete with a cancelled state, allowing the application to handle the cancellation properly.

**Actual Behavior:**

The application crashes immediately with "Lost connection to device" - no Dart exception is caught, indicating a native-level crash in the Amplify SDK that bypasses Dart's exception handling.

**Reproduction Steps:**

Download Operation (downloadData):
```dart
final operation = Amplify.Storage.downloadData(
  path: StoragePath.fromString(key),
  onProgress: (progress) {
    print('Progress: ${progress.transferredBytes}/${progress.totalBytes}');
  },
);

// Start download, let it transfer data for a few seconds
// Then call cancel - app crashes
await operation.cancel();
```

Download Operation (downloadFile):
```dart
final operation = Amplify.Storage.downloadFile(
  localFile: AWSFile.fromPath(localFilePath),
  path: StoragePath.fromString(key),
  onProgress: (progress) {
    print('Progress: ${progress.transferredBytes}/${progress.totalBytes}');
  },
);

// Start download, let it transfer data for a few seconds
// Then call cancel - app crashes
await operation.cancel();
```

Upload Operation (uploadData):
```dart
final payload = StorageDataPayload.bytes(bytes);
final operation = Amplify.Storage.uploadData(
  path: StoragePath.fromString(key),
  data: payload,
  onProgress: (progress) {
    print('Progress: ${progress.transferredBytes}/${progress.totalBytes}');
  },
);

// Start upload, let it transfer data for a few seconds
// Then call cancel - app crashes
await operation.cancel();
```

Upload Operation (uploadFile):
```dart
[...]

PR #7353, opened 2026-08-28
fix(storage): Prevent crash when cancelling an active S3 download

VarshithaPamisetty:
## Description

Cancelling an active S3 download could complete its result `Completer` twice: once from `cancel()` or an error path, and again from the body stream's terminal event. The second completion threw `Bad state: Future already completed`, an uncaught async error that could crash the app. This guards every completion site in `S3DownloadTask` so the result completes at most once.

It also makes `cancel()` await the download starting, so cancelling before the response returns actually stops the download instead of finishing as success.

## Testing

- Regression test: an error-then-done on the body stream no longer double-completes (fails without the change).
- Cancelling before the getObject response returns actually stops the download and ends `canceled` (fails without the change).
- `amplify_storage_s3_dart` green on VM and web; `dart analyze --fatal-infos --fatal-warnings` and `dart format` clean.
- Verified e2e on iOS against live S3.

## Follow-up

The same completion-race guard will be applied to `S3UploadTask` in a separate PR.

Related to #6401


packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_download_task.dart
```diff
@@ -242,9 +242,11 @@ class S3DownloadTask {
     _bytesSubscription = null;
 
     _emitTransferProgress();
-    _downloadCompleter.completeError(
-      s3_exception.s3ControllableOperationCanceledException,
-    );
+    if (!_downloadCompleter.isCompleted) {
```
ekjotmultani, line 248: `cancel()` can run before the `getObject` response comes back, and at that point `_bytesSubscription` is still null so there's nothing to cancel. `start()` then resumes and subscribes to the stream anyway, so the download just keeps going after the cancel and `onDone` marks it success. The crash used to be the only symptom of this, so the guard makes it silent. I think `cancel()` needs to `await _getObjectInitiated` first like `pause()` does.

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_download_task_test.dart
```diff
@@ -654,6 +654,62 @@ void main() {
         },
       );
 
+      test('result should complete only once when the body stream emits an '
```
ekjotmultani, line 715: Can we add a test that cancels mid-stream? That path isn't covered.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_download_task.dart
VarshithaPamisetty: Yes true, thanks for pointing it out. Made changes so that cancel() now awaits `_getObjectInitiated` first.

**Heads-up:** I tested another scenario and found that if the `download fails` before its stream is set up and you then call `cancel()`, cancel hangs forever (the test timed out) since it's waiting on a `started` signal that never fires. `pause()` has the same issue since it awaits the same thing. 
Might need a new PR for the fix.

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_download_task_test.dart
VarshithaPamisetty: Added one

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_download_task.dart
```diff
@@ -288,19 +293,21 @@ class S3DownloadTask {
...
-                  // retrieve via a separate `HeadObject` call.
-                  // To unify the behavior on `downloadOptions.getProperties`
-                  // we hide the metadata from the result on VM if this parameter
-                  // is set to `false`.
-                  _s3PluginOptions.getProperties
-                      ? _downloadedS3Item
-                      : S3Item(path: _downloadedS3Item.path),
-                );
+                if (!_downloadCompleter.isCompleted) {
+                  _downloadCompleter.complete(
```
cadivus, line 316: Guard the terminal outcome before callbacks run

Because the body stream keeps listening after an error, its error and done callbacks can both run. The done callback sets success, invokes completion work, and emits progress before this guard, so a trailing stream error can still produce success side effects and race the failure result. Select and guard the terminal outcome before callbacks mutate state.

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_download_task_test.dart
```diff
@@ -654,6 +712,62 @@ void main() {
...
+            options: defaultTestOptions,
+            onProgress: (_) {},
+          );
+
+          await downloadTask.start();
+          // Listen so the (error) result is never itself uncaught.
+          unawaited(downloadTask.result.then((_) {}, onError: (_) {}));
+
+          // The body subscription uses the default `cancelOnError: false`,
+          // so both onError and onDone fire and each completes the result.
```
cadivus: Describe the guarded completion behavior

Both callbacks can run, but the new completion guards allow only the first callback to complete the result. The comment still describes the previous double-completion behavior.

```suggestion
          // With `cancelOnError: false`, both callbacks run; only the first may
          // complete the result.
```

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_download_task.dart
```diff
@@ -288,19 +293,21 @@ class S3DownloadTask {
...
+                    // On VM, download operation gets object metadata directly
+                    // from the underlying `GetObject` call.
+                    // On Web, download operation is done by browser download from
+                    // object presigned URL, where object metadata needs to be
+                    // retrieve via a separate `HeadObject` call.
+                    // To unify the behavior on `downloadOptions.getProperties`
+                    // we hide the metadata from the result on VM if this parameter
+                    // is set to `false`.
```
cadivus: Suggestion: Condense the metadata compatibility comment

This comment repeats platform and option details that can be stated more directly, obscuring the compatibility rationale.

```suggestion
                    // VM gets metadata from GetObject; web browser downloads use presigned URLs
                    // and require HeadObject. Omit VM metadata when getProperties is false for
                    // consistent behavior.
```

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_download_task.dart
VarshithaPamisetty: YEs, moved the guard to the top of `onDone` and the error path, so nothing runs before the outcome is picked.

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_download_task_test.dart
VarshithaPamisetty: Yes, made changes

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_download_task.dart
VarshithaPamisetty: Yes, made changes

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_download_task.dart
```diff
@@ -228,9 +233,10 @@ class S3DownloadTask {
   ///
   /// A canceled [S3DownloadTask] is not resumable.
   Future<void> cancel() async {
-    if (_state == StorageTransferState.canceled ||
-        _state == StorageTransferState.success ||
-        _state == StorageTransferState.failure) {
+    // ensure the task has actually started before cancelling
+    await _getObjectInitiated;
```
cadivus: Record cancellation before waiting for GetObject

An early cancel now waits for the GetObject response before marking the task canceled. If that request hangs, cancel and result never complete; if it fails first, the caller receives the request failure instead of the documented cancellation outcome.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_download_task.dart
```diff
@@ -314,21 +335,38 @@ class S3DownloadTask {
...
+          ..onError(_terminateDownloadOnError);
 
     // After setting up the body stream listener, we consider the task is fully
     // started, and can be paused etc.
-    _getObjectCompleter?.complete();
+    _completeGetObjectCompleter();
+  }
+
+  // in parallel the stream error callback and the start/resume error paths can
+  // all invoke this, use this to avoid a second terminal outcome
```
cadivus: Suggestion: Condense the terminal-outcome rationale

This run-on comment can state the only non-obvious rationale, guarding competing error paths, in one short sentence.

```suggestion
  // Prevent competing error paths from recording a second terminal outcome.
```

packages/storage/amplify_storage_s3_dart/test/storage_s3_service/task/s3_download_task_test.dart
```diff
@@ -413,6 +413,100 @@ void main() {
...
+          },
+        );
+
+        unawaited(downloadTask.start());
+        final expectation = expectLater(
+          downloadTask.result,
+          throwsA(isA<StorageOperationCanceledException>()),
+        );
+        // Cancel before the getObject response returns; the download must
+        // actually stop (its body stream is cancelled), not run to success.
```
cadivus: Suggestion: Remove the repeated cancellation narration

The test title already states the timing, and the next assertions verify canceled state and body-stream cancellation. This block adds no information.

PR #7354, opened 2026-09-01
fix(storage): Prevent crash when cancelling or failing an active S3 upload

VarshithaPamisetty:
## Description

Cancelling or failing an active S3 upload could complete its result `Completer` twice: once from `cancel()` or a part-failure path, and again from another concurrent completion (for multipart, two parts failing at once). The second completion threw `Bad state: Future already completed`, an uncaught async error that could crash the app. This guards every completion site in `S3UploadTask` so the result completes at most once.

## Testing

- Regression test: concurrent multipart part failures no longer double-complete (fails without the change).
- `amplify_storage_s3_dart` green on VM and web; `dart analyze --fatal-infos --fatal-warnings` and `dart format` clean.

Related to #6401


packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_upload_task.dart
```diff
@@ -772,7 +778,9 @@ class S3UploadTask {
     await _s3Client.abortMultipartUpload(request).result;
```
cadivus, line 778: Probably just a follow-up: Do we need to set state to `failure` after this await?

cadivus: Or before that await, not sure (concurrency).

VarshithaPamisetty: Yes, I believe setting `failure` before the await would let the guard dedupe the parallel aborts. If we set it after (as now), they all pass the guard during the await and each fires its own abort. 
I'd do it as a follow-up PR since it's an existing multipart concurrency issue, not part of the current fix.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_upload_task.dart
```diff
@@ -787,7 +795,9 @@ class S3UploadTask {
   void _completeUploadWithError(Object error, [StackTrace? stackTrace]) {
     _state = StorageTransferState.failure;
     _emitTransferProgress();
-    _uploadCompleter.completeError(error, stackTrace);
+    if (!_uploadCompleter.isCompleted) {
+      _uploadCompleter.completeError(error, stackTrace);
+    }
```
cadivus, line 800: Concurrent abort failures can escape

The guards prevent duplicate completer writes, but concurrent callers can all pass the earlier failure-state check before abort completes. Each caller then sends AbortMultipartUpload, and a later abort failure can escape the unawaited termination future as an uncaught error. Serialize termination with a shared Future (e.g. a simple lock), or claim the failure state before awaiting abort. Marking the operation in progress before the await must also preserve the original failure for all callers.

VarshithaPamisetty: Yeah, kept this PR to just the `isCompleted` guards. 
The concurrency part is done as a follow-up in #7384. It sets failure before the abort await so the parallel failures dedupe to a single abort, and wraps the abort in a try/catch, so if the abort itself fails it doesn't throw an uncaught error, and the caller still gets the original failure.

packages/storage/amplify_storage_s3_dart/lib/src/storage_s3_service/service/task/s3_upload_task.dart
cadivus: Claim termination before awaiting the abort request

Parallel part failures can all pass the state guard before the first abort completes because the failure state is assigned only afterward. Each handler can then send an abort request, and a later duplicate failure can escape from its unawaited termination future.

cadivus (approved):

Issue #7355, opened 2026-09-02
Uncaught 'Bad state: No element' from worker.stream.first in auth state machine (Flutter web)


raffaji:
### Description

On Flutter **web**, shortly after `Amplify.Auth.signIn` completes (while the AppSync subscription WebSocket is concurrently cycling — our app subscribes to a custom `onAlertTriggered` subscription at startup), an **uncaught async error** surfaces in the root zone:

```
RethrownDartError: Bad state: No element
    dart-sdk/lib/_internal/js_dev_runtime/patch/core_patch.dart 749:28                get current
dart-sdk/lib/async/stream.dart 1653:32                                            <fn>
dart-sdk/lib/async/zone.dart 879:9                                                runGuarded
dart-sdk/lib/async/stream_impl.dart 434:7                                         sendDone
dart-sdk/lib/async/stream_impl.dart 444:7                                         [_sendDone]
dart-sdk/lib/async/stream_impl.dart 592:13                                        perform
dart-sdk/lib/async/stream_impl.dart 678:10                                        handleNext
dart-sdk/lib/async/stream_impl.dart 649:7                                         <fn>
dart-sdk/lib/async/schedule_microtask.dart 40:34                                  _microtaskLoop
dart-sdk/lib/async/schedule_microtask.dart 49:5                                   _startMicrotaskLoop
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 118:77  tear
dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 183:69               <fn>
```

The sign-in itself succeeds and the session works; the error is uncaught noise, but it escapes Amplify entirely and hits the app's `runZonedGuarded`/root error handler, which we treat as a crash signal.

### Analysis

`dart-sdk/lib/async/stream.dart:1653` is the `onDone` callback of **`Stream.first`** — when a stream closes without emitting, `.first` completes its future with `IterableElementError.noElement()`. So something is awaiting `.first` on a stream that only ever fires `done`.

The only `Stream.first` calls active at that moment in the dependency tree are the worker_bee response streams:

- `amplify_auth_cognito_dart` — `lib/src/state/machines/sign_in_state_machine.dart` lines 143, 445, 485, 1127: `worker.stream.first` (SRP init, password verifier, SMS MFA, device confirm)
- `amplify_secure_storage_dart` — `lib/src/amplify_secure_storage_dart.dart:126`: `await _worker.stream.first`

[...]

PR #7374, opened 2026-09-04
fix(auth): Handle worker streams which close without a response

VarshithaPamisetty:
## Description

During sign-in, `SignInStateMachine` awaited `worker.stream.first` for each worker's response. If that stream closes without emitting, `.first` throws `StateError` ("Bad state: No element"), which is an `Error` rather than an `Exception`, so `resolveError` let it escape to the app's root zone uncaught.

The fix adds `WorkerBeeCommon.nextResponse`, which surfaces the worker's real error instead, wrapped as an `UnknownException` at all four sign-in call sites so callers get an `AuthException`. The request send sits inside the guard too, so a send to an already-closed worker ("Cannot add event after closing") is wrapped rather than thrown raw.

All workers on the sign-in path were tested. `ConfirmDeviceWorker` reproduced the reported behavior most closely, since it runs after Cognito has returned tokens, so the user ends up signed in while the error still surfaces.

## Testing

- Regression tests: signing in with a `ConfirmDeviceWorker` which never responds, and one which reports an `Error`, both now fail with an `AuthException`. Without the change both fail with `Bad state: No element`.
- `worker_bee` unit tests for `nextResponse` cover a normal response, an empty close, a worker-reported exception, and a worker-reported `Error`.
- `worker_bee` suite green on VM and Chrome; `amplify_auth_cognito_test` and `amplify_secure_storage_test` green.
- `dart analyze --fatal-infos --fatal-warnings`, `dart format` and `dart fix` clean.
- Verified e2e on Flutter web against live Cognito: the uncaught `Bad state: No element` no longer occurs, `signIn` throws a catchable `AuthException`, and the session remains valid.

Refs #7355

### DIFF
--- a/packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_in_state_machine.dart
+++ b/packages/auth/amplify_auth_cognito_dart/lib/src/state/machines/sign_in_state_machine.dart
@@ -136,11 +136,32 @@ final class SignInStateMachine
         return worker;
       });
 
+  /// Awaits [response] from the worker named [workerName], surfacing a failed
+  /// worker as an [AuthException].
+  Future<R> _workerResponse<R>(
+    String workerName,
+    Future<R> Function() response,
+  ) async {
+    try {
+      return await response();
+    } on Object catch (e, st) {
+      Error.throwWithStackTrace(
+        UnknownException(
+          'Could not get a response from the $workerName worker',
+          underlyingException: e,
+        ),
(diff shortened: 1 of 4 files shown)


cadivus:
The changes look good, thanks! I am just not completely sure about the failing workflows (just the usual flakiness?). Please  look into that!

VarshithaPamisetty:
Yes, the failing tests are just flaky ones.

PR #7375, opened 2026-09-04
chore: Regenerate autogenerated files

cadivus:


packages/api/amplify_api/example/macos/Flutter/GeneratedPluginRegistrant.swift
```diff
@@ -10,7 +10,6 @@ import amplify_secure_storage
 import connectivity_plus
 import device_info_plus
 import package_info_plus
-import path_provider_foundation
```
VarshithaPamisetty, line 13: Noticed this only removes `path_provider_foundation` from 2 of the macOS example registrants. 9 others still have it, including the `templates/flutter-package` brick. Did the regen run across all the examples? If we're dropping it, we should probably regen the rest here too so they don't drift.

VarshithaPamisetty (approved):

PR #7402, opened 2026-09-15
ci: skip heavy workflows on draft PRs

ekjotmultani:
## Description

Gates the heavy CI workflows (the 10 E2E `*_example` workflows and 9 native Android/iOS build-test workflows) so they don't run on draft PRs. Analyze, format, and unit test workflows still run on drafts. Marking a PR ready for review triggers the full suite; any gated workflow can also be run on a draft manually via `workflow_dispatch`.

Workflows are generated, so the change lives in the aft generator and the regenerated workflows are committed. Gated workflows get `ready_for_review` in their `pull_request` types and this on every job:

```yaml
if: github.event_name != 'pull_request' || github.event.pull_request.draft == false
```

`push`, `schedule`, and `workflow_dispatch` triggers are unaffected.

## Testing

- `dart test` in `packages/aft`: 117/117 pass, `dart analyze` and `dart format` clean
- `aft generate workflows --set-exit-if-changed`: workflows in sync
- Parser check over all workflow files: exactly 19 gated, every job in each carries the condition


cadivus:
Shouldn't we at least run one kind of e2e tests? I think Linux could be good here, these are the cheapest.

PR #7420, opened 2026-09-21
fix(authenticator): Drop a repeated dial code from the phone number field

VarshithaPamisetty:
## Description

Selecting a dial code and typing it again submitted it twice: `94771234567` entered under `+94` reached Cognito as `+9494771234567`, so signing in with the real number failed. `formatPhoneNumber` prefixed the dial code without checking whether the entry already began with it.

The entry is now resolved with `phone_numbers_parser` and a repeated dial code or trunk prefix is dropped. A value that does not resolve to a number the selected region assigns is left as typed, so the doubled entry is the only output that changes. Germany, Indonesia, Sweden and Kosovo assign the doubled form as a real number, so it stays untouched there.

## Tests

- Unit tests for a repeated dial code, two trunk prefixes, a NANP area code, and a national number that legitimately starts with its own dial code (KZ); each fails without the change.
- Widget tests assert the value handed to Cognito: five spellings of the same number converge on `+94771234567`.
- Across all 240 regions, no valid number the parser publishes is altered and none is reduced to empty.
- Suite green on VM and Chrome; `flutter analyze --fatal-infos --fatal-warnings` and `dart format` clean. Verified e2e against live Cognito.

Related to #4079


VarshithaPamisetty:
The failing e2e checks are known flakes, unrelated to the change. The datastore `android/ios` ones are the API-sync `setUpAll` timeout, and the web ones are the pre-existing browser-attach hang where all attempts time out before any test runs. 
`amplify_authenticator` itself and all non-e2e checks are green.

ekjotmultani (dismissed):

cadivus (approved):

